PR #8, opened 2019-10-07
Custom generator code in the react sample

samelhusseini:
Add custom generation code to the react sample, and added links to documentation.

### DIFF
--- a/blockly-react/src/App.js
+++ b/blockly-react/src/App.js
@@ -27,15 +27,25 @@ import './App.css';
 import logo from './logo.svg';
 
 import BlocklyComponent, { Block, Value, Field, Shadow } from './Blockly';
+
+import BlocklyJS from 'blockly/javascript';
+
 import './blocks/customblocks';
+import './generator/generator';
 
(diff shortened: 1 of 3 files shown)


blockly-react/src/blocks/customblocks.js
```diff
@@ -1,11 +1,52 @@
...
+      "type": "field_react_component",
+      "name": "FIELD",
+      "text": "Click me"
+    },
+  ],
+  "previousStatement": null,
+  "nextStatement": null,
+};
 
 Blockly.Blocks['test_react_field'] = {
```
rachel-fenichel, line 47: FWIW I think you could use defineBlocksWithJsonArray here (assuming json supports style)

rachel-fenichel (approved):

blockly-react/src/blocks/customblocks.js
samelhusseini: Yeah, I wanted to keep the two there for demonstration purposes.

PR #15, opened 2019-10-24
Add advanced usage example by Vue

abbychau:
This PR adds a Vue project to show how Blockly can be "componentized" in a more realistic scenario.


blockly-vue-advanced/src/components/BlocklyComponent.vue
```diff
@@ -0,0 +1,225 @@
...
+ */
+
+import Blockly from "blockly";
+import Swal from "sweetalert2";
+
+var renameVar = function(name) {
+  return name;
+};
+
+Blockly.prompt = function(msg, defaultValue, callback) {
```
samelhusseini: Can we organize this a little. My suggestion would be to move blocks into a separate file. And possibly the prompt into a blockly_configure.js file or something like that?

samelhusseini:
@abbychau, I'm not certain this warrants having a separate sample for it. I'm quite happy moving this into the blockly-vue sample. Do you see any reason not to do that?

blockly-vue-advanced/src/components/BlocklyComponent.vue
abbychau: This would be a good idea. I am on a trip so I can move it a few days later.

abbychau:
> @abbychau, I'm not certain this warrants having a separate sample for it. I'm quite happy moving this into the blockly-vue sample. Do you see any reason not to do that?

I just thought that blockly-* should had meant to be with more consistent scope of demonstration. I think it can be merging into blockly-vue as well. I have no way to contact with you so I opened this PR for a more direct communication xD

blockly-vue-advanced/src/components/BlocklyComponent.vue
samelhusseini: No problem, take your time, and thanks again!

samelhusseini (commented):
Yeah I think we could use these samples to demonstrate both how to integrate Blockly into each of these frameworks, plus give a little guidance of how to structure code for blocks, fields, and generators.

Opening a PR is definitely the way to go! Thank you!

abbychau:
> Yeah I think we could use these samples to demonstrate both how to integrate Blockly into each of these frameworks, plus give a little guidance of how to structure code for blocks, fields, and generators.
> 
> Opening a PR is definitely the way to go! Thank you!

Now ``blockly-[a-z]`` is sharing very similar scope of demonstration. So for the purity and the ease of grasping the basic usage of blockly, it may be good to keep a bare-minimal project.

Although I think that this can be replaced but it is up to you, do you prefer to replace ``blockly-vue`` with this project? if yes, I will update it.

abbychau:
@samelhusseini

samelhusseini:
I think it depends, but for what you're adding here, which is a couple of sample blocks, I would keep it in the same sample, just split up into separate files so each file can be described with what it's purpose is. Does that make sense?

abbychau:
Yes.  I removed `<slot></slot>` and `<block>` doms in App.vue.

I think that is also useful to guide usage of vue.

abbychau:
I merged them. Have a look? @samelhusseini

blockly-vue/src/App.vue
```diff
@@ -49,30 +56,47 @@
 
 /**
  * @fileoverview Main Vue component that includes the Blockly component.
- * @author samelh@google.com (Sam El-Husseini)
+ * @author abbychau@gmail.com (Abby Chau)
```
samelhusseini, line 59: Just add your name on another line if you'd like.

blockly-vue/src/components/BlocklyComponent.vue
```diff
@@ -29,30 +26,197 @@
...
+    allowOutsideClick: () => !Swal.isLoading()
+  }).then(result => {
+    if (result.value) {
+      // console.log(result);
+      callback(renameVar(result.value));
+    }
+  });
+};
+
+Blockly.Blocks["stock_buy_simple"] = {
```
samelhusseini, line 58: Can we split this into another file under ``/src/blocks/``?

samelhusseini: see the react example as a reference: https://github.com/google/blockly-samples/blob/master/blockly-react/src/blocks/customblocks.js

blockly-vue/src/components/BlocklyComponent.vue
```diff
@@ -29,30 +26,197 @@
...
+      return Blockly.JavaScript.workspaceToCode(this.workspace);
     }
   },
   mounted() {
-    this.workspace = Blockly.inject(document.getElementById('blocklyDiv'), {
-      toolbox: document.getElementById('toolbox')
-    })
+
+
+    var xml = `<xml>
```
samelhusseini, line 168: Can you accept the xml as a prop to make this component extensible?

blockly-vue/src/components/BlocklyComponent.vue
```diff
@@ -29,30 +26,197 @@
...
+    <category name="Stocks" colour="%{BKY_LOOPS_HUE}">
+      <block type="stock_buy_simple"></block>
+      <block type="stock_buy_prog"></block>
+      <block type="stock_fetch_price"></block>
+      
+    </category>
+    </xml>`;
+    //console.log(xml)
+    this.workspace = Blockly.inject(document.getElementById("blocklyDiv"), {
+      toolbox: Blockly.Xml.textToDom(xml) //document.getElementById('toolbox')
```
samelhusseini, line 210: remove comment.

samelhusseini (commented):
@abbychau I'm interested if we're able to create a single component that can: 
- accept xml as a prop. If so, use that as the toolbox
- otherwise, use the slot for the toolbox

blockly-vue/vue.config.js
```diff
@@ -1,6 +1,6 @@
 // vue.config.js
 module.exports = {
   devServer: {
-    port: 3000
+    port: 3001
```
samelhusseini, line 4: Why the change in port?

samelhusseini:
Closing in favour of https://github.com/google/blockly-samples/pull/17. Thank you @abbychau!

PR #16, opened 2019-11-01
Initialize SQlite database for realtime collab

navilperez:


blockly-rtc/src/db.js
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview The SQlite database object.
+ * @author navil@google.com (Navil Perez)
+ */
+const sqlite3 = require('sqlite3');
+
+let db = new sqlite3.Database("./events.sqlite", function(error) {
```
samelhusseini: const

blockly-rtc/src/db_utils.js
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Util scripts for interacting with the database.
```
samelhusseini, line 19: Add @author

blockly-rtc/src/db_utils.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+var db = require('./db');
```
samelhusseini: const

blockly-rtc/src/db_utils.js
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Query the database for rows since the given server id.
+ * @param {number} serverId serverId for the lower bound of the query.
+ * @return {!Array.<!Object>} The rows since the last given server id.
+ * @public
+ */
+function queryDatabase(serverId) {
+  var sql = `SELECT * from events WHERE serverId > ${serverId};`;
+  db.all(sql, [], (err, rows) => {
```
samelhusseini: You can probably lose the ``sql`` variable here.

blockly-rtc/src/db_utils.js
```diff
@@ -0,0 +1,85 @@
...
+};
+
+/**
+ * Add rows to the database.
+ * Rows are added transactionally.
+ * @param {!Array.<!Object>} rows The rows to be added to the database.
+ * @public
+ */
+function addToServer(rows) {
+  var insertQueries = [];
```
samelhusseini: const

blockly-rtc/src/db_utils.js
```diff
@@ -0,0 +1,85 @@
...
+
+    db.run('COMMIT;');
+  });   
+
+};
+
+/**
+ * Create a SQlite query script to insert a new row.
+ * @param {!Object} row serverId for the lower bound of the query.
+ * @return {!String} The SQlite query.
```
samelhusseini: lowercase string and strings are non-nullable by default, so ``{string}``

samelhusseini (commented):
What's in the binary? Does it need to be checked in or can it be generated by the service when it starts?

navilperez:
> What's in the binary? Does it need to be checked in or can it be generated by the service when it starts?

It is the binary for the database since I am using an in-disk database. I just ran the node server without it and it seems as if the service can generate it, so it does not need to be checked in. I will go ahead and remove it. Thank you for drawing my attention to that.

blockly-rtc/src/db.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+/**
+ * @fileoverview The SQlite database object.
+ * @author navil@google.com (Navil Perez)
+ */
+
+const sqlite3 = require('sqlite3');
+
+const db = new sqlite3.Database("./events.sqlite", (err) => {
```
samelhusseini: Following the style guide, use single quotes here.

samelhusseini (commented):
One last nit but otherwise lgtm. My preference would be for this to go into a branch, until the demo the ready for release. Placing these files there on there own might confuse people using this samples repo, especially without a README.

samelhusseini (approved):

samelhusseini:
LGTM.

PR #17, opened 2019-11-02
Extend Blockly Vue Component to accept workspace options 

samelhusseini:
@abbychau I opened this PR to share my ideas on having the one Vue BlocklyComponent that can accept both workspace options (which the toolbox can be a part of), and falls back on using the ``<slot/>`` for the toolbox if the option isn't passed in.

Extension of https://github.com/google/blockly-samples/pull/15


abbychau:
This could work. LGTM

PR #20, opened 2019-11-14
Client-side logic for querying the database.

navilperez:
WorkspaceClient.js: Adds logic for querying events from the database and
resolving conflicts between the order of events on the server and on the
client.

index.js: Logic for triggering a query request and running a series of
events as a result of the query response.

### DIFF
--- a/blockly-rtc/src/WorkspaceClient.js
+++ b/blockly-rtc/src/WorkspaceClient.js
@@ -32,6 +32,7 @@ import Blockly from 'blockly';
 export default class WorkspaceClient {
     constructor(workspaceId) {
         this.workspaceId = workspaceId;
+        this.lastSync = 0;
         this.inProgress = [];
         this.notSent = [];
         this.activeChanges = [];
@@ -43,7 +44,7 @@ export default class WorkspaceClient {
      * Create and entry from an event and add it to active changes.
      * An entry is of the form {"event": Blockly.Event, "entryId": string} where
      * event is an event created on the workspace and entryId is of the form
-     * {workspaceId}:{counter}.
+     * {workspaceId}{counter}.
      * @param {!Object} event The Blockly.Event JSON created by the client.
      * @public
      */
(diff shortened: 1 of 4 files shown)


blockly-rtc/src/WorkspaceClient.js
```diff
@@ -108,4 +109,93 @@ export default class WorkspaceClient {
...
+      };
+    };
+
+    /**
+     * Compare the order of events in the rows retrieved from the database to
+     * the stacks of local-only changes and provide a series of steps that
+     * will allow the server and local workspace to converge.
+     * @param {<!Array<!Object>} rows Rows of event entries retrieved by
+     * querying the database.
+     * @returns {!Array<!Object>} eventQueue An array of events and the
```
rachel-fenichel: Can you make these annotations more specific?  Using Object is correct but not very specific.  

You may need to use record types or typedefs.  More information here: https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System

rachel-fenichel: And a nit on annotations: we use `{!Array.<!Object>}`.  Notice the extra period.  Both are valid, so this is just about consistency.

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -108,4 +109,93 @@ export default class WorkspaceClient {
...
+     * @private
+     */
+    processQueryResults_(rows) {
+        const eventQueue = [];
+
+        if (rows.length == 0) {
+            return eventQueue;
+        };
+    
+        this.lastSync = rows[rows.length-1].serverId;
```
rachel-fenichel: Nit: spaces around the `-`

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -108,4 +109,93 @@ export default class WorkspaceClient {
...
+        if (this.inProgress.length > 0 && rows[0].entryId == this.inProgress[0].entryId) {
+            rows = rows.slice(this.inProgress.length);
+            this.inProgress = [];
+        };
+    
+        if (rows.length > 0) {
+            // Undo local events.
+            this.notSent.slice().reverse().forEach((row) => {
+                eventQueue.push({
+                    event: row.event,
```
rachel-fenichel: Sometimes you're calling `JSON.parse(row.event)` and not others.  What's the difference?

rachel-fenichel: And if they should all be one way or the other, you can pull the callback out into a separate function since it's the same for each of these `forEach` calls.

rachel-fenichel (commented):
Preliminary comments.

blockly-rtc/src/WorkspaceClient.js
navilperez: The difference is that the events are stored differently. There is actually a little more background on this that might be worth discussing offline if you have a chance.

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -108,4 +109,108 @@ export default class WorkspaceClient {
         this.writeInProgress = false;
     };
 
+    /**
+     * A row from the database.
```
rachel-fenichel: Keep your typedefs together and at the top of the file.

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -108,4 +109,108 @@ export default class WorkspaceClient {
...
+     * @typedef {Object} WorkspaceEvent
+     * @property {string} event The JSON of a Blockly event.
+     * @property {boolean} forward Indicates the direction to run an event.
+     */
+
+    /**
+     * Compare the order of events in the rows retrieved from the database to
+     * the stacks of local-only changes and provide a series of steps that
+     * will allow the server and local workspace to converge.
+     * @param {<!Array<!Row>} rows Rows of event entries retrieved by
```
rachel-fenichel: `<!Array.<!Row>>`

Add the period everywhere else you have arrays, too.

blockly-rtc/src/index.js
```diff
@@ -58,10 +59,51 @@ document.addEventListener('DOMContentLoaded', () => {
...
         });
     };
+
+    /**
+     * Periodically signal the WorkspaceClient to query the database and call
+     * runEvents_() with the result of the query.
+     * @private
+     */
+    function pollServer_() {
+        if (! workspaceClient.writeInProgress) {
```
rachel-fenichel: Nit: remove extra space between `!` and `workspaceClient`

rachel-fenichel (commented):
A few nits, then lgtm.

Per offline discussion, transactions currently require strings for the event objects, which is why there's some extra parsing going on.  In a change that Navil is working on now, she's fixing that requirement on both client and server side.

rachel-fenichel (approved):

PR #21, opened 2019-11-20
Batch events for writing to the database.

navilperez:
Currently, each event is written to the database as an individual entry.
With this change, the client batches any available events and writes
them to the database as a single entry. By writing all events to the
database in a single entry this removes the need to use a transaction.

Other:
WorkspaceClient.js: this.inProgress was an array of events, now
this.inProgress is an array of LocalEntry objects. Tracking a LocalEntry
object facilitates merging client and server events. Currently, only one
LocalEntry event will exist at a time, however, once websockets are
implemented, inProgress will hold multiple LocalEntry objects.

### DIFF
--- a/blockly-rtc/src/WorkspaceClient.js
+++ b/blockly-rtc/src/WorkspaceClient.js
@@ -22,7 +22,6 @@
  */
 
 import {getEvents, writeEvents} from './api';
-import Blockly from 'blockly';
 
 /**
  * An action to be performed on the workspace.
@@ -32,9 +31,18 @@ import Blockly from 'blockly';
  */
 
 /**
+ * A local representation of an entry in the database.
+ * @typedef {Object} LocalEntry
+ * @property {<!Array.<!Object>>} events An array of Blockly Events in JSON
+ * format.
+ * @property {string} entryId The id assigned to an event by the client.
+ */
+
+/**
  * A row from the database.
  * @typedef {Object} Row
- * @property {!Object} event The JSON of a Blockly event.
+ * @property {<!Array.<!Object>>} events An array of Blockly Events in JSON
+ * format.
  * @property {string} entryId The id assigned to an event by the client.
  * @property {string} serverId The id assigned to an event by the server.
  */
@@ -46,7 +54,7 @@ import Blockly from 'blockly';
  */
 export default class WorkspaceClient {
     constructor(workspaceId) {
-        this.workspaceId = workspaceId;
+      this.workspaceId = workspaceId;
         this.lastSync = 0;
         this.inProgress = [];
         this.notSent = [];
@@ -56,29 +64,12 @@ export default class WorkspaceClient {
     };
 
     /**
-     * Create and entry from an event and add it to active changes.
-     * An entry is of the form {"event": Blockly.Event, "entryId": string} where
-     * event is an event created on the workspace and entryId is of the form
-     * {workspaceId}{counter}.
-     * @param {!Object} event The Blockly.Event JSON created by the client.
-     * @public
-     */
-    addEvent(event) {
-        var entryId = this.workspaceId.concat(this.counter);
-        this.counter += 1;
-        this.activeChanges.push({
-            event: event,
-            entryId: entryId
-        });
-    };
-
-    /**
-     * Move events in a Blockly.Events group from activeChanges to notSent.
+     * Add the events in activeChanges to notSent.
      * @public
      */
     flushEvents() {
-        this.notSent = this.notSent.concat(this.activeChanges);
-        this.activeChanges = [];
+      this.notSent = this.notSent.concat(this.activeChanges);
+      this.activeChanges = [];
     };
 
     /**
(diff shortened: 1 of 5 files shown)


blockly-rtc/src/WorkspaceClient.js
```diff
@@ -87,39 +78,46 @@ export default class WorkspaceClient {
...
+     * Change writeInProgress to true. If write was successful, the LocalEntry
+     * remains in inProgress, otherwise, the LocalEntry is deleted and the
+     * events move back to the front of notSent.
      * @param {boolean} success Indicates the success of the database write.
      * @private
      */
     endWrite_(success) {
         if (!success) {
-            this.notSent = this.inProgress.concat(this.notSent);
+            this.notSent = this.inProgress[0].events.concat(this.notSent);
```
RoboErikG: This could be dangerous in the unlikely event that the server received the request, the response was lost, and new events were added in the meantime. In that case, you would resend the update with the same id, but with extra events. The server would recognize it as the same update and throw it away, losing the new events.

Instead, you'll want to keep the in progress update as is and retry it on the next iteration.

RoboErikG (commented):
Didn't take a close look at all the code, but noted an edge case to watch out for.

blockly-rtc/src/WorkspaceClient.js
navilperez: Thanks for pointing out this concern :)
I will create a task for implementing retry functionality and add it as a TODO.
I would like the focus of this PR to remain on the redesign of the storage schema, so I will address the TODO in a following PR.

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -155,60 +153,69 @@ export default class WorkspaceClient {
...
       };
-  
+
       this.lastSync = rows[rows.length - 1].serverId;
   
       // No local changes.
       if (this.notSent.length == 0 && this.inProgress.length == 0) {
         rows.forEach((row) => {
-          eventQueue.push(this.createWorkspaceAction_(row.event, true));
+          eventQueue.push.apply(
```
rachel-fenichel, line 171: Why do you need to use `push.apply` instead of `push` here?

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -155,60 +153,69 @@ export default class WorkspaceClient {
...
-          eventQueue.push(this.createWorkspaceAction_(row.event, false));
-        });
-        this.inProgress.slice().reverse().forEach((row) => {
-          eventQueue.push(this.createWorkspaceAction_(row.event, false));
-        });
+        eventQueue.push.apply(
+          eventQueue,
+          this.createWorkspaceActions_(this.notSent.slice().reverse(), false));
+        if (this.inProgress.length > 0) {
+          eventQueue.push.apply(eventQueue,this.createWorkspaceActions_(
```
rachel-fenichel: Nit: missing space after `eventQueue,`

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -155,60 +153,69 @@ export default class WorkspaceClient {
...
-        this.inProgress.forEach((row) => {
-          eventQueue.push(this.createWorkspaceAction_(row.event, true));
-        });
-        this.notSent.forEach((row) => {
-          eventQueue.push(this.createWorkspaceAction_(row.event, true));
-        });
+        if (this.inProgress.length > 0) {
+          eventQueue.push.apply(
+            eventQueue,
+            this.createWorkspaceActions_(this.inProgress[0].events, true));
```
rachel-fenichel, line 206: Is the queue guaranteed to be of length 1 here?

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -155,60 +153,69 @@ export default class WorkspaceClient {
...
+     * @returns {<Array.<!WorkspaceEvent>>} An array of actions to be performed
+     * on the workspace.
      * @private
      */
-    createWorkspaceAction_(event, forward) {
-      return {
-        event: event,
-        forward: forward
-      };
+    createWorkspaceActions_(events, forward) {
```
rachel-fenichel, line 222: :+1:

blockly-rtc/src/index.js
```diff
@@ -43,7 +43,7 @@ document.addEventListener('DOMContentLoaded', () => {
         if (event instanceof Blockly.Events.Ui) {
             return;
         };
-        workspaceClient.addEvent(event.toJson());
+        workspaceClient.activeChanges.push(event.toJson());
```
rachel-fenichel: Why did you get rid of `addEvent` and have the workspace client expose its `activeChanges` array instead?

rachel-fenichel (commented):
A few questions but looks good overall.

blockly-rtc/src/WorkspaceClient.js
navilperez: `this.createWorkspaceAction` would return one object at a time, so I added each object to the `eventQueue` using `push`.
Now, `this.createWorkspaceActions` returns an array of objects, so I am using `push.apply` to add the objects to `eventQueue` individually rather than using `push` which would add an array.

blockly-rtc/src/index.js
navilperez: I created `addEvent` to take in an event, create an object of the form `{event: event, entryId: entryIdString}` and add that object to `notSent`.
Now, that we are not creating an object, just adding an event to `activeChanges` I removed it. However, I do like the idea of keeping the function `addEvent` (just changing its functionality) as to not expose the `activeChanges` array.
What do you think?

rachel-fenichel: I think it's a good idea to keep `addEvent` to avoid exposing the extra array.

blockly-rtc/src/WorkspaceClient.js
navilperez: Yes, at this place in the code the eventQueue is guaranteed to be of at least length 1.
The only case I can think of for it not being at least 1 is if the all the events it got from the server were created by the client, in which case the workspace is up to date.

Or were you talking about `this.inProgress` being of length 1? In which case yes. At this point `this.inProgress` can only be of length 0 or 1. If it is of length 0, the function would have exit out at [L165](https://github.com/google/blockly-samples/pull/21/files/a799c08f6dd2860a3fc67b85e212152fe5780e92#diff-039d24c9cc7f1f76cd5c17b18c7cdbb7L165)

rachel-fenichel: I meant `this.inProgress`.  I see where it'll exit if it's of length 0.  If the length is only ever 0 or 1, why store it as an array instead of a nullable object?

navilperez: Discussed offline.

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -78,14 +87,14 @@ export default class WorkspaceClient {
...
     async writeToDatabase() {
-      this.beginWrite_();
-      try {
-        await writeEvents(this.inProgress[0]);
-        this.endWrite_(true);
-      } catch {
-        this.endWrite_(false);
-        throw Error('Failed to write to database.');
-      };
+        this.beginWrite_();
```
rachel-fenichel: Nit: get rid of the extra indentation here.

rachel-fenichel (commented):
One nit, then lgtm.

rachel-fenichel (approved):

PR #27, opened 2019-12-02
Set up websocket server and middleware.

navilperez:
This change adds a websocket server and middleware handlers for adding
data to the server and polling the server.
A following PR will implement broadcasting.


blockly-rtc/server/websocket_server.js
```diff
@@ -0,0 +1,43 @@
+/**
```
samelhusseini, line 1: Add copyright header.

blockly-rtc/server/websocket_server.js
```diff
@@ -0,0 +1,43 @@
...
+ * @fileoverview Node.js HTTP server for realtime collaboration.
+ * @author navil@google.com (Navil Perez)
+ */
+
+const socket = require('socket.io');
+const http = require('http');
+
+const sql = require('./db_utils');
+
+const PORT = 3001;
```
samelhusseini: rename to WS_PORT so it's clearer this is the web socket port.

blockly-rtc/src/websocket_handlers.js
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Realtime collaboration demo using Node.js and SQlite.
+ * @author navil@google.com (Navil Perez)
+ */
+
+import io from 'socket.io-client';
```
samelhusseini, line 24: Does ``socket.io-client`` need to be added to package.json?

samelhusseini: or is it included in the ``socket.io`` package?

navilperez: Discussed offline. It is imported in the socket.io package, but we are still including it in package.json to convey that it does need to be imported/

samelhusseini (approved):

blockly-rtc/src/index.js
```diff
@@ -22,7 +22,7 @@
  */
 
 import * as Blockly from 'blockly';
-import {getEvents, writeEvents} from './api';
```
rachel-fenichel, line 25: Quick question since I'm out of date: where is http_handlers referenced now?

navilperez: http_handlers is currently not reference right now. WorkspaceClient can be initiated with either http_handlers or websocket_handlers and we are using websocket_handlers as the default for now.

PR #29, opened 2019-12-05
Return serverId when an entry is added to the database.

navilperez:
After an entry is written to the database return its assigned serverId.
This will support the broadcasting of event once it is written to the
database.

### DIFF
--- a/blockly-rtc/server/db_utils.js
+++ b/blockly-rtc/server/db_utils.js
@@ -23,6 +23,14 @@
 const db = require('./db');
 
 /**
+ * A local representation of an entry in the database.
+ * @typedef {Object} LocalEntry
+ * @property {<!Array.<!Object>>} events An array of Blockly Events in JSON
+ * format.
+ * @property {string} entryId The id assigned to an event by the client.
+ */
+
+/**
  * Query the database for rows since the given server id.
  * @param {number} serverId serverId for the lower bound of the query.
  * @return {!Promise} Promise object represents the rows since the last given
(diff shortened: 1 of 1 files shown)


blockly-rtc/server/db_utils.js
```diff
@@ -45,22 +53,31 @@ function queryDatabase(serverId) {
...
-        console.error(err.message);
-        reject(err);
-      };
+    db.serialize(() => {
+      db.run('INSERT INTO eventsdb(events, entryId) VALUES(?,?)',
+          [JSON.stringify(entry.events), entry.entryId], (err) => {
+        if (err) {
+          console.error(err.message);
+          reject(err);
+        };
```
samelhusseini, line 71: What happens if the second SQL query completes before the first? I think you would resolve before having completed the ``INSERT``

samelhusseini: What is the relationship between the two queries, should they be chained?

navilperez: Yes, the two are executing sequentially since they are wrapped by db.serialize. More info on db.serialize here: https://www.sqlitetutorial.net/sqlite-nodejs/statements-control-flow/. Based on this, I do not believe the second query can resolve before the first is completed. Let me know if this is incorrect.

samelhusseini: Yep, that makes sense. Thanks for clarifying.

samelhusseini (approved):

PR #31, opened 2019-12-06
Refactor db_util to Database.

navilperez:
Database is an object that takes as a parameter a SQLite database object. The functions from db_utils are now methods of Database.


blockly-rtc/server/http_server.js
```diff
@@ -23,14 +23,16 @@
 const http = require('http'); 
 const url = require('url');
 
-const sql = require('./db_utils');
+const Database = require('./Database').Database;
+const db = require('./db');
```
rachel-fenichel: Why is this require out here instead of in the database constructor?

blockly-rtc/server/http_server.js
```diff
@@ -23,14 +23,16 @@
...
 
+const database = new Database(db);
 const PORT = 3001;
 
 http.createServer((req, res) => {
   const parsedUrl = url.parse(req.url, true);
   if (req.method === 'GET' && parsedUrl.pathname === '/api/events/query') {
-    sql.queryDatabase(parsedUrl.query.serverId)
+    database.queryDatabase(parsedUrl.query.serverId)
```
rachel-fenichel: I suggest renaming the function to `query` now that `database` is already in the call.

rachel-fenichel (commented):
I have a few questions, but overall it looks good.

blockly-rtc/server/http_server.js
navilperez: I am having trouble understanding what having db.js in the constructor would look like. Would you mind elaborating?
My thought was that a the server could create a Database object with any sqlitedb (db.js). Is the suggestion for the Database constructor to not take parameters and already have this.db = db?

blockly-rtc/server/http_server.js
navilperez: Good point! Done.

blockly-rtc/server/http_server.js
rachel-fenichel: You would move the `const db = require('./db');` line into the start of the database file, and then do `this.db = db;` in the constructor.  

I ask because it looks like everywhere that you're passing in a db as a parameter, it comes directly from this require.  So it's unclear to me why you have the require in multiple files instead of minimizing its scope.

navilperez: Ah yes, this const db = 'require('./db'); was originally in the database file prior to this change. I will move it back.
I believe I initially took it out so that we could test the Database wrapper functions without actually connecting to a database, however that might have been an unnecessary precaution.

rachel-fenichel (approved):

PR #32, opened 2019-12-06
Store client entryIds to validate database writes.

navilperez:
A second table in the database will be used to track information about
each client. Current use is for keeping track of the last entryId of
each client. The server will use this information to verify if an entry
it recieved is in the correct order before adding it to the database.

### DIFF
--- a/blockly-rtc/server/Database.js
+++ b/blockly-rtc/server/Database.js
@@ -55,29 +55,98 @@ class Database {
   };
 
   /**
-   * Add an entry to the database.
+   * Add entry to the database if the entry is a valid next addition.
+   * For each client, an addition is valid if the entryId is greater than the
+   * entryId of its last added entry.
+   * @param {!LocalEntry} entry The entry to be added to the database.
+   * @return {!Promise} Promise object with the serverId the entry was written
+   * to the database.
+   * @public
+   */
+  async addToServer(entry) {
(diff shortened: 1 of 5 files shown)


blockly-rtc/server/Database.js
```diff
@@ -55,33 +55,102 @@ class Database {
...
+  /**
+   * Run query to add an entry to the database.
    * @param {!LocalEntry} entry The entry to be added to the database.
    * @return {!Promise} Promise object with the serverId for the entry if the
    * write succeeded.
    * @public
    */
-  addToServer(entry) {
-
+  runInsertQuery_(entry) {
```
samelhusseini, line 93: Is this in fact @public? If so it shouldn't have an ``_`` in the name.

blockly-rtc/server/Database.js
```diff
@@ -55,33 +55,102 @@ class Database {
...
   };
+
+  /**
+   * Get the numerical part of the last added entryId for a given client.
+   * @param {!string} workspaceId The workspaceId of the client.
+   * @return {!Promise} Promise object with the the numerical part of the
+   * entryId.
+   * @public
+   */
+  getLastEntryIdNumber_(workspaceId) {
```
samelhusseini, line 121: Similar to above, ``_`` in the name for protected or private methods.

samelhusseini (approved):

PR #33, opened 2019-12-07
Logic for recieving events via broadcasting.

navilperez:
websocket_handlers: new handler for listening to events broadcast by the
server. When it recieves an event it is able to notify the
WorkspaceClient via a callback.

WorkspaceClient and index.js: Logic for trigger interactions with the
server moved from index.js to WorkspaceClient.

### DIFF
--- a/blockly-rtc/src/WorkspaceClient.js
+++ b/blockly-rtc/src/WorkspaceClient.js
@@ -53,272 +53,274 @@ import EventEmitter from 'events';
  * client corresponds to.
  */
 export default class WorkspaceClient {
-    constructor(workspaceId, getEventsHandler, addEventsHandler, broadcastEventsHandler) {
-      this.workspaceId = workspaceId;
-      this.lastSync = 0;
-      this.inProgress = [];
-      this.notSent = [];
-      this.activeChanges = [];
-      this.writeInProgress = false;
-      this.counter = 0;
-      this.serverEvents = [];
-      this.updateInProgress = false;
-      this.getEventsHandler = getEventsHandler;
-      this.addEventsHandler = addEventsHandler;
-      this.broadcastEventsHandler = broadcastEventsHandler;
-      this.listener = new EventEmitter();
-    };  
+  constructor(workspaceId, getEventsHandler, addEventsHandler, broadcastEventsHandler) {
+    this.workspaceId = workspaceId;
+    this.lastSync = 0;
+    this.inProgress = [];
+    this.notSent = [];
+    this.activeChanges = [];
+    this.writeInProgress = false;
+    this.counter = 0;
+    this.serverEvents = [];
+    this.updateInProgress = false;
+    this.getEventsHandler = getEventsHandler;
+    this.addEventsHandler = addEventsHandler;
+    this.broadcastEventsHandler = broadcastEventsHandler;
+    this.listener = new EventEmitter();
+  };  
 
 
-    /**
-     * Initiate the workspace by running all stored events and activating the
-     * handling of recieving events from the server.
-     * @public
-     */
-    async initiateWorkspace() {
-      // TODO: Get an SVG of the current blocks on the workspace instead of
-      // replaying history.
-      const events = await this.getEventsHandler(0);
-      this.addServerEvents_(events);
-      if (this.broadcastEventsHandler) {
-        this.broadcastEventsHandler(this.addServerEvents_.bind(this));
-      } else {
-        this.pollServer_();
-      };
+  /**
+   * Initiate the workspace by running all stored events and activating the
+   * handling of recieving events from the server.
+   * @public
+   */
+  async initiateWorkspace() {
+    // TODO: Get an SVG of the current blocks on the workspace instead of
+    // replaying history.
+    const events = await this.getEventsHandler(0);
+    this.addServerEvents_(events);
+    if (this.broadcastEventsHandler) {
+      this.broadcastEventsHandler(this.addServerEvents_.bind(this));
+    } else {
+      this.pollServer_();
     };
+  };
 
-    /**
-     * Add an event to activeChanges.
-     * @param {!Object} event The Blockly.Event JSON created by the client.
-     * @public
-     */
-    addEvent(event) {
-      this.activeChanges.push(event);
+  /**
+   * Add an event to activeChanges.
+   * @param {!Object} event The Blockly.Event JSON created by the client.
+   * @public
+   */
+  addEvent(event) {
+    this.activeChanges.push(event);
+  };
+
+  /**
+   * Add the events in activeChanges to notSent. Initiates process for sending
+   * local changes to the database.
+   * @public
+   */
+  flushEvents() {
+    this.notSent = this.notSent.concat(this.activeChanges);
+    this.activeChanges = [];
+    this.updateServer_();
+  };
+
+  /**
+   * Send local changes to the server. Continuously runs until all local changes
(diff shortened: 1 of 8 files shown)


blockly-rtc/src/websocket_handlers.js
```diff
@@ -33,10 +48,29 @@ export async function getEvents(serverId) {
...
+
+/**
+ * Listen for events broadcast by the server.
+ * @param {!Function} callback The callback handler that passes the events to
+ * the workspace.
+ * @return {!Promise} Promise object that represents the success of the write.
+ * @public
+ */
+export function getBroadcast(callback) {
+  socket.on('broadcastEvents', (rows)=> {
```
rachel-fenichel: Naming question: everywhere else you're calling it your variable `entry`.  Here you're calling the variable `rows`.  Why?

blockly-rtc/src/index.js
```diff
@@ -33,65 +33,29 @@ import WorkspaceClient from './WorkspaceClient';
...
-            console.error('Failed to write to database.');
-        });
-    };
-
-    /**
-     * Periodically signal the WorkspaceClient to query the database and call
-     * runEvents_() with the result of the query.
-     * @private
-     */
-    function pollServer_() {
```
rachel-fenichel: Your doc says that you will retain polling for use in cases where broadcasting isn't working.  You may need to keep this function around for that.

blockly-rtc/src/http_handlers.js
```diff
@@ -20,10 +20,18 @@
  * @author navil@google.com (Navil Perez)
  */
 
+/**
```
rachel-fenichel, line 23: Is there a single place that you can put all of these typedefs, so that you don't need to have them repeated in multiple files?

blockly-rtc/src/http_handlers.js
```diff
@@ -20,10 +20,18 @@
...
+ * @property {<!Array.<!Object>>} events An array of Blockly Events in JSON
+ * format.
+ * @property {string} entryId The id assigned to an event by the client.
+ */
+
 /**
  * Query the database for rows since the given server id.
  * @param {number} serverId serverId for the lower bound of the query.
- * @return {!Array.<!Object>} Rows of events since the given serverId.
+ * @return {<!Array.<!Object>>} Rows of entries since the given serverId.
```
rachel-fenichel: Can you make `!Object` more specific here?  Is it `!LocalEntry`?

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -51,51 +53,92 @@
  * client corresponds to.
  */
 export default class WorkspaceClient {
-    constructor(workspaceId, getEventsHandler, addEventsHandler) {
```
rachel-fenichel: Nit: Why is so much of this a diff?  It looks like maybe you have inconsistent sizing on your tabs, so that some are 2 spaces and some are 4.  I think this diff will be much smaller if you fix that.

rachel-fenichel (commented):
I started by looking at files other than WorkspaceClient and the tests, and will loop back to those later today.

blockly-rtc/src/WorkspaceClient.js
navilperez: Yes, there is inconsistent indentation in this file. I fixed those inconsistencies in this PR, but it seems to create a lot of noise and hide the relevant changes. My plan is to go back to inconsistent formatting for this PR and upload another PR with format changes that I will merge here. The diff should be smaller now. Thanks for your patience.

blockly-rtc/src/index.js
navilperez: Yes, we do still keep polling. This logic has been migrated over to WorkspaceClient which should be easier to see now that the diff is more representative of the change.

blockly-rtc/src/websocket_handlers.js
navilperez: Good catch. It used to be `rows` when each Blockly Event was written to the database in a separate row. Now, Blockly Events are grouped and added to the database as one row, or `entry`.

I go through looking for any places where it is still called `rows` and change it to `entry` in a following commit to this PR.

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -126,26 +169,75 @@ export default class WorkspaceClient {
...
 
+  /**
+   * Periodically query the database for new server events and add them to
+   * serverEvents.
+   * @private
+   */
+  async pollServer_() {
+    const rows = await this.queryDatabase_();
+    await this.addServerEvents_(rows);
+    this.pollServer_();
```
rachel-fenichel: This will run more or less continuously.  What's the plan for knocking it down to a lower frequency?

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -126,26 +169,75 @@ export default class WorkspaceClient {
...
     } catch {
       return [];
     };
   };
 
+  /**
+   * Add newServerEvents to the end of this.serverEvents and initiate process of
+   * applying server events to the workspace if the newServerEvents are recieved
+   * in the correct order.
+   * @param {<!Array.<!row>>} newServerEvents The events recieved from the server.
```
rachel-fenichel: Capital R Row?

blockly-rtc/src/WorkspaceClient.js
```diff
@@ -126,26 +169,75 @@ export default class WorkspaceClient {
...
+   * @private
+   */
+  async addServerEvents_(newServerEvents) {
+    if (newServerEvents.length == 0) {
+      return;
+    }
+    if (newServerEvents[0].serverId != this.lastSync + 1) {
+      newServerEvents = await this.queryDatabase_();
+    };
+    this.lastSync = newServerEvents[newServerEvents.length-1].serverId;
```
rachel-fenichel: Nit: spaces around `-`

blockly-rtc/src/http_handlers.js
navilperez: They are a bit different. LocalEntry is how they are stored locally before being written to the database where they then assigned a serverId.
LocalEntry
```
{
  events:
  entryId:
}
```

This object (tentatively ServerEntry)
```
{
  events:
  entryId:
  serverId:
}
```
 I can make another typedef for this object and make it more specific.

blockly-rtc/src/http_handlers.js
navilperez: Not having them repeated would be ideal. Based on a quick Google search it seems that putting them all in a single `config.js` file can be done.

Is there a convention for this that you recommend?

blockly-rtc/src/http_handlers.js
navilperez: Actually, there already is a typedef for this called "Row" I will change the JSDoc accordingly.

blockly-rtc/src/http_handlers.js
rachel-fenichel: That sounds good.  I would call it `typedefs.js` instead of `config.js`.  You can add that to your task list and do it in a separate PR.

blockly-rtc/src/WorkspaceClient.js
navilperez: For the time being I will add a timeout to poll every couple of seconds.
The long term solution is to move from continuous polling to long polling.

rachel-fenichel (approved):

PR #39, opened 2019-12-18
MarkerUpdate data model.

navilperez:
Creates a MarkerUpdate class for passing marker information between the
clients and the server.
MarkerUpdate has all the information needed to create a marker and is
serializable to faciliate storage in the database.

Also in this change, Blockly is imported from the develop branch on github instead of npm as this change is dependent on unpublished changes to Markers.

PR #40, opened 2019-12-18
Add MarkerUpdates to database

navilperez:
Functions for adding and retrieving MarkerUpdates (#39) from the database.


navilperez:
Relevant parts of the design doc are under
- Marker Manager > MarkerUpdate Data Model
- Database > Clients Table.

blockly-rtc/server/Database.js
```diff
@@ -174,6 +174,80 @@ class Database {
       });
     });
   };
+
+  /**
+   * Add client to clients if it is not already in the table.
```
rachel-fenichel, line 179: This isn't called in this PR.

navilperez: This function is not yet called. It will be called by MarkerManager when a client first connects to the server.
The purpose is to ensure that a client is in clients table otherwise the SQL queries updating the client's markerLocation will throw errors if it has not been added to the table.

rachel-fenichel (approved):

PR #42, opened 2019-12-19
Middleware for MarkerUpdates.

navilperez:
APIs and server functions for sending MarkerUpdates between the client
and server.

### DIFF
--- a/blockly-rtc/server/http_server.js
+++ b/blockly-rtc/server/http_server.js
@@ -34,11 +34,9 @@ http.createServer(async (req, res) => {
     await queryEventsHandler_(res, parsedUrl.query.serverId);
   } else if (req.method === 'POST' && parsedUrl.pathname === '/api/events/add') {
     await addEventsHandler_(req, res);
-  } else if (req.method === 'GET' && parsedUrl.pathname === '/api/markers/query') {
+  } else if (req.method === 'GET' && parsedUrl.pathname === '/api/clients/query') {
     await getMarkerUpdatesHandler_(res, parsedUrl.query.workspaceId);
-  } else if (req.method === 'POST' && parsedUrl.pathname === '/api/markers/add') {
-    await addClientHandler_(res, parsedUrl.query.workspaceId);
-  } else if (req.method === 'PUT' && parsedUrl.pathname === '/api/markers/update') {
+  } else if (req.method === 'PUT' && parsedUrl.pathname === '/api/clients/update') {
     await updateMarkerHandler_(req, res);
   } else {
     res.statusCode = 404;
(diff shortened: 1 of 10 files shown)


navilperez:
Relevant parts of the design doc are under:

Endpoint API Server/Middleware > Marker Manager
Websocket Server/Middleware > MarkerManager

blockly-rtc/server/http_server.js
```diff
@@ -34,6 +34,12 @@ http.createServer(async (req, res) => {
     await queryEventsHandler_(res, parsedUrl.query.serverId);
   } else if (req.method === 'POST' && parsedUrl.pathname === '/api/events/add') {
     await addEventsHandler_(req, res);
+  } else if (req.method === 'GET' && parsedUrl.pathname === '/api/markers/query') {
```
RoboErikG: I'd recommend calling this clients. What you're actually requesting is information about the active clients, which includes their location. Markers are the UI element that represents the client. So /api/clients/query

blockly-rtc/server/http_server.js
```diff
@@ -34,6 +34,12 @@ http.createServer(async (req, res) => {
...
+    await getMarkerUpdatesHandler_(res, parsedUrl.query.workspaceId);
+  } else if (req.method === 'POST' && parsedUrl.pathname === '/api/markers/add') {
+    await addClientHandler_(res, parsedUrl.query.workspaceId);
```
RoboErikG: You can combine the add and update actions by using an [UPSERT sql command](https://www.sqlite.org/lang_UPSERT.html) and only sending updates. This will also remove the need for clients to keep track of their registered state on the server or having repeated adds for the same client.

blockly-rtc/server/websocket_server.js
```diff
@@ -59,4 +59,20 @@ async function onConnect(user) {
     const entries = await database.query(serverId);
     callback(entries);
   });
+
```
RoboErikG: You should probably document these functions.

blockly-rtc/server/http_server.js
navilperez: This is a valid point. In reality what we are passing is information about the location of a user. For this change, I updated the API paths.

In a following change I will rename MarkerUpdates to something more along the lines of ClientLocation and also update the names of the handlers in the APIs and server.

I will also switch the language I use from client to user in a subsequent change.

blockly-rtc/server/websocket_server.js
navilperez: Yes, this would be helpful to the user. I have now created handler functions and added JSDocs.

blockly-rtc/server/http_server.js
navilperez: This is amazing! Thanks you for pointing me towards this.

RoboErikG (approved):

PR #43, opened 2019-12-19
MarkerManager class.

navilperez:
Class for displaying markers for other users. Sends and recieves
MarkerUpdates.

### DIFF
--- a/blockly-rtc/src/MarkerManager.js
+++ b/blockly-rtc/src/MarkerManager.js
@@ -25,13 +25,12 @@ import MarkerUpdate from './MarkerUpdate';
 
 export default class MarkerManager {
   constructor(workspaceId, sendMarkerUpdate, getMarkerUpdates,
-      addClient, getBroadcastMarkerUpdates) {
+      getBroadcastMarkerUpdates) {
     this.workspaceId = workspaceId
     this.colours =  [
         '#fcba03', '#03fc20', '#03f0fc', '#035efc', '#5603fc', '#fc03d2'];
     this.sendMarkerUpdate = sendMarkerUpdate;
     this.getMarkerUpdates = getMarkerUpdates;
-    this.addClient = addClient;
     this.broadcastMarkerHandler = getBroadcastMarkerUpdates;
   };
 
@@ -41,17 +40,16 @@ export default class MarkerManager {
    * @public
    */
   async initMarkers() {
-    await this.addClient(this.workspaceId);
     const markerUpdates = await this.getMarkerUpdates();
     markerUpdates.forEach((markerUpdate) => {
       this.createMarker_(markerUpdate);
-      if (this.broadcastMarkerHandler) {
-        this.broadcastMarkerHandler(this.updateMarkerLocations_.bind(this));
-      } else {
-        this.pollServer_();
-      };
     });
-  };
+    if (this.broadcastMarkerHandler) {
+      this.broadcastMarkerHandler(this.updateMarkerLocations_.bind(this));
+    } else {
+      this.pollServer_();
+    };
+};
 
   /**
    * Create a MarkerUpdate from a Blockly event and send it to the server.
@@ -91,15 +89,16 @@ export default class MarkerManager {
    * @private
    */  
   getMarkerManager_() {
-    return this.getWorkspace_().getMarkerManager();
+    return this.getWorkspace_() ?
+        this.getWorkspace_().getMarkerManager(): null;
   };
 
   /**
-   * Get a unique color to assign to a new client marker.
+   * Get a color to assign to a new client marker.
    * @returns {string} The string encoding a colour.
    * @private
    */  
-  setColour_() {
+  getColour_() {
     return this.colours.pop();
   };
 
@@ -111,8 +110,11 @@ export default class MarkerManager {
    * @private
    */
   createMarker_(markerUpdate) {
+    if (!this.getMarkerManager_()) {
+      return;
+    };
     const marker = markerUpdate.toMarker(this.getWorkspace_());
-    marker.colour = this.setColour_();
+    marker.colour = this.getColour_();
     this.getMarkerManager_().registerMarker(markerUpdate.id, marker)
     marker.setCurNode(markerUpdate.createNode(this.getWorkspace_()));
     return marker;
@@ -125,7 +127,8 @@ export default class MarkerManager {
    * @private
    */  
   getMarker(workspaceId) {
-    return this.getMarkerManager_().getMarker(workspaceId);
+    return this.getMarkerManager_() ?
+        this.getMarkerManager_().getMarker(workspaceId) : null;
   };
 
   /**
@@ -139,10 +142,10 @@ export default class MarkerManager {
         markerUpdate => markerUpdate.id != this.workspaceId);
     filteredMarkerUpdates.forEach((markerUpdate) => {
       const node = markerUpdate.createNode(this.getWorkspace_());
-      if (this.getMarker(markerUpdate.id) == undefined) {
-        this.createMarker_(markerUpdate).setCurNode(node);
-      } else {
+      if (this.getMarker(markerUpdate.id)) {
         this.getMarker(markerUpdate.id).setCurNode(node);
+      } else {
+        this.createMarker_(markerUpdate).setCurNode(node);        
       };
     });
   };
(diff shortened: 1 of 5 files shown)


navilperez:
Relevant parts of the design doc under Marker Manager.

blockly-rtc/src/MarkerManager.js
```diff
@@ -0,0 +1,149 @@
...
+  };
+
+  /**
+   * Get the MarkerManager for the workspace specified by MarkerUpdate.
+   * @return {Blockly.MarkerManager} The sought after MarkerManager or null if not
+   * found.
+   * @private
+   */  
+  getMarkerManager_() {
+    return this.getWorkspace_().getMarkerManager();
```
rachel-fenichel: This will break if `getWorkspace_()` returned `null`.

blockly-rtc/src/MarkerManager.js
```diff
@@ -0,0 +1,149 @@
...
+  };
+
+  /**
+   * Get a unique color to assign to a new client marker.
+   * @returns {string} The string encoding a colour.
+   * @private
+   */  
+  setColour_() {
+    return this.colours.pop();
```
rachel-fenichel: What happens when you run out of colours?  

Also, this function should be `getColour_` or `getUniqueColour_` since it's not setting anything.

blockly-rtc/src/MarkerManager.js
```diff
@@ -0,0 +1,149 @@
...
+   * Create a Marker with a unique color and register it.
+   * @param {!MarkerUpdate} markerUpdate The MarkerUpdate for which to create
+   * a Marker.
+   * @returns {!Marker} marker The Marker represented by the MarkerUpdate.
+   * @private
+   */
+  createMarker_(markerUpdate) {
+    const marker = markerUpdate.toMarker(this.getWorkspace_());
+    marker.colour = this.setColour_();
+    this.getMarkerManager_().registerMarker(markerUpdate.id, marker)
```
rachel-fenichel, line 122: `getMarkerManager_` can return null.

blockly-rtc/src/index.js
```diff
@@ -33,15 +35,20 @@ document.addEventListener('DOMContentLoaded', () => {
...
   const workspaceClient = new WorkspaceClient(
       workspace.id, getEvents, writeEvents);
-
   workspaceClient.listener.on('runEvents', (eventQueue) => {
     runEvents_(eventQueue);
   });
-
   workspaceClient.initiateWorkspace();
 
+  const markerManager = new MarkerManager(workspace.id, sendMarkerUpdate,
```
rachel-fenichel, line 43: Why do you import these functions here and then pass them through to the marker manager, instead of just importing them there?

navilperez: I want to be able to make MarkerManager an interface that supports markers for realtime collaboration regardless of how the handlers are implemented. Index.js will be the glue where developers can create an instance of MarkerManager by passing in their own handlers.

This is also how WorkspaceClient is implemented.

blockly-rtc/src/MarkerManager.js
navilperez: Added a check.

blockly-rtc/src/MarkerManager.js
navilperez: Renamed.

As written, the Marker Colour would be set to undefined once we run out of colours any new marker created will appear with the default colour set in marker_svg. Therefore, makers will still render, they just will not appear with a unique colour.

An alternative would be to cycle through the colours. This will also mean that the colour would not be unique, but would create a more even distribution among users who have a certain colour marker.

blockly-rtc/src/MarkerManager.js
navilperez: Done.

blockly-rtc/src/MarkerManager.js
```diff
@@ -0,0 +1,152 @@
...
+    if (this.broadcastMarkerHandler) {
+      this.broadcastMarkerHandler(this.updateMarkerLocations_.bind(this));
+    } else {
+      this.pollServer_();
+    };
+};
+
+  /**
+   * Create a MarkerUpdate from a Blockly event and send it to the server.
+   * @param {!Blockly.Event} The event for which to create a MarkerUpdate.
```
samelhusseini: add the name of the param here: 
```
 @param {!Blockly.Event} event The event for which to create a MarkerUpdate.
```

blockly-rtc/src/MarkerManager.js
samelhusseini: I'm of the preference of cycling through colours if we are using a limited colour set.

blockly-rtc/src/MarkerManager.js
```diff
@@ -0,0 +1,152 @@
...
+    marker.colour = this.getColour_();
+    this.getMarkerManager_().registerMarker(markerUpdate.id, marker)
+    marker.setCurNode(markerUpdate.createNode(this.getWorkspace_()));
+    return marker;
+  };
+
+  /**
+   * Get the Marker that corresponds to the given client.
+   * @param {!string} workspaceId The workspaceId of the client.
+   * @returns {Marker} The Marker for the client if it exists, otherwise null.
```
samelhusseini: Is this meant to be Blockly.Marker?

blockly-rtc/src/MarkerManager.js
navilperez: Done.

blockly-rtc/src/MarkerManager.js
navilperez: Yes. Changed.

blockly-rtc/src/MarkerManager.js
```diff
@@ -0,0 +1,154 @@
...
+ */
+
+import * as Blockly from 'blockly/dist';
+import MarkerUpdate from './MarkerUpdate';
+
+export default class MarkerManager {
+  constructor(workspaceId, sendMarkerUpdate, getMarkerUpdates,
+      getBroadcastMarkerUpdates) {
+    this.workspaceId = workspaceId
+    this.colours =  [
```
samelhusseini: nit: remove extra space after ``this.colours``.

blockly-rtc/src/MarkerManager.js
```diff
@@ -0,0 +1,154 @@
...
+ * @author navil@google.com (Navil Perez)
+ */
+
+import * as Blockly from 'blockly/dist';
+import MarkerUpdate from './MarkerUpdate';
+
+export default class MarkerManager {
+  constructor(workspaceId, sendMarkerUpdate, getMarkerUpdates,
+      getBroadcastMarkerUpdates) {
+    this.workspaceId = workspaceId
```
samelhusseini: nit: add semicolon after workspaceId.

blockly-rtc/src/MarkerManager.js
samelhusseini: I think this issue still exists.

blockly-rtc/src/MarkerManager.js
```diff
@@ -0,0 +1,154 @@
...
+    const marker = markerUpdate.toMarker(this.getWorkspace_());
+    marker.colour = this.getColour_();
+    this.getMarkerManager_().registerMarker(markerUpdate.id, marker)
+    marker.setCurNode(markerUpdate.createNode(this.getWorkspace_()));
+    return marker;
+  };
+
+  /**
+   * Get the Marker that corresponds to the given client.
+   * @param {!string} workspaceId The workspaceId of the client.
```
samelhusseini: type ``{string}`` is ``{!string}`` by default. (ie: not null). The ``!`` is redundant. You only need to add the ``!`` for objects.

samelhusseini (commented):
Overall looks good. Few more unresolved comments.

blockly-rtc/src/MarkerManager.js
navilperez: Done.

blockly-rtc/src/MarkerManager.js
navilperez: Done.

blockly-rtc/src/MarkerManager.js
navilperez: Done.

blockly-rtc/src/MarkerManager.js
navilperez: I now changed it so that it throws an error if MarkerManager returns null and added a test for this.

samelhusseini (approved):

PR #48, opened 2020-01-15
Refactoring MarkerManager to UserDataManager.

navilperez:
MarkerManager is renamed UserDataManager in order to allow the module to be expanded to track a wider range of user metadata in addition to locations and markers.

MarkerUpdate is renamed  Location/LocationUpdate in order to more accurately convey the information being transmitted and stored.

The clients table is renamed users. Currently the word client and user are both used in the code base to refer to someone connected to a Blockly Workspace; this change makes it so one term is used.


navilperez:
Due to the amount of code changed with the renaming, the "Files Changed" view shows a few files as new files rather than renames. Please refer to the individual commits to more accurately view the diffs.

blockly-rtc/src/Location.js
```diff
@@ -16,96 +16,64 @@
  */
 
 /**
- * @fileoverview Object for communicating an update to a Marker's location.
+ * @fileoverview Object representing a user's location.
```
samelhusseini: location is a little ambiguous as it could mean perhaps a user's physical location. Can we make that clear here? Perhaps, ``Object representing a user's cursor location``?

navilperez: Discussed offline: Location > Position and a more descriptive file overview.

samelhusseini (approved):

PR #52, opened 2020-01-24
Separate modules for server and middleware.

navilperez:
Create separate modules for the handlers that interact with the
database. events_handlers interact with the events table to enable the
core realtime collaboration function. users_handlers enable markers on
the workspace.

### DIFF
--- a/blockly-rtc/server/http/events_handlers.js
+++ b/blockly-rtc/server/http/events_handlers.js
@@ -1,0 +1,72 @@
+/**
+ * @license
+ * 
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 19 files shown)


blockly-rtc/server/http/events_handlers.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+async function queryEventsHandler_(res, serverId) {
+  try {
+    const entries = await database.query(serverId);
+    res.setHeader('Content-Type', 'application/json');
+    res.statusCode = 200;
+    res.write(JSON.stringify({ entries }));
+    res.end();
+  } catch {
+    console.log('why am i here');
```
samelhusseini: Is this intended?

blockly-rtc/server/http/events_handlers.js
```diff
@@ -0,0 +1,72 @@
...
+      res.statusCode = 200;
+      res.end();  
+    });
+  } catch {
+    res.statusCode = 401;
+    res.end();
+  };
+};
+
+module.exports.queryEventsHandler_ = queryEventsHandler_;
```
samelhusseini: Perhaps if it's being exported publicly it shouldn't have an ``_`` in the name? or at least the exported name shouldn't

blockly-rtc/server/http/events_handlers.js
navilperez: Yikes. Leftover from debugging. Thanks for the catch.

blockly-rtc/server/http/events_handlers.js
navilperez: Good point.

samelhusseini (approved):

PR #54, opened 2020-01-30
Create a snapshot of the workspace for users to load upon connecting.

navilperez:
Create and store a snapshot of the current state of the workspace
whenever a new user connects. A snapshot object contains the XML of the
workspace and the serverId of the last entry represented on the
workspace.

### DIFF
--- a/blockly-rtc/src/WorkspaceClient.js
+++ b/blockly-rtc/src/WorkspaceClient.js
@@ -30,7 +30,7 @@ import EventEmitter from 'events';
  * client corresponds to.
  */
 export default class WorkspaceClient {
-  constructor(workspaceId, getEventsHandler, addEventsHandler, broadcastEventsHandler) {
+  constructor(workspaceId, getSnapshotHandler, getEventsHandler, addEventsHandler, broadcastEventsHandler) {
     this.workspaceId = workspaceId;
     this.lastSync = 0;
     this.inProgress = [];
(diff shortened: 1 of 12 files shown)


blockly-rtc/src/WorkspaceClient.js
```diff
@@ -48,15 +49,22 @@ export default class WorkspaceClient {
...
    * @public
    */
-  async initiateWorkspace() {
-    // TODO: Get an SVG of the current blocks on the workspace instead of
-    // replaying history.
-    const events = await this.getEventsHandler(0);
+  async start() {
+    // Load the current state of the workspace.
+    const workspace = Blockly.Workspace.getById(this.workspaceId);
+    const snapshot = await this.getSnapshotHandler();
```
rachel-fenichel, line 59: Will snapshot ever be null?

rachel-fenichel (commented):
One Q, mostly lgtm

blockly-rtc/src/WorkspaceClient.js
navilperez: It should not be, the server has a valid default value if it has not been set. I can only think of the case where there are errors retrieving it from the server in which case this.getSnapshotHandler() will reject. Is this a case I should account for?

rachel-fenichel: Nope, that sounds fine.

rachel-fenichel (approved):

PR #55, opened 2020-01-30
Detect user disconnects and dispose markers for websocket sample.

navilperez:
Websocket server will attach the workspaceId to each user in order track
users and send a broadcast to users when someone disconnects. Upon
disconnecting, the user is removed from the database and the event is
broadcast to the other users.

UserDataManager disposes of markers that have disconnected.

### DIFF
--- a/blockly-rtc/server/websocket/users_handlers.js
+++ b/blockly-rtc/server/websocket/users_handlers.js
@@ -84,7 +84,6 @@ async function connectUserHandler(user, workspaceId, callback) {
  * @public
  */
 async function disconnectUserHandler(workspaceId, callback) {
-  console.log('User disconnected.');
   await database.deleteUser(workspaceId);
   callback();
 };
--- a/blockly-rtc/server/websocket/users_handlers.js
+++ b/blockly-rtc/server/websocket/users_handlers.js
@@ -53,6 +53,43 @@ async function getPositionUpdatesHandler(workspaceId, callback) {
   const positionUpdates = await database.getPositionUpdates(workspaceId);
   callback(positionUpdates);
 };
-  
+
+/**
+ * Handler for a connectUser message. Attach the workspaceId to the user and
+ * add the user to the users table.
+ * @param {!Object} user The user connecting.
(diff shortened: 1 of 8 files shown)


blockly-rtc/server/websocket/users_handlers.js
```diff
@@ -53,6 +53,43 @@ async function getPositionUpdatesHandler(workspaceId, callback) {
...
+
+/**
+ * Handler for a disconnect. Delete the user from the users table.
+ * @param {string} workspaceId The workspaceId for the disconnecting user.
+ * @param {!Function} callback The callback that broadcasts the disconnect to
+ * the connected users.
+ * @public
+ */
+async function disconnectUserHandler(workspaceId, callback) {
+  console.log('User disconnected.');
```
rachel-fenichel: Remove log.

blockly-rtc/src/UserDataManager.js
```diff
@@ -49,7 +50,24 @@ export default class UserDataManager {
     } else {
       this.pollServer_();
     };
-};
+    if (this.getUserDisconnects) {
+      this.getUserDisconnects(this.disposeMarker_.bind(this));
```
rachel-fenichel, line 54: Won't this always be null here?

rachel-fenichel (commented):
One question, otherwise lgtm.

blockly-rtc/src/UserDataManager.js
navilperez: The current flow is that `enablePresence()` assigns `this.getUserDisconnects`. And then `start()` is called which calls the function.

Another approach is to assign and call the function `this.getUserDisconnects` in `enablePresence` and remove L53-54 from `start`.

I am open and happy with either option.

navilperez: Correction, when I say `enablePresence` above, I am actually referring to `setPresenceHandlers`

rachel-fenichel:
Approve once you remove the leftover `console.log`.

PR #56, opened 2020-01-30
Create positions for fields.

navilperez:
Allow Blockly Change events to indicate a user position on a field.

### DIFF
--- a/blockly-rtc/src/Position.js
+++ b/blockly-rtc/src/Position.js
@@ -37,10 +37,12 @@ export default class Position {
    * @public
    */  
   static fromEvent(event) {
-    if (event instanceof Blockly.Events.Ui) {
-      return this.fromUiEvent_(event);
-    } else if (event instanceof Blockly.Events.Change) {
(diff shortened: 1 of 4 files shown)


blockly-rtc/src/Position.js
```diff
@@ -37,14 +37,41 @@ export default class Position {
...
     return new Position(type, blockId, fieldName);  
   };
 
+  /**
+   * Create a Position from a Blockly Change event.
+   * @param {!Blockly.Event.Change} event The event that creates a Position.
+   * @return {!PositionUpdate} The Position representative of the event.
+   * @private
+   */  
+  static fromChangeEvent_(event) {
```
rachel-fenichel: What about non-field change events?

navilperez: Did more research into change events, now I made sure this function is only called on field change events and index.js is now only passing in field change events into this function.

rachel-fenichel (approved):

PR #66, opened 2020-02-14
Implementing search functionality.

moniika:
#### Features added
- basic search (with logic for collapsed blocks)
- adding/removing highlighting for matching blocks through adding class (class currently fills block black)
- next/previous functionality (with wrapping)
- clear search functionality
- keydown/oninput event binding for search bar

#### Demo
e265e3db-b529-400d-86ea-d985a896e309


blockly-ws-search/index.html
```diff
@@ -58,7 +58,7 @@
   var ws = Blockly.inject('blocklyDiv' + suffix, options);
   var workspaceSearch = new WorkspaceSearch(ws);
   workspaceSearch.init();
-  workspaceSearch.setVisible(true);
+  workspaceSearch.open(true);
```
alschmiedt: I think we can remove the true here.

blockly-ws-search/style.css
```diff
@@ -35,3 +35,7 @@
 .workspaceSearchBar button {
   border: none;
 }
+
+path.blocklyPath.searchHighlight {
+  fill: black;
+}
```
alschmiedt: Add a new line

blockly-ws-search/workspace_search.js
```diff
@@ -114,44 +128,90 @@ class WorkspaceSearch {
...
+    }
+    this.selectBlock_(this.blockIndex_ + 1);
+    console.log("Get next value");
+  }
+
+  /**
+   * Selects the block at the given index.
+   * @param {number} index Index of block to select. Number is wrapped.
+   */
+  selectBlock_(index) {
```
alschmiedt, line 206: If this is meant to be public remove the _ after selectBlock

alschmiedt, line 206: Not necessary for this PR, but I think we should either pass in the list of blocks here, or we should make this.blocks_ public. That way people can easily override if they want to change behavior.

alschmiedt (approved):
This looks really good! Just a few small comments then lgtm

blockly-ws-search/workspace_search.js
moniika: Good point, we should probably revisit a lot of the methods. I forgot subclasses aren't supposed to access private variables.

blockly-ws-search/style.css
moniika: Is that a style thing?

blockly-ws-search/index.html
moniika: Yes, will remove

blockly-ws-search/style.css
alschmiedt: Yeah, in Blockly we try to have a new line at the end of each file. I think it would probably be good to continue over here.

blockly-ws-search/workspace_search.js
moniika: Added protected.

PR #69, opened 2020-02-15
Adds shortcut to open the search bar

alschmiedt:
Adds a listener on the injection div to listen for control/command F and open the search bar. 
Focuses on the textInput when search bar is opened.
Focuses on the workspace when the search bar is closed.
**This might be changed if/when we add a way to easily add shortcuts to Blockly.

### DIFF
--- a/blockly-ws-search/index.html
+++ b/blockly-ws-search/index.html
@@ -58,6 +58,7 @@ function startBlocklyInstance(suffix, rtl, horizontalLayout, position,
   var ws = Blockly.inject('blocklyDiv' + suffix, options);
   var workspaceSearch = new WorkspaceSearch(ws);
   workspaceSearch.init();
+  workspaceSearch.open();
 }
 </script>
 
--- a/blockly-ws-search/index.html
+++ b/blockly-ws-search/index.html
@@ -58,7 +58,6 @@ function startBlocklyInstance(suffix, rtl, horizontalLayout, position,
   var ws = Blockly.inject('blocklyDiv' + suffix, options);
   var workspaceSearch = new WorkspaceSearch(ws);
   workspaceSearch.init();
-  workspaceSearch.open();
 }
 </script>
 
--- a/blockly-ws-search/workspace_search.js
+++ b/blockly-ws-search/workspace_search.js
@@ -67,14 +67,14 @@ class WorkspaceSearch {
 
     /**
      * HTML container for the workspace search bar.
-     * @type {?Element}
+     * @type {?HTMLElement}
      * @private
      */
     this.HtmlDiv_ = null;
 
     /**
      * The input for the search text.
-     * @type {?HTMLElement}
+     * @type {?HTMLTextInput}
      * @private
      */
     this.textInput_ = null;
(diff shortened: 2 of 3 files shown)


blockly-ws-search/workspace_search.js
```diff
@@ -64,6 +64,20 @@ class WorkspaceSearch {
      * @type {boolean}
      */
     this.searchOnInput = true;
+
+    /**
+     * HTML container for the workspace search bar.
+     * @type {?Element}
```
moniika: since html is also in the name of this, I think type should change to HTMLElement

blockly-ws-search/index.html
```diff
@@ -58,7 +58,6 @@
   var ws = Blockly.inject('blocklyDiv' + suffix, options);
   var workspaceSearch = new WorkspaceSearch(ws);
   workspaceSearch.init();
-  workspaceSearch.open();
```
moniika: I think it would be nice to open on page load, just for convenience of our own testing.

blockly-ws-search/workspace_search.js
```diff
@@ -242,13 +276,15 @@ class WorkspaceSearch {
...
+    this.workspace_.markFocused();
     this.clearBlocks();
     console.log("Close search bar");
   }
 
   /**
    * Shows or hides the workspace search bar.
-   * @param {boolean} show True to set the search bar as visible. False otherwise. 
+   * @param {boolean} show True to set the search bar as visible. False
+   *    otherwise.
```
moniika: Optional suggestion: 
`{boolean} show Whether to set the search bar as visible.`

blockly-ws-search/workspace_search.js
```diff
@@ -165,6 +183,21 @@ class WorkspaceSearch {
...
+    /**
+   * Add listener on the workspace to open the search bar when Control F or
+   * Command F are used.
+   * TODO: We might want Blockly to be able to deal with setting shortcuts on
+   * workspaces.
+   * @param {KeyboardEvent} e The key down event.
+   * @private
+   */
+  onWorkspaceKeyDown_ = function(e) {
+    if ((e.ctrlKey || e.metaKey) && e.keyCode == Blockly.utils.KeyCodes.F) {
```
moniika: The keyCode property is deprecated (on KeyboardEvent) so I think its recommended to use the key or code property instead (I noticed because my IDE gave warning for using keyCode). I'm not certain what the right approach is, it probably depends on what browsers we want to support. Currently I'm using key in onKeyDown_.

blockly-ws-search/workspace_search.js
```diff
@@ -64,6 +64,20 @@ class WorkspaceSearch {
...
+    /**
+     * HTML container for the workspace search bar.
+     * @type {?Element}
+     * @private
+     */
+    this.HtmlDiv_ = null;
+
+    /**
+     * The input for the search text.
+     * @type {?HTMLElement}
```
moniika: Could update to HTMLTextInput type.

moniika (approved):

PR #70, opened 2020-02-19
Adding scrolling to current highlighted block.

moniika:


blockly-ws-search/workspace_search.js
```diff
@@ -395,4 +401,75 @@ class WorkspaceSearch {
...
+    // Workspace scale, used to convert from workspace coordinates to pixels.
+    const scale = this.workspace_.scale;
+
+    // Block bounds in pixels relative to the workspace origin (0,0 is centre).
+    const top = blockTop * scale;
+    const bottom = blockBottom * scale;
+    const left = blockLeft * scale;
+    const right = blockRight * scale;
+    const width = blockWidth * scale;
+    const height = blockHeight * scale;
```
alschmiedt: I would maybe consider combining these and then you could just have one set of constants and have bock in the name.

blockly-ws-search/workspace_search.js
```diff
@@ -395,4 +401,75 @@ class WorkspaceSearch {
...
+
+    let targetLeft = metrics.viewLeft;
+
+    const overflowLeft = left < metrics.viewLeft;
+    const overflowRight = right > metrics.viewLeft + metrics.viewWidth;
+    const hasXOverflow = overflowLeft || overflowRight;
+    const wideBlock = width > metrics.viewWidth;
+
+    if ((overflowLeft && !wideBlock) ||
+        (wideBlock && !this.workspace_.RTL && hasXOverflow)) {
```
alschmiedt: do we need the check for hasXOverflow here? It seems like if wide > metrics.viewWidth is true then there will always be an overflow?

alschmiedt (approved):
Looks great! Two small things then lgtm : )

blockly-ws-search/workspace_search.js
moniika: Cleaned it up. It looks nicer now.

blockly-ws-search/workspace_search.js
moniika: Yeah, not sure why I thought it was necessary. Removing it as I can't think of any case where it wouldn't be true.

blockly-ws-search/workspace_search.js
```diff
@@ -395,4 +401,62 @@ class WorkspaceSearch {
...
+
+    // Block bounds in pixels relative to the workspace origin (0,0 is centre).
+    const width = block.width * scale;
+    const height = block.height * scale;
+    const top = xy.y * scale;
+    const bottom = (xy.y + block.height) * scale;
+    // In RTL the block's position is the top right of the block, not top left.
+    const left = this.workspace_.RTL ? xy.x * scale - width: xy.x * scale;
+    const right = this.workspace_.RTL ? xy.x * scale : xy.x * scale +  width;
+
```
alschmiedt, line 428: Looks great : ) On another PR I would maybe make it blockTop, blockBottom etc. just for readability sake

PR #74, opened 2020-02-20
Comment and formatting cleanup

moniika:
- Updating param name in jsdoc to match variable name
- Updating references to  `this.HTMLDiv` to use `this.HTMLDiv_` property
- Changing type of `textInput_` to `HTMLInputElement` (`HTMLTextInput` appears to not be recognized type)
- Moving `textInput_` and `HTMLDiv_` up in declaration to be next to the other private properties

### DIFF
--- a/blockly-ws-search/workspace_search.js
+++ b/blockly-ws-search/workspace_search.js
@@ -39,6 +39,13 @@ class WorkspaceSearch {
     this.svgGroup_ = null;
 
     /**
+     * The input for the search text.
+     * @type {?HTMLInputElement}
+     * @private
+     */
+    this.textInput_ = null;
+
+    /**
      * A list of blocks that came up in the search
      * @type {!Array.<Blockly.Block>}
      * @protected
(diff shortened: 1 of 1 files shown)


blockly-ws-search/workspace_search.js
```diff
@@ -64,20 +71,6 @@ class WorkspaceSearch {
      * @type {boolean}
      */
     this.searchOnInput = true;
-
-    /**
-     * HTML container for the workspace search bar.
-     * @type {?HTMLElement}
-     * @private
-     */
-    this.HtmlDiv_ = null;
```
alschmiedt, line 86: Instead of getting rid of this, I would like to just update references in the code to be use this.HtmlDiv_ instead of HtmlDiv.

moniika: ah, that's what happened. I've updated to use HtmlDiv_

alschmiedt (approved):
lgtm thanks for updating!

PR #75, opened 2020-02-20
Keyboard accessibility

alschmiedt:
If the marker is not already being used marks the current location before moving the cursor to the currently selected block. 

Added keydown handler to the buttons so it does not bubble up to Blockly's keydown handler. 

Use addEventListener instead of the private Blockly.bindEventWithChecks_ 

Fixes bug with this being undefined in isBlockMatch_

### DIFF
--- a/blockly-ws-search/workspace_search.js
+++ b/blockly-ws-search/workspace_search.js
@@ -146,6 +146,8 @@ class WorkspaceSearch {
     var btn = document.createElement('button');
     Blockly.utils.dom.addClass(btn, className);
     btn.addEventListener('click', onClickFn);
+    // TODO: Reivew Blockly's key handling to see if there is a way to avoid
+    // this.
     btn.addEventListener('keydown', function(e) {
       if (e.key === "Enter") {
         onClickFn(e);
--- a/blockly-ws-search/workspace_search.js
+++ b/blockly-ws-search/workspace_search.js
@@ -78,7 +78,6 @@ class WorkspaceSearch {
      * @private
      */
     this.textInput_ = null;
-    this.id = Math.random();
   }
 
   /**
--- a/blockly-ws-search/workspace_search.js
+++ b/blockly-ws-search/workspace_search.js
@@ -252,6 +252,10 @@ class WorkspaceSearch {
       const currBlock = this.blocks_[this.currentBlockIndex_];
       const currPath = currBlock.pathObject.svgPath;
       Blockly.utils.dom.addClass(currPath, 'searchCurrent');
+      if (this.workspace_.keyboardAccessibilityMode) {
+        let currAstNode = Blockly.navigation.getTopNode(currBlock);
+        this.workspace_.getCursor().setCurNode(currAstNode);
+      }
       // TODO: scroll to block if it is not visible on workspace
     }
   }
--- a/blockly-ws-search/workspace_search.js
+++ b/blockly-ws-search/workspace_search.js
@@ -78,6 +78,7 @@ class WorkspaceSearch {
      * @private
      */
     this.textInput_ = null;
+    this.id = Math.random();
   }
 
   /**
@@ -101,10 +102,9 @@ class WorkspaceSearch {
         this.onWorkspaceKeyDown_);
 
     // Add all the buttons for the search bar
-    var upBtn = this.createBtn_('upBtn', 'Find previous', this.previous_);
-    var downBtn = this.createBtn_('downBtn', 'Find next', this.next_);
-    var closeBtn = this.createBtn_('closeBtn', 'Close search bar', this.close);
-
+    var upBtn = this.createBtn_('upBtn', 'Find previous', this.previous_.bind(this));
+    var downBtn = this.createBtn_('downBtn', 'Find next', this.next_.bind(this));
+    var closeBtn = this.createBtn_('closeBtn', 'Close search bar', this.close.bind(this));
     this.HtmlDiv = document.createElement('div');
     Blockly.utils.dom.addClass(this.HtmlDiv, 'workspaceSearchBar');
 
@@ -146,7 +146,16 @@ class WorkspaceSearch {
     // Create the button
     var btn = document.createElement('button');
     Blockly.utils.dom.addClass(btn, className);
-    Blockly.bindEventWithChecks_(btn, 'click', this, onClickFn);
+    btn.addEventListener('click', onClickFn);
+    btn.addEventListener('keydown', function(e) {
+      if (e.key === "Enter") {
+        onClickFn(e);
+        e.preventDefault();  
+      } else if (e.key === "Escape") {
+        this.close();
+      }
+      e.stopPropagation();
+    }.bind(this));
     btn.append(textSpan);
     return btn;
   }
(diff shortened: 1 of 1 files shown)


blockly-ws-search/workspace_search.js
```diff
@@ -146,7 +145,18 @@ class WorkspaceSearch {
     // Create the button
     var btn = document.createElement('button');
     Blockly.utils.dom.addClass(btn, className);
-    Blockly.bindEventWithChecks_(btn, 'click', this, onClickFn);
+    btn.addEventListener('click', onClickFn);
+    // TODO: Reivew Blockly's key handling to see if there is a way to avoid
```
moniika: Reivew -> Review

blockly-ws-search/workspace_search.js
```diff
@@ -146,7 +145,18 @@ class WorkspaceSearch {
     // Create the button
     var btn = document.createElement('button');
     Blockly.utils.dom.addClass(btn, className);
-    Blockly.bindEventWithChecks_(btn, 'click', this, onClickFn);
+    btn.addEventListener('click', onClickFn);
```
moniika, line 164: Since all the `onClickFn` need to bind `this`, I think you should deal with that binding here in `createBtn_`.
Also I found that in ES6 there is an easier way to deal with `this` by using arrow functions.
ES5:
`btn.addEventListener('click', onClickFn.bind(this));`
ES6
`btn.addEventListener('click', evt => onClickFn(evt));`

The same approach should be able to be used for L151

blockly-ws-search/workspace_search.js
```diff
@@ -258,10 +268,22 @@ class WorkspaceSearch {
...
     }
   }
 
+  /**
+   * Update the location of the cursor if the user is in keyboard naviation
+   * mode.
+   */
+  updateCursor_(currBlock) {
+    if (this.workspace_.keyboardAccessibilityMode) {
+      let currAstNode = Blockly.navigation.getTopNode(currBlock);
```
moniika: can use `const` here

blockly-ws-search/workspace_search.js
```diff
@@ -302,13 +327,25 @@ class WorkspaceSearch {
...
+    this.updateMarker_();
     this.textInput_.focus();
     if (this.searchText_) {
       this.search();
     }
     console.log("Open search bar");
   }
 
+  /**
+   * Mark the user's current position when opening the search bar.
```
moniika: Mark -> Marks

blockly-ws-search/workspace_search.js
```diff
@@ -302,13 +327,25 @@ class WorkspaceSearch {
...
+   */
+  updateMarker_() {
+    var marker = this.workspace_.getMarker(Blockly.navigation.MARKER_NAME);
+    if (this.workspace_.keyboardAccessibilityMode && !marker.getCurNode()) {
```
moniika: Since `getMarker` can return `null`, should we add a check in the `if`? or is it sufficient that the workspace is in `keyboardAccessibilityMode`?

blockly-ws-search/workspace_search.js
```diff
@@ -286,6 +300,17 @@ class WorkspaceSearch {
     }
   }
 
+  /**
+   * Update the location of the cursor if the user is in keyboard naviation
```
moniika: Update -> Updates

blockly-ws-search/workspace_search.js
```diff
@@ -302,13 +327,25 @@ class WorkspaceSearch {
...
     console.log("Open search bar");
   }
 
+  /**
+   * Mark the user's current position when opening the search bar.
+   */
+  updateMarker_() {
+    var marker = this.workspace_.getMarker(Blockly.navigation.MARKER_NAME);
+    if (this.workspace_.keyboardAccessibilityMode && !marker.getCurNode()) {
+      var curNode = this.workspace_.getCursor().getCurNode();
```
moniika: `var` -> `const`

blockly-ws-search/workspace_search.js
```diff
@@ -108,16 +108,18 @@ class WorkspaceSearch {
...
     // Add all the buttons for the search bar
-    var upBtn = this.createBtn_('upBtn', 'Find previous', this.previous_);
-    var downBtn = this.createBtn_('downBtn', 'Find next', this.next_);
-    var closeBtn = this.createBtn_('closeBtn', 'Close search bar', this.close);
-
+    var upBtn = this.createBtn_('upBtn', 'Find previous', 
+        this.previous_.bind(this));
+    var downBtn = this.createBtn_('downBtn', 'Find next', 
+        this.next_.bind(this));
+    var closeBtn = this.createBtn_('closeBtn', 'Close search bar',
```
moniika: Since touching this already `var` -> `const`

blockly-ws-search/workspace_search.js
```diff
@@ -108,16 +108,18 @@ class WorkspaceSearch {
...
     textInput.setAttribute('placeholder', 'Search');
     this.textInput_ = textInput;
-    Blockly.bindEventWithChecks_(textInput, 'keydown', this, this.onKeyDown_);
-    Blockly.bindEventWithChecks_(textInput, 'input', this, this.onInput_);
-    Blockly.bindEventWithChecks_(svg.parentNode, 'keydown', this,
-        this.onWorkspaceKeyDown_);
+    textInput.addEventListener('keydown', this.onKeyDown_.bind(this));
+    textInput.addEventListener('input', this.onInput_.bind(this));
+    svg.parentNode.addEventListener('keydown',
+        this.onWorkspaceKeyDown_.bind(this));
```
moniika: Recommend using arrow function to deal with binding this. See lower review comment for more details.

blockly-ws-search/workspace_search.js
```diff
@@ -387,7 +424,7 @@ class WorkspaceSearch {
       return;
     }
     const searchGroup = this.getSearchPool_();
-    const isBlockMatch = this.isBlockMatch_;
+    const isBlockMatch = this.isBlockMatch_.bind(this);
```
moniika: This bind is not necessary, isBlockMatch does not use `this`.
However, we could employ arrow function to avoid needing to make a variable for isBlockMatch, but I'll address in in another PR (because I should also refactor a bit in isBlockMatch_).

moniika: Backtrack slightly, I forgot I add in `this.caseSensitive` to function. I'm making a PR that would fix. If trying to fix it here, I would recommend instead using arrow function, or leaving as I'll be fixing it.

alschmiedt: Yeah I was trying to fix bug I found in that. And ok I can revert and leave for you to fix in a different PR

blockly-ws-search/workspace_search.js
alschmiedt: We always create this marker in createDom for the workspace so it should be fine, but it won't hurt so I will add.

blockly-ws-search/workspace_search.js
alschmiedt: I would rather leave that to the caller to decide since it allows more flexibility. For example if we want to create a btn but not bind it to this. Updated all the binds to use the arrow function instead of .bind.

moniika (approved):

PR #77, opened 2020-02-20
Update on click

alschmiedt:
Redo search when a user clicks on the input. This deals with when a user adds new blocks or deletes blocks.

When a user deletes the current block it finds the next in the list. 
When a user deletes any other block it updates itself to have the new index of the old block. 

In the future we will want to probably change to onfocus instead of onclick but can't do right now since focus is a bit messed up because of Blockly's scroll function.

7f1a706b-654f-4c24-a69e-6fc2b53a91f7

### DIFF
--- a/blockly-ws-search/workspace_search.js
+++ b/blockly-ws-search/workspace_search.js
@@ -110,7 +110,7 @@ class WorkspaceSearch {
     this.textInput_ = textInput;
     Blockly.bindEventWithChecks_(textInput, 'keydown', this, this.onKeyDown_);
     Blockly.bindEventWithChecks_(textInput, 'input', this, this.onInput_);
-    Blockly.bindEventWithChecks_(textInput, 'click', this, this.onClick_);
+    textInput.addEventListener('click', evt => this.onInputClick_(evt));
     Blockly.bindEventWithChecks_(svg.parentNode, 'keydown', this,
         this.onWorkspaceKeyDown_);
 
@@ -183,12 +183,16 @@ class WorkspaceSearch {
    * Handles clicking on the input value in search bar.
    * @param {Event} e The onclick event.
    */
-  onClick_(e) {
-    const inputValue = e.target.value;
-    const currIdx = this.currentBlockIndex_;
-    this.setSearchText_(inputValue);
+  onInputClick_() {
+    const oldBlock = this.currentBlock_;
+    let oldIdx = this.currentBlockIndex_;
     this.search();
-    this.setCurrentBlock_(currIdx);
+    const currentBlockIdx = this.blocks_.indexOf(oldBlock);
+    if (currentBlockIdx > -1) {
+      this.setCurrentBlock_(currentBlockIdx);
+    } else {
+      this.setCurrentBlock_(oldIdx);
+    }
   }
 
   /**
--- a/blockly-ws-search/workspace_search.js
+++ b/blockly-ws-search/workspace_search.js
@@ -197,15 +197,7 @@ class WorkspaceSearch {
    * @param {Event} e The onclick event.
    */
   onInputClick_() {
-    const oldBlock = this.currentBlock_;
-    let oldIdx = this.currentBlockIndex_;
-    this.search();
-    const currentBlockIdx = this.blocks_.indexOf(oldBlock);
-    if (currentBlockIdx > -1) {
-      this.setCurrentBlock_(currentBlockIdx);
-    } else {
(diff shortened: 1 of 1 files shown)


blockly-ws-search/workspace_search.js
```diff
@@ -191,6 +192,22 @@ class WorkspaceSearch {
     }
   }
 
+  /**
+   * Handles clicking on the input value in search bar.
+   * @param {Event} e The onclick event.
+   */
+  onInputClick_() {
+    const oldBlock = this.currentBlock_;
+    let oldIdx = this.currentBlockIndex_;
```
moniika: `let` -> `const`

blockly-ws-search/workspace_search.js
```diff
@@ -191,6 +192,22 @@ class WorkspaceSearch {
...
+    this.search();
+    const currentBlockIdx = this.blocks_.indexOf(oldBlock);
+    if (currentBlockIdx > -1) {
+      this.setCurrentBlock_(currentBlockIdx);
+    } else {
+      this.setCurrentBlock_(oldIdx);
```
moniika: Search will already set a new current block. If we want to change which one is selected, then I think it should only happen if the old block exists in new search results (i.e., remove else).

Additionally, this is behavior I think should go inside of `search` (so it always happens), possibly with a flag to allow turning off for performance optimization. That way if for example you are one a block that matches and you add/remove letters to search text, and that same "current" block is still in search pool it will remain highlighted. Also by moving it into `search` we can avoid calling `setCurrentBlock_` twice in one search unnecessarily.

blockly-ws-search/workspace_search.js
```diff
@@ -282,6 +293,9 @@ class WorkspaceSearch {
       Blockly.utils.dom.addClass(currPath, 'searchCurrent');
       this.updateCursor_(this.currentBlock_);
       this.scrollToVisible_(this.currentBlock_);
+      // Blockly.WidgetDiv.hide called in scroll is taking away focus.
+      // TODO: review setFocused call in Blockly.WidgetDiv.hide.
+      this.textInput_.focus();
```
moniika: I had put this call inside of `next_` because I wanted to decouple the text input logic with the search behavior. Instead of moving it here, I'd like to suggest instead adding this logic to `scrollToVisible_` (and remove `this.textInput_.focus() `entirely):
```
const activeEl = document.activeElement;
this.workspace_.scroll(-targetLeft, -targetTop);
if (activeEl) {
  // Blockly.WidgetDiv.hide called in scroll is taking away focus.
  // TODO: review setFocused call in Blockly.WidgetDiv.hide.
  activeEl.focus();
}
```
I tested it locally and it appears to work the same. That way, where the focus goes back to is a bit more flexible and its moved closer to the `scroll` call that is culprit of taking the focus.

blockly-ws-search/workspace_search.js
alschmiedt: I don't really agree with the first part. I think it is much nicer functionality to start searching where we left off. I think for a small ws it doesn't really matter but if someone has a lot of blocks it would be pretty jarring to be editing code in one part and then be taken to the start of the workspace when we click in the box. 

However, I updated it to be the functionality you described for this part and will take a look at adding extra functionality in another PR.

blockly-ws-search/workspace_search.js
alschmiedt: updated

blockly-ws-search/workspace_search.js
```diff
@@ -515,9 +522,14 @@ class WorkspaceSearch {
       // Scroll to show bottom of block
       targetTop = bottom - metrics.viewHeight;
     }
-
+    const activeEl = document.activeElement;
```
moniika: Move to inside the if statement, we don't need to assign this if we aren't going to scroll.

blockly-ws-search/workspace_search.js
```diff
@@ -367,12 +370,16 @@ class WorkspaceSearch {
 
   /**
    * Searches the workspace for the current search term.
+   * @param {Blockly.Block=} opt_block Optional block to use as the current
```
moniika: Looking at code search, it doesn't appear common to use "optional" in description for optional params. Would like to also add note that this is block to use as current if it is included in new search blocks or something like the block to try to use as the current.

moniika (approved):
Some minor comment, otherwise lgtm

PR #85, opened 2020-02-26
Injecting css for workspace search.

moniika:


blockly-ws-search/src/css.js
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * @fileoverview Styling for workspace search.
+ * @author kozbial@google.com (Monica Kozbial)
+ */
+
+/**
+ * CSS for highlighting blocks.
+ */
+Blockly.Css.register([
```
alschmiedt: Why are there two separate Css.register calls?

alschmiedt (approved):

blockly-ws-search/src/css.js
moniika: More for organization, but I can combine them.

PR #94, opened 2020-02-28
Adding helper npm scripts to web-script example.

moniika:


blockly-ws-search/example/web-script/package.json
```diff
@@ -2,8 +2,18 @@
   "name": "blockly-ws-search-web-script",
   "version": "1.0.0",
   "description": "Example of workspace search with script tag.",
+  "private": true,
+  "scripts": {
+    "preinstall": "npm install ../../../blockly-ws-search",
```
alschmiedt, line 7: If this is already in the devDependencies why do we need it here?

alschmiedt (approved):

blockly-ws-search/example/web-script/package.json
moniika: `blockly-ws-search` is a dependency, but without this `preinstall`, `blockly-ws-search` isn't installed **before** this module causing an error "no such file or directory" (but the npm install works the second time you call it).
Adding this line ensures that npm install works the first time. I think this is only a problem only because `blockly-ws-search` is a file path.

rachel-fenichel: I don't understand this.  Can you explain further or link to a reference you looked at?

PR #95, opened 2020-02-28
Fixing css injection logic.

moniika:
Removing reliance on Blockly.Css.register in order to:
- avoid issues of referencing the "correct" Blockly
- allowing for css to be injected at init, rather than needing to happen before Blockly workspace is injected.
- injecting css in style tag with id specific to workspace search

### DIFF
--- a/blockly-ws-search/src/WorkspaceSearch.js
+++ b/blockly-ws-search/src/WorkspaceSearch.js
@@ -10,7 +10,7 @@
  */
 
 
-import './css.js';
+import { injectSearchCss } from './css.js';
 import * as Blockly from 'blockly/core';
 
 export class WorkspaceSearch {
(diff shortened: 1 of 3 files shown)


alschmiedt (approved):
lgtm

rachel-fenichel (commented):
+1 to moving the injection call into init() and making sure it only happens once, but what's the reason for injecting css directly instead of through blockly?

moniika:
> +1 to moving the injection call into init() and making sure it only happens once, but what's the reason for injecting css directly instead of through blockly?

Blockly doesn't currently expose a way to inject css other than through `Blockly.Css.register`. Calls to `register` only work before Blockly is injected (since it is injected then and it should only happen once). Not depending on `register` gives more freedom on when we can inject the css (in `init` and it doesn't need to happen before workspace `inject`) and makes it easier to use `WorkspaceSearch`.
The other benefit is that we have control over the id of the style tag.

PR #99, opened 2020-02-29
Allowing for search algorithm overriding

moniika:
- renaming `search` to `searchAndHighlight` AND `populateBlocks` to `search`
- adding function `setSearchFunction` to allow for overriding default search behavior (and logic in to use the custom search if available).
- changing `Block.Block` to `Block.BlockSvg`
- moving dom creation to `createDom` method

### DIFF
--- a/blockly-ws-search/src/WorkspaceSearch.js
+++ b/blockly-ws-search/src/WorkspaceSearch.js
@@ -42,7 +42,7 @@ export class WorkspaceSearch {
 
     /**
      * A list of blocks that came up in the search
-     * @type {!Array.<Blockly.Block>}
+     * @type {!Array.<Blockly.BlockSvg>}
      * @protected
      */
     this.blocks_ = [];
@@ -69,6 +69,13 @@ export class WorkspaceSearch {
     this.searchText_ = '';
 
     /**
+     * An optional custom search function to override default behavior.
+     * @type {?function(Blockly.WorkspaceSvg, string): Array.<Blockly.BlockSvg>}
+     * @protected
+     */
+    this.customSearchFunction_ = null;
+
+    /**
      * Whether to search as input changes as opposed to on enter.
      * @type {boolean}
      */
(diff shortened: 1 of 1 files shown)


blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -42,7 +42,7 @@ export class WorkspaceSearch {
 
     /**
      * A list of blocks that came up in the search
-     * @type {!Array.<Blockly.Block>}
+     * @type {!Array.<Blockly.BlockSvg>}
```
alschmiedt, line 45: Why the change to BlockSvg?

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -68,6 +68,13 @@ export class WorkspaceSearch {
      */
     this.searchText_ = '';
 
+    /**
+     * An optional custom search function to override default behavior.
+     * @type {?function(Blockly.WorkspaceSvg, string): Array.<Blockly.BlockSvg>}
+     * @protected
+     */
+    this.customSearchFunction_ = null;
```
alschmiedt, line 76: We might want to make this static. Blockly.WorkspaceSearch.customSearchFunction. Doesn't seem like there would be a large use case for changing on object by object basis.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -409,27 +422,40 @@ export class WorkspaceSearch {
...
+  }
+
+  /**
+   * Searches the workspace for the current search term and highlights found
+   * blocks.
+   * @param {Blockly.BlockSvg=} opt_block Block to use as the current
    *     block if it is included in new search blocks.
    */
-  search(opt_block) {
+  searchAndHighlight(opt_block) {
```
alschmiedt, line 440: Could this be private? Since we now allow them to set the search function, and they can change how the block is highlighted through css?

moniika: Leaving this public gives the freedom of triggering/binding it outside of the search bar.

blockly-ws-search/src/WorkspaceSearch.js
moniika: Keeping track Block vs BlockSvg started to seem like too much overhead, especially after introducing a way for users to custom set the search function (and we are expecting the blocks returned to be BlockSvg). Since we take in a WorkspaceSvg, we're only sort of supporting non-rendered workspaces. We should probably talk more offline.

PR #103, opened 2020-03-04
Refactoring and type cleanup.

moniika:
- Cleaning up comments
  - Fixing typos
  - Rewording comments
  - Moving TODOs from method description to inside method
- Cleaning up methods
  - Reorder of declaration for some methods.
  - Reviewing private/protected annotations (following naming convention of _ for protected)
  - Removing unused parameters
  - Renaming of dom creation methods from get to create (to be more clear)
  - Removing `dispose()` as it was not actually disposing of html element
- Removing partial support for headless
  - Removing checks for `rendered`
  - Changing types from `Blockly.Block` to `Blockly.BlockSvg`
  - Removing now unnecessary casts
- Minor refactor of search preserving current block
  - Changing param from block to boolean for easier use
  - Preserving current block for more cases other than onclick
  - Exposing property to allow for turning off preserving block
- Exposing search bar placeholder text with setter method

### DIFF
--- a/blockly-ws-search/src/WorkspaceSearch.js
+++ b/blockly-ws-search/src/WorkspaceSearch.js
@@ -124,7 +124,7 @@ export class WorkspaceSearch {
     searchContainer.append(searchContent);
 
     const inputWrapper = this.createInputDiv_();
-    this.textInput_ = this.createTextInput();
+    this.textInput_ = this.createTextInput_();
     inputWrapper.append(this.textInput_);
     searchContent.append(inputWrapper);
 
@@ -178,7 +178,7 @@ export class WorkspaceSearch {
    * Creates the text input for the search bar.
    * @protected
    */
-  createTextInput() {
+  createTextInput_() {
     let textInput = document.createElement('input');
     textInput.type = 'text';
     textInput.setAttribute('placeholder', this.textInputPlaceholder_);
--- a/blockly-ws-search/src/WorkspaceSearch.js
+++ b/blockly-ws-search/src/WorkspaceSearch.js
@@ -7,6 +7,7 @@
 /**
  * @fileoverview Object responsible for workspace search.
  * @author aschmiedt@google.com (Abby Schmiedt)
+ * @author kozbial@google.com (Monica Kozbial)
  */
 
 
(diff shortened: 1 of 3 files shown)


blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -463,8 +479,9 @@ export class WorkspaceSearch {
 
   /**
    * Populates block list with blocks that match the search text.
+   * @protected
    */
-  populateBlocks() {
+  populateBlocks_() {
```
alschmiedt, line 484: Should we have this return the set of blocks and then set this.block_ in the parent function? That way if someone overwrites this then they can just return a set of blocks?

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -408,26 +420,30 @@ export class WorkspaceSearch {
...
+   * @param {boolean=} preserveCurrent Whether to preserve the current block
+   *    if it is included in the new search block.
    */
-  search(opt_block) {
+  search(preserveCurrent) {
+    let oldCurrentBlock = this.currentBlock_;
     this.clearBlocks();
-    this.populateBlocks();
+    this.populateBlocks_();
     this.highlightBlocks();
```
alschmiedt, line 430: Similar to the below comment we could take in the blocks that we want to highlight here. That way someone who is overriding this function knows how to get the blocks they need to highlight

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -102,88 +123,73 @@ export class WorkspaceSearch {
...
     return inputWrapper;
   }
 
   /**
-   * Gets the text input for the search bar.
+   * Creates the text input for the search bar.
+   * @protected
    */
-  getTextInput() {
+  createTextInput() {
```
alschmiedt: add _

blockly-ws-search/src/WorkspaceSearch.js
moniika: I feel like the name of the function `populateBlocks`  hints at how it sets the `blocks_` array and someone overriding this function would have access to the array since its been made protected as well.
So, if we were to change to have it pass around array, then I'd also want a rename. 
The pattern of passing around the blocks array sounds like it could be nice, but I'd rather not introduce these changes now and save it for a future version of the plugin.

blockly-ws-search/src/WorkspaceSearch.js
moniika: Similar to previous comment, I'd prefer to keep as is for now.

blockly-ws-search/src/WorkspaceSearch.js
moniika: Done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: If we wait until later and someone has already overridden this, then it will break them because they won't be returning the correct value. Would rather decide earlier rather than later how we want to implement this.

moniika: Discussed offline. Will merge as is and update API in follow up PR still targeting for getting in for first version.

PR #106, opened 2020-03-05
Update Package.json

alschmiedt:
•	Everything equal to or above 3.20200123.0.
•	Tested with earlier versions but doesn't work because the Marker Manager was introduced in last release.

### DIFF
--- a/blockly-ws-search/package.json
+++ b/blockly-ws-search/package.json
@@ -30,7 +30,8 @@
   ],
   "scripts": {
     "clean": "rm -rf $npm_package_directories_dist",
-    "prepare": "webpack --display-error-details --config webpack.config.js\n",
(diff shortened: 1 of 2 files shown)


blockly-ws-search/package.json
```diff
@@ -45,7 +46,11 @@
     "webpack-dev-server": "^3.10.3"
   },
   "peerDependencies": {
-    "blockly": "^3.20200123.1"
+    "blockly": ">=3.20200123.0"
```
moniika: we probably want to add an upper bound, but the explicit >= (and then < to be added) could be clearer than using "^" we had before.

moniika (approved):
Just note on peerDependencies version, otherwise looks good.

PR #109, opened 2020-03-05
Update APIs for highlights.

moniika:
Updating APIs for highlighting of blocks matching search as well as current block to make it easier for developers when overriding.

### DIFF
--- a/blockly-ws-search/src/WorkspaceSearch.js
+++ b/blockly-ws-search/src/WorkspaceSearch.js
@@ -349,7 +349,9 @@ export class WorkspaceSearch {
     if (!this.blocks_.length) {
       return;
     }
-    this.unhighlightCurrentSelection_(this.currentBlock_);
+    if (this.currentBlock_) {
+      this.unhighlightCurrentSelection_(this.currentBlock_);
+    }
     this.currentBlockIndex_ =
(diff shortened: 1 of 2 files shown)


blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -344,42 +344,18 @@ export class WorkspaceSearch {
...
+    }
     this.currentBlockIndex_ =
         (index % this.blocks_.length + this.blocks_.length) %
         this.blocks_.length;
     this.currentBlock_ = this.blocks_[this.currentBlockIndex_];
-    const path = this.currentBlock_.pathObject.svgPath;
-    Blockly.utils.dom.addClass(path, 'search-current');
-    this.updateCursor_(this.currentBlock_);
+    this.highlightCurrentSelection_(this.currentBlock_);
+    this.updateCursor_();
```
alschmiedt: Why are we not passing in the block here?

alschmiedt (approved):

blockly-ws-search/src/WorkspaceSearch.js
moniika: I've changed to back to accepting block as param as it seems more consistent

PR #110, opened 2020-03-05
Updating search algorithm APIs

moniika:


blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -463,10 +466,13 @@ export class WorkspaceSearch {
   /**
    * Returns whether the given block matches the search text.
    * @param {!Blockly.BlockSvg} block The block to check.
+   * @param {string} searchText The search text. Note if the search is case
+   *    insensitive, this will be passed already converted to lowercase letters.
```
alschmiedt, line 467: Not sure if I understand this comment, why will it be passed already converted to lowercase?

alschmiedt (approved):

blockly-ws-search/src/WorkspaceSearch.js
moniika: Our setSearchText converts it to lowercase, so at this point it would be lower case. I wanted to communicate that.
However, if someone overrode setSearchText_ then this would no longer be the case, so possibly the call to lower() should be moved to here. What do you think?

alschmiedt: sorry was reading case sensitive instead of case insensitive. Makes sense

moniika: I've moved the logic for calling setLowerCase on searchTest  as discussed offline.

PR #112, opened 2020-03-06
Update create dom to make it easier to override

alschmiedt:
- Add event listeners in the createDom method
- Check if the btn exists before adding event listeners and appending
- Add addActionBtn method so a user can add a button and we do the event binding for them. 
- For the above feature made createBtn public so that the buttons they create will be accessible.
- Added create previous and next button methods


blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -34,6 +34,13 @@ export class WorkspaceSearch {
      */
     this.HtmlDiv_ = null;
```
moniika: Since we are touching code with this. I think that according to style guide, this should be: `this.htmlDiv_`

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -141,18 +148,55 @@ export class WorkspaceSearch {
...
     this.textInput_ = this.createTextInput_();
-    inputWrapper.append(this.textInput_);
+    if (this.textInput_) {
+      this.textInput_.addEventListener('keydown',
+          evt => this.onKeyDown_(/** @type {KeyboardEvent} */ evt));
+      this.textInput_.addEventListener('input', () => this.onInput_());
+      this.textInput_.addEventListener('click',
+          () => this.searchAndHighlight(this.preserveSelected));
+      inputWrapper.append(this.textInput_);
+    }
```
moniika: I don't think we should support/allow for createTextInput_ returning null (and communicate through the jsdoc annotation things that can be null).

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -162,30 +206,25 @@ export class WorkspaceSearch {
...
-    textInput.addEventListener('click',
-        () => this.searchAndHighlight(this.preserveSelected));
     return textInput;
   }
 
   /**
-   * Creates the div that holds all of the search bar actions.
-   * @return {!HTMLDivElement} A div holding search bar actions.
+   * Creates the button used to get the next block in the list.
+   * @return {HTMLButtonElement} The next button.
```
moniika: I'd like to add `?` to be `?HTMLButtonElement` to make intent clearer for this and also for `createPreviousBtn_` and `createCloseBtn_`.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -141,18 +148,55 @@ export class WorkspaceSearch {
...
+    if (previousBtn) {
+      this.addBtnListener_(previousBtn, () => this.previous());
+      this.actionDiv_.append(previousBtn);
+    }
+
+    const closeBtn = this.createCloseBtn_();
+    if (closeBtn) {
+      this.addBtnListener_(closeBtn, () => this.close())
+      searchContainer.append(closeBtn);
+    }
```
moniika: Can we loosen the requirement to allow for these elements to be `HTMLElement` instead of `HTMLButtonElement`?
Also, we should use `this.addActionBtn` in the if statements.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -194,20 +233,16 @@ export class WorkspaceSearch {
...
    * Creates a button for the workspace search bar.
    * @param {string} className The class name for the button.
    * @param {string} text The text to display to the screen reader.
-   * @param {!Function} onClickFn The function to call when the user clicks on
-   *    the button.
    * @return {HTMLButtonElement} The created button.
-   * @private
    */
-  createBtn_(className, text, onClickFn) {
+  createBtn(className, text) {
```
moniika, line 241: I think we can leave `createBtn` private or at least protected. I don't think it makes sense as a public api for WS as its just a convenience method.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -216,6 +251,21 @@ export class WorkspaceSearch {
...
+   * Add event listener for clicking and keydown on the given button.
+   * @param {HTMLButtonElement} btn The button to add the event listener to.
+   * @param {!Function} onClickFn The function to call when the user clicks on 
+   *     or hits enter on the button.
+   * @private
+   */
+  addBtnListener_(btn, onClickFn) {
+    if (!btn) {
+      return;
+    }
```
moniika: We check for btn to be non-null in all cases where we use this (except for addActionBtn which I think we should document as requiring non-nul) so I think we should say this function only takes non-null button and remove check.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -141,18 +148,55 @@ export class WorkspaceSearch {
...
 
     this.HtmlDiv_.append(searchContainer);
 
     parentSvg.parentNode.insertBefore(this.HtmlDiv_, parentSvg);
   }
 
+  /**
+   * Add a button to the action div. This must be called after the init function
+   * has been called.
+   * @param {HTMLButtonElement} btn The button to add the event listener to.
```
moniika: I think this should be non-nullable `!HTMLButtonElement` and also possibly allow for this to be `HTMLElement` so that there is a way to append to the action div.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -141,18 +148,55 @@ export class WorkspaceSearch {
...
+   * @param {!Function} onClickFn The function to call when the user clicks on
```
moniika, line 188: I think it might make sense to allow for `onClickFn` to be passed as null if someone wanted to just append something to the div.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -228,8 +278,6 @@ export class WorkspaceSearch {
       }
```
moniika, line 271: I think this binding of the escape key to close is problematic for if someone doesn't want close on escape key.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -141,18 +148,55 @@ export class WorkspaceSearch {
     const inputWrapper = document.createElement('div');
     Blockly.utils.dom.addClass(inputWrapper, 'ws-search-input');
     this.textInput_ = this.createTextInput_();
```
moniika: I would like for us to change from restricting to textinput to just input (rename textInput to inputElement) so that we are more flexible. I don't believe it would break anything.

blockly-ws-search/src/WorkspaceSearch.js
moniika: I suppose we were never supporting not closing with Escape. We could add another property on WS search (such as: `closeOnEscape` maybe?) that we reference in here for whether to close on escape.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: Talked about offline, leaving public.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: Saving this to implement at a later date, or if we see any need for it.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: That seems like a rare use case and not going to support for this version.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: As discussed offline and in a comment before, not implementing for this pass.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -162,52 +202,43 @@ export class WorkspaceSearch {
...
+    return this.createBtn('close-btn', 'Close search bar');
   }
 
   /**
    * Creates a button for the workspace search bar.
    * @param {string} className The class name for the button.
    * @param {string} text The text to display to the screen reader.
-   * @param {!Function} onClickFn The function to call when the user clicks on
-   *    the button.
    * @return {HTMLButtonElement} The created button.
```
moniika: !HTMLButtonElement 
I think we can say that this method shouldn't return null, even if overriden.

moniika (approved):

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: ok, done.

PR #114, opened 2020-03-09
Workspace search

alschmiedt:
Adding workspace search demo to blockly samples. 

Examples folder:
- These are ways for us to test with the bundle we created to make sure everything works
Test folder:
 - What we are using to test ws search during development


blockly-ws-search/README.md
```diff
@@ -0,0 +1,30 @@
+[Home](../README.md)
+
+# blockly-ws-search [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
+
+This package provides an implementation of workspace search for a Blockly workspace.
+
+## Installation
+
+```
+npm install blockly-ws-search
```
samelhusseini: @blockly/plugin-workspace-search?

blockly-ws-search/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+## Usage
+
+```js
+var WorkSpaceSearch = require('blockly-ws-search');
```
samelhusseini: Same here, update to new package name.

blockly-ws-search/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+var workspace = Blockly.inject('blocklyDiv' + suffix, options);
```
samelhusseini: What's a suffix?

blockly-ws-search/example/node/index.js
```diff
@@ -0,0 +1,51 @@
+/**
+ * @license
+ * 
+ * Copyright 2019 Google LLC
```
samelhusseini: 2020. and can we instead use:
```
/**
 * @license
 * Copyright 2020 Google LLC
 */
```

blockly-ws-search/example/node/index.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Example of including Blockly in Node (Common JS)
```
samelhusseini: Update example description.

blockly-ws-search/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "author": "Blockly Team",
```
samelhusseini: The custom is name, version and description first, then everything else can be ordered however.

blockly-ws-search/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "name": "@blockly/plugin-workspace-search",
+  "version": "1.20200304.0",
+  "description": "Implementation of workspace search for Blockly.",
```
samelhusseini: This description will show up on npm. Perhaps something like:
"A Blockly plugin that adds workspace search support."

blockly-ws-search/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "license": "Apache-2.0",
+  "main": "./dist/workspace-search.umd.js",
+  "unpkg": "./dist/workspace-search.umd.js",
+  "directories": {
+    "dist": "dist",
+    "src": "src"
+  },
+  "files": [
+    "dist",
+    "src"
```
samelhusseini, line 29: Do we still need to publish sources?

blockly-ws-search/package.json
```diff
@@ -0,0 +1,56 @@
...
+  ],
+  "scripts": {
+    "clean": "rm -rf $npm_package_directories_dist",
+    "build": "webpack --display-error-details --env.mode=development",
```
samelhusseini, line 33: What does --display-error-details do?

blockly-ws-search/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "prepublishOnly": "npm run clean && webpack --display-error-details --env.mode=production",
+    "test": "webpack-dev-server"
```
samelhusseini: Change ``test`` to ``start``. webpack-dev-server doesn't run your tests.

blockly-ws-search/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "webpack": "^4.41.6",
+    "webpack-cli": "^3.3.11",
+    "webpack-dev-server": "^3.10.3"
+  },
+  "peerDependencies": {
+    "blockly": ">=3.20200123.0 <4"
+  },
+  "publishConfig": {
+    "access": "public",
+    "registry": "https://wombat-dressing-room.appspot.com"
```
samelhusseini, line 53: Why add this here?

blockly-ws-search/src/css.js
```diff
@@ -0,0 +1,115 @@
...
+  '.ws-search-content {',
+    'display: flex;',
+  '}',
+];
+
+/**
+ * Injects CSS for workspace search.
+ */
+export const injectSearchCss = (function() {
+  let executed = false;
```
samelhusseini, line 96: Why not use Blockly.Css.register ?

blockly-ws-search/test/main.js
```diff
@@ -0,0 +1,62 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Example of using WorkspaceSearch with Blockly.
```
samelhusseini: Is this description correct for this file?

blockly-ws-search/webpack.config.js
```diff
@@ -0,0 +1,52 @@
...
+    return {
+        target: 'web',
+        mode: env.mode,
+        entry: './src/WorkspaceSearch.js',
+        devtool: 'source-map',
+        output: {
+            path: path.resolve(__dirname, 'dist'),
+            filename: 'workspace-search.umd.js',
+            libraryTarget: 'umd',
+            umdNamedDefine: true,
```
samelhusseini: What does umdNamedDefine do?

blockly-ws-search/webpack.config.js
```diff
@@ -0,0 +1,52 @@
...
+            globalObject: 'this'
```
samelhusseini, line 25: What does globalObject: this do?

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Object responsible for workspace search.
```
samelhusseini: class?

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+ */
+
+
+import { injectSearchCss } from './css.js';
+import * as Blockly from 'blockly/core';
+
+export class WorkspaceSearch {
+  /**
+   * Class for workspace search.
+   * @param {!Blockly.WorkspaceSvg} workspace
```
samelhusseini: nit: Param description.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+     * @private
+     */
+    this.inputElement_ = null;
+
+    /**
+     * The placeholder text for the search bar input.
+     * @type {string}
+     * @private
+     */
+    this.textInputPlaceholder_ = 'Search';
```
samelhusseini, line 57: Just a thought, and not merge blocking, but what's the localization story here? Can this be integrated with Blockly.Msg.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+
+    /**
+     * A list of blocks that came up in the search
```
samelhusseini: nit: Add full stop.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+    /**
+     * Whether search should preserve the currently selected block by default.
+     * @type {boolean}
+     */
+    this.preserveSelected = true;
+  }
+
+  /**
+   * Initializes the workspace search bar.
+   */
```
samelhusseini, line 101: Do we need an associated dispose? 
How does one get rid of this and its event handlers without leaking memory?

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+
+    const closeBtn = this.createCloseBtn_();
+    if (closeBtn) {
+      this.addBtnListener_(closeBtn, () => this.close())
+      searchContainer.append(closeBtn);
+    }
+
+    this.htmlDiv_.append(searchContainer);
+
+    parentSvg.parentNode.insertBefore(this.htmlDiv_, parentSvg);
```
samelhusseini: Instead of parentSvg.parentNode, why not use injectionDiv. ``workspace.getInjectionDiv``

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+    return this.createBtn('close-btn', 'Close search bar');
+  }
+
+  /**
+   * Creates a button for the workspace search bar.
+   * @param {string} className The class name for the button.
+   * @param {string} text The text to display to the screen reader.
+   * @return {!HTMLButtonElement} The created button.
+   */
+  createBtn(className, text) {
```
samelhusseini: Why is this public?

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+        this.close();
+      }
+      e.stopPropagation();
+    });
+  }
+
+  /**
+   * Positions the search bar based on where the workspace's toolbox is.
+   */
+  positionSearchBar() {
```
samelhusseini: Should this be public?

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+    }
+  }
+
+  /**
+   * Sets search text.
+   * @param {string} text
+   * @protected
+   */
+  setSearchText_(text) {
+    this.searchText_ = text.trim();
```
samelhusseini: Why is the purpose of this. It's never used without calling searchAndHighlight. Can the two be combined.

blockly-ws-search/src/css.js
moniika: We're not using Blockly.Css.register so our css goes in a separate style tag than common that we can name 'block-ws-search-style' and so that we can inject the css when WorkspaceSearch init is called (which could happen after inject), rather than needing to do so when this script loads.

blockly-ws-search/package.json
alschmiedt: Didn't you say we needed them for source maps?

blockly-ws-search/test/main.js
moniika: I don't think its necessarily incorrect, but this could be updated to "Testing playground for WorkspaceSearch."

blockly-ws-search/example/web-imports/dist/index.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE html>
```
rachel-fenichel: I thought dist was for built files, and therefore gitignored.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+  /**
+   * Opens the search bar when Control F or Command F are used on the workspace.
+   * @param {KeyboardEvent} e The key down event.
+   * @private
+   */
+  onWorkspaceKeyDown_(e) {
+    // TODO: Look into handling keyboard shortcuts on workspace in Blockly.
+    if ((e.ctrlKey || e.metaKey) && e.key === "f") {
+      this.open();
+      e.preventDefault();
```
samelhusseini, line 327: Should probably stop propagation since you're handling it.

blockly-ws-search/package.json
alschmiedt: It's recommended in docs so you don't have to add to the command line every time you want to publish.

blockly-ws-search/package.json
moniika: We named it test because it opens our "test" playground. I'm fine with rename to start.

blockly-ws-search/package.json
moniika: It's like a verbose that displays additional details if there are errors.

blockly-ws-search/webpack.config.js
alschmiedt: https://github.com/webpack/webpack/issues/6522#issuecomment-557451468

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+  setCurrentBlock_(index) {
+    if (!this.blocks_.length) {
+      return;
+    }
+    if (this.currentBlock_) {
+      this.unhighlightCurrentSelection_(this.currentBlock_);
+    }
+    this.currentBlockIndex_ =
+        (index % this.blocks_.length + this.blocks_.length) %
+        this.blocks_.length;
```
samelhusseini, line 373: Simply: ``index % this.blocks_.length``

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+      return;
+    }
+    this.setCurrentBlock_(this.currentBlockIndex_ - 1);
+  }
+
+  /**
+   * Selects the next block.
+   */
+  next() {
+    if (!this.blocks_.length) {
```
samelhusseini: No need for this, you're not doing any real validation here. Validation is already happening in setCurrentBlock_.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+      this.open();
+      e.preventDefault();
+    }
+  }
+
+  /**
+   * Selects the previous block.
+   */
+  previous() {
+    if (!this.blocks_.length) {
```
samelhusseini: No need for this, you're not doing any real validation here. Validation is already happening in setCurrentBlock_.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+    this.workspace_.markFocused();
+    this.clearBlocks();
+  }
+
+  /**
+   * Shows or hides the workspace search bar.
+   * @param {boolean} show Whether to set the search bar as visible.
+   */
+  setVisible(show) {
+    this.htmlDiv_.style.display = show ? 'flex' : 'none';
```
samelhusseini, line 421: Why would a developer use this, if there's open and close? If no use, then maybe it should be private.

blockly-ws-search/README.md
moniika: It comes from multi-workspace demo. It should be removed.
`var workspace = Blockly.inject('blocklyDiv', options);`

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+    this.markCurrentPosition_();
+    this.inputElement_.focus();
+    if (this.searchText_) {
+      this.searchAndHighlight();
+    }
+  }
+
+  /**
+   * Marks the user's current position when opening the search bar.
+   */
```
samelhusseini, line 396: Add visibility

blockly-ws-search/README.md
rachel-fenichel: Also, options is not declared.  Either declare it with a value or add a comment explaining what it is.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+     * @protected
+     */
+    this.currentBlockIndex_ = -1;
+
+    /**
+     * Currently "selected" block.
+     * @type {Blockly.BlockSvg}
+     * @protected
+     */
+    this.currentBlock_ = null;
```
samelhusseini: I don't see uses for this where ``this.block_[this.currentBlockIndex_]]`` can't be used.

blockly-ws-search/src/css.js
```diff
@@ -0,0 +1,115 @@
...
+    'data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC' +
+    '9zdmciIGhlaWdodD0iMTQiIHZpZXdCb3g9IjAgMCAyNCAyNCIgd2lkdGg9IjE0Ij48cGF0aC' +
+    'BkPSJNNy40MSAxNS40MUwxMiAxMC44M2w0LjU5IDQuNThMMTggMTRsLTYtNi02IDZ6Ii8+PH' +
+    'BhdGggZD0iTTAgMGgyNHYyNEgweiIgZmlsbD0ibm9uZSIvPjwvc3ZnPg==';
+
+/**
+ * CSS for workspace search.
+ * @type {Array.<string>}
+ */
+const CSS_CONTENT = [
```
samelhusseini, line 48: In the spirit of not conflicting with user CSS. Can we prefix all these with blockly-

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+    blocks.forEach(function(block) {
+      const blockPath = block.pathObject.svgPath;
+      Blockly.utils.dom.addClass(blockPath, 'search-highlight');
+    });
+  }
+
+  /**
+   * Removes highlight from the provided blocks.
+   * @param {!Array.<Blockly.BlockSvg>} blocks The blocks to unhighlight.
+   * @protected
```
samelhusseini, line 575: @protected, use underscore.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+    const path = currentBlock.pathObject.svgPath;
+    Blockly.utils.dom.removeClass(path, 'search-current');
+  }
+
+  /**
+   * Adds highlight to the provided blocks.
+   * @param {!Array.<Blockly.BlockSvg>} blocks The blocks to highlight.
+   * @protected
+   */
+  highlightSearchGroup(blocks) {
```
samelhusseini: @protected, use underscore.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+    const path = currentBlock.pathObject.svgPath;
+    Blockly.utils.dom.addClass(path, 'search-current');
+  }
+
+  /**
+   * Removes "current selection" highlight from provided block.
+   * @param {Blockly.BlockSvg} currentBlock The block to unhighlight.
+   * @protected
+   */
+  unhighlightCurrentSelection_(currentBlock) {
```
samelhusseini, line 555: @protected, use underscore.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+      this.workspace_.getCursor().setCurNode(currAstNode);
+    }
+  }
+
+  /**
+   * Adds "current selection" highlight to the provided block.
+   * Highlights the provided block as the "current selection".
+   * @param {!Blockly.BlockSvg} currentBlock The block to highlight.
+   * @protected
+   */
```
samelhusseini, line 544: @protected, use underscore.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+    Blockly.utils.dom.removeClass(path, 'search-current');
+  }
+
+  /**
+   * Adds highlight to the provided blocks.
+   * @param {!Array.<Blockly.BlockSvg>} blocks The blocks to highlight.
+   * @protected
+   */
+  highlightSearchGroup(blocks) {
+    blocks.forEach(function(block) {
```
samelhusseini: use arrow function, don't mix unless you absolutely need to.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+    });
+  }
+
+  /**
+   * Removes highlight from the provided blocks.
+   * @param {!Array.<Blockly.BlockSvg>} blocks The blocks to unhighlight.
+   * @protected
+   */
+  unhighlightSearchGroup(blocks) {
+    blocks.forEach(function(block) {
```
samelhusseini: same here: use arrow function

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+  }
+
+  /**
+   * Updates the location of the cursor if the user is in keyboard accessibility
+   * mode.
+   * @param {!Blockly.BlockSvg} block The block to set the cursor to.
+   * @protected
+   */
+  updateCursor_(block) {
+    if (this.workspace_.keyboardAccessibilityMode) {
```
samelhusseini, line 533: FYI, this is package in Blockly. Should be public if we expect developers to use it.

blockly-ws-search/webpack.config.js
alschmiedt: Originally, it was used to fix one of our examples. But it seems to be working without it so will remove.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+  getMatchingBlocks_(workspace, searchText, caseSensitive) {
+    if (!searchText) {
+      return [];
+    }
+    if (!this.caseSensitive) {
+      searchText = searchText.toLowerCase();
+    }
+    const searchGroup = this.getSearchPool_(workspace);
+    return searchGroup.filter(
+        block => this.isBlockMatch_(block, searchText, caseSensitive));
```
samelhusseini, line 510: What is the purpose of splitting this method into another method. Do we expect to use isBlockMatch_ elsewhere? This shallow splitting of methods is actually making it harder to follow the code IMO.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: Since that is the only translation that would need to be done, we side stepped the issue a bit by having a setSearchPlaceholder so developers can set it. We can look into this more if people are interested, however we were not originally planning on it.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: Because this adds some nice properties for accessibility. If we are giving users the option to addActionBtn then I also wanted to give them an easy way to create an accessible button.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: Probably not. Will update to protected.

blockly-ws-search/webpack.config.js
samelhusseini: That makes sense, curious why we're not defining ``library``.

blockly-ws-search/README.md
alschmiedt: done

blockly-ws-search/README.md
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,660 @@
...
+  /**
+   * Creates a button for the workspace search bar.
+   * @param {string} className The class name for the button.
+   * @param {string} text The text to display to the screen reader.
+   * @return {!HTMLButtonElement} The created button.
+   */
+  createBtn(className, text) {
+    // Create a span holding text to be used for accessibility purposes.
+    const textSpan = document.createElement('span');
+    textSpan.innerText = text;
```
samelhusseini: What is this pattern based on? Why not use aria-label?

blockly-ws-search/package.json
samelhusseini: Right..

blockly-ws-search/package.json
samelhusseini: Yes it's more conventional to use start there. In addition, you can type ``npm start`` to run the start script.

blockly-ws-search/package.json
samelhusseini: Awesome!

blockly-ws-search/README.md
alschmiedt: done

blockly-ws-search/example/node/index.js
alschmiedt: done

blockly-ws-search/example/node/index.js
alschmiedt: done

blockly-ws-search/package.json
alschmiedt: done

blockly-ws-search/package.json
alschmiedt: done

blockly-ws-search/package.json
alschmiedt: done

blockly-ws-search/src/css.js
alschmiedt: done

blockly-ws-search/test/main.js
alschmiedt: done

blockly-ws-search/webpack.config.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: I didn't write this, but IMO the splitting of methods makes it easier to follow. I appreciate being able to read isBlockMatch vs just seeing a large chunk of code. However, if people have strong feelings I can definitely update.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: Good point, will make a PR in blockly

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: Both this one and the one below have _?

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/webpack.config.js
alschmiedt: For the script example you would need to use libraryName.WorkspaceSearch for the constructor, which I don't love. Also, libraries seem to be for providing multiple functions, we are really only providing the single class so didn't seem like there was a large need.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: Yeah, goooooood point. Don't know why I didn't do that before. done.

blockly-ws-search/src/WorkspaceSearch.js
moniika: It should be fine to replace `this.currentBlock_` with `this.block_[this.currentBlockIndex_]`

blockly-ws-search/src/WorkspaceSearch.js
moniika: That simplification doesn't work when index is negative (which we are trying to handle).

blockly-ws-search/example/web-imports/dist/index.html
alschmiedt: We still need this index file so just changing this to be public and then copying over this into the build folder when we run webpack.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: Sorry, didn't see this before I changed. Lmk and I will update.

blockly-ws-search/src/WorkspaceSearch.js
moniika: There used to be more logic in `setSearchText_` but also some motivation behind it was to make a way to decouple search and interacting with the `inputElement_`.

Proposal:
We pass search text as a parameter to `searchAndHighlight`
- `onInput` and `onKeyDown` would call pass in `this.inputElement_.value.trim()` while other calls pass `this.searchText_`

AND

`searchAndHighlight` is responsible for updating `this.searchText_` with new value.
OR
move the logic for `setSearchText_` (which is just trim and set searchText) to `onKeyDown` and `onInput`.

blockly-ws-search/src/WorkspaceSearch.js
moniika: I split it up because I thought it was easier to read.

blockly-ws-search/webpack.config.js
samelhusseini: Okay great, just checking.

blockly-ws-search/src/WorkspaceSearch.js
samelhusseini: I think making this public is scope creep. Developers can create a button, set a class name and add an aria-label themselves, we don't need to support that functionality.

blockly-ws-search/src/WorkspaceSearch.js
samelhusseini: I'm okay with the first option, having the ``searchAndHighlight`` sanitize is input is appropriate regardless from which entry point its called.

blockly-ws-search/src/WorkspaceSearch.js
samelhusseini: Okay this is going to be hard to reason about as its very subjective, but since both of you think it reads better then keep as is.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: ok done.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: Changed back

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: Ok updated

blockly-ws-search/README.md
```diff
@@ -0,0 +1,32 @@
+[Home](../README.md)
```
samelhusseini: Remove this link, since this is a published package, we don't want this around in the npm readme

blockly-ws-search/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+# blockly-ws-search [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
```
samelhusseini: Use ``@blockly/plugin-workspace-search`` here to ensure consistency when referring to this package.

blockly-ws-search/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+This package provides an implementation of workspace search for a Blockly workspace.
```
samelhusseini:
```
A [Blockly](https://www.npmjs.com/package/blockly) plugin that adds workspace search support.
```

blockly-ws-search/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Installation
+
+```
+npm install @blockly/plugin-workspace-search
```
samelhusseini:
```
npm install @blockly/plugin-workspace-search --save
```

blockly-ws-search/README.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+## Usage
+
+```js
+var WorkSpaceSearch = require('@blockly/plugin-workspace-search');
```
samelhusseini: Default to ES6 imports here, especially that this isn't a package aimed at Node.js developers.

```js
import Blockly from 'blockly';
import { WorkspaceSearch } from '@blockly/plugin-workspace-search';

const workspace = Blockly.inject('blocklyDiv');
const workspaceSearch = new WorkspaceSearch(workspace);

workspaceSearch.init();
```

blockly-ws-search/README.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+## API
+
+### `blockly-ws-search`
+
+- `init`
+- `dispose`
+- `open`
+- `close`
```
samelhusseini, line 33: Add License: 
```md

## License

Apache 2.0
```

blockly-ws-search/README.md
```diff
@@ -0,0 +1,32 @@
...
+// Dictionary of options used to initialize the workspace.
+var options  = {};
+var workspace = Blockly.inject('blocklyDiv', options);
+var workspaceSearch = new WorkspaceSearch(workspace);
+workspaceSearch.init();
+```
+
+## API
+
+### `blockly-ws-search`
```
samelhusseini: Why do we need this subheading?

blockly-ws-search/webpack.config.js
```diff
@@ -0,0 +1,51 @@
...
+    return {
+        target: 'web',
+        mode: env.mode,
+        entry: './src/WorkspaceSearch.js',
+        devtool: 'source-map',
+        output: {
+            path: path.resolve(__dirname, 'dist'),
+            filename: 'workspace-search.umd.js',
+            libraryTarget: 'umd',
+            globalObject: 'this'
```
samelhusseini, line 25: Since you're referencing the build file ``../dist/workspace-search.umd.js`` from your test ``index.html`` you'll need to tell webpack where to serve its build files when they're in-memory (ie: when webpack-dev-server) is running. You can do that by adding ``publicPath: '/dist/'`` here.

samelhusseini: To repro the issue, try running ``npm run clean`` and then ``npm start``.

blockly-ws-search/webpack.config.js
```diff
@@ -0,0 +1,51 @@
...
+                use: {
+                    loader: "babel-loader",
+                    options: {
+                        presets: ["@babel/preset-env"]
+                    }
+                }
+            }]
+        },
+        devServer: {
+            openPage: 'test',
```
samelhusseini, line 40: Configure the devServer port to 3000. Keeps it consistent with the rest of the repo. 
```
        port: 3000,
```

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,655 @@
...
+    this.htmlDiv_ = document.createElement('div');
+    Blockly.utils.dom.addClass(this.htmlDiv_, 'blockly-ws-search');
+    this.positionSearchBar_();
+
+    const searchContainer = document.createElement('div');
+    Blockly.utils.dom.addClass(searchContainer, 'blockly-ws-search-container');
+
+    const searchContent = document.createElement('div');
+    Blockly.utils.dom.addClass(searchContent, 'blockly-ws-search-content');
+    searchContainer.append(searchContent);
```
samelhusseini: ``append`` not available on IE11. 
Babel only transforms syntax. If you have a really good reason to use ``append`` instead of ``appendChild`` then you'll need to add a polyfill.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,655 @@
...
+        input.fieldRow.forEach(field => {
+          topBlockText.push(field.getText());
+        });
+      });
+      blockText = topBlockText.join(' ').trim();
+    }
+    if (!caseSensitive) {
+      blockText = blockText.toLowerCase();
+    }
+    return blockText.includes(searchText);
```
samelhusseini: string.``includes`` not available on IE11. https://caniuse.com/#feat=es6-string-includes
Either use ``indexOf`` or include the polyfill https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Polyfill

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,655 @@
...
+    }
+  }
+
+  /**
+   * Handles a key down for the search bar.
+   * @param {KeyboardEvent} e The key down event.
+   * @private
+   */
+  onKeyDown_(e) {
+    if (e.key === 'Escape') {
```
samelhusseini: The key value for escape is "Esc" instead of "Escape" on IE11. I assume you wanted to avoid keyCode because its deprecated. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode#Browser_compatibility

I think since Blockly uses keyCode, we could continue to use it here. Alternately, you could check for both e.key == 'Esc' and 'Escape', although I don't know what other non-standard key values IE uses, so it might be best to just go with keyCode to be on the safe side.

blockly-ws-search/README.md
alschmiedt: done

blockly-ws-search/README.md
alschmiedt: done

blockly-ws-search/README.md
alschmiedt: done

blockly-ws-search/README.md
alschmiedt: done

blockly-ws-search/README.md
alschmiedt: done

blockly-ws-search/README.md
alschmiedt: done

blockly-ws-search/webpack.config.js
alschmiedt: done

blockly-ws-search/webpack.config.js
alschmiedt: sweet thanks!

blockly-ws-search/src/css.js
```diff
@@ -0,0 +1,115 @@
...
+  'path.blocklyPath.search-highlight.search-current {',
+    'fill: grey;',
+  '}',
+  '.blockly-btn-text {',
+    'height: 1px;',
+    'overflow: hidden;',
+    'position: absolute;',
+    'width: 1px;',
+  '}',
+  '.blockly-close-btn {',
```
samelhusseini: Why is it that some are prefixed with ws-search, and others are not?

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: updated to use keycode.

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

blockly-ws-search/README.md
alschmiedt: removed.

blockly-ws-search/src/css.js
alschmiedt: Yeah looking back I was trying to follow some good conventions for css naming and then just didn't apply that to the buttons. Planning on adding blockly-ws-search as prefix to the buttons. Sound ok?

blockly-ws-search/src/css.js
```diff
@@ -0,0 +1,111 @@
...
+    'BkPSJNNy40MSAxNS40MUwxMiAxMC44M2w0LjU5IDQuNThMMTggMTRsLTYtNi02IDZ6Ii8+PH' +
+    'BhdGggZD0iTTAgMGgyNHYyNEgweiIgZmlsbD0ibm9uZSIvPjwvc3ZnPg==';
+
+/**
+ * CSS for workspace search.
+ * @type {Array.<string>}
+ */
+const CSS_CONTENT = [
+  /* eslint-disable indent */
+  'path.blocklyPath.search-highlight {',
```
samelhusseini: One last css that doesn't have the blockly-ws-search prefix.

blockly-ws-search/src/WorkspaceSearch.js
```diff
@@ -0,0 +1,658 @@
...
+   * Disposes of workspace search.
+   * Unlink from all DOM elements to prevent memory leaks.
+   */
+  dispose() {
+    if (this.htmlDiv_) {
+      this.htmlDiv_.remove();
+      this.htmlDiv_ = null;
+    }
+    this.actionDiv_ = null;
+    this.inputElement_ = null;
```
samelhusseini: I have a couple thoughts about this: 
- Should this also call workspace_.dispose as workspace search is usually wrapped around a workspace.
- You aren't disposing of event handlers that you've registered.

samelhusseini (approved):
Couple of nits. For dispose, I think you should just remove it and create a bug to tackle it in a separate change. In order to properly dispose of the events you've created (since they're anonymous functions), you'll have to reorganize your code a fair amount.

blockly-ws-search/src/css.js
alschmiedt: done

blockly-ws-search/src/WorkspaceSearch.js
alschmiedt: done

PR #122, opened 2020-03-11
Add template for Blockly codelabs, and sample asset.

rachel-fenichel:
The template calls out sections that we expect to exist, including "What you'll learn", "What you'll build", and "Setup".


codelabs/template.md
```diff
@@ -0,0 +1,56 @@
...
+categories: codelab,blockly,MORE CATEGORIES
+status: Draft
+Feedback Link: https://github.com/google/blockly-samples/issues/new
+
+# CODELAB TITLE
+
+## Codelab overview
+
+### What you'll learn
+Describe the skills or tools the developers will learn by the end of this codelab.
```
alschmiedt: Might be nice if this was a quick to read list. So just change from Describe -> List.

codelabs/template.md
```diff
@@ -0,0 +1,56 @@
...
+
+### What you'll build
+Describe what the finished product will look like.  Include screenshots or gifs.
+
+### Prerequisites
```
alschmiedt: From the other codelabs i've look at they all have "What you'll need" Not sure if we want to stick with that convention or not.

codelabs/template.md
```diff
@@ -0,0 +1,56 @@
+author: YOUR NAME
+summary: CODELAB SUMMARY
+id: TEMPLATE-CODELAB-ID
```
alschmiedt, line 3: Is there any kind of format you want for the codelab id? For example do we want Blockly in front of all of them? Or does it matter?

codelabs/template.md
```diff
@@ -0,0 +1,56 @@
...
+Describe what the finished product will look like.  Include screenshots or gifs.
+
+### Prerequisites
+List any terms, skills, or tools the developer should be familiar with when starting this codelab.
+
+## Setup
+
+In this section, tell the developer how to get set up to write and run their code.
+
+## Step One
```
alschmiedt: Is this going to be the text the user sees on the left hand side of the screen? If so, could we have some indicator that we are suppose to change this to a meaningful title?

alschmiedt (approved):
Added a few questions/suggestions but overall lgtm!

codelabs/template.md
rachel-fenichel: Changed to "Step One Title"

codelabs/template.md
rachel-fenichel: I'm not sure yet.  I think it doesn't matter too much since it isn't going to into the more general codelabs list.

codelabs/template.md
rachel-fenichel: Done

codelabs/template.md
rachel-fenichel: Done

PR #140, opened 2020-03-23
Adding unit tests to slider field

moniika:
Adding tests based on FieldNumber.


rachel-fenichel (commented):
Why is everything async?

moniika:
> Why is everything async?

I was going based off on another test suite in samples, but there's no reason for them to be async that I see, so I've removed the async.

rachel-fenichel (approved):

PR #155, opened 2020-03-25
Typed Variable Plugin

alschmiedt:
Guide to this very large PR (sorry): 
Main Files: 
Modal.js - I pulled out the main functionality for a modal, so if we want we can pull this into its own plugin. At the very least if we want to do something similar again we will have a n easy starting point.
TypedModal.js - Extends Modal and adds features necessary for a typed modal. 

Test Folder:
InputModal - Just my way of checking to see how easy it was to change
multi-playgorund.html
playground.html

Still In Progress: 
1. Tests
2. Centering on the workspace - I want to add this on the modal but what I need isn't in metrics so still need to figure out how to get what I need from Blockly.

Note: If you want to test multiple workspaces you will need to build the blockly.min.js file and then copy it into the node_modules folder in test, since this depends on the latest version of Blockly.
Screen Shot 2020-03-25 at 3 12 20 PM


packages/plugins/typed-modal/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "typed variables"
+  ],
+  "homepage": "https://github.com/google/blockly-samples/tree/master/packages/plugins/typed-modal#readme",
+  "bugs": {
+    "url": "https://github.com/google/blockly-samples/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/google/blockly-samples.git",
+    "directory": "packages/plugins/blockly-typed-modal"
```
samelhusseini: packages/plugins/typed-modal

packages/plugins/typed-modal/src/Modal.js
```diff
@@ -0,0 +1,366 @@
...
+    this.addEvent_(/** @type{!HTMLDivElement} */ this.htmlDiv_, 'keydown',
+        this, this.handleKeyDown_);
+  }
+
+  /**
+   * Inject the css for a Blockly modal.
+   * @protected
+   */
+  injectCss_() {
+    injectCss('blockly-modal-css', `
```
samelhusseini: I still think there's little reason not to use Blockly.CSS.register. We're shipping css injection logic in every plugin and that doesn't make sense.

packages/plugins/typed-modal/src/Modal.js
```diff
@@ -0,0 +1,366 @@
...
+      this.htmlDiv_.remove();
+      this.htmlDiv_ = null;
+    }
+  }
+
+  /**
+   * Shows the Blockly modal and focus on the first focusable element.
+   */
+  show() {
+    this.focusableEls = this.htmlDiv_.querySelectorAll('a[href],' +
```
samelhusseini: Test on IE, querySelectorAll doesn't have full support there.

packages/plugins/typed-modal/src/Modal.js
```diff
@@ -0,0 +1,366 @@
...
+   */
+  handleKeyDown_(e) {
+    if (e.keyCode === Blockly.utils.KeyCodes.TAB) {
+      if (this.focusableEls.length <= 1) {
+        e.preventDefault();
+        e.stopPropagation();
+        return;
+      }
+      if (e.shiftKey) {
+        this.handleBackwardTab_(e);
```
samelhusseini: Seems like we're doing more work than we need to here. You're taking a snapshot of all the focusable elements, and then handling tab navigation on each and every one of them, rather: 

Couldn't you just handle "wrap" tab navigation for the first and last element, to tab lock and ensure the user doesn't end up outside of the modal.

Good resource for implementing a dialog here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role

packages/plugins/typed-modal/package.json
alschmiedt: done

packages/plugins/typed-modal/src/Modal.js
alschmiedt: I did try to use Blockly.Css.Register, but I get a "CSS already injected" error. It looks like because the css is injected during createDom and the modal is created after this. However,I agree it is weird to keep including. Will look into alternative solution.

packages/plugins/typed-modal/src/Modal.js
alschmiedt: I did test and it seems to be working. I think the only bug I found was if there is nothing in the modal it throws an error. However all modals should have at least the x in the corner so seems like a corner case.

packages/plugins/typed-modal/src/Modal.js
alschmiedt: I think that is what I am doing. The handleForwardTab and handlBackwardTab only check if I need to wrap, if I don't then they don't do anything. 
Looking at it more I don't need to hold all the focusableEls so updated to have a firstFocusableEl and lastFocusableEl

packages/plugins/typed-modal/.eslintrc.json
```diff
@@ -0,0 +1,99 @@
+{
```
rachel-fenichel: Can you move this up to `packages/plugins/` so that you don't have to recreate it for each plugin?

packages/plugins/typed-modal/src/Modal.js
samelhusseini: You should be able to add the CSS to be registered during page load, rather than during modal initialization, like the toolbox in Blockly for example: 
https://github.com/google/blockly/blob/master/core/toolbox.js#L741

packages/plugins/typed-modal/.eslintrc.json
alschmiedt: done

packages/plugins/typed-modal/src/Modal.js
```diff
@@ -0,0 +1,369 @@
...
+     * HTML container for the modal.
+     * @type {HTMLDivElement}
+     * @private
+     */
+    this.htmlDiv_ = null;
+
+    /**
+     * Array holding info needed to unbind events.
+     * Used for disposing.
+     * Ex: [[node, name, func], [node, name, func]].
```
rachel-fenichel: API thought: should we expose `Blockly.boundEvent` or similar?

packages/plugins/typed-modal/src/TypedModal.js
```diff
@@ -0,0 +1,371 @@
...
+   * Dispose of the typed modal.
+   * @override
+   */
+  dispose() {
+    super.dispose();
+    this.workspace_.removeButtonCallback(this.btnCallBackName_);
+    this.variableNameInput_ = null;
+    this.variableTypesDiv_ = null;
+    this.firstTypeInput_ = null;
+    // TODO: Remove style tag.
```
rachel-fenichel: Remove todo

packages/plugins/typed-modal/src/TypedModal.js
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  /**
+   * Dispose of the typed modal.
+   * @override
+   */
+  dispose() {
+    super.dispose();
+    this.workspace_.removeButtonCallback(this.btnCallBackName_);
+    this.variableNameInput_ = null;
```
rachel-fenichel: Explicitly setting these to null is an old blockly pattern, but I don't think you need it for these--the garbage collector will get them--and we're phasing it out in blockly more generally.

rachel-fenichel (commented):
Looks pretty solid overall.  A naming nit: I think it should be TypedVariableModal, not TypedModal, so that the adjective is modifying the correct thing.

packages/plugins/typed-modal/src/TypedModal.js
alschmiedt: done

packages/plugins/typed-modal/src/TypedModal.js
alschmiedt: done

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
```diff
@@ -0,0 +1,368 @@
...
+   * @param {string} btnCallbackName The name used to register the button
+   *     callback.
+   * @param {Array<Array<string>>} types An array holding arrays with the
+   *     display name as the first value and the type as the second.
+   *     Ex: [['Penguin', 'PENGUIN'], ['Giraffe', 'GIRAFFE']].
+   * @param {TypedVarModalMessages} opt_messages The messages for a typed variable
+   *     modal.
+   */
+  constructor(workspace, btnCallbackName, types, opt_messages) {
+    const title = opt_messages ? opt_messages["TYPED_VAR_MODAL_TITLE"] :
```
samelhusseini: Why don't we use Blockly.Msg to house these instead?

packages/plugins/typed-modal/src/Modal.js
alschmiedt: We use it in both plugins so far, and I can see us using it in future plugins so it would probably be nice to not have to keep duplicating.

packages/plugins/typed-variable-modal/test/InputModal.js
```diff
@@ -0,0 +1,58 @@
...
+    this.firstTypeInput_ = selectDiv.querySelector('.input-modal-type');
+    return selectDiv;
+  }
+
+  /**
+   * Get the selected type.
+   * @return {string} The selected type.
+   */
+  getSelectedType() {
+    const selectDiv = this.variableTypesDiv_;
```
samelhusseini, line 32: What's the use case for this?

packages/plugins/typed-variable-modal/webpack.config.js
```diff
@@ -0,0 +1,52 @@
+/**
```
samelhusseini: You can now use ``../../webpack.config.js``.

packages/plugins/typed-variable-modal/src/Modal.js
```diff
@@ -0,0 +1,368 @@
...
+    this.htmlDiv_.setAttribute('role', 'dialog');
+    // End creating the container
+
+    // Create the header
+    const modalHeader = document.createElement('header');
+    Blockly.utils.dom.addClass(modalHeader, 'blockly-modal-header');
+
+    this.renderHeader_(modalHeader);
+
+    const exitButton = document.createElement('button');
```
samelhusseini, line 275: It's a best practice to place the exit button at the bottom of the DOM, otherwise its always going to be the first focusable element (which isnt a great user experience for a user using a screen reader)

packages/plugins/typed-variable-modal/test/InputModal.js
alschmiedt: In the InputModal I use an Input instead of a list of types to get the selected type. I needed a  way to give the selectedType. Probably doesn't need to be public though, so I can change that.

samelhusseini: Yeah, I just don't see a use case for this outside of the 'protected' case.

packages/plugins/typed-variable-modal/webpack.config.js
alschmiedt: done

packages/plugins/typed-variable-modal/src/Modal.js
alschmiedt: Ok, I removed the X in the top right corner and changed the cancel button to be in the modal.

samelhusseini: Curious why you decided to remove it?

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
alschmiedt: I think the idea was to make it standalone. We don't necessarily know what Blockly is going to do with its messages in the future and by making this separate it gives us the benefit of not having to update in the future. Also, because we are not planning on translating any of these messages I don't see the benefit we get from having it on Blockly.Msg.

samelhusseini: Fair enough.
If Blockly changes messages in the future, I would imagine we would continue to be backwards compatible with Blockly.Msg. 
The pro of using Blockly.Msg that I think you're missing with this is it makes it so the plugin is compatible with a developer's localization flow. 

Blockly uses Blockly.Msg to house localized messages, internally, but if a developer has additional blocks, fields, etc and wants to localize those, they would do so with their own localization flow (translatewiki, crowdin, etc), and the output of these processes is a JSON that is dumped into Blockly.Msg. In other words, if they add these additional messages to their flow, everything would automagically work if you use Blockly.Msg.

alschmiedt: Ok that makes sense. I changed so that the typed modal uses Blockly.Msg.

packages/plugins/typed-variable-modal/src/Modal.js
alschmiedt: Talked about offline a bit. In the end I put the 'x' button back and kept it in the header. When I show the modal I skip over the 'x' button and focus the next item in the list if one exists. The reason I didn't position absolutely is because I didn't have all the positioning info I needed. Also, if I positioned absolutely and someone had a very long title then the button would overlap with it.

packages/plugins/typed-variable-modal/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+Create your workspace.
+```js
+workspace = Blockly.inject('blocklyDiv', {
+    toolbox: toolbox
+});
+```
+
+Add a custom category to your toolbox.
+```js
```
samelhusseini: xml

packages/plugins/typed-variable-modal/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "name": "@blockly/plugin-typed-var-modal",
+  "version": "1.0.0",
+  "description": "A Blockly plugin to create a modal for creating typed variables.",
+  "author": "Blockly Team",
+  "keywords": [
+    "blockly",
+    "modal",
+    "typed variables"
+  ],
+  "homepage": "https://github.com/google/blockly-samples/tree/master/packages/plugins/typed-var-modal#readme",
```
samelhusseini: rename to new directory name: typed-variable-modal

packages/plugins/typed-variable-modal/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "bugs": {
+    "url": "https://github.com/google/blockly-samples/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/google/blockly-samples.git",
+    "directory": "packages/plugins/typed-var-modal"
```
samelhusseini: same here: typed-variable-modal

packages/plugins/typed-variable-modal/src/Modal.js
```diff
@@ -0,0 +1,372 @@
...
+     * @private
+     */
+    this.boundEvents_ = [];
+
+    /**
+     * If true close the modal when the user clicks outside the modal.
+     * Otherwise, only close when user hits the 'X' button or escape.
+     * @type {boolean}
+     */
+    this.closeOnClick = true;
```
samelhusseini: Maybe rename to shouldCloseOnOverlayClick?

samelhusseini: Also, should we add one for shouldCloseOnEsc?

packages/plugins/typed-variable-modal/src/Modal.js
```diff
@@ -0,0 +1,372 @@
...
+    this.htmlDiv_.appendChild(modalFooter);
+  }
+
+  /**
+   * Render content for the modal header.
+   * @param {HTMLElement} headerContainer The modal's header div.
+   * @protected
+   */
+  renderHeader_(headerContainer) {
+    const modalTitle = document.createElement('H2');
```
samelhusseini: nit: lowercase h2.

packages/plugins/typed-variable-modal/src/Modal.js
```diff
@@ -0,0 +1,372 @@
...
+   * Render content for the modal footer.
+   * @param {HTMLElement} _footerContainer The modal's footer div.
+   * @protected
+   */
+  renderFooter_(_footerContainer) {
+    // No-op on the base class.
+  }
+}
+Blockly.Css.register([`
+    .blockly-modal-container {
```
samelhusseini: I'm seeing different styles of CSS here. In blockly we use camelcase, and not `-`. Any reason to not stick with the blockly style?

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * An array holding arrays with the name of the type and the display name
+     *     for the type. Ex: [['Penguin', 'PENGUIN'], ['Giraffe', 'GIRAFFE']].
+     * @type {Array<Array<string>>}
+     * @protected
+     */
+    this.types_ = types;
+
+    const messages = opt_messages || {
```
samelhusseini: Might be better to do something like. 
Blockly.utils.object.mixin({
 "TYPED_VAR_MODAL_TITLE": "... ",
 ... 
}, opt_messages);

That way you ensure that we have at least the EN messages when querying Blockly.Msg['TYPED_VAR...']

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
```diff
@@ -0,0 +1,385 @@
...
+   * @param {string} btnCallbackName The name used to register the button
+   *     callback.
+   * @param {Array<Array<string>>} types An array holding arrays with the
+   *     display name as the first value and the type as the second.
+   *     Ex: [['Penguin', 'PENGUIN'], ['Giraffe', 'GIRAFFE']].
+   * @param {TypedVarModalMessages} opt_messages The messages for a typed variable
+   *     modal.
+   */
+  constructor(workspace, btnCallbackName, types, opt_messages) {
+    const title = opt_messages ? opt_messages["TYPED_VAR_MODAL_TITLE"] :
```
samelhusseini: If you're passed in a opt_messages bag that doesn't contain 'TYPED_VAR_MODAL_TITLE', you get no title. 

Alternately you could do: 
```
(opt_messages && opt_messages['TYPED_VAR_MODAL_TITLE']) || 'Create Typed Variable';
```

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
```diff
@@ -0,0 +1,385 @@
...
+  }
+
+  /**
+   * Set the messages for the typed variable modal.
+   * Used to change the location.
+   * @param {!TypedVarModalMessages} messages The messages needed to create a
+   *     typed modal.
+   */
+  setLocale(messages) {
+    Object.keys(messages).forEach(function(k) {
```
samelhusseini: nit: Use arrow function.

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
```diff
@@ -0,0 +1,385 @@
...
+   * @override
+   */
+  renderContent_(contentContainer) {
+    const varNameContainer = this.createVarNameContainer_();
+    this.variableNameInput_ = varNameContainer
+        .querySelector('.typed-modal-variable-name-input');
+
+    const typedVarDiv = document.createElement('div');
+    Blockly.utils.dom.addClass(typedVarDiv, 'typed-modal-types');
+    typedVarDiv.innerText = Blockly.Msg["TYPED_VAR_MODAL_TYPES_LABEL"];
```
samelhusseini: use textContent instead.

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
```diff
@@ -0,0 +1,385 @@
...
+   * @override
+   */
+  renderFooter_(footerContainer) {
+    footerContainer.appendChild(this.createConfirmBtn_());
+    footerContainer.appendChild(this.createCancelBtn_());
+  }
+
+  /**
+   * Create button in charge of creating the variable.
+   * @return {!HTMLButtonElement}The button in charge of creating a variable.
```
samelhusseini: nit: extra space after }

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
```diff
@@ -0,0 +1,385 @@
...
+   * @protected
+   */
+  createConfirmBtn_() {
+    const confirmBtn = document.createElement('button');
+    Blockly.utils.dom.addClass(confirmBtn, 'blockly-modal-btn');
```
samelhusseini: I would avoid using ``Blockly.utils.dom.addClass`` unless you have a need to add / remove a class. In this case you're creating the element right there, simply setting ``className`` to ``blockly-modal-btn  blockly-modal-btn-primary`` is clearer.

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
```diff
@@ -0,0 +1,385 @@
...
+    Blockly.utils.dom.addClass(confirmBtn, 'blockly-modal-btn-primary');
+    confirmBtn.innerText = Blockly.Msg['TYPED_VAR_MODAL_CONFIRM_BUTTON'];
```
samelhusseini: In general go with textContent. It doesn't have security implications like innerHTML (security), and doesn't parse HTML like innerText (performance).

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
```diff
@@ -0,0 +1,385 @@
...
+
+  /**
+   * Create button in charge of cancelling.
+   * @return {!HTMLButtonElement} The button in charge of cancelling.
+   * @protected
+   */
+  createCancelBtn_() {
+    const cancelBtn = document.createElement('button');
+    Blockly.utils.dom.addClass(cancelBtn, 'blockly-modal-btn');
+    cancelBtn.innerText = Blockly.Msg['TYPED_VAR_MODAL_CANCEL_BUTTON'];
```
samelhusseini: Same as above. Use className = '...', and textContent

packages/plugins/typed-variable-modal/test/InputModal.js
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * @override
+   */
+  renderFooter_(footerContainer) {
+    footerContainer.appendChild(this.createConfirmBtn_());
+    const randomBtn = document.createElement('button');
+    Blockly.utils.dom.addClass(randomBtn, 'blockly-modal-btn');
+    randomBtn.innerText = 'Create random name';
+    this.addEvent_(randomBtn,'click',this, ()=> this.randomName());
```
samelhusseini: nit, spacing after 'click'

packages/plugins/typed-variable-modal/README.md
alschmiedt: done

packages/plugins/typed-variable-modal/package.json
alschmiedt: done

packages/plugins/typed-variable-modal/package.json
alschmiedt: done

packages/plugins/typed-variable-modal/src/Modal.js
alschmiedt: done

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
alschmiedt: done

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
alschmiedt: done

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
alschmiedt: done

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
alschmiedt: done

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
alschmiedt: done

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
alschmiedt: done

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
alschmiedt: done

packages/plugins/typed-variable-modal/src/TypedVariableModal.js
alschmiedt: done

packages/plugins/typed-variable-modal/test/InputModal.js
alschmiedt: done

samelhusseini (approved):
Overall looks good, a few nits and bits worth attending to.

packages/plugins/typed-variable-modal/src/Modal.js
alschmiedt: Added both

packages/plugins/typed-variable-modal/src/Modal.js
alschmiedt: We use the - in workspace search plugin. It seems like that is the most agreed upon style for css naming in the community and what the google style guide says as well. I am fine either way, but would prefer that we keep all the plugins the same. Depending on what we decide will update workspace search to match.

samelhusseini: Agreed on consistency, and I'm okay with diverging from blockly style guide (to google style guide) if we're consistent about it. Problem here is things like field_date and field_slider would have to change (only because they live in blockly-samples), and particularly for fields, some of the field CSS will follow one pattern and others a different pattern only because it lives in a different repo. 

I would recommend we follow the blockly style guide for things that exist in Blockly. Obviously for ES6 style things, we would have to match the Google style guide.

packages/plugins/typed-variable-modal/src/Modal.js
```diff
@@ -0,0 +1,378 @@
...
+    this.boundEvents_.push(event);
+  }
+
+  /**
+   * Create all the dom elements for the modal.
+   */
+  render() {
+    /*
+     * Creates the Modal. The generated modal looks like:
+     * <div class="blockly-modal-container" role="dialog">
```
samelhusseini: I know that we've split up the "content" of the modal into a header / content / footer, but these are just semantic divisions. 
The "content" of the modal (ie: the div that gets styled to be in the center) in this case is the container div, and i don't see a div for the overlay (in case you wanted to style that differently). 

Could we have an overlay div and a content div.
```
<div class="blocklyModalOverlay">
  <div class="blocklyModalContent">
      <header class="blocklyModalHeader"></header>
      <main class="blocklyModalMainContent"></main>
      <footer class="blocklyModalHeader"></footer>
  </div>
</div>
```
or similar.

This lets you style the modal like this: 
Screen Shot 2020-04-01 at 2 43 50 PM

alschmiedt: I think the "overlay" is the widget div. When we show the modal we add it into the widget div. I could add a class to the widget div so they could do something similar to that?

samelhusseini: (In summary), there's no way to style the overlay.

samelhusseini: You could add a style to the widget div, but I think and I'm seeing this in the widgetDispose, that's adding a lot of extra complexity on widget cleanup.
Why not just add another div, say blocklyModalOverlay and set all event handlers, and styling on that.

Making the widget div just a container for housing things, that is simply cleared on hideChaff. The other benefit of that, is you wouldn't have to compete with .blocklyWidgetDiv CSS elsewhere in the editor for what you're trying to achieve.

packages/plugins/typed-variable-modal/src/Modal.js
```diff
@@ -0,0 +1,378 @@
...
+   */
+  renderFooter_(_footerContainer) {
+    // No-op on the base class.
+  }
+}
+Blockly.Css.register([`
+    .blockly-modal-container {
+      background-color: white;
+      border: 1px solid gray;
+      max-width: 50%;
```
samelhusseini: why set max-width to 50%, and then width to 400px. That doesn't make sense to me?

alschmiedt: whoops, left over from a previous version. Removed.

packages/plugins/typed-variable-modal/src/Modal.js
alschmiedt: Ok updated.

packages/plugins/typed-variable-modal/src/Modal.js
```diff
@@ -0,0 +1,381 @@
...
+  /**
+   * Render content for the modal header.
+   * @param {HTMLElement} headerContainer The modal's header div.
+   * @protected
+   */
+  renderHeader_(headerContainer) {
+    const modalTitle = document.createElement('h2');
+    modalTitle.className = 'blocklyModalHeaderTitle';
+    modalTitle.appendChild(document.createTextNode(this.title_));
+    this.htmlDiv_.setAttribute('aria-labelledby', this.title_);
```
samelhusseini: This now needs to go on the modalContainer, where the role is set.

packages/plugins/typed-variable-modal/test/index.html
```diff
@@ -0,0 +1,83 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>Workspace Search Playground</title>
```
samelhusseini: Rename to typed variable playground.

packages/plugins/typed-variable-modal/src/Modal.js
alschmiedt: updated

packages/plugins/typed-variable-modal/test/index.html
alschmiedt: done

Issue #157, opened 2020-03-26
Base webpack config for plugins


rachel-fenichel:
Where it lives: `packages/plugins/`
What it contains:

```
 env => {
    return {
        target: 'web',
        mode: env.mode,
        entry: './src/index.js',
        devtool: 'source-map',
        module: {
            rules: [{
                test: /\.js$/,
                exclude: /(node_modules)/,
                use: {
                    loader: "babel-loader",
                    options: {
                        presets: ["@babel/preset-env"]
                    }
                }
            }]
        },
        devServer: {
            openPage: 'test',
            port: 3000,
            open: true
        },
        externals: {
            'blockly/core': {
                root: 'Blockly',
                commonjs: 'blockly/core',
                commonjs2: 'blockly/core',
                amd: 'blockly/core'
            },
        }
    }
};
```

Open question: 
Do we pick a standard `output` and include it, or require each package to add it?  Sample `output`:
```
        output: {
            path: path.resolve(__dirname, 'dist'),
            publicPath: '/dist/',
            filename: 'typed-modal.umd.js',
            libraryTarget: 'umd',
            globalObject: 'this'
        },
```

PR #162, opened 2020-03-27
Add a common webpack config file to be used by all packages.

samelhusseini:
This should reduce duplication of webpack configuration. A package can instead call: 
``webpack --config ../../webpack.config.js`` to use this config instead of its own. 

The config does the following: 
- Builds src with babel-loader. It expects a src/index.js and outputs to dist/index.js
- If the ``--env.buildTest=true`` variable is passed in, it will also build a test directory, expecting a ``test/index.js`` and outputs that into ``build/test_bundle.js``. 

This is what I would expect the new package.jsons to configure their scripts to do: 
```
    "build": "webpack --config ../../webpack.config.js --display-error-details --env.mode=development",
    "prepublishOnly": "npm run clean && webpack --config ../../webpack.config.js --display-error-details --env.mode=production",
    "start": "webpack-dev-server --config ../../webpack.config.js --env.buildTest=true --env.mode=development",
```

### DIFF
--- a/packages/webpack.config.js
+++ b/packages/webpack.config.js
@@ -1,0 +1,88 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+/**
+ * @fileoverview Webpack base configuration file.
+ * @author samelh@google.com (Sam El-Husseini)
+ */
+
+const path = require('path');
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):

samelhusseini:
Resolves https://github.com/google/blockly-samples/issues/157

PR #164, opened 2020-03-27
Upgrade webpack field slider

samelhusseini:
Upgrade field slider to not need a webpack.config and use the common one.


packages/fields/field-slider/src/index.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './field_slider';
```
moniika: add newline

moniika (approved):

PR #168, opened 2020-03-27
Webpack dev server config

samelhusseini:
Webpack dev server configs need to be in the first compile.
Adjust to add test page if we are building test/

### DIFF
--- a/packages/webpack.config.js
+++ b/packages/webpack.config.js
@@ -18,6 +18,15 @@ const packageJson = require(resolveApp('package.json'));
 console.log(`Building ${packageJson.name}`);
 
 module.exports = env => {
+    const devServer = {
+        port: 3000,
+        host: '0.0.0.0'
+    };
+    if (env.buildTest) {
+        devServer.openPage = 'test';
+        devServer.open = true;
+    }
+
     const src = {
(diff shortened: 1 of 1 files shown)


packages/webpack.config.js
```diff
@@ -18,6 +18,15 @@ const packageJson = require(resolveApp('package.json'));
 console.log(`Building ${packageJson.name}`);
 
 module.exports = env => {
+    const devServer = {
+        port: 3000,
+        host: '0.0.0.0'
+    };
+    if (env.buildTest) {
+        devServer.openPage = 'test';
```
alschmiedt, line 26: What is this part used for?

alschmiedt (approved):

packages/webpack.config.js
samelhusseini: openPage opens the test page if we're building a /test directory.
Or do you mean what is it used for if we aren't?

alschmiedt: I guess I mean how does it know what test page to open.

samelhusseini: It assumes there's an index.html. And just opens the test directory.

Issue #156, opened 2020-03-26
Plugin template


rachel-fenichel:
Files and folders needed for a plugin template:
- `eslintrc` (unless we move it up a level)
- `README`
- `package.json`
- `src/`
- `test/`
- `test/playground.html`
- `test/multi_playground.html`
- `webpack.config.js`

rachel-fenichel:
After discussion, `eslintrc` is not needed because we're moving it up a level to `packages/plugins`

samelhusseini:
``webpack.config.js`` is also no longer needed if ``@blockly/scripts`` is used.

PR #172, opened 2020-03-30
Add +/- Mutator Blocks

BeksOmega:
### Description

Adds +/- mutators for the following blocks:
* controls if
* text join
* list create
* procedure def

Blocks are based on OpenRoberta's mutators, and some [discussion](https://groups.google.com/d/msg/blockly/jktzy12YND4/C4JsrulhBQAJ) on the forums. Currently they look like this:
Blocks

Currently it is only slightly module-ized. This was just so that I could test it on mocha easily haha.

### Testing
I wrote unit tests covering all of the mutator behavior I could think of. I also tried to thoroughly manually test the blocks, paying special attention to procedures and how they interact with variables.

### ToDo
- [X] Do you guys like the design? any changes?
- [X] How do you want to handle the untranslated 'variable' text?
- [X] What do you want to directory structure to look like?
- [X] How do you want to package this up?
- [X] Decide if any playground functions should go in dev-tools.
- [X] Decide how to handle FieldPlus and FieldMinus.
- [X] Decide if we should / how we should add pictures to the readme.
- [X] Possibly move all plus buttons defined in JSON to helper functions.
- [X] Delete pre-edit variable

Also please tell me if you guys would like me to update the demo on my site.

### Additional Info
This has been a fun project, thanks for letting me give it a shot :D Now please tear my code to pieces!


rachel-fenichel:
I just created #175 for discussion of expected directory structure (similar to #156 for plugins).

blocks/alternative_mutators/plus_minus/main.js
```diff
@@ -0,0 +1,178 @@
...
+
+document.addEventListener("DOMContentLoaded", function () { start() });
+
+var options = {};
+
+var workspace = null;
+
+
+// Playground specific functions.
+function parseArguments() {
```
rachel-fenichel: @samelhusseini to consider whether any of these functions (or similar) should go in the `dev-tools` package.

blocks/alternative_mutators/plus_minus/webpack.config.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+/**
+ * @fileoverview Webpack configuration file.
+ * @author aschmiedt@google.com (Abby Schmiedt)
+ */
+
+const path = require('path');
+const webpack = require('webpack');
+
+module.exports = env => {
```
rachel-fenichel: @samelhusseini to look into making a root `webpack.config.js`, similar to the one for plugins.

blocks/alternative_mutators/plus_minus/src/field_plus.js
```diff
@@ -0,0 +1,61 @@
...
+      this, 'media/plus.svg', 15, 15, '');
+};
+Blockly.utils.object.inherits(FieldPlus, Blockly.FieldImage);
+
+FieldPlus.fromJson = function(options) {
+  return new FieldPlus(options['args']);
+};
+
+FieldPlus.prototype.showEditor_ = function() {
+  // TODO: This is a dupe of the mutator code, anyway to unify?
```
rachel-fenichel: Can you make a function that's `doMutation(add)` that takes in a boolean to know whether to add or remove an input?

blocks/alternative_mutators/plus_minus/src/field_minus.js
```diff
@@ -0,0 +1,61 @@
...
+
+  block.minus(this.args_);
+
+  var newMutationDom = block.mutationToDom();
+  var newMutation = newMutationDom && Blockly.Xml.domToText(newMutationDom);
+
+  if (oldMutation != newMutation) {
+    Blockly.Events.fire(new Blockly.Events.BlockChange(
+        block, 'mutation', null, oldMutation, newMutation));
+    // Ensure that any bump is part of this mutation's event group.
```
rachel-fenichel: Your recent PR said that this call to `bumpNeighbours` isn't needed, because anything that needs a bump should call it directly.  True here as well?

blocks/alternative_mutators/plus_minus/src/field_minus.js
```diff
@@ -0,0 +1,61 @@
...
+  return FieldMinus.superClass_.constructor.call(
+      this, 'media/minus.svg', 15, 15, '');
+};
+Blockly.utils.object.inherits(FieldMinus, Blockly.FieldImage);
+
+FieldMinus.fromJson = function(options) {
+  return new FieldMinus(options['args']);
+};
+
+FieldMinus.prototype.showEditor_ = function() {
```
rachel-fenichel: Why use `showEditor_` instead of the image field's `onClick` handling?

blocks/alternative_mutators/plus_minus/webpack.config.js
samelhusseini: Should this be published? If so, then we would want it under packages/ in which case it would share the ``webpack.config.js`` of all other packages.

blocks/alternative_mutators/plus_minus/src/field_plus.js
```diff
@@ -0,0 +1,61 @@
...
+
+// TODO: Refactor into class?
+// TODO: Re-add plusMinus namespace prefix?
+/**
+ * Class for a plus button used for mutation.
+ * @param opt_args Arguements to pass to the 'plus' function when the button
+ *    is clicked.
+ * @constructor
+ */
+export default function FieldPlus(opt_args) {
```
samelhusseini: Perhaps these two should be published as separate fields. Making it more reusable in other blocks. Plus / Minus fields seem quite generic to me.

blocks/alternative_mutators/plus_minus/src/blocks.js
```diff
@@ -0,0 +1,1040 @@
...
+  /**
+   * Removes the argument associated with the given argument ID from the block.
+   * @param {string} argId An ID used to track arguments on the block.
+   * @private
+   */
+  removeArg_: function(argId) {
+    if (!this.getInput(argId)) {
+      return;
+    }
+    // TODO: My life would be so much happier if this returned a boolean instead
```
rachel-fenichel: 🤔

rachel-fenichel: Here's the offending implementation:
```
/**
 * Remove an input from this block.
 * @param {string} name The name of the input.
 * @param {boolean=} opt_quiet True to prevent error if input is not present.
 * @throws {Error} if the input is not present and opt_quiet is not true.
 */
Blockly.Block.prototype.removeInput = function(name, opt_quiet) {
  for (var i = 0, input; (input = this.inputList[i]); i++) {
    if (input.name == name) {
      if (input.type == Blockly.NEXT_STATEMENT) {
        this.statementInputCount--;
      }
      input.dispose();
      this.inputList.splice(i, 1);
      return;
    }
  }
  if (!opt_quiet) {
    throw Error('Input not found: ' + name);
  }
};
```

blocks/alternative_mutators/plus_minus/src/field_minus.js
BeksOmega: Yes indeed! Thanks for the catch :D

blocks/alternative_mutators/plus_minus/src/field_minus.js
BeksOmega: Derp, fixed!

blocks/alternative_mutators/plus_minus/src/field_plus.js
BeksOmega: I'm cool with that! The only caveat is that these fields do fire mutation events. So whatever is used to document them should say explicitly that they should only be used for mutation.

blocks/alternative_mutators/plus_minus/src/field_plus.js
BeksOmega: I could, but I'm not sure that this TODO is important to fix right now.

I think this is something that should probably be fixed in core, rather than in this package, because anything that does mutation is going to look like:
```
  Blockly.Events.setGroup(true);
  var oldMutationDom = block.mutationToDom();
  var oldMutation = oldMutationDom && Blockly.Xml.domToText(oldMutationDom);

  // Do the mutating.

  var newMutationDom = block.mutationToDom();
  var newMutation = newMutationDom && Blockly.Xml.domToText(newMutationDom);
  if (oldMutation != newMutation) {
    Blockly.Events.fire(new Blockly.Events.BlockChange(
        block, 'mutation', null, oldMutation, newMutation));
  }
  Blockly.Events.setGroup(false);
```
And that doesn't just go for icons or +/- buttons, but also whatever other crazy UI people come up with.

I think a template or wrapper method that existed in core would make everything easier to maintain. But I also think that for now, leaving it like this is fine.

Idk though, whatever you think is good I'm cool with (┏－‿ ◕)┏

blocks/alternative_mutators/plus_minus/src/field_plus.js
rachel-fenichel: Now that it's an image field with an on-click, I think you don't need a custom field type at all for it.  You just need to add the onclick handler in your block code. That matches with the fact that you're just calling the superclass constructor with the specified image, and that the onclick is mutator-specific so you wouldn't want it in a generic implementation anyway.

rachel-fenichel:
- I like the block design.
- Packaging and directory structure: Sam is currently putting together templates for packages that we will publish, including blocks.  The appropriate directory will be `packages/blocks/`, once it exists.

BeksOmega:
So the timeline is: #178 goes in, then the templates branch gets merged into master, then I rebase and reorganize? Or would it be better if I rebased onto templates as it is now and reorganized?

samelhusseini:
There are still some changes brewing with templates. I think for you @BeksOmega my recommendation would be to reorganize your changes here based on the current blocks template: 
https://github.com/google/blockly-samples/tree/templates/packages/blocks/__template

It doesn't have to be re-based or anything, the primary structure really just has to be: 
- src/index.js 
- src/... rest of your files ...
- test/index.js
- test/index.html
- test/... rest of your test files ...
- package.json (Make sure that @blockly/dev-tools, @blockly/dev-scripts and blockly are dev dependencies. Also make sure your build / start npm scripts are calling "blockly-scripts build", "blockly-scripts start".
- a README.md file with readme stuff

And that should be the grunt of it. The above basically takes care of building (with webpack), and running your dev server (with webpack dev server). Mind you all of that is done behind the scenes.

So yeah, templates are still and will continue to evolve, but you don't have to wait for that branch to be ready to get this in. If your stuff matched the rest of the plugins / fields currently checked in, and by extension matches the current blocks template, then I think we'd be happy to update the package if we make changes to the blocks template. 

ps: your package is the first blocks package we're hoping to publish, so thank you for doing that, and for bearing with us as we figure out the details!

BeksOmega:
Ok I think I've got everything moved, tested, linted, etc!

Two little things:
1) I decided to go with plus-minus for the package name, but I'm totally open to changing that.
2) I think it would be nice to include some example images in the README. Would that be possible? And if so, how do you want me to store the images?

packages/blocks/plus_minus/README.md
```diff
@@ -0,0 +1,87 @@
+
+# @blockly/plus-minus [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
```
samelhusseini: The naming convention we're going with is: @blockly/(field|block|plugin)-NAME.
Since this is a block package, that would be: 
``@blockly/block-plus-minus``

packages/blocks/plus_minus/package.json
```diff
@@ -0,0 +1,58 @@
...
+    },
+    "main": "./dist/index.js",
+    "unpkg": "./dist/index.js",
+    "author": "Beka Westberg",
+    "keywords": [
+        "blockly",
+        "block",
+        "mutator"
+    ],
+    "homepage": "https://github.com/google/blockly-samples/tree/master/packages/blocks/plus-minus$readme",
```
samelhusseini: #readme instead of $readme.

samelhusseini:
@BeksOmega structure looks good to me :) Thanks for updating.

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1056 @@
...
+  /**
+   * Number of else-if inputs on this block.
+   * @type {number}
+   */
+  elseIfCount_: 0,
+  /**
+   * Number of else inputs on this block. Max should be 1.
+   * @type {number}
+   */
+  elseCount_: 0,
```
rachel-fenichel: This should be a boolean.

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1056 @@
...
+  // To properly keep track of indices we have to increment before/after adding
+  // the inputs, and decrement the opposite.
+  // Because we want our first elseif to be IF1 (not IF0) we increment first.
+
+  /**
+   * Adds an else-if and a do input to the bottom of the block.
+   * @this Blockly.Block
+   * @private
+   */
+  addPart_: function() {
```
rachel-fenichel: Naming suggestion: `addElseIf_` and `removeElseIf_`.

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1056 @@
...
+    // Handle if-elseif-else block.
+    if (this.getInput('ELSE')) {
+      this.moveInputBefore('ELSE', /* put at end */ null);
+    }
+  },
+
+  /**
+   * Appears to remove the input at the given index. Actually shifts attached
+   * blocks and then removes the input at the bottom of the block. This is to
+   * keep input names accurate.
```
rachel-fenichel: > This is to keep input names accurate.

Accurate meaning that input names are IF1, IF2, etc. no matter how many times inputs have been added or removed, right?

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1056 @@
...
+  removePart_: function(index) {
+    // The strategy for removing a part at an index is to:
+    //  - Kick any blocks connected to the relevant inputs.
+    //  - Move all connect blocks from the other inputs up.
+    //  - Remove the last input.
+    // This makes sure all of our indices are correct.
+    if (index !== undefined) {
+      let foundInput = false;
+      const name = 'IF' + index;
+      for (let i = 0, input; (input = this.inputList[i]); i++) {
```
rachel-fenichel: Why does this loop start at zero instead of at `index`?

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1056 @@
...
+        if (input.name == name) {
+          foundInput = true;
+          let connection = input.connection;
+          // TODO: Is there any reason we need to throw errors when the
```
rachel-fenichel: Is this TODO because that's the behaviour on the existing mutators in core Blockly?

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1056 @@
...
+            connection.disconnect();
+          }
+          connection = this.inputList[++i].connection; // Do connection.
+          if (connection.isConnected()) {
+            connection.disconnect();
+          }
+          this.bumpNeighbours();
+          continue;
+        }
+        if (!foundInput) {
```
rachel-fenichel: Instead of 
```
if (!foundInput) {
  continue;
}
if (input.name == 'ELSE') {
  break;
}
// other stuff
```

Do
```
if (foundInput && input.name != 'ELSE') {
  // other stuff
}
```

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1056 @@
...
+      for (let i = 0, input; (input = this.inputList[i]); i++) {
+        if (input.name == name) {
+          foundInput = true;
+          let connection = input.connection;
+          // TODO: Is there any reason we need to throw errors when the
+          //  connection isn't currently connected?
+          if (connection.isConnected()) {
+            connection.disconnect();
+          }
+          connection = this.inputList[++i].connection; // Do connection.
```
rachel-fenichel: Add a longer comment pointing out + explaining the `++i`.  It's easy to miss.

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1056 @@
...
+    this.removeInput('DO' + this.elseIfCount_);
+    this.elseIfCount_--;
+  },
+};
+
+/**
+ * Adds the initial plus button to the if block.
+ * @this Blockly.Block
+ */
+const controlsIfHelper = function() {
```
rachel-fenichel: Why is the initial plus added dynamically here, but for a few other blocks you redefined the block entirely and added the plus in the json?

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1056 @@
...
+  /**
+   * Adds an input to the end of the block. If the block currently has no
+   * inputs it updates the top 'EMPTY' input to receive a block.
+   * @this Blockly.Block
+   * @private
+   */
+  addPart_: function() {
+    if (this.itemCount_ == 0) {
+      if (this.getInput('EMPTY')) {
+        // TODO: I don't think this should throw errors. It would be nice if
```
rachel-fenichel: Filed as https://github.com/google/blockly/issues/3803

rachel-fenichel (commented):
Here's a partial review, mostly focusing on the if mutator so far.

packages/blocks/plus_minus/src/index.js
BeksOmega: I can do that! It will still need to be stored as an int in the XML though b/c that's how core handles it.

packages/blocks/plus_minus/src/index.js
BeksOmega: Yes! I could change it to... "To keep input names in number order" maybe. Idk the best way to describe it, what do you think?

packages/blocks/plus_minus/src/index.js
BeksOmega: It is the existing behavior in core blockly, but I don't think any mutators use this. I believe they just let the blocks get kicked automatically when the input gets removed. But because we're faking which input gets removed, we have to kick them manually.

packages/blocks/plus_minus/src/index.js
BeksOmega: I think that in an ideal world I wouldn't have to define any JSON, and all of the plus buttons would be added this way. I added json for the other blocks because in core they are defined using js and I figured it was better to move to the new standard.

It might be worth removing the + buttons from the JSON and adding them via helpers, so that it's easier to remove the json if core ever moves these blocks to json.

BeksOmega (commented):
Thanks for the review! I'll get on fixing stuff : D Also I updated the OP's todo list so that everybody knows where we're at.

packages/blocks/plus_minus/src/index.js
rachel-fenichel: "The inputs are always IF1, IF2, etc. with no gaps."

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+   * @private
+   */
+  removeElseIf_: function(opt_index) {
+    // The strategy for removing a part at an index is to:
+    //  - Kick any blocks connected to the relevant inputs.
+    //  - Move all connect blocks from the other inputs up.
+    //  - Remove the last input.
+    // This makes sure all of our indices are correct.
+    if (opt_index !== undefined && opt_index!= this.elseIfCount_) {
+      let inputIndex = opt_index * 2;
```
rachel-fenichel: Now that you're not doing a loop for this section, you're only using this index in a few places and the `++`es are not needed.  

Instead, here I would use:
```
// Each else-if is two inputs on the block: the else-if connection and the do connection.
let elseIfIndex = opt_index * 2;
```  

On line 254, use `elseIfIndex + 1`.

On line 260, use `let i = elseIfIndex + 2`.

This way the relationship with the first input where you're making a change is clear.

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+        //   It returned a boolean instead.
+        this.removeInput('EMPTY');
+      }
+      this.topInput_ = this.appendValueInput('ADD' + this.itemCount_)
+          .appendField(new FieldPlus(), 'PLUS')
+          .appendField(Blockly.Msg['TEXT_JOIN_TITLE_CREATEWITH']);
+    } else {
+      this.appendValueInput('ADD' + this.itemCount_);
+    }
+    this.itemCount_++;
```
rachel-fenichel: Put a comment here that says "// Input names are zero-indexed (ADD0, ADD1, etc)" and another before line 389 that says the same.

And remove the comment from 356-358

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+      this.topInput_.removeField('MINUS');
+    }
+  },
+};
+
+/**
+ * Adds the quotes mixin to the block. Also updates the shape so that if no
+ * mutator is provided the block has two inputs.
+ * @this Blockly.Block
+ * @constructor
```
rachel-fenichel: Why the constructor annotation?  It's called in the constructor, but it's not a constructor fn on its own.

packages/blocks/plus_minus/src/index.js
rachel-fenichel: Similar story for the related comments for the if block--delete the comment hanging out between functions (lines 206-209) in favor of comments on the lines that need to be explained.

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+const textJoinHelper = function() {
+  this.mixin(Blockly.Constants.Text.QUOTE_IMAGE_MIXIN);
+  this.updateShape_(2);
+};
+
+Blockly.Extensions.unregister('text_join_mutator');
+Blockly.Extensions.registerMutator('text_join_mutator',
+    textJoinMutator, textJoinHelper);
+
+const listCreateMutator = {
```
rachel-fenichel: Note: this mutator is very similar to the previous one, differing mostly in the `addPart_` and `removePart_` functions.

rachel-fenichel: I'm thinking of ways to combine them but they all add another layer of abstraction and I'm not convinced that it's work adding that to avoid the code duplication.

@samelhusseini thoughts?  Look at `mutationToDom`, `domToMutation`, `updateShape_`, `plus`, `minus`, and `updateMinus_` on this and on the previous chunk.

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+  /**
+   * Used by the context menu to create a caller block.
+   * @type {string}
+   */
+  callType_: 'procedures_callreturn',
+};
+
+Blockly.Extensions.registerMixin('get_procedure_def_return', getDefReturn);
+
+const procedureContextMenu = {
```
rachel-fenichel: I like that you split the procedure code into multiple extensions like this, because it's easy to understand what each extension does.

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+    this.hasStatements_ = xmlElement.getAttribute('statements') !== 'false';
+    if (!this.hasStatements_) {
+      this.removeInput('STACK');
+    }
+
+    const names = [];
+    const ids = [];
+    for (let i = 0, childNode; (childNode = xmlElement.childNodes[i]); i++) {
+      if (childNode.nodeName.toLowerCase() != 'arg') {
+        continue;
```
rachel-fenichel: I prefer 
```
if (childNode.nodeName.toLowerCase() == 'arg') {
  names.push(childNode.getAttribute('name'));
  ids.push(childNode.getAttribute('varid') ||
      childNode.getAttribute('varId'));
}
```

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+  /**
+   * Adds arguments to the block until it matches the targets.
+   * @param {!Array<string>} names An array of argument names to display.
+   * @param {!Array<string>} varIds An array of variable IDs associated with
+   *     those names.
+   * @this Blockly.Block
+   * @private
+   */
+  updateShape_: function(names, varIds) {
+    // In this case it is easiest to just reset and build from scratch.
```
rachel-fenichel: In which case?

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+    for (let i = this.argIds_.length - 1; i >= 0; i--) {
+      this.removeArg_(this.argIds_[i]);
+    }
+
+    this.arguments_ = [];
+    this.varIds_ = [];
+    this.argumentVarModels_ = [];
+    this.argIds_ = [];
+
+    const length = varIds.length;
```
rachel-fenichel: Check that `varids.length` and `names.length` are the same, and error if not, unless guaranteed elsewhere.

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+    if (!this.argIds_.length) {
+      return;
+    }
+    this.removeArg_(argId);
+    Blockly.Procedures.mutateCallers(this);
+  },
+
+  /**
+   * Adds an argument to the block and updates the block's parallel tracking
+   * tracking arrays as appropriate.
```
rachel-fenichel: Nit: "tracking" is in here twice.

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+  /**
+   * Removes the argument associated with the given argument ID from the block.
+   * @param {string} argId An ID used to track arguments on the block.
+   * @private
+   */
+  removeArg_: function(argId) {
+    if (!this.getInput(argId)) {
+      return;
+    }
+    // TODO: My life would be so much happier if this returned a boolean instead
```
rachel-fenichel: Remove TODO, since it's now tracked in blockly core

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+    //  of throwing an error. I'm saying at least 15% more joy.
+    this.removeInput(argId);
+    if (this.arguments_.length == 1) { // Becoming argumentless.
+      this.getInput('TOP').removeField('WITH');
+    }
+
+    const index = this.argIds_.indexOf(argId);
```
rachel-fenichel: There are four arrays that you're tracking together and keeping in sync.  Instead, use a single array that contains objects, each of which has `argument`, `varId`, `varmodel`, `argId`

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+   */
+  addVarInput_: function(name, argId) {
+    const nameField = new Blockly.FieldTextInput(name, this.validator_);
+    nameField.onFinishEditing_ = this.finishEditing_.bind(nameField);
+    nameField.createdVarIds_ = [];
+
+    this.appendDummyInput(argId)
+        .setAlign(Blockly.ALIGN_RIGHT)
+        .appendField(new FieldMinus(argId))
+        .appendField('variable:') // Untranslated!
```
rachel-fenichel: You use Blockly.Msg for most of the field text in this PR.  Is this one untranslated because we don't have a message in Blockly that matches what you need?

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+    newName = newName.replace(/[\s\xa0]+/g, ' ').replace(/^ | $/g, '');
+    if (!newName) {
+      return null;
+    }
+
+    // The field name (aka id) is always equal to the arg id.
+    const index = sourceBlock.argIds_.indexOf(this.name);
+    const caselessName = newName.toLowerCase();
+    for (let i = 0, name; (name = sourceBlock.arguments_[i]); i++) {
+      if (i == index) {
```
rachel-fenichel: As mentioned above, I prefer 
```
if (i != index && caselesName == name.toLowerCase()) {
  return null;
}
```
instead of having an extra `if (condition) { continue; }`

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+
+  /**
+   * Validates text entered into the argument name field.
+   * @param {string} newName The new text entered into the field.
+   * @return {?string} The field's new value.
+   * @this Blockly.FieldTextInput
+   */
+  validator_: function(newName) {
+    const sourceBlock = this.getSourceBlock();
+    newName = newName.replace(/[\s\xa0]+/g, ' ').replace(/^ | $/g, '');
```
rachel-fenichel: For regexes, add a comment saying what it's checking.

packages/blocks/plus_minus/src/index.js
```diff
@@ -0,0 +1,1048 @@
...
+        continue;
+      }
+      if (caselessName == name.toLowerCase()) {
+        return null; // It matches, so it is invalid.
+      }
+    }
+
+    sourceBlock.arguments_[index] = newName;
+
+    // TODO: Maybe delete the pre-edit variable if it has no other uses.
```
rachel-fenichel: You can still get the variable with `workspace.getVariableById`

rachel-fenichel (commented):
Okay, I've looked through the rest of your main file now and put comments in.

A few higher-level comments:
- I am still not sure that the plus and minus fields need to be a new type of field, instead of just adding an image field with the correct onclick whenever you want one.
- I think you should split index.js into multiple files: one for procedures, one for if-blocks, etc.  They can all be required in index.js, but they don't need to be in the same file since they don't inteact.
- I'm still thinking about the JSON vs javascript thing for how the blocks are defined/redefined.

packages/blocks/plus_minus/src/index.js
BeksOmega: How does "Because item inputs are 0-index we decrement first, increment last" sound to you?

I want to get across the idea that we need to do one (decrementing or incrementing) at one end, and the other at the other end.

This is because when I was writing this I would drop it for a few weeks and come back later. I kept having this problem where I would be like "why didn't I make them match!?" so I would move them and break everything. I'd like to keep a reminder that they need to be done opposite eachother in there somewhere.

packages/blocks/plus_minus/src/index.js
BeksOmega: [This](http://bekawestberg.me/blog/mutator-api) might be of interest. I'm not super attached to the idea, but I think it would be one way to go if you're interested in API/reusability stuffs.

Basic idea is to abstract mutators into objects (think strategy pattern style interchangability). Users would have to pass in a top input, extra input, and empty input to initialize the mutator.

packages/blocks/plus_minus/src/index.js
BeksOmega: Yeah, there are [lots of strings](https://github.com/google/blockly/blob/07762ff4da3c713f7592c80052b1fa7cadb461a2/msg/messages.js#L143) that /contain/ the world variable, but none that just say 'variable' :/

packages/blocks/plus_minus/src/index.js
BeksOmega: Problem: I'm not exactly sure what it's checking. I just stole it from the core one haha. This check dates back to the [dawn of time](https://github.com/google/blockly/blame/19969117efaa16096a0913e6a00ddbc33d0be43a/blocks/procedures.js#L496), and there's no comment on it :/

My best guess is the first bit is replacing white space (tabs, non-breaking, etc) with normal spaces, and the second bit is removing any spaces at the start or end. -- I added that.

packages/blocks/plus_minus/src/index.js
BeksOmega: oh sweet you're right! Do you want me to add this? The behavior will be different than core, but I think it will be better, and it won't affect serialization or anything.

packages/blocks/plus_minus/src/index.js
BeksOmega: Am I allowed to use .map() and .find() and things?

BeksOmega:
Changing the procedures data structure caused the following behavior:
CreateDef

Because of [this line](https://github.com/google/blockly/blob/07762ff4da3c713f7592c80052b1fa7cadb461a2/blocks/procedures.js#L839) in core. It access def.arguements_ which no longer exists.

Edit: Should be fixed by https://github.com/google/blockly/pull/3814

packages/blocks/plus_minus/src/index.js
rachel-fenichel: Yes, you are!  (In this repo)

packages/blocks/plus_minus/src/index.js
rachel-fenichel: Okay, dawn of time regexes are excluded from that.

packages/blocks/plus_minus/src/index.js
rachel-fenichel: Sure, sounds good.

rachel-fenichel:
Remaining items from the todo list at the top:

- How do you want to handle the untranslated 'variable' text?
@alschmiedt how did you end up handling this in the typed variable modal?

- Decide if any playground functions should go in dev-tools.
Let's leave them in this package--no harm, since it's not in the module itself.

- Decide how to handle FieldPlus and FieldMinus.
I have a slight preference for not having these as fields registered in Blockly as a whole, given that they expect callbacks.

- Decide if we should / how we should add pictures to the readme.
Yes, the readme should have pictures.  I propose `/readme-media`.  @samelhusseini opinions?

- Possibly move all plus buttons defined in JSON to helper functions.
No opinion from me.

- Mutator code reusability proposal
I still need to finish reading/parsing.

alschmiedt:
For the typed modal I have a [setLocale](https://github.com/google/blockly-samples/blob/6cb6de6bfea412ae8f58a7ca2bfe912f853b54a1/packages/plugins/typed-variable-modal/src/TypedVariableModal.js#L133) method that adds the messages given to it to the Blockly.Msg object. I then reference that throughout the code and mention in my README all the values that someone would need to change in order to translate the typed modal. I'm not entirely sure, but I think in this case just using Blockly.Msg would be sufficient.

BeksOmega:
Updated!

- Removed field class and added the createMinusField an createPlusField functions instead. We can always change it back laters.
- Removed all plus fields from JSON definitions b/c of the above.
- Added information about assigning to Blockly.Msg['PROCEDURE_VARIABLE'] in the readme.
- Added images to the readme. -- I used absolute references b/c I think that's necessary for the npm page to work.
- Added deleting the pre-edit variable as long as it is only used by blocks associated with the procedure (i.e. the def and callers of the def). If it is access by a var block or by a different procedure def, it will not be deleted.
Variables

rachel-fenichel (approved):

PR #176, opened 2020-04-03
Add blockly scripts.

samelhusseini:
A package to house all dev scripts for Blockly packages. 

Current scripts: 
- build (prod): build dev / prod
```
 * This script:
 *   - Uses webpack to build the src directory in development mode if no
 *   additional arguments are passed.
 *   - Uses webpack to build the src directory in production mode if
 *   ``blockly-scripts build prod`` is called.
```
- start: run dev server
```
 * This script: 
 *   Builds the src and test directories in development mode.
 *   Starts the webpack dev server which watches changes to these directories.
 *   Opens the test page 0.0.0.0:3000/test
 *   Hot reloads the page if any changes are made to the source files.
```
- lint: run eslint on src and test directories
```
 * This script:
 *   Runs eslint on the src and test directories.
```
- clean: remove dist and build directories
```
 * This script:
 *   Deletes the dist and build directories if they exist.
```

Showcasing the use of it in the dev/tools package.

PR #178, opened 2020-04-03
Add field template

samelhusseini:
- [x] README.md
- [x] package.json
- [x] src/index.js
- [x] test/index.js
- [x] test/index.html
- [ ] Mocha tests?

Relies on blockly-scripts: https://github.com/google/blockly-samples/pull/176


rachel-fenichel (commented):
Let's also have mocha tests in the template, with a todo describing basic test cases.

packages/fields/__template/test/index.js
```diff
@@ -0,0 +1,45 @@
...
+          'type': 'field_label',
+          'text': 'NO_SLIDER_FIELD',
+        },
+      },
+    ],
+    'style': 'math_blocks',
+  }]);
+
+/**
+ * DOM loaded.
```
moniika, line 34: I would like for us to follow the google style guide for wording method comments, although we are not consistently following this rule in Blockly Core. ("written in the third person, as if there is an implied This method ... before it")
https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments
In this case, could reword to: Loads the DOM.

PR #190, opened 2020-04-07
Adding travis

samelhusseini:
- [x] Run build on all packages
- [x] Run lint on all packages
- [x] Run test on all packages

@rachel-fenichel, @moniika, @alschmiedt anything else?


samelhusseini:
@rachel-fenichel I believe I have to check this in to get travis to go online. I've enabled only the build for now, will enable lint and test in separate PRs.

.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: node_js
+node_js:
```
rachel-fenichel, line 4: 7 is not a good version--odd-numbered releases are unsupported after 6 months.  I think we should be testing on 10 and 12, which are the current active releases.

samelhusseini: Done.

samelhusseini:
Looks like travis got enabled :)
I'll continue working on the travis configuration in this PR. Changing to WIP.

samelhusseini:
And we're passing, build test and lint. I think this is good for now.

PR #194, opened 2020-04-07
Update @blockly/dev-scripts package

samelhusseini:
Add a README for the @blockly/dev-scripts package.
Fix up package.json


packages/dev/scripts/README.md
```diff
@@ -0,0 +1,25 @@
+# @blockly/dev-scripts [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
+
+This package includes scripts and configuration used by Blockly extension
```
alschmiedt: configuration -> configurations

alschmiedt: maybe just choose between Blockly extensions or Blockly packages?

alschmiedt (approved):

packages/dev/scripts/README.md
samelhusseini: Good point.

samelhusseini: Actually I agree with the standardization, but I actually think we should call them: 
``extension packages. ``

alschmiedt: Ok, any reason? I'm fine either way, just thinking that we have everything under packages and don't really seem to use extension anywhere else.

alschmiedt (approved):

packages/dev/scripts/README.md
samelhusseini: Blockly extension, could be confused core Blockly extensions: https://developers.google.com/blockly/guides/create-custom-blocks/extensions

Maybe ``Blockly package`` ain't too bad. They are npm packages afterall.
Under packages, we're referring to reach item by its type (plugin, field, block).

samelhusseini: Problem with Blockly package is that it doesn't really tell me anything about it being an extension on top of Blockly, hence why I landed on ``extension packages``

samelhusseini: I'm open to ideas.

samelhusseini:
Will hold off on merging this until we come to a consensus about naming.

samelhusseini:
Closing in favour of https://github.com/google/blockly-samples/pull/207

PR #201, opened 2020-04-10
Update package.json

alschmiedt:
npm run prod doesn't look like a command. I think it is suppose to be npm run dist. If it is then I can go and update in other places?


samelhusseini:
Yep, please update everywhere else.

samelhusseini (approved):

PR #203, opened 2020-04-13
slider color update falls back to block color if block has no style

seldomU:
The slider field currently assumes that the source block has a style property. This change lets it tolerate blocks without style by falling back to their colour_ property.

### DIFF
--- a/packages/fields/field-slider/src/field_slider.js
+++ b/packages/fields/field-slider/src/field_slider.js
@@ -87,8 +87,10 @@ export class FieldSlider extends Blockly.FieldNumber {
 
     Blockly.DropDownDiv.getContentDiv().appendChild(editor);
 
-    Blockly.DropDownDiv.setColour(this.sourceBlock_.style.colourPrimary,
-        this.sourceBlock_.style.colourTertiary);
+    Blockly.DropDownDiv.setColour(
+        this.sourceBlock_.style ? this.sourceBlock_.style.colourPrimary : this.sourceBlock_.colour_, 
+        this.sourceBlock_.style ? this.sourceBlock_.style.colourTertiary : this.sourceBlock_.colour_, 
+    );
 
     Blockly.DropDownDiv.showPositionedByField(
         this, this.dropdownDispose_.bind(this));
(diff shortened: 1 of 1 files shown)


packages/fields/field-slider/src/field_slider.js
```diff
@@ -87,8 +87,10 @@ export class FieldSlider extends Blockly.FieldNumber {
 
     Blockly.DropDownDiv.getContentDiv().appendChild(editor);
 
-    Blockly.DropDownDiv.setColour(this.sourceBlock_.style.colourPrimary,
-        this.sourceBlock_.style.colourTertiary);
+    Blockly.DropDownDiv.setColour(
+        this.sourceBlock_.style ? this.sourceBlock_.style.colourPrimary : this.sourceBlock_.colour_,
```
samelhusseini: use ``this.sourceBlock.getColour()`` instead, colour_ is private.

packages/fields/field-slider/src/field_slider.js
```diff
@@ -87,8 +87,10 @@ export class FieldSlider extends Blockly.FieldNumber {
 
     Blockly.DropDownDiv.getContentDiv().appendChild(editor);
 
-    Blockly.DropDownDiv.setColour(this.sourceBlock_.style.colourPrimary,
-        this.sourceBlock_.style.colourTertiary);
+    Blockly.DropDownDiv.setColour(
+        this.sourceBlock_.style ? this.sourceBlock_.style.colourPrimary : this.sourceBlock_.colour_, 
+        this.sourceBlock_.style ? this.sourceBlock_.style.colourTertiary : this.sourceBlock_.colour_,
```
samelhusseini: same here.

samelhusseini:
Thanks for your PR! One minor change, and it looks good to me.

samelhusseini (approved):

samelhusseini:
Apologies I thought this was doing something else than I re-read it, and my understanding was that a block should always have a style. When the colour is set explicitly a style (auto) property is created for it.

https://github.com/google/blockly/blob/master/core/block_svg.js#L1255

I'm curious why this is needed?

seldomU:
My bad. My Blockly fork is outdated and doesn't set the style prop in setColour. For up-to-date Blockly the slider field shouldn't cause any issue.

samelhusseini:
Okay great, I'll close this then.

BeksOmega:
So how far back are packages supposed to be compatible? Is it just the latest release?

samelhusseini:
That would be package dependent, every package lists its min requirements in the peerDependency part of the package.json.

PR #206, opened 2020-04-15
Update to be able to use path

alschmiedt:
Not sure if we want this but thought it would be nice. 

Allows you to do: create plugin path/to/package/package-name

Currently, it makes the name of the plugin path/to/package/package-name.


samelhusseini:
Can we figure this out automatically??

Path to root of git repo

samelhusseini:
Get the relative path between current directory and ``git rev-parse --show-toplevel``

plugins/dev-create/bin/create.js
```diff
@@ -20,6 +20,7 @@ const path = require('path');
 const execSync = require('child_process').execSync;
 
 const root = process.cwd();
+const pluginDirName = 'plugins';
```
samelhusseini: Can we make this more generic. I don't think this one needs to be a requirement here. ie: I want it so that if "we, the Blockly team" use this, we would create our package under the plugin directory, but that can be a loose requirement, so others can just add it in their own repo. 

The logic I would have expected is, find where the root of the repo is, get the relative path between current directory and the root, and add the packageName to the end of that.

PR #208, opened 2020-04-15
Update @blockly/block-plug-minus to fix image links

samelhusseini:


samelhusseini:
Same as https://github.com/google/blockly-samples/pull/204 but re-based.

PR #216, opened 2020-04-17
Add the theme template and update create script

samelhusseini:
Add a new template for themes.

Update create script to support themes, and use the following logic for naming the folder: 
- If it's a theme, use theme-* where * is the name of the theme.
- If it's a field, use field-* where * is the name of the field.
- If it's a block, use block-* where * is the name of the block.
- If it's a plugin, use * where * is the name of the plugin.

Fix block template by moving template.json to the right location.


plugins/dev-create/templates/theme/template.json
```diff
@@ -0,0 +1,17 @@
...
+    "prepublishOnly": "npm run clean && npm run dist",
+    "start": "blockly-scripts start"
+  },
+  "devDependencies": {
+    "@blockly/dev-tools": "^0.20200313.1"
+  },
+  "peerDependencies": {
+    "blockly": ">=2"
+  }
+}
```
alschmiedt: nit: newline

plugins/dev-create/templates/theme/template.json
```diff
@@ -0,0 +1,17 @@
...
+    "dist": "blockly-scripts build prod",
+    "lint": "blockly-scripts lint",
+    "prepublishOnly": "npm run clean && npm run dist",
+    "start": "blockly-scripts start"
+  },
+  "devDependencies": {
+    "@blockly/dev-tools": "^0.20200313.1"
+  },
+  "peerDependencies": {
+    "blockly": ">=2"
```
alschmiedt: I think we changed the themes api in November so this would have to be > 3.30300123.0

alschmiedt (approved):

plugins/dev-create/templates/theme/template.json
samelhusseini: True, thanks.

PR #217, opened 2020-04-17
Add the Blockly modern theme plugin.

samelhusseini:
Moving the modern theme into a plugin.


plugins/theme-modern/README.md
```diff
@@ -0,0 +1,30 @@
+# @blockly/theme-modern [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
+
+A [Blockly](https://www.npmjs.com/package/blockly) modern theme.
```
alschmiedt: Maybe a bit more description here? " A Blockly theme that has the same classic colors with no highlight" Or something like that.

plugins/theme-modern/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@blockly/theme-modern",
+  "version": "0.20200417.0",
```
alschmiedt: Are you planning on bumping to 1 before publishing?

plugins/theme-modern/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "dist",
+    "src"
+  ],
+  "devDependencies": {
+    "blockly": "^3.20200402.1",
+    "@blockly/dev-scripts": "^0.20200403.0",
+    "@blockly/dev-tools": "^0.20200313.1"
+  },
+  "peerDependencies": {
+    "blockly": ">=2"
```
alschmiedt: same as comment in template review, I think this needs to be a later blockly version.

alschmiedt (approved):

plugins/theme-modern/package.json
samelhusseini: Yeah I wonder if we should have all the plugins start at 1.

plugins/theme-modern/README.md
samelhusseini: Yeah I struggled to come up with a description of the modern theme.
Same classic colours but sets the secondary and tertiary colours of blocks, but would developers really know what that entails?

Dark / light paths are controlled by the renderer, in geras this would still have these paths, we are just controlling the colours of these paths with secondary / tertiary colours.

Thoughts on what this could say?

alschmiedt: You could  just add a [link](https://developers.google.com/blockly/guides/configure/web/themes#block_style) to the description of the secondary and tertiary theme. 

Or you could say something like "A modern theme has the same classic colours, but with a darker border. This is mainly meant for use with thrasos or zelos renderer." Added the last bit since it doesn't look great with geras but up to you whether you want to include.

plugins/theme-modern/package.json
alschmiedt: 1.date.0? Or just 1.0.0?

samelhusseini: I think we standardized on major.date.patch to match google-closure / Blockly.

plugins/theme-modern/package.json
samelhusseini: Fixed.

plugins/theme-modern/README.md
samelhusseini: Updated.

samelhusseini:
@alschmiedt for a final LGTM on README description when you get a chance :)

alschmiedt (approved):

Issue #209, opened 2020-04-15
Plugin restructuring v3000


samelhusseini:
- [x] Move packages around
- [x] Update their package.json
- [x] Update READMEs
- [x] Republish all plugins
- [x] Update templates
- [x] Updating issue templates (Question about what the feature request it is about)
- [x] Move modern theme into a plugin
- [x] Add browser support in package.json
- [ ] Move screenshot to devtools
- [x] Typescript template?

rachel-fenichel:
- [ ] Add plugin usage to framework integration examples

rachel-fenichel:
I'll update issue templates today.

samelhusseini:
https://github.com/google/blockly-samples/pulls and https://github.com/google/blockly-samples/pull/228 for Typescript support.

samelhusseini:
https://github.com/google/blockly-samples/issues/234 tracking eslint changes

rachel-fenichel:
https://github.com/google/blockly-samples/issues/245 tracks the last item, screenshot downloading in dev-tools.

PR #220, opened 2020-04-18
Add browserlist

samelhusseini:
Part of https://github.com/google/blockly-samples/issues/209

Add browserlist to all packages and templates.

Using defaults: https://browserl.ist/?q=defaults

I explicitly added IE 11 and IE_MOB 11, so that we can easily choose to remove support for it by doing:
```
  browserslist: [
    'defaults',
    'not IE 11',
    'not IE_Mob 11',
  ],
```

### DIFF
--- a/plugins/dev-create/bin/create.js
+++ b/plugins/dev-create/bin/create.js
@@ -116,6 +116,11 @@ const packageJson = {
   engines: {
     'node': '>=8.17.0',
   },
+  browserslist: [
+    'defaults',
+    'IE 11',
+    'IE_Mob 11',
+  ],
 };
 
 // Add dev dependencies.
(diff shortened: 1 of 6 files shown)


rachel-fenichel (approved):
Is this information visible on npm?

samelhusseini:
Yes, it's published with the package.

samelhusseini:
If you're asking if the npm UI shows this info, then I don't believe they've integrated with it, here where it's used according to the browserlist README: 
```md
It is used in:

Autoprefixer
Babel
postcss-preset-env
eslint-plugin-compat
stylelint-no-unsupported-browser-features
postcss-normalize
obsolete-webpack-plugin
```

Issue #219, opened 2020-04-18
Bump block-plus-minus Blockly Dependency (post next release)


BeksOmega:
I just remembered that the blockly dependency for the block-plus-minus package will need to be bumped after the next release because it relies on https://github.com/google/blockly/pull/3814 for the procedures to work properly.

Maybe it can be bumped ahead of time? Not sure.

Just wanted to put up an issue to make sure this doesn't get lost!

samelhusseini:
Thanks for pointing that out @BeksOmega. The package has not been published and since it depends on the next release, we’ll wait till that happens before we publish it.

Let’s use this issue to track that.

alschmiedt:
Blockly dependency for block-plus-minus has been set to >=3.20202625.1 in #315

PR #227, opened 2020-04-22
More updating

alschmiedt:
Update all packages. Going to wait to merge until I see changes reflected in npm

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -7,6 +7,9 @@
   "command": {
     "bootstrap": {
       "forceLocal": true
+    },
+    "publish": {
+      "registry": "https://wombat-dressing-room.appspot.com"
     }
   }
 }
--- a/plugins/block-plus-minus/package.json
+++ b/plugins/block-plus-minus/package.json
@@ -38,8 +38,8 @@
         "src"
     ],
     "devDependencies": {
-        "@blockly/dev-scripts": "^0.20200403.0",
-        "@blockly/dev-tools": "^0.20200313.1",
+        "@blockly/dev-scripts": "^0.20200403.1",
+        "@blockly/dev-tools": "^0.20200313.2",
         "blockly": "^3.20200123.1",
         "chai": "^4.2.0",
         "mocha": "^7.1.0",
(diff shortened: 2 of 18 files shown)


lerna.json
```diff
@@ -7,6 +7,9 @@
   "command": {
     "bootstrap": {
       "forceLocal": true
+    },
+    "publish": {
+      "registry": "https://wombat-dressing-room.appspot.com"
```
samelhusseini, line 12: Does this mean we don't have to include this in every package.json?

plugins/block-plus-minus/package.json
```diff
@@ -38,8 +38,8 @@
         "src"
     ],
     "devDependencies": {
-        "@blockly/dev-scripts": "^0.20200403.0",
-        "@blockly/dev-tools": "^0.20200313.1",
+        "@blockly/dev-scripts": "^0.20200403.1",
```
samelhusseini, line 41: FYI, we should be holding off on publishing this one. due to https://github.com/google/blockly-samples/issues/219

plugins/theme-modern/package.json
```diff
@@ -38,9 +38,9 @@
     "src"
   ],
   "devDependencies": {
-    "blockly": "^3.20200402.1",
-    "@blockly/dev-scripts": "^0.20200403.0",
-    "@blockly/dev-tools": "^0.20200313.1"
+    "@blockly/dev-scripts": "^0.20200403.1",
+    "@blockly/dev-tools": "^0.20200313.2",
+    "blockly": "^3.20200402.1"
```
samelhusseini, line 43: We should update the template script to sort these automatically.

samelhusseini (approved):

lerna.json
alschmiedt: I am not sure. This needed to be added so that I could log in via the command line.

plugins/block-plus-minus/package.json
alschmiedt: Unfortunately, too late. This gets generated as part of the publish script.

PR #228, opened 2020-04-22
Typescript support in build and lint

samelhusseini:
- Lint as part of the build (prod and dev)
- Use babel for TS compilation
- Run TS type checker for TS directories
- Use eslint for TS linting

Will update templates in a separate PR.


samelhusseini:
I've had to fix linting and tests in block-plus-minus for the build to work, as we're enabling linting on all packages. 
see https://github.com/google/blockly-samples/pull/228/commits/51fca14f6dac80647d6b330a94486dc55d486960

alschmiedt (approved):

PR #238, opened 2020-04-27
Eslint for plugins

### LINKED ISSUE #234: Eslint for plugins
**Category**
Plugins.

**Component**

@blockly/eslint-config

**Describe the bug**

Eslint is peculiar about how it likes its config and where it looks for eslint plugins (eslint-config-google, @typescript-eslint, etc).

We want to have third party plugins using ``@blockly/dev-scripts`` to get our eslint configs so ``npm run lint`` and editor linting would work. That leads us to creating our own eslint sharable plugin (@blockly/eslint-config).

It gets a little more tricky because we (for first party plugins), use lerna, which doesn't install package "dependencies" in a plugin that depends on a locally linked package. It only installs a symlink to that package, so adding @blockly/eslint-config isn't sufficient for editor eslint to work.

Options: 
- Create the eslint-config outside of the plugins folder and have it published separately.
That would have lerna plugins and third party plugins using the same eslint configuration.

- Use a different eslint setup for first party vs third party plugins, and keep all eslint plugins installed in blockly-samples's package root.

samelhusseini:
Resolves https://github.com/google/blockly-samples/issues/234

Supports both first party and third party plugins.

- Create a sharable eslint config for Blockly plugins ``@blockly/eslint-config``.
- Remove eslintrc config.
- Add ``eslintConfig`` into package.json so code editors pick up on the config
- Update root package.json to use the same eslint config from npm.

Tested that ``npm run start``, ``npm run lint``, and eslint editors work for both first party plugins in blockly-samples, and third party plugins created in a custom repo using ``npx @blockly/create-package``


plugins/eslint-config/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "typescript": {
+      "optional": true
+    }
+  },
+  "publishConfig": {
+    "access": "public",
+    "registry": "https://wombat-dressing-room.appspot.com"
+  },
+  "engines": {
+    "node": ">=8.0.0"
```
alschmiedt, line 47: Can this be removed?

alschmiedt (approved):

plugins/dev-create/bin/create.js
```diff
@@ -59,7 +59,7 @@ const pluginType = program.type || 'plugin';
 const pluginDir = program.dir ||
     (pluginType == 'plugin' ? pluginName : `${pluginType}-${pluginName}`);
 const pluginPath = path.join(root, pluginDir);
-const pluginAuthor = program.author || 'Blockly Team';
+const pluginAuthor = program.author || '';
```
alschmiedt: Could we have this still be Blockly team if it is a first party plugin?

plugins/eslint-config/package.json
samelhusseini: Possibly, do you mean specifically for this package or for all packages?
I hadn't done much digging into whether this is necessary, I just assumed so as it was in the original workspace-search package.json

plugins/dev-create/bin/create.js
samelhusseini: Good point. Yes, will fix.

samelhusseini: done.

plugins/eslint-config/package.json
alschmiedt: I meant specifically for this package, but thinking about it more it might be good to revisit whether it should be in the template. We had this in workspace search since we planned on supporting node for that package and it was recommended to have a baseline version.

samelhusseini: Yeah, there's a chance this isn't relevant, but I want to keep our package.jsons consistent (incase someone copies them as an example), so if the template has it (which it currently does), then lets keep it there.

PR #251, opened 2020-05-01
Add field date tests

samelhusseini:
Move field date tests from Blockly.
https://github.com/google/blockly/blob/master/tests/mocha/field_date_test.js

Run tests using mocha, custom script as field-date doesn't using ``blockly-scripts``.
Update build script to use externs instead of the goog.provide hack.

### DIFF
--- a/plugins/field-date/package.json
+++ b/plugins/field-date/package.json
@@ -7,7 +7,7 @@
         "clean": "rm -rf $npm_package_directories_dist",
         "prepublishOnly": "npm run build",
         "test": "npm run build && npm run test:mocha",
-        "test:mocha": "mocha --ui tdd test/test.mocha.js"
+        "test:mocha": "mocha --ui tdd test/field_date_test.mocha.js"
     },
     "main": "dist/date_compressed.js",
     "author": "Blockly Team",
--- a/plugins/field-date/package.json
+++ b/plugins/field-date/package.json
@@ -5,7 +5,9 @@
     "scripts": {
         "build": "gulp build",
         "clean": "rm -rf $npm_package_directories_dist",
-        "prepublishOnly": "npm run build"
+        "prepublishOnly": "npm run build",
+        "test": "npm run build && npm run test:mocha",
+        "test:mocha": "mocha --ui tdd test/test.mocha.js"
     },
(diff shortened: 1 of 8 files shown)


plugins/field-date/package.json
```diff
@@ -5,7 +5,9 @@
     "scripts": {
         "build": "gulp build",
         "clean": "rm -rf $npm_package_directories_dist",
-        "prepublishOnly": "npm run build"
+        "prepublishOnly": "npm run build",
+        "test": "npm run build && npm run test:mocha",
+        "test:mocha": "mocha --ui tdd test/test.mocha.js"
```
moniika: Should this be using the blockly-scripts?

plugins/field-date/test/test.mocha.js
```diff
@@ -0,0 +1,159 @@
+/**
```
moniika: Other tests in samples like workspace search and slider are using the naming convention for test to be:
field_date_test.mocha.js
I'm not sure if just test or the longer name is better, could update the other ones. I'd just like to be consistent

plugins/field-date/test/test.mocha.js
```diff
@@ -0,0 +1,159 @@
...
+ * @license
+ * Copyright 2019 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const assert = require('assert');
+require('blockly');
```
moniika: Why is this needed? and if it is, then maybe field_slider should have it too?

plugins/field-date/package.json
samelhusseini: Oh yes.

samelhusseini: No, as blockly-scripts is very much catered for the standard build environment. Field date uses closure to compile not webpack so it's unique.

plugins/field-date/test/test.mocha.js
samelhusseini: Sure, happy to switch.

plugins/field-date/test/test.mocha.js
samelhusseini: Not needed, removed.

Issue #252, opened 2020-05-01
Field slider should inherit parent block if shadow


samelhusseini:
**Category**

Plugins

**Component**

field-slider

**Describe the bug**

The slider field should inherit the parent block colour if it's a shadow block. 

**To Reproduce**
<img width="217" alt="Screen Shot 2020-05-01 at 11 26 59 AM" src="https://user-images.githubusercontent.com/16690124/80830994-d6058100-8b9e-11ea-9a4f-e90e9baec798.png">

**Expected behavior**

The slider colour should be the same as the parent. 
Order of precedence should favour the parent block's color if it's a shadow block.

**Screenshots**

See above

**Additional context**

To work on this issue, first read the [background information](https://developers.google.com/blockly/guides/contribute/samples) about how to contribute to samples and verify your changes.

The slider field sets its color in the `showEditor_` function. Currently, it only checks the source block's color, but if the source block is a shadow block, this fails as shown above. To fix this, do the following steps:

1. Run `npm run start` in the field-slider plugin directory to start the playground.
2. In the right-hand configuration settings panel.
3. Drag one of the "parent" blocks out of the toolbox and click the number to open the slider field. Note how the colors are inconsistent.
4. Change the slider field's `showEditor_` function to take into account if the block is a shadow block. You can check if it is with `this.sourceBlock_.isShadow()`. If it is a shadow block, then instead of setting the color to the source block's color, set it to the parent block's color. You can get the parent's block color with `this.sourceBlock_.getParent().getColour()`.
5. Test your changes by repeating step 3 and ensuring the colors match now.

For more information about shadow blocks, see [this page](https://developers.google.com/blockly/guides/configure/web/toolbox?hl=en#shadow_blocks)

BeksOmega:
I'm not sure that we actually want to do this, because it only applies to zelos. The dropdown field [handles this by using a special constant](https://github.com/google/blockly/blob/da8e486995df8f0b7b08a086ffb1b4bd22b21963/core/field_dropdown.js#L288) in the [zelos renderer sets to true](https://github.com/google/blockly/blob/da8e486995df8f0b7b08a086ffb1b4bd22b21963/core/renderers/zelos/constants.js#L292). We could use the same flag to trigger the functionality in the slider field, but it sucks that the constant is prefixed with `FIELD_DROPDOWN`, and not something more generic.

maribethb:
[...]

PR #253, opened 2020-05-04
add inherit background color for field slider

michiboo:
Address issue #252


samelhusseini:
Thanks for your PR @michiboo, this isn't quite what I meant by that issue. The dropdown div color is set when the field is shown https://github.com/google/blockly-samples/blob/master/plugins/field-slider/src/field_slider.js#L90. If it's a single field block, we would want it to inherit the color of the parent block.

PR #258, opened 2020-05-05
Less noisy build

samelhusseini:
Webpack messages are too noisy, and aren't very helpful. 

Adding hooks to streamline webpack messages to a simple: 
- Successful build
- Successful with warnings / errors
- Build failure

eslint, webpack and Typechecker messages are outputted.

### DIFF
--- a/plugins/dev-scripts/scripts/start.js
+++ b/plugins/dev-scripts/scripts/start.js
@@ -18,16 +18,27 @@
 
 const path = require('path');
 const fs = require('fs');
+const os = require('os');
 
 const webpack = require('webpack');
 const WebpackDevServer = require('webpack-dev-server');
 const webpackConfig = require('../config/webpack.config');
 const webpackDevServerConfig = require('../config/webpackDevServer.config');
 
+const chalk = require('chalk');
+const codeFrame = require('@babel/code-frame').codeFrameColumns;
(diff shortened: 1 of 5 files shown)


plugins/dev-scripts/scripts/start.js
```diff
@@ -18,24 +18,160 @@
...
+const chalk = require('chalk');
+const codeFrame = require('@babel/code-frame').codeFrameColumns;
+const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
+
 const appDirectory = fs.realpathSync(process.cwd());
 const resolveApp = (relativePath) => path.resolve(appDirectory, relativePath);
 
+const clearConsole = () => process.stdout.write(
+    process.platform === 'win32' ? '\x1B[2J\x1B[0f' : '\x1B[2J\x1B[3J\x1B[H'
+);
```
alschmiedt: Usually when I clear my console it will continue to let me scroll up to past output. This looks like it completely removes all past output, which is pretty frustrating. Could we either remove this part? Or change it so it doesn't remove all past output?

alschmiedt (approved):

plugins/dev-scripts/scripts/start.js
samelhusseini: Okay removing..

PR #260, opened 2020-05-05
Add field test blocks generator, and update field-slider to use it.

samelhusseini:
Add a dev tool to generate field test blocks, for each group, generate:
- A single block field
- A parent block that adds the single block field as a shadow
- A block with the field on it
- A parent block that adds that block with the field on it as a shadow

Screen Shot 2020-05-05 at 9 59 45 AM

### DIFF
--- a/plugins/dev-tools/src/generateFieldTestBlocks.js
+++ b/plugins/dev-tools/src/generateFieldTestBlocks.js
@@ -39,9 +39,10 @@ export default function generateFieldTestBlocks(fieldName, options) {
     }
 
     // Define a single field block.
+    const singleFieldBlock = `${++id}test_${fieldName}_single`;
     blocks.push(
         {
-          'type': `${++id}test_${fieldName}_single`,
+          'type': singleFieldBlock,
           'message0': '%1',
           'args0': [
             {
@@ -57,14 +58,15 @@ export default function generateFieldTestBlocks(fieldName, options) {
           'output': null,
           'style': 'math_blocks',
         });
-    toolboxXml += `<block type="${id}test_${fieldName}_single"></block>`;
+    toolboxXml += `<block type="${singleFieldBlock}"></block>`;
     toolboxXml += `<sep gap="10"></sep>`;
 
     // Define a block and add the 'single field block' as a shadow.
+    const parentBlock = `${++id}test_${fieldName}_parent`;
     blocks.push(
         {
-          'type': `${++id}test_${fieldName}_parent`,
-          'message0': 'in parent %1',
+          'type': `${parentBlock}`,
+          'message0': 'parent %1',
           'args0': [
             {
               'type': 'input_value',
@@ -85,10 +87,11 @@ export default function generateFieldTestBlocks(fieldName, options) {
     toolboxXml += `<sep gap="10"></sep>`;
 
     // Define a block with the field on it.
+    const blockWithField = `${++id}test_${fieldName}_block`;
     blocks.push(
         {
-          'type': `${++id}test_${fieldName}_block`,
-          'message0': 'on block %1',
+          'type': blockWithField,
+          'message0': 'block %1',
           'args0': [
             {
               'type': fieldName,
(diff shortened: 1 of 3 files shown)


plugins/dev-tools/src/generateFieldTestBlocks.js
```diff
@@ -0,0 +1,126 @@
...
+ */
+
+import * as Blockly from 'blockly/core';
+
+
+/**
+ * Generates a number of field testing blocks for a specific field and returns
+ * the toolbox xml string.
+ * @param {string} fieldName The name of the field.
+ * @param {Object|Array.<Object>=} options An options object containing a label
```
moniika: Based on [this](https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System) the dot syntax ".<>" seems to be deprecated syntax

plugins/dev-tools/src/generateFieldTestBlocks.js
```diff
@@ -0,0 +1,126 @@
...
+ *     and an args map that is passed to the field during initialization.  If an
+ *     array is passed, multiple groups of blocks are created each with
+ *     different initialization arguments.
+ * @return {string} The toolbox XML string.
+ */
+export default function generateFieldTestBlocks(fieldName, options) {
```
moniika: I thought we wanted to avoid export default according to style guide.

plugins/dev-tools/src/generateFieldTestBlocks.js
```diff
@@ -0,0 +1,126 @@
...
+    const singleFieldBlock = `${++id}test_${fieldName}_single`;
+    blocks.push(
+        {
+          'type': singleFieldBlock,
+          'message0': '%1',
+          'args0': [
+            {
+              'type': fieldName,
+              'name': 'FIELDNAME',
+              ...m.args,
```
moniika, line 51: I'm curious what does  ".." do?

plugins/dev-tools/src/generateFieldTestBlocks.js
```diff
@@ -0,0 +1,126 @@
...
+export default function generateFieldTestBlocks(fieldName, options) {
+  if (!Array.isArray(options)) {
+    options = [options || {}];
+  }
+
+  let id = 0;
+  let toolboxXml = '';
+  const blocks = [];
+
+  options.forEach((m) => {
```
moniika, line 35: My first impression was that it might be easier to read with a different variable name, but nothing in particular came to mind as a better option.

plugins/dev-tools/src/generateFieldTestBlocks.js
samelhusseini: Yes, good point we should move away from that syntax in this repo.

plugins/dev-tools/src/generateFieldTestBlocks.js
samelhusseini: Good point.

plugins/dev-tools/src/generateFieldTestBlocks.js
samelhusseini: This is known as a spread. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

Basically it copies over all properties from m.args and adds them to that object.

plugins/dev-tools/src/generateFieldTestBlocks.js
samelhusseini: For options?, I agree, I ended up settling for that. but yeah open for suggestions.

plugins/dev-tools/src/generateFieldTestBlocks.js
samelhusseini: Done.

rachel-fenichel: +1 to avoiding export default.

PR #261, opened 2020-05-05
Refactoring field tests.

moniika:
Created field test template and test helpers.
Refactored field-slider and field-date tests to fit template.


plugins/field-slider/test/field_slider_test_helpers.js
```diff
@@ -0,0 +1,91 @@
...
+// Monica's note
+// These were moved out to clean up the test file to make it easier to read.
+
+/**
+ * Assert a field's value is the same as the expected value.
+ * @param {Field} field The field.
+ * @param {*} expectedValue The expected value.
+ * @param {string=} opt_expectedText The expected text.
+ */
+function assertFieldValue(field, expectedValue, opt_expectedText) {
```
moniika: This is a function I would expect would be common between all field tests.

plugins/field-slider/test/field_slider_test_helpers.js
```diff
@@ -0,0 +1,91 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const assert = require('chai').assert;
+const FieldSlider = require('../dist/index').FieldSlider;
+
+// Monica's note
```
samelhusseini: Suggestion: add your notes behind a TODO, that way eslint will remind you they're there. eg: 
```
// TODO (Monica): ... 
```

plugins/field-slider/test/field_slider_test_helpers.js
```diff
@@ -0,0 +1,91 @@
+/**
```
samelhusseini, line 1: Is the idea that these helpers would end up in dev-tools eventually? If so, I think the name of all these methods needs to be more generic.

plugins/field-slider/test/field_slider_test_helpers.js
```diff
@@ -0,0 +1,91 @@
...
+function assertSliderFieldDefault(sliderField) {
+  assertSliderField(sliderField, -Infinity, Infinity, 0, 0);
+}
+/**
+ * Create a simple slider field with constructor values that are all the same.
+ * @param {number} value The constructor value.
+ * @return {FieldSlider} A new field slider.
+ */
+function createSliderFieldSameValuesConstructor(value) {
+  return new FieldSlider(value, value, value, value);
```
samelhusseini: This scenario seems oddly specific for slider fields, ie: why would any other field need to do this?

plugins/field-slider/test/field_slider_test_v2.mocha.js
```diff
@@ -0,0 +1,119 @@
...
+            helpers.assertFieldValue(this.field, run.expectedValue);
+          });
+        }
+      });
+    });
+    suite('Vale -> New Value', function() {
+      setup(function() {
+        this.field = new FieldSlider(1);
+      });
+      runs.forEach(function(run) {
```
samelhusseini: A helper method that abstracts all this away into a simple ``runTests`` could help with readability.
eg: 
```
runTests(() => {
   ... 
});
```

plugins/field-slider/test/field_slider_test_v2.mocha.js
```diff
@@ -0,0 +1,119 @@
...
+        if (run.expectDefault) {
+          test(run.title, function() {
+            this.field.setValue(run.value);
+            // Expect value to be unchanged.
+            helpers.assertFieldValue(this.field, 1);
+          });
+        } else {
```
samelhusseini: Maybe these should be split into different test types, the ones that expect default and the ones that don't, split into two separate calls to a run helper.
```
runTestsThatExpectDefaults(() => {
 ... // What to do for each test.
});

runTestThatDontExpectDefaults(() => {
 /// other
});
```

(obviously with better names)

plugins/field-slider/test/field_slider_test_helpers.js
moniika: Only the first one I was initially intending to be in dev-tools, the others are currently too specific to slider/fieldnumber.

plugins/field-slider/test/field_slider_test_helpers.js
moniika: Thanks. I'm removing a lot of them for now, but I'll probably use the TODO for any future comments while I'm working.

plugins/field-slider/test/field_slider_test_helpers.js
moniika: This is specific to slider. To clarify my comment, I refactored out the slider field specific helper methods to make the test easier to read as well. I was thinking each field may need their own implementation of certain helper methods.

plugins/field-slider/test/field_slider_test_v2.mocha.js
moniika: I've split back into 2 arrays for valid and invalid values and done a larger refactor to generalize.

plugins/field-slider/test/field_slider_test_helpers.js
moniika: To keep track of things easier, I've made a field_slider_helpers and a field_helpers now

plugins/field-slider/test/field_slider_test_old.js
```diff
@@ -0,0 +1,534 @@
+/**
```
moniika: This is here only for comparison, it will be deleted.

plugins/field-slider/test/field_slider_test.mocha.js
```diff
@@ -4,531 +4,139 @@
...
-      assertSliderFieldSameValues(sliderField, -Infinity);
-    });
+suite('FieldTemplate', function() {
+  const invalidValueRuns = [
+    {title: 'Undefined', value: undefined},
+    {title: 'Null', value: null},
+    {title: 'NaN', value: NaN},
+    {title: 'Non-Parsable String', value: 'bad'},
+  ];
+  const validValueRuns = [
```
samelhusseini, line 32: Nice, defining test cases in this way is great!
More re-usable and readable!

plugins/field-slider/test/field_test_helpers.js
```diff
@@ -0,0 +1,87 @@
...
+      const field = new TestedField();
+      assertFieldDefault(field);
+    });
+    invalidValueRuns.forEach(function(run) {
+      test(run.title, function() {
+        const field = new TestedField(...run.args);
+        assertFieldDefault(field);
+      });
+    });
+    validValueRuns.forEach(function(run) {
```
samelhusseini: The main difference I see between the invalid runs and the valid runs is the assertion method used, can that perhaps be passed in? 

The main difference I see between creation via constructor vs via json is the construction method. 
Can that perhaps also be included in the object definition of these tests, eg: 

What if you had something like this
```
var createWithJS = (run) => new FieldSlider(...run.args);
var createWithJson = (run) => FieldSlider.fromJson(run.json);

var assertInvalid = (field, run) => assertFieldDefault(field);
var assertValid = (field, run) => assertField(field, run.expectedValue);

var tests = {
  {title: 'Null', value: null, creation: createWithJS, assertion: assertInvalid},
  {title: 'Float', value: 1.5, expectedValue: 1.5, creation: createWithJS, assertion: assertValid},
};

runAllTests(tests);

function runAllTests(tests) {
   suite('RunAllTests', function() {
      tests.forEach(function(run) {
          test(run.title, function() {
              const field = run.creation(run);
              run.assertion(field, run);
          });
      });
   });
}

```

samelhusseini (commented):
Overall direction looks good to me, we're essentially creating a simple way to define our tests and abstracting away all the duplicate code, great!

plugins/field-slider/package.json
```diff
@@ -41,6 +41,8 @@
     "@blockly/dev-scripts": "^0.20200403.6",
     "@blockly/dev-tools": "^0.20200313.7",
     "blockly": "^3.20200123.1",
+    "chai": "^4.2.0",
+    "mocha": "^7.1.2",
```
samelhusseini: There should be no need to add mocha here, as we're using blockly-scripts tests to run the mocha tests. Are you seeing something different?

plugins/field-slider/test/field_slider_test_helpers.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const assert = require('chai').assert;
+const assertFieldValue = require('../test/field_test_helpers').assertFieldValue;
```
samelhusseini: It might be cleaner, and useful if you add more imports here from the same file to use object destructuring instead. ie: 
```
const {assertFieldValue} = require('../test/field_test_helpers')
```

plugins/field-slider/test/field_test_helpers.js
```diff
@@ -0,0 +1,138 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const assert = require('chai').assert;
+const Blockly = require('blockly');
```
samelhusseini: Did we figure out whether we need this?

plugins/field-slider/test/field_test_helpers.js
```diff
@@ -0,0 +1,138 @@
...
+
+
+/**
+ * @typedef {{
```
samelhusseini, line 11: Nice, add comment for what this is a typedef for.

plugins/field-slider/test/field_test_helpers.js
```diff
@@ -0,0 +1,138 @@
...
+  assert.equal(actualValue, expectedValue, 'Value');
+  assert.equal(actualText, opt_expectedText, 'Text');
+}
+
+/**
+ * Runs provided creation test cases.
+ * @param {Array<Run>} runs The test cases to run.
+ * @param {function(T, Run)} assertion The assertion to use.
+ * @param {function(new:T,Run):T} creation The creation method to use.
+ * @template {!Blockly.Field} T
```
samelhusseini: Any reason for the template here? All fields inherit from Blockly.Field, and simply ``function(Blockly.Field, Run)`` should work.

samelhusseini (approved):
A few more comments to resolve, but otherwise lgtm.

plugins/field-slider/test/field_test_helpers.js
moniika: I actually added this to make my IDE happy for lint reasons, but it works without it (and also there's other lint things my IDE is complaining about). So, I'm going to remove.

plugins/field-slider/test/field_test_helpers.js
moniika: Done.

plugins/field-slider/test/field_test_helpers.js
moniika: I was trying to get fancy and say that the type used is a subtype of Field.
It seemed to be more correct to use generics as I could use the same type in all the params.
I'm fine with changing back to Blockly.Field as I'm having some trouble recalling any specific benefits to this.

plugins/field-slider/package.json
moniika: I think it should work without mocha here because blockly-scripts tests pulls in that dependency for us.

plugins/field-slider/test/field_slider_test_helpers.js
moniika: I didn't realize that syntax was valid. I'll update.

PR #271, opened 2020-05-12
Refactoring shared test helpers into dev tools.

moniika:


plugins/dev-tools/package.json
```diff
@@ -38,7 +38,9 @@
   ],
   "dependencies": {
     "@types/dat.gui": "^0.7.5",
-    "dat.gui": "^0.7.7"
+    "chai": "^4.2.0",
+    "dat.gui": "^0.7.7",
+    "mocha": "^7.1.0"
```
samelhusseini, line 43: Mocha is already a dependency of dev-scripts, so I don't see why it needs to be a dependency here.

samelhusseini: Mocha is already a dependency of `dev-script`, unclear why it needs to be a dependency here?

plugins/dev-tools/src/index.js
```diff
@@ -6,14 +6,23 @@
 
 import toolboxCategories from './toolboxCategories';
 import toolboxSimple from './toolboxSimple';
-import addGUIControls from './addGUIControls';
+let addGUIControls;
+if (typeof window !== 'undefined') {
+  addGUIControls = require('./addGUIControls').default;
+}
+import {assertFieldValue, runConstructorSuiteTests, runFromJsonSuiteTests,
```
samelhusseini: Can we just export all of these under a single namespace? eg: 
```js
import * as testHelpers from './field_test_helpers.mocha';

export {
 ....
 testHelpers,
 ....
}
```
and then in the test code: 
```js
import {testHelpers} from '@blockly/dev-tools';
const {assertFieldValue, runConstructorSuiteTests, runFromJsonSuiteTests, runSetValueTests} = testHelpers;

```

Just in an effort of more grouping, and reduction of clutter.

plugins/field-slider/test/field_slider_test.mocha.js
```diff
@@ -6,11 +6,10 @@
 
 const {assert} = require('chai');
 const sinon = require('sinon');
-const {assertSliderField,
-  assertSliderFieldDefault} = require('../test/field_slider_test_helpers');
+const {assertSliderField,assertSliderFieldDefault} = require(
```
samelhusseini: nite: space between asserFieldSlider and assertSliderFieldDefault

plugins/dev-tools/package.json
moniika: It seems to build fine without this mocha dependency. I thought this would be needed because the test helpers were compiled and built, as opposed to just being run by dev-script.

plugins/field-slider/test/field_slider_test.mocha.js
moniika: Done

plugins/dev-tools/src/index.js
moniika: Updated in dev-tools. I've updated the requires in the test code as well to use the exported testHelpers (since we don't currently support es6 modules for the test code).

samelhusseini (approved):

PR #272, opened 2020-05-12
Populate random

rachel-fenichel:
- Add two functions to dev-tools: `populateRandom` and `populateRandomButton`.
- Document in the README.
- Use in the blocks-plus-minus test page.
- Rename an input in blocks-plus-minus so it won't conflict with another id on the page

Points for review:
- This is probably not the best way to add a button
- (Edit: fixed) _In blocks-plus-minus, this works for the first time the workspace is injected, but fails when the workspace is changed through the gui controls. (The import/export code also breaks in that case.)_

Related to https://github.com/google/blockly/issues/3882

### DIFF
--- a/plugins/dev-tools/README.md
+++ b/plugins/dev-tools/README.md
@@ -56,15 +56,5 @@ import {populateRandom} from '@blockly/dev-tools';
 populateRandom(workspace, 10);
 ```
 
-The `populateRandomButton` function adds a button to the page with `populateRandom` as its onClick function.
-The first parameter is a function that returns the workspace blocks should be added to.
-The second parameter says where to add the button on the document.
-The third parameter is how many blocks to add.
-```js
-import {populateRandomButton} from '@blockly/dev-tools';
-
-populateRandomButton(Blockly.getMainWorkspace, document.getElementById('myElement'), 10);
-
-````
 ## License
 Apache 2.0
--- a/plugins/dev-tools/README.md
+++ b/plugins/dev-tools/README.md
@@ -47,5 +47,20 @@ addGUIControls((options) => {
 }, defaultOptions);
 ```
 
+### Populate Random
+
+The `populateRandom` function adds random blocks to a workspace. Blocks are selected from the full set of defined blocks. Pass in a worskpace and how many blocks should be created.
+```js
+import {populateRandom} from '@blockly/dev-tools';
+// Add 10 random blocks to the workspace.
+populateRandom(workspace, 10);
+```
+
+The `populateRandomButton` function adds a button to the page with `populateRandom` as its onClick function. The button is appended to the document after the element passed as the second parameter.
+```js
+import {populateRandomButton} from '@blockly/dev-tools';
+populateRandomButton(workspace, document.getElementById('myElement'), 10);
+
+````
 ## License
 Apache 2.0
--- a/plugins/dev-tools/README.md
+++ b/plugins/dev-tools/README.md
@@ -56,10 +56,14 @@ import {populateRandom} from '@blockly/dev-tools';
 populateRandom(workspace, 10);
 ```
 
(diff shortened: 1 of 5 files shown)


plugins/dev-tools/README.md
```diff
@@ -47,5 +47,24 @@ addGUIControls((options) => {
...
+### Populate Random
+
+The `populateRandom` function adds random blocks to a workspace. Blocks are selected from the full set of defined blocks. Pass in a worskpace and how many blocks should be created.
+```js
+import {populateRandom} from '@blockly/dev-tools';
+// Add 10 random blocks to the workspace.
+populateRandom(workspace, 10);
+```
+
+The `populateRandomButton` function adds a button to the page with `populateRandom` as its onClick function.
```
samelhusseini, line 59: This sounds good for now, although as the playground comes alive it will have functionality to add a button and this will become redundant.

plugins/dev-tools/README.md
```diff
@@ -47,5 +47,24 @@ addGUIControls((options) => {
 }, defaultOptions);
 ```
 
+### Populate Random
+
+The `populateRandom` function adds random blocks to a workspace. Blocks are selected from the full set of defined blocks. Pass in a worskpace and how many blocks should be created.
```
samelhusseini, line 52: The playground might also want to have a button that does this by default. Although not against still exporting it (similar to what we do with the DebugRenderer)

plugins/dev-tools/src/populateRandom.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Populate the workspace with a random set of blocks, for testing.
+ * @param {Blockly.Workspace} workspace The workspace to populate.
+ * @param {number} count How many blocks to create.
+ */
+export function populateRandom(workspace, count) {
+  const names = [];
+  for (const blockName in Blockly.Blocks) {
+    if (Object.prototype.hasOwnProperty.call(Blockly.Blocks, blockName)) {
```
samelhusseini, line 21: Curious why we're calling ``hasOwnProperty`` like this?  Why not:
```
Object.hasOwnProperty(...)
```

samelhusseini, line 21: For in + hasOwnProperty is a good pattern, so +1. However, why are we calling hasOwnProperty like this? ie: why not:
```
Object.hasOwnProperty(...)
```

plugins/dev-tools/README.md
samelhusseini: This is okay for now, but just keep in mind this will become redundant once the playground has support for adding buttons.

plugins/dev-tools/README.md
samelhusseini, line 52: Similar to other comment, we would want this to be a default button in the playground, but I'm not opposed to exporting it anyway. (as it could be used in a different setting)

samelhusseini (approved):
LGTM

plugins/dev-tools/README.md
rachel-fenichel: The idea is to export it for use outside of our playgrounds. However, I don't want to add the button function and then remove it immediately, so I'm just going to take it out of this PR. It already served its purpose, which was in testing.

plugins/dev-tools/src/populateRandom.js
rachel-fenichel: hasOwnProperty is forced by this: https://eslint.org/docs/rules/guard-for-in
This style of call is forced by https://eslint.org/docs/rules/no-prototype-builtins

We could also just turn off the latter rule. Thoughts?

plugins/dev-tools/README.md
samelhusseini: Makes sense.

plugins/dev-tools/src/populateRandom.js
samelhusseini: Ah, yeah we should fix that on the lint side, let me look into that.

PR #276, opened 2020-05-15
Adding ability to skip tests.

### LINKED ISSUE #269: Add ability to skip test cases to test helpers
**Category**
Plugins

**Component**
Tests

The same run configurations is often used between the contructor/fromJson/setValue tests. The newly created test helpers take run configuration options, but there isn't currently a way to easily skip (and reference TODO) for individual test cases in the event that there's a bug that is under process of being fixed.
In the test helpers there isn't currently a way to skip test cases There is

A way to mark a run option to be skipped for certain test types (like constructor or fromJson only).

Commenting out test cases, but since the run configuration is shared between the constructor/fromJson/setValue tests commenting out the run skips it for all the test, when I might want to just skip for constructor tests, for example.

**Additional context**

Field test refactor in #261 is what introduces the new field helpers that don't yet handle skipping test cases.

moniika:
- Refactor of test helpers, creating common_test_helpers for non-field specific test helpers and test_helpers as a common entry point for test helpers
- Added a method to run test cases with logic to skip or only the test call based on passed run information

Fixes #269

### DIFF
--- a/plugins/dev-tools/src/common_test_helpers.mocha.js
+++ b/plugins/dev-tools/src/common_test_helpers.mocha.js
@@ -25,7 +25,7 @@ export let TestCase;
 export function runTestCases(testCases, testFn) {
   testCases.forEach((testCase) => {
     let testCall = (testCase.skip ? test.skip : test);
-    testCall = (testCase.only ? test.skip : testCall);
+    testCall = (testCase.only ? test.only : testCall);
     testCall(testCase.title, testFn(testCase));
   });
 }
--- a/plugins/dev-tools/src/common_test_helpers.mocha.js
+++ b/plugins/dev-tools/src/common_test_helpers.mocha.js
@@ -5,7 +5,7 @@
  */
 
 /**
- * Run configuration information.
+ * Test case configuration information.
  * @typedef {{
  *            title:string,
  *            value:*
@@ -14,18 +14,18 @@
  *            expectedValue:?
  *          }}
  */
-let Run;
+export let TestCase;
 
 /**
  * Runs provided test cases.
- * @param {Array<Run>} runs The test cases to run.
- * @param {function(Run):function} testFn Function that returns test callback.
- * @private
+ * @param {Array<TestCase>} testCases The test cases to run.
+ * @param {function(TestCase):function} testFn Function that returns test
+ *  callback.
  */
-export function runTestCases(runs, testFn) {
-  runs.forEach((run) => {
-    let testCall = (run.skip ? test.skip : test);
-    testCall = (run.only ? test.skip : testCall);
-    testCall(run.title, testFn(run));
+export function runTestCases(testCases, testFn) {
+  testCases.forEach((testCase) => {
+    let testCall = (testCase.skip ? test.skip : test);
+    testCall = (testCase.only ? test.skip : testCall);
+    testCall(testCase.title, testFn(testCase));
   });
 }
--- a/plugins/dev-tools/src/common_test_helpers.mocha.js
+++ b/plugins/dev-tools/src/common_test_helpers.mocha.js
@@ -1,0 +1,31 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Run configuration information.
+ * @typedef {{
+ *            title:string,
+ *            value:*
+ *            skip:(boolean|undefined),
+ *            only:(boolean|undefined),
+ *            expectedValue:?
+ *          }}
+ */
+let Run;
+
+/**
+ * Runs provided test cases.
+ * @param {Array<Run>} runs The test cases to run.
+ * @param {function(Run):function} testFn Function that returns test callback.
+ * @private
+ */
+export function runTestCases(runs, testFn) {
+  runs.forEach((run) => {
+    let testCall = (run.skip ? test.skip : test);
+    testCall = (run.only ? test.skip : testCall);
+    testCall(run.title, testFn(run));
+  });
+}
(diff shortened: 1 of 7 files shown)


plugins/field-slider/test/field_slider_test.mocha.js
```diff
@@ -59,13 +58,23 @@ suite('FieldSlider', function() {
...
+
   runConstructorSuiteTests(
       FieldSlider, validValueRuns, invalidValueRuns, validRunAssertField,
       assertSliderFieldDefault);
 
   runFromJsonSuiteTests(FieldSlider, validValueRuns, invalidValueRuns,
       validRunAssertField, assertSliderFieldDefault);
 
+
+  // TODO(https://github.com/google/blockly/issues/3903): Remove skip=false
```
rachel-fenichel, line 74: Why do you have to reset `skip`? Is the object being used again?

plugins/dev-tools/src/common_test_helpers.mocha.js
```diff
@@ -0,0 +1,31 @@
...
+ * Run configuration information.
+ * @typedef {{
+ *            title:string,
+ *            value:*
+ *            skip:(boolean|undefined),
+ *            only:(boolean|undefined),
+ *            expectedValue:?
+ *          }}
+ */
+let Run;
```
rachel-fenichel: `Run` isn't a very specific/descriptive name. I prefer `TestRun` or `TestCase`.

plugins/dev-tools/src/common_test_helpers.mocha.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Runs provided test cases.
+ * @param {Array<Run>} runs The test cases to run.
+ * @param {function(Run):function} testFn Function that returns test callback.
+ * @private
+ */
+export function runTestCases(runs, testFn) {
+  runs.forEach((run) => {
+    let testCall = (run.skip ? test.skip : test);
```
rachel-fenichel: What is `test` here?

plugins/dev-tools/src/test_helpers.mocha.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as commonHelpers from './common_test_helpers.mocha';
+import * as fieldHelpers from './field_test_helpers.mocha';
+
+const {Run} = commonHelpers;
```
rachel-fenichel: You're only using one thing from `commonHelpers`, so you can just do

`import {Run} from './common_test_helpers.mocha';`

plugins/field-date/test/field_date_test.mocha.js
```diff
@@ -45,6 +44,12 @@ suite('FieldDate', function() {
     assertFieldValue(field, run.value);
   };
 
+  // TODO(https://github.com/google/blockly/issues/3903): Re-enable test cases
+  //  after fixing.
+  invalidValueRuns[3].skip = true;
```
rachel-fenichel: Why set skip here instead of in the definition of the test cases? This is counting on the index of the cases staying the same, which seems unstable.

plugins/dev-tools/src/common_test_helpers.mocha.js
moniika: `test` is the mocha function used for defining a test case. `only` and `skip` are similarly part of mocha api

plugins/field-slider/test/field_slider_test.mocha.js
moniika: Yes, it is used in the setValueTests L89. I need to reset skip so that the test case is not skipped for setValue tests.

plugins/dev-tools/src/test_helpers.mocha.js
moniika: I expect for this deconstruction to become larger as more common helpers are added.
I'm already planning on also adding to the list runTestCases in/for another PR I'm working on.
So, I think it makes sense to leave it like this for this reason.

plugins/dev-tools/src/common_test_helpers.mocha.js
moniika: I like TestCase, I can do that refactor. A few more files will end up being added, but it should be *mostly* JSDoc changes.

plugins/field-date/test/field_date_test.mocha.js
moniika: In this test suite I want skip `true` before the construction tests and skip `false` after (before `setValue` tests).
I think using the index for setting true and for false also makes it easier to make sure the setting the skip flag `true`/`false` line up (possibly making it more obvious if there is an error with index). Also, the skip true right before the call to run the creation tests seems a little clearer for what the skip is being set for.

I agree using index feels a bit unstable, but I feel like its not too much overhead to update the index as its not expected for the testCases to be expanded frequently, if at all.

plugins/dev-tools/src/common_test_helpers.mocha.js
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Runs provided test cases.
+ * @param {Array<TestCase>} testCases The test cases to run.
+ * @param {function(TestCase):function} testFn Function that returns test
+ *  callback.
+ */
+export function runTestCases(testCases, testFn) {
+  testCases.forEach((testCase) => {
+    let testCall = (testCase.skip ? test.skip : test);
+    testCall = (testCase.only ? test.skip : testCall);
```
rachel-fenichel: skip->only

rachel-fenichel (commented):
One change and then LGTM

plugins/dev-tools/src/common_test_helpers.mocha.js
moniika: Fixed

PR #280, opened 2020-05-30
The Blockly playground

samelhusseini:
Create a playground that lays out the Blockly workspace on the left, options / actions in the top right corner and the code editor in the bottom right.
Screen Shot 2020-05-29 at 5 56 39 PM

Add tabs for XML and each language generator. Clicking on each tab generates the code for that language. While the (auto) generate checkbox is checked, any changes to the workspace trigger a code generation. If the XML tab is selected, xml is generated.

While on the XML tab, a developer can "Right click -> Import" to import from xml. On all other tabs, the only option available is "Right click -> Export"

Added helper methods for loading the monaco editor, this is exported in the case a plugin developer wants to create their own playground. First attempt to load it from dev-tools node_modules, then the local plugin node_modules, and finally fall back on loading it from a CDN.

Re-organized GUI controls into: 
- Options
- Debug
- Actions

The `createPlayground` method initializes the playground and returns a promise to the playground API, which currently consists of: 
- addAction (to add additional actions)
- addGenerator (to add additional generator tabs)


BeksOmega:
I'm not sure if it's worth adding, but I'd like to request a "Clean XML" tab. I have this option on my personal playground, and it strips out things like coordinates and ids. This can be useful for sharing XML in an issue, or posting it on the forums.

I find it helpful, but I'm not sure if it's worth the screen space it would take up.

P.S. Love the auto checkmark!

samelhusseini:
Yay, thanks for trying it out! And for the feedback. I think that would be a good editor command to add, so it doesn’t clutter the UI, and just shows up in the context menu. Is that something you’d like to add once this is in? Otherwise, if you share your regex I’d be happy to add.

BeksOmega:
> thanks for trying it out! And for the feedback.

Thanks for making such a cool playground :P

> Is that something you’d like to add once this is in?

I would love to! But I'm pretty swamped with other projects at the moment so I don't think I'll have time :/

> Otherwise, if you share your regex I’d be happy to add.

I was actually cleaning up the nodes before converting them to text hehe. I just never think to use regex, but yeah that does seem like the better solution. It seems like this works?
```
/\s*(?:id|x|y)="[^"]*"/g
```
but I'm not regexpert =)

Thanks for taking the suggestion into consideration!

plugins/dev-tools/src/addGUIControls.js
```diff
@@ -90,6 +89,10 @@ export default function addGUIControls(createWorkspace, defaultOptions) {
     onChangeInternal();
   };
 
+  // Options folder.
```
rachel-fenichel: This function is huge--please break it up into helper functions.

plugins/dev-tools/src/playground/index.js
```diff
@@ -0,0 +1,304 @@
...
+import * as Blockly from 'blockly/core';
+import * as BlocklyJS from 'blockly/javascript';
+import * as BlocklyPython from 'blockly/python';
+import * as BlocklyLua from 'blockly/lua';
+import * as BlocklyDart from 'blockly/dart';
+import * as BlocklyPHP from 'blockly/php';
+
+import {renderPlayground, renderCheckbox, renderCodeTab} from './ui';
+import {addCodeEditor} from './monaco';
+import addGUIControls from '../addGUIControls';
```
rachel-fenichel: Consider changing addGUIControls to no longer be a default export, to match everything else.

plugins/dev-tools/src/playground/index.js
```diff
@@ -0,0 +1,304 @@
...
+    theme: 'vs-dark',
+    scrollBeyondLastLine: false,
+    automaticLayout: true,
+  }, vsEditorPath).then((editor) => {
+    let workspace;
+
+    // Create a model for displaying errors.
+    const errorModel = window.monaco.editor.createModel('');
+
+    // TODO(samelh): Sync playground settings with localStorage.
```
rachel-fenichel: Use an issue number instead of your name.

plugins/dev-tools/src/playground/index.js
```diff
@@ -0,0 +1,304 @@
...
+  return addCodeEditor(monacoDiv, {
+    model: null,
+    language: 'xml',
+    minimap: {
+      enabled: false,
+    },
+    theme: 'vs-dark',
+    scrollBeyondLastLine: false,
+    automaticLayout: true,
+  }, vsEditorPath).then((editor) => {
```
rachel-fenichel, line 75: Use helper functions--this single function is hard to read.

samelhusseini: Done.

plugins/dev-tools/src/playground/index.js
samelhusseini: Done.

plugins/dev-tools/src/playground/index.js
samelhusseini: Done.

plugins/dev-tools/src/addGUIControls.js
samelhusseini: Split up.

rachel-fenichel:
LGTM

PR #285, opened 2020-06-04
Update React example with createRef

ryandrew14:
This PR changes the uses of React's `ref` prop to use the more modern `React.createRef` function rather than directly setting a class variable. More on this here if you're interested: https://reactjs.org/docs/refs-and-the-dom.html

I was wrong earlier, they actually do describe "callback refs" here which is how the repo was using them, but it's a much less common syntax and doesn't need to be used here (we don't need the fine-grained control it offers).

### DIFF
--- a/examples/blockly-react/src/App.js
+++ b/examples/blockly-react/src/App.js
@@ -1,6 +1,6 @@
 /**
  * @license
- * 
+ *
  * Copyright 2019 Google LLC
  *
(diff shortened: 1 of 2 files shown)


ryandrew14:
CLA signed (edit: oops)

ryandrew14:
@googlebot I signed it!

samelhusseini (approved):

samelhusseini:
Thanks for the PR @ryandrew14!

PR #286, opened 2020-06-04
Get list of renderers from global registry

samelhusseini:
The global registry has moved where the list of renderers are kept. This makes it so the playground supports both.

### DIFF
--- a/plugins/dev-tools/src/addGUIControls.js
+++ b/plugins/dev-tools/src/addGUIControls.js
@@ -252,7 +252,7 @@ function populateRendererOption(folder, options, onChange) {
   // rendererMap_, whereas newer versions that use the global registry get their
   // list of renderers from somewhere else.
   const renderers = Blockly.blockRendering.rendererMap_ ||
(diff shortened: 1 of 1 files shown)


plugins/dev-tools/src/addGUIControls.js
```diff
@@ -243,6 +241,22 @@ function populateBasicOptions(basicFolder, options, onChange) {
...
+ * @param {dat.GUI} folder The dat.GUI folder.
+ * @param {Blockly.Options} options Blockly options.
+ * @param {function(string, string):void} onChange On Change method.
+ */
+function populateRendererOption(folder, options, onChange) {
+  // Get the list of renderers. Previous versions of Blockly used the
+  // rendererMap_, whereas newer versions that use the global registry get their
+  // list of renderers from somewhere else.
+  const renderers = Blockly.blockRendering.rendererMap_ ||
+    Blockly.registry.typeMap_['renderer'];
```
alschmiedt: should this be indented 2 more spaces?

alschmiedt (approved):

PR #287, opened 2020-06-05
Add playground support for a theme registry

### LINKED ISSUE #263: gui controls should find all registered renderers and themes
**Category**
Plugins

**Component**

`@blockly/dev-tools`

I want to be able to add and test a new theme or renderer in a playground that uses `addGUIControls`.

Find all registered themes and renderers and add them to the corresponding dropdowns on GUI controls automatically.

Expose an API for setting what the dropdown values could be.

samelhusseini:
Get the list of themes from the theme registry if we're using a version of Blockly that supports it.

Fixes https://github.com/google/blockly-samples/issues/263

### DIFF
--- a/plugins/dev-tools/src/addGUIControls.js
+++ b/plugins/dev-tools/src/addGUIControls.js
@@ -123,18 +123,7 @@ export function addGUIControls(createWorkspace, defaultOptions) {
   populateRendererOption(optionsFolder, options, onChange);
 
   // Theme.
-  const themes = {
-    'classic': Blockly.Themes.Classic,
-    'dark': Blockly.Themes.Dark,
-    'deuteranopia': Blockly.Themes.Deuteranopia,
-    'highcontrast': Blockly.Themes.HighContrast,
-    'tritanopia': Blockly.Themes.Tritanopia,
-  };
-  if (defaultOptions.theme) {
-    themes[defaultOptions.theme.name] = defaultOptions.theme;
-  }
-  optionsFolder.add(options.theme, 'name', Object.keys(themes)).name('theme')
-      .onChange((value) => onChange('theme', themes[value]));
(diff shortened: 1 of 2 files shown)


plugins/dev-tools/src/addGUIControls.js
```diff
@@ -252,11 +241,40 @@ function populateRendererOption(folder, options, onChange) {
...
+ * @param {dat.GUI} folder The dat.GUI folder.
+ * @param {Blockly.Options} options Blockly options.
+ * @param {Blockly.Options} defaultOptions Default Blockly options.
+ * @param {function(string, string):void} onChange On Change method.
+ */
+function populateThemeOption(folder, options, defaultOptions, onChange) {
+  let themes;
+  if (Blockly.registry && Blockly.registry.typeMap_['theme']) {
+    // Using a version of Blockly that registers themes.
+    themes = Blockly.registry.typeMap_['theme'];
```
alschmiedt, line 260: Should we make a public getter for this?

alschmiedt (approved):

plugins/dev-tools/src/addGUIControls.js
samelhusseini: Aside from dev purposes, which is what we're doing here. What would be the use case for a public getter?

alschmiedt: Someone could want to display all the themes to the user so they can choose in a drop down which one they want.

samelhusseini: Sure, could be useful. Although in cases like that I would resort to maintaining my own list of themes I want to display to my users. eg, just because the Dark theme is "available" (ie; in the registry), doesn't mean the developer wants to show that to their users.

Filed issue https://github.com/google/blockly/issues/3935

PR #297, opened 2020-06-08
Fix resizing GUI control in playground

samelhusseini:
Fix issue with resizing of GUI controls when right toolbox side is selected.

@alschmiedt this fixes the issue you came across.

### DIFF
--- a/plugins/dev-tools/src/playground/index.js
+++ b/plugins/dev-tools/src/playground/index.js
@@ -36,13 +36,17 @@ let CreateWorkspaceFn;
  */
 let PlaygroundTab;
 
+
 /**
- * @typedef {{
- *     auto:?boolean,
- *     toolboxes:Array<Blockly.utils.toolbox.ToolboxDefinition>,
- * }}
+ * @typedef {Blockly.utils.toolbox.ToolboxDefinition} BlocklyToolbox
  */
-let PlaygroundConfig;
+
+/**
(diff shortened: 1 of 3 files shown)


plugins/dev-tools/src/playground/index.js
```diff
@@ -267,7 +267,10 @@ export function createPlayground(container, createWorkspace,
     // Move the GUI Element to the gui container.
     const guiElement = gui.domElement;
     guiElement.removeChild(guiElement.firstChild);
-    guiElement.style.top = '0';
+    guiElement.style.top = '';
+    guiElement.style.left = '';
+    guiElement.style.right = '';
+    guiElement.style.bottom = '';
```
alschmiedt: What does setting to empty string do?

alschmiedt (approved):

plugins/dev-tools/src/playground/index.js
samelhusseini: It clears it, the reason behind this is when the GUI control is first initialized (as it can be initialized without the playground), it calls resize (which sizes it depending on where the toolbox is).
This clears these properties set (as it it sized with position=relative) instead of absolute positioning when placed in the playground.

samelhusseini: On second thought, I think this is better dealt with a config now that we pass configs to the GUI control. (rather than patching it up afterwards).

PR #299, opened 2020-06-08
Playground - persist theme

samelhusseini:
Persist the theme by name in the playground.

Similar to how the toolbox is persisted.

### DIFF
--- a/plugins/dev-tools/src/addGUIControls.js
+++ b/plugins/dev-tools/src/addGUIControls.js
@@ -158,7 +158,7 @@ export function addGUIControls(createWorkspace, defaultOptions, config) {
   populateRendererOption(optionsFolder, options, onChange);
 
   // Theme.
-  populateThemeOption(optionsFolder, guiState, defaultOptions, defaultThemeName,
+  populateThemeOption(optionsFolder, guiState, themes, defaultThemeName,
       onChange);
 
   // Toolbox.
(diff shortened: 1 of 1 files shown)


plugins/dev-tools/src/addGUIControls.js
```diff
@@ -292,6 +300,9 @@ function saveGUIState(guiState, defaultToolboxName) {
   if (guiState.toolboxName !== defaultToolboxName) {
     hashGuiState.toolbox = guiState.toolboxName;
   }
+  if (guiState.themeName !== defaultThemeName) {
```
alschmiedt, line 303: Why do you have to check this? Why can't you always set it, even if it is the default value?

alschmiedt (approved):

plugins/dev-tools/src/addGUIControls.js
samelhusseini: If the value is set, we add url parameters (eg: theme=classic), and open any relevant GUI controls folders.

Same reasoning for not including all options unless selected by the user, eg: rtl=false, toolboxPosition=start, etc.

Does that seem reasonable? Or do you feel the theme option I different?

alschmiedt: Seems reasonable. I was just curious, thanks!

alschmiedt (approved):

PR #308, opened 2020-06-12
Add support for testing fields that throw on invalid values.

moniika:
Some fields, such as `Blockly.FieldImage`, throw when an invalid value is passed at creation, rather than reverting to a default. This change would add support in helpers to test field creation for fields that throw on invalid values.

### DIFF
--- a/plugins/dev-tools/src/field_test_helpers.mocha.js
+++ b/plugins/dev-tools/src/field_test_helpers.mocha.js
@@ -23,10 +23,15 @@ FieldValueTestCase.prototype.value = undefined;
  */
 FieldValueTestCase.prototype.expectedValue = undefined;
 /**
- * @type {*} The expected text (if not specified, default is
+ * @type {string|undefined} Optional expected text (if not specified, default is
  *    String(expectedValue).
  */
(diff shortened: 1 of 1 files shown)


plugins/dev-tools/src/field_test_helpers.mocha.js
```diff
@@ -86,6 +91,30 @@ function runCreationTests_(testCases, assertion, creation) {
...
 
+/**
+ * Runs provided creation test cases.
+ * @param {Array<FieldCreationTestCase>} testCases The test cases to run.
+ * @param {function(new:Blockly.Field,FieldCreationTestCase):Blockly.Field
+ *    } creation A function that returns an instance of the field based on the
+ *    provided test case.
+ * @private
+ */
+function runCreationTestsAssertThrows_(testCases, creation) {
```
moniika, line 102: Alternatively, I have considered adding a property on the test case. However, the only benefit I see there is that we would be able to test a field where some values throw and some revert to default, which I think seems unlikely (I would expect it to be one or the other, not both).

samelhusseini (approved):

PR #331, opened 2020-07-20
Test blocks and Test toolbox

### LINKED ISSUE #210: Test blocks and test block toolbox in devtools
Move the test blocks from the playground, and the test block toolbox, into devtools.

Why: they're useful for testing all plugins, not just code in core.

We can still use them in the core playground with unpkg and a script tag.

samelhusseini:
Move the test blocks and test toolbox from Blockly into dev tools.
Fixes https://github.com/google/blockly-samples/issues/210

The change divides up each toolbox category into its own file, and places subcategories into folders.

Each category file contains: 
- The Blocks it defines
- Mutators / Extensions it defines
- A category `const` that is exported. This is the category definition in JSON. 
- An `onInit` method that is called with the workspace when the workspace is initialized. This allows us to register button callbacks specific to a category.

The playground is initialized to include (amongst category and simple toolboxes), the test blocks toolbox.
`toolboxTestBlocks` and `toolboxTestBlocksInit` are exported for use outside the playground.


alschmiedt (approved):

alschmiedt:
We might want to wait to merge these for a bit since the toolbox json definition is still a bit in flux.

samelhusseini:
Discussed offline. We will release this without enabling the test blocks toolbox by default, but the test blocks will be defined and their init methods called.

We will also remove the test blocks from blockly as part of this release. We will wait until the JSON is finalized for the toolbox before adding the test blocks toolbox to the playground by default.

PR #333, opened 2020-07-22
Strict connections

rachel-fenichel:
A plugin that imposes stricter typing rules than the default checker in Blockly, but uses the same rules for dragging and safety.

This checker still expects nullable arrays of string for connection type checks, and still looks for intersections in the arrays. Unlike the default checker, null checks arrays are only compatible with other null arrays.

### Resolves

https://github.com/google/blockly/issues/1357

### Test Coverage

I copied over the mocha tests I had for the connection checker, fixed lint, and replaced the checker object in setup. I only needed to change one test ('One Typed, One Null').

### Additional Information

The new `IConnectionChecker` API is in develop, so I `npm link`ed to my own instance of Blockly to test it.

This PR is intended to show how to extend the new API--I hope to see an inheritance-based checker from the community.

The package is currently private, and will not be published on npm until after the Q3 release.


plugins/strict-type-checker/README.md
```diff
@@ -0,0 +1,43 @@
+# blockly-plugin-strict-type-checker [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
```
samelhusseini: Any reason not to match the naming guide with `@blockly/plugin-strict-type-checker`?

Also, should this be called a strict-connection-checker instead as it's overriding a connection checker.

plugins/strict-type-checker/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "blockly-plugin-strict-type-checker",
+  "version": "0.20200717.0",
+  "description": "A Blockly plugin.",
```
samelhusseini: Add description

plugins/strict-type-checker/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "clean": "blockly-scripts clean",
+    "lint": "blockly-scripts lint",
+    "prepublishOnly": "npm run clean && npm run build",
+    "start": "blockly-scripts start",
+    "test": "blockly-scripts test"
+  },
+  "main": "./dist/index.js",
+  "module": "./src/index.js",
+  "unpkg": "./dist/index.js",
+  "author": "Rachel Fenichel",
```
samelhusseini: We have generally left this one as "Blockly Team" for other plugins.

plugins/strict-type-checker/test/index.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Plugin test.
```
samelhusseini: Update description.

plugins/strict-type-checker/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "blockly-plugin-strict-type-checker",
+  "version": "0.20200717.0",
```
samelhusseini: I think the create-package script never got updated after we switched to semver, so this should be 0.0.0 for now.

Filed https://github.com/google/blockly-samples/issues/334

samelhusseini (approved):
Overall looks good, minor comments. Only question is the choice of module and package name.

PR #335, opened 2020-07-24
Support acquiring Blockly through Git instead of npm

samelhusseini:
Support being able to acquire Blockly from Git instead of npm for the purpose of depending on changes that have not yet been released.

How: 
When we detect that Blockly was acquired from Git instead of npm:
- Add a postinstall script the goes into the Blockly node_module directory and installs, builds and packages blockly.
- Add a `dist` at the end of the blockly module alias (as we package everything out of the 'dist' directory)

Using this for the strict-connection-checker package.
Enabled tests for strict-connection-checker (which were previously failing as they relied on pre-released bits)

@rachel-fenichel let me know what you think of this strategy.
Travis build should succeed.


plugins/dev-scripts/config/webpack.config.js
```diff
@@ -55,6 +57,16 @@ module.exports = (env) => {
     target = 'node';
   }
 
+  // Add 'dist' to the end of the blockly module alias if we have acquired
+  // blockly from git instead of npm.
+  let blocklyAliasSuffix = '';
+  const blocklyDependency =
+    (packageJson.dependencies && packageJson.dependencies['blockly']) ||
+    (packageJson.devDependencies && packageJson.devDependencies['blockly']);
+  if (blocklyDependency && blocklyDependency.indexOf('git://') === 0) {
```
rachel-fenichel, line 66: This is not guaranteed to be a good check: https://docs.npmjs.com/files/package.json#github-urls

However, we could just require that you use this prefix. nbd.

rachel-fenichel (approved):
Looks good if we just require that they use the git:// form of the dependency.

Travis errors seem to just be lint.

plugins/dev-scripts/config/webpack.config.js
samelhusseini: Yeah, I didn't want to make the check`!semver`, because you can technically do tags and things like `latest` in that field, so it's simpler to just limit it to `git://` for this use case.

Issue #344, opened 2020-08-15
Outdated code in custom generator codelab step 3


moniika:
**Category**

Codelabs

**Component**

Custom Generator

**Describe the bug**

[Step 3 of the custom generator codelab](https://blocklycodelabs.dev/codelabs/custom-generator/index.html?index=..%2F..index#2) is unclear as to where to put the code (for the toJSON button) and outdated.
The instructions do not work because significant changes have been been made to the advanced playground and the `workspace` variable is not easily accessible and `taChange` function no longer exists (sample code uses these).

The sample code and instructions for step 3 need to be updated and the codelab as a whole should be reviewed to make sure nothing else needs changing.

**Additional context**

n/a

PR #347, opened 2020-08-25
Set a dynamic generator as the current tab if it is active

samelhusseini:
For generators added through `addGenerators` after the playground is loaded we aren't able to set them as active until they are added.

This adds a check once a generator is added to see if it's active, if so, it sets it as the current tab.


rachel-fenichel (approved):

rachel-fenichel:
Related to https://github.com/google/blockly-samples/issues/344

PR #348, opened 2020-08-26
Update custom generator codelab for advanced playground

### LINKED ISSUE #344: Outdated code in custom generator codelab step 3

rachel-fenichel:
Fixes #344

Depends on publishing #347 

Updated instructions for adding a generator--instead of explicitly adding a button and callback, just add a generator to the playground.

Updated description of where output will display.

### DIFF
--- a/codelabs/custom_generator/custom_generator.md
+++ b/codelabs/custom_generator/custom_generator.md
@@ -153,7 +153,7 @@ A language generator has a single entry point: `workspaceToCode`. This function 
 
 The first step is to define and call your language generator.
 
-A custom language generator is simply an instance of `Blockly.Generator`. Call the constructor, passing in your generator's name, and store the result.
+A custom language generator is simply an instance of `Blockly.Generator`. In `custom_generator.js` call the constructor, passing in your generator's name, and store the result.
 
 ```js
 const codelabGenerator = new Blockly.Generator('JSON');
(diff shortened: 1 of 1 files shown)


codelabs/custom_generator/custom_generator.md
```diff
@@ -153,43 +153,28 @@ A language generator has a single entry point: `workspaceToCode`. This function
...
 ### Generate code
 
-Add a button to the playground to generate JSON:
-
-```html
-<input type="button" value="To JSON" onclick="toJSON()">
-```
-
-Implement `toJSON` so that it generates code and outputs it to two places: the console, and the playground's text area.
+Next, register your new generator with the playground. Find the `configurePlayground` function in `advanced_playground.html` and add this line:
```
samelhusseini, line 164: Not exactly a recommendation, but do you think some people might get lost at where to put the code above (initialization) if the code below is going in configurePlayground.

If I'm following right, both of them should go in configurePlayground.

samelhusseini (approved):

codelabs/custom_generator/custom_generator.md
rachel-fenichel: Yeah I had that code going in `custom_generator.js`, so obviously we have some confusion here.

I prefer having it in custom_generator.js so that it hangs out with the rest of the generator definition.

wulph111:
Section 9 still references the nonexistent button, which was mildly confusing. 
image

PR #349, opened 2020-08-27
Make the flyout part of the deletable area

maribethb:
When dragging a block out of the flyout, you now have to drag it sufficiently out of the flyout area to drop it onto the workspace. This is done if autoClose = false, or in other words if the flyout will still be open while you're dragging. Otherwise, the normal toolbox rules apply.

### DIFF
--- a/plugins/continuous-toolbox/src/ContinuousToolbox.js
+++ b/plugins/continuous-toolbox/src/ContinuousToolbox.js
@@ -63,6 +63,15 @@ export class ContinuousToolbox extends Blockly.Toolbox {
       this.flyout_.stepScrollAnimation();
     }
   }
+
+  /** @override */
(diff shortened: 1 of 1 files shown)


plugins/continuous-toolbox/src/ContinuousToolbox.js
```diff
@@ -63,6 +63,15 @@ export class ContinuousToolbox extends Blockly.Toolbox {
       this.flyout_.stepScrollAnimation();
     }
   }
+
+  /** @override */
+  getClientRect() {
+    // If the flyout never closes, it should be in the deletable area.
```
alschmiedt: nit: remove in

alschmiedt (approved):

Issue #346, opened 2020-08-23
Nominal type checker w/ higher types & generics


BeksOmega:
Codename: Project Static

**Category**

 - Plugins

**Is your feature request related to a problem? Please describe.**

Blockly's built-in type checker makes it hard to model languages which have strong static typing like C, C#, Java, Swift, Rust, etc. Currently, modeling subtyping requires connection checks to be very verbose and WET. There is also no support for generics in the current system.

**Describe the solution you'd like**

A nominal static type checker with support for higher types and generics. This would act as an optional replacement for Blockly's built in type checker.

**Additional**

Are you guys still ok with me taking on this project? If so, is this issue a good place to organize conversation?

I've been pretty excited about this project so I started work on a [design document](https://docs.google.com/document/d/1QKYkmWjkle1JWCi3O8jXr8-7Toazh1pW_4EaVVgB_OI/edit?usp=sharing) discussing how I think it should work from the perspective of someone integrating Blockly into their application. I think it would be good to get input from people on the forums about it once it's done, because this requires so many design decisions. How do you guys feel about that?

I also have a pretty good idea of how I would to go about implementing the system, but I didn't want to bog the doc down with all of those details.

All that being said, I'm still totally cool if you think someone else should take this on hehe.

rachel-fenichel:
Yes, you should definitely work on this, and I agree about getting input from people on the forums after getting the design sorted and before starting the serious lifting on implementation. Let's use your doc for back-and-forth on the design, and then post a summary here once you're happy with the design.

BeksOmega:
Ok, I think the [doc](https://docs.google.com/document/d/1QKYkmWjkle1JWCi3O8jXr8-7Toazh1pW_4EaVVgB_OI/edit?usp=sharing) does a pretty good job explaining what I was imagining now =) Anyone can feel free to take a look. All comments and suggestions are welcome, regardless of whether they be about the design or the doc itself :D

[Edit: Added a section on [Multiple generic input checks](https://docs.google.com/document/d/1QKYkmWjkle1JWCi3O8jXr8-7Toazh1pW_4EaVVgB_OI/edit#heading=h.ch88lvpv9s55), and a section on [Constraints containing generic type names](https://docs.google.com/document/d/1QKYkmWjkle1JWCi3O8jXr8-7Toazh1pW_4EaVVgB_OI/edit#heading=h.hvbpl1k6gwzi)]

BeksOmega:
[...]

PR #350, opened 2020-09-03
Setup for project static

BeksOmega:
### Relevant Issue

#346 

### Description

* Creates a package for project static. I decided go with the package name nominal-connection-checker, to mirror strict-connection-checker, but tell me if there's a name you like better.
* Sets up a new playground that will hopefully make testing easier. It allows you to define the type hierarchy and the blocks in the browser using JSON. It also saves the current state of the workspace, but it doesn't have the ability to load arbitrary XML via a code editor.
StaticPlayground

### Question

Where would I go to learn how to host this playground somewhere? I think it would be cool to host it somewhere so that people could try it out as I add stuff. But I don't even know what search terms to use to find out how to do that haha. I'm like... "webpack?"... "npm?"... "node?" I'm completely out of my depth here. If someone could point me in the right direction I would appreciate that so much.

### DIFF
--- a/plugins/nominal-connection-checker/GETSTARTED.md
+++ b/plugins/nominal-connection-checker/GETSTARTED.md
@@ -1,0 +1,24 @@
+This package was bootstrapped with [@blockly/create-package](https://www.npmjs.com/package/@blockly/create-package).
+
+## Available Scripts
+
+In this directory, you can run:
+
+### `npm start`
+
+Runs the package in development mode.
+
+Open [http://localhost:3000/test](http://localhost:3000/test) to view the test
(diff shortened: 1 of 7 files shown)


samelhusseini:
In terms of hosting, I would recommend using github pages. We could add some tooling to make it easier for folks to deploy their package to github pages. https://pages.github.com/

There's also an [npm package](https://www.npmjs.com/package/gh-pages) that we could build in to make it easy for third party developers to deploy their packages to gh pages.

rachel-fenichel (approved):
LGTM for initial project setup.

PR #351, opened 2020-09-09
Update the toolbox definition

alschmiedt:
Update the toolbox definition for test blocks to work with the September 2020 release of blockly. Since dev tools is snapped to a certain version pushing this should not cause a problem.

### DIFF
--- a/plugins/dev-tools/src/testblocks/index.js
+++ b/plugins/dev-tools/src/testblocks/index.js
@@ -24,15 +24,17 @@ import {category as styleCategory, onInit as initStyle} from './style';
 /**
  * The Test blocks toolbox.
  */
-export const toolbox = [
-  alignCategory,
-  basicCategory,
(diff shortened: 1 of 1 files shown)


samelhusseini (approved):

plugins/dev-tools/src/testblocks/index.js
```diff
@@ -24,15 +24,17 @@ import {category as styleCategory, onInit as initStyle} from './style';
...
-  alignCategory,
-  basicCategory,
-  connectionsCategory,
-  dragCategory,
-  fieldsCategory,
-  mutatorsCategory,
-  styleCategory,
-];
+export const toolbox = {
+  "contents": [
```
samelhusseini: nit: use single quote instead of double.

alschmiedt: done

PR #352, opened 2020-09-09
use throw for handling errors

geekyahmed:
I used the throw statement to provide a proper way of showing errors rather than just logging them into the console.

### DIFF
--- a/examples/blockly-node/index.js
+++ b/examples/blockly-node/index.js
@@ -45,5 +45,5 @@ try {
   console.log(code);
 }
 catch (e) {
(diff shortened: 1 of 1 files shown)


samelhusseini:
I don't really understand the desired outcome here?

a - we can just get rid of the try / catch if the purpose is to throw.
b - you're hiding away the error message.

geekyahmed:
I think it is better to hide the error message and just throw it instead

rachel-fenichel:
Closing as stale, and because we don't want to hide the messages in the sample.

PR #357, opened 2020-09-12
Add basic subtype support to project static

BeksOmega:
### Description

Adds basic subtyping support to project static. This has no parameterized types or generic types, just a very simple TypeHierarchy.

### Hierarchy def validation

Currently the TypeHierarchy doesn't do any validation on the hierarchyDef it gets passed, which means that it is possible to define circular dependencies.

I haven't added validation yet because I'm worried that it is going to become really expensive once the type system gets more complex. (Eg, making sure that the parameter variance of parameters is compatible with their super types.) If in most cases devs will be defining valid hierarchies, it doesn't seem worth it to me atm.

I'm wondering if maybe it would be nice to add a validation function devs could run while they're constructing their hierarchy? That way they can test it if they want, but it's not required on start-up.

Or maybe it is just better to always validate. I would love your opinions on this :D

### Testing

Manually tested using the playground.
Wrote unit tests for all of the public functions on the TypeHierarchy.
Wrote unit tests for the connection checker to cover basic subtyping.

---

[Edit: I removed support for multiple checks on connections because that doesn't really makes sense when we go to bind generic type names to explicit types. Anything you can do with multiple type checks you should be able to do with single type checks by defining (via the type hierarchy) that both types fulfill a common super type.]

### DIFF
--- a/plugins/nominal-connection-checker/src/type_hierarchy.js
+++ b/plugins/nominal-connection-checker/src/type_hierarchy.js
@@ -39,9 +39,10 @@ export class TypeHierarchy {
     // NOTE: This does not do anything to stop a developer from creating a
     // cyclic type hierarchy (eg Dog <: Mammal <: Dog). They are expected to
     // not do that.
-    const typeNames = Object.keys(hierarchyDef);
-    for (let i = 0, typeName; (typeName = typeNames[i]); i++) {
-      this.types_[typeName] = new TypeDef(typeName, hierarchyDef[typeName]);
+    for (const typeName of Object.keys(hierarchyDef)) {
+      const lowerCaseName = typeName.toLowerCase();
+      this.types_[lowerCaseName] = new TypeDef(
+          lowerCaseName, hierarchyDef[typeName]);
     }
   }
 
@@ -53,7 +54,7 @@ export class TypeHierarchy {
    * otherwise.
    */
   typeExists(name) {
-    return !!this.types_[name];
+    return !!this.types_[name.toLowerCase()];
   }
 
   /**
@@ -64,7 +65,7 @@ export class TypeHierarchy {
    *     otherwise.
    */
   typeIsExactlyType(name1, name2) {
-    return name1 == name2;
+    return name1.toLowerCase() == name2.toLowerCase();
   }
 
   /**
@@ -78,15 +79,17 @@ export class TypeHierarchy {
    *     in the type hierarchy definition. False otherwise.
    */
   typeFulfillsType(subName, superName) {
-    if (this.typeIsExactlyType(subName, superName)) {
+    const caselessSub = subName.toLowerCase();
+    const caselessSup = superName.toLowerCase();
+    if (this.typeIsExactlyType(caselessSub, caselessSup)) {
       return true;
     }
-    const subType = this.types_[subName];
-    if (subType.hasDirectSuper(superName)) {
+    const subType = this.types_[caselessSub];
+    if (subType.hasDirectSuper(caselessSup)) {
       return true;
     }
     return subType.someSuper(
-        (name) => this.typeFulfillsType(name, superName), this);
+        (name) => this.typeFulfillsType(name, caselessSup), this);
   }
 }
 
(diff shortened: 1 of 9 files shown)


plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -0,0 +1,167 @@
...
+   * Returns true if the given function returns a truthy value for at least one
+   * super type of this type. False otherwise.
+   * @param {function(string, number=, !Array=):boolean} callback A function
+   *     used to test each super type.
+   * @param {!Object=} thisArg A value to use as `this` when executing callback.
+   * @return {boolean} True if the given function returns a truthy value for
+   *     at least one super type of this type. False otherwise.
+   */
+  someSuper(callback, thisArg) {
+    return this.fulfills_.some(callback, thisArg);
```
rachel-fenichel, line 156: TIL about `array.some`

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -0,0 +1,167 @@
...
+ */
+export class TypeHierarchy {
+  /**
+   * Constructs the TypeHierarchy, initializing it with the given hierarchyDef.
+   * @param {!Object} hierarchyDef The definition of the type hierarchy.
+   */
+  constructor(hierarchyDef) {
+    /**
+     * Map of type names to TypeDefs.
+     * @type {!Object<!string, !TypeDef>}
```
rachel-fenichel: Why not use a Map?

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -0,0 +1,167 @@
...
+    if (subType.hasDirectSuper(superName)) {
+      return true;
+    }
+    return subType.someSuper(function(name) {
+      return this.typeIsSuperOfType(superName, name);
+    }, this);
+  }
+
+  /**
+   * Returns true if the first type is a subt type of the second type. False
```
rachel-fenichel: typo: subt -> sub

plugins/nominal-connection-checker/test/type_hierarchy_test.mocha.js
```diff
@@ -0,0 +1,247 @@
...
+        'A': { },
+      });
+      chai.assert.isTrue(hierarchy.typeExists('A'));
+    });
+
+    test('Case', function() {
+      const hierarchy = new TypeHierarchy({
+        'A': { },
+      });
+      chai.assert.isFalse(hierarchy.typeExists('a'),
```
rachel-fenichel: What benefit do you see in case-sensitivity?  I learn toward case-insensitivity for this application, again with the goal of reducing the impact of typos.

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Returns true if the first type is a super type of the second type. False
+   * otherwise.
+   * @param {string} superName The name of the possible super type.
+   * @param {string} subName The name of the possible sub type.
+   * @return {boolean} True if the first type is a super type of the second
+   *     type. False otherwise.
+   */
+  typeIsSuperOfType(superName, subName) {
```
rachel-fenichel: This name feels awkward. Can you phrase it in terms of `fulfills` instead?

rachel-fenichel (commented):
Looks solid, though I left a few questions and comments.

On validation: I think you should make a validator that is easy to use, and expect developers to manually run their hierarchy through the validator while they're developing it. That lets you avoid running it on every workspace creation.

Although I'm curious about how expensive "expensive" can mean in this cas.e

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: TIL about Map :P

BeksOmega:
> Although I'm curious about how expensive "expensive" can mean in this case.

I tabulated it up, and it turns out it probably wouldn't be that expensive. I think the slowest thing would actually be the circular dependency check, which has a time of O(V+E) and a space of O(V). But since peoples type hierarchies should be pretty sparse it shouldn't be that bad.

I think that I'll implement the validation in a separate PR (since it will need its own tests and things). I'll probably leave it separate for now, but after it's built it will probably be easier to decide if we want to add validation as part of initialization or not.

rachel-fenichel (approved):

PR #363, opened 2020-09-18
Add test blocks back in

alschmiedt:
Found this while looking through the code. #210 has been closed so this should be uncommented. I'm a bit confused how test blocks are working in the advanced playground.


samelhusseini:
They’re working in the playground because in Blockly we override the list of toolboxes. And test blocks is one of them.

samelhusseini (approved):

PR #365, opened 2020-09-18
Add type hierarchy validation - project static

BeksOmega:
### Description

Adds an exported function `validateHierarchy` that takes in an object an validates it. This function logs an informative error to the console using `console.error` whenever it encounters a problem with the hierarchy definition. I chose to use `console.error` rather than `throw` so that I could log multiple errors.

#### Checks
Validation currently runs three checks:
* **Conflicting types.**
  For example the following hierarchy definition:
  ```json
  {
    "typeA": { },
    "TYPEA": { }
  }
  ```
  would throw an error because the type hierarchy is caseless, which causes typeA and TYPEA to conflict.
* **Undefined supers.**
  For example the following hierarchy definition:
  ```json
  {
    "typeA": {
      "fulfills": ["typeB"]
    }
  }
  ```
  would log an error, because the super type typeB is not defined in the type hierarchy.
* **Circular dependencies.**
  For example the following hierarchy definition:
  ```json
  {
    "typeA": {
      "fulfills": ["typeB"]
    },
    "typeB": {
      "fullfills": ["typeA"]
    }
  }
  ```
  would log an error, because typeA fulfills typeB fulfills typeA, which doesn't make sense.

More checks will be added as project static becomes more complex.

### Testing

I added unit tests for all the different kinds of errors :D

### DIFF
--- a/plugins/nominal-connection-checker/src/hierarchy_validation.js
+++ b/plugins/nominal-connection-checker/src/hierarchy_validation.js
@@ -1,0 +1,57 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Defines functions for validation a type hierarchy definition.
+ */
+'use strict';
+
+/**
+ * Validates the given hierarchy definition. Does checks for duplicate types,
+ * circular dependencies, etc. Errors are logged to the console (not thrown).
+ * @param {!Object} hierarchyDef The definition of the type hierarchy.
+ */
+export function validateHierarchy(hierarchyDef) {
+  if (!hierarchyDef ||
+      typeof hierarchyDef != 'object' ||
+      Array.isArray(hierarchyDef)) {
+    console.error('The hierarchy definition should be an object.');
+    return;
+  }
+
+  checkConflictingTypes(hierarchyDef);
+}
+
+/**
+ * Checks the hierarchyDef for conflicting types.
+ * @param {!Object} hierarchyDef The definition of the type hierarchy.
+ */
+function checkConflictingTypes(hierarchyDef) {
+  const conflictMsg = 'The type name \'%s\' conflicts with the type name(s) %s';
+
+  // Map of caseless names to cased names.
+  const definedTypes = new Map();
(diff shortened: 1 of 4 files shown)


plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -0,0 +1,168 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Defines functions for validation a type hierarchy definition.
```
rachel-fenichel: Nit: validation *of*

plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -0,0 +1,168 @@
...
+ * }
+ * @param {!Object} hierarchyDef The definition of the type hierarchy.
+ */
+function checkCircularDependencies(hierarchyDef) {
+  /**
+   * Logs an informative error about a cyclic dependency.
+   * @param {!Array<string>} cycleArray An array of all the types that have been
+   *     visited since our entry node.
+   */
+  function logCycle(cycleArray) {
```
rachel-fenichel: Why is this defined inside `checkCircularDependencies` instead of separately? It would not be exported in either case.

BeksOmega: Oh yeah I got used to doing this from working in Scheme hehe. The idea is that because the `logCycle` function is only used by the `checkCircularDependencies` function you define it as in an inner function so that (a) you remember that and (b) you can collapse the outer function and hide everything at once.

It's really just a code org thing, so idc how it ends up =) Do you want me to move the inner funcs to top-level definitions?

rachel-fenichel: I have a slight preference for having it at the top level. Let mek now which you pick and when you're ready to merge.

BeksOmega: I moved `logCycle` out and renamed it to `logCircularDependency` to be more clear. I left `searchForCyclesRec` as an inner function because I forgot it was a closure.

Should be gtg :D

PR #376, opened 2020-09-22
Add super basic generics - project static

BeksOmega:
### Preface

This is a bit of a weird PR because it's mostly puzzle pieces that will hopefully come together later. But I'll do my best to explain where my brain is going.

Also, I know this PR description is super long, and I still don't think I did a good job explaining. So if there's anything that's unclear or anything I can do to make it more clear please tell me.

### Basic Idea

Let's start with a simple example of generics. Below there is a generic Identity block with a connection check T on its input and output. After we connect the Dog block to the Identity block we want the T check to start acting like a Dog check.
ConnectDog

Now say we connect annother Identity block to the first Identity block. We now want that block's T check to /also/ start acting like a Dog check.
ConnectIdentity

There are two ways we could implement this.
  1) Whenever we connect a connection with a generic check to another connection, we create a "pointer" to the connection check of the other connection. Then when we go to get the explicit type of the generic check, we do a recursive lookup to see if it is bound to an explicit type anywhere.
  So if we were to connect these two Identity blocks their T types would be bound to eachother.
  IdentityBlocks

  2) We only "bind" a generic check to another check if the other check is an explicit type, or has an explicit type associated with it.
  So if were were to connect these two Identity blocks their T types would /not/ be bound to eachother.
  IdentityBlocks
  But if we added a Dog block to the second one, then our bindings would update to look like this:
  IdentityWithDog

I decide to go with option 2. The logic will be more complicated, but I believe it will be more efficient.

#### Efficiency

For example, say we implement option 1 and then I build the following set of blocks:

Each time I connect an Identity block, the growing blob of blocks would need to recursively check every connected block to see if it is bound to an explicit anywhere.
ConnectBlob

Whereas option 2 wouldn't require any of that, because none of the blocks are bound yet. It would only require a recursive update if/when I were to eventually connect an explicit type.
AddDogToBlob

#### Complicated logic

To implement option 2 we need to do different things after two connections connect, depending on the "kinds" of the connections.

So currently a connection check can have three different "kinds"
A) It can be explicit, eg Dog - **E**
B) It can be generic, and not currently bound to an explicit type - **GU**
C) It can be generic, and bound to an explicit type eg Dog - **GB**

Note that we **do** want to bind already bound generics (**GB**s) to **E**s and other **GB**s. The reasoning for this is explained below in the PriorityQueueMap section.

| Parent | Child | Action |
|----------|---------|----------|
|E|E|Bind neither|
|E|GU|Bind child to parent|
|E|GB|Bind child to parent|
|GU|E|Bind parent to child|
|GU|GU|Bind neither|
|GU|GB|Bind parent to child|
|GB|E|Bind parent to child|
|GB|GU|Bind child to parent|
|GB|GB|Bind both to each other|

This is what the logic [here](https://github.com/google/blockly-samples/pull/376/files#diff-150b7d992bf47b4060c87ae4d301ee6cR122) is meant to do.

We also need to do similar unbinding when the two blocks disconnect, except it's more complicated, because once an unbound connection is connected, it becomes bound.

| Parent | Child | Action |
|----------|---------|----------|
|E|E|Unbind Neither|
|~~E~~|~~GU~~|------|
|E|GB|Unbind child from parent|
|~~GU~~|~~E~~|------|
|GU|GU|Unbind neither|
|~~GU~~|~~GB~~|------|
|GB|E|Unbind parent from child|
|~~GB~~|~~GU~~|------|
|GB|GB|Unbind both to each other|

Luckily the unbinding logic pretty much works even though connections that were previously GUs are now GBs (since they've been bound by connecting). We just have to make sure that the GenericMap (see below) fails graceful if we try to delete a binding and the binding doesn't exist.

#### Current logic limitiations

Note that currently the logic doesn't handle updating the checks that are dependent on a block when a block updates.

So if you connect the explicit block first, and then continue connecting generic blocks all of the bindings "flow through" the connections.
CorrectFlow

But if you connect the generic blocks together first, and then connect the explicit block, that's not handled yet.
BadFlow

#### PriorityQueueMap

This is a new data structure I created that works sort of like a map, except that keys are mapped to priority queues of values. So a key can be bound to multiple values, but it's actual value(s) is the value with the highest priority.

This new data structure is used to bind generic type names to their explicit type names, within the context of a block. It will probably also be used for generic type constraints, once I get there.

The reason we cannot implement type binding using a simple map is twofold:
1) We want outside devs to be able to override type bindings using fields or other UI they create (hence the priority).
2) We want to be able to bind generic types to multiple explicit types so that they can easily adapt to changing connections.

For example right now the generic type (T) of the Select Random block is unbound:
Unbound

Now it should be bound to Dog:
Dog

Now Mammal:
Mammal1
Mammal2
Mammal3

And now Cat:
Cat

Allowing a type to have multiple bindings feels to me like the best way to handle any binding being removed at any time, and forcing the generic type to act like one of its other types.

#### Generic Map

The NominalConnectionChecker decides when to bind and unbind types, but the GenericMap handles the consequences of that decision (basically).

The generic map has two things it keeps track of.
1) For a given generic type on a given block, what explicit type is it currently bound to?
2) For a given generic type on a given block, what generic types on other blocks are dependent on the explicit type of this type?

(1) is useful for finding the explicit type of a generic type so that we can perform connection checks. (2) is useful for unbinding types. It will also allow us to have bindings “flow” through dependent types, once I add that.

### Problems related to blockly APIs

So currently all of the binding is handled through event listeners, which causes two problems:
1) Generic checks don't act bound until we wait 1 ms for the event delay.
2) Generic checks don't get bound when loading from XML

Adding a way for a connection checker to get a callback immediately after connect & disconnect would solve both problems, but that's really only useful if you want to implement generics.

If we decide to stick with the events system I'll probably end up adding a Promises system to fix (1) and hooking into the FinishedLoading event for (2)

### DIFF
--- a/plugins/nominal-connection-checker/src/generic_map.js
+++ b/plugins/nominal-connection-checker/src/generic_map.js
@@ -36,8 +36,6 @@ export class GenericMap {
    *     to.
    */
   constructor(workspace) {
-    // TODO: Move to maps.
-
     /**
      * The workspace this GenericMap belongs to.
      * @type {!Blockly.Workspace}
--- a/plugins/nominal-connection-checker/src/generic_map.js
+++ b/plugins/nominal-connection-checker/src/generic_map.js
@@ -26,6 +26,18 @@ export const INPUT_PRIORITY = 100;
 export const OUTPUT_PRIORITY = 200;
 
 /**
+ * The minimum priority that can/should be passed to the GenericMap.
+ * @type {number}
+ */
+export const MIN_PRIORITY = 0;
+
+/**
+ * The maximum priority that can/should be passed to the GenericMap.
+ * @type {number}
+ */
+export const MAX_PRIORITY = Number.MAX_SAFE_INTEGER;
+
+/**
  * Defines a map where generic types can be bound to explicit types in
  * a given environment.
  */
--- a/plugins/nominal-connection-checker/src/generic_map.js
+++ b/plugins/nominal-connection-checker/src/generic_map.js
@@ -1,0 +1,239 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Defines the GenericMap.
+ */
+'use strict';
+
+import {PriorityQueueMap} from './priority_queue_map';
+
+/**
+ * The priority for binding an explicit type to a generic type based on an input
+ * of the block with the generic type.
+ * @type {number}
+ */
+export const INPUT_PRIORITY = 100;
+
+/**
+ * The priority for binding an explicit type to a generic type based on the
+ * output of the block with the generic type.
+ * @type {number}
+ */
+export const OUTPUT_PRIORITY = 200;
+
+/**
+ * Defines a map where generic types can be bound to explicit types in
+ * a given environment.
+ */
+export class GenericMap {
+  /**
+   * Constructs the GenericMap.
+   * @param {!Blockly.Workspace} workspace The workspace this GenericMap belongs
+   *     to.
+   */
+  constructor(workspace) {
+    // TODO: Move to maps.
+
+    /**
+     * The workspace this GenericMap belongs to.
+     * @type {!Blockly.Workspace}
+     * @private
+     */
+    this.workspace_ = workspace;
+
+    /**
+     * Map of block ids to PriorityQueMaps that define what the generic types
+     * of the block are bound to.
+     * @type {!Map<string, !PriorityQueueMap>}
+     * @private
+     */
+    this.dependenciesMap_ = new Map();
+
+    /**
+     * Map of block ids to objects that map generic type names to arrays of
+     * DependerInfo specifying what other blocks and generic types on those
+     * blocks depend on the generic type of the lookup block.
+     * @type {!Map<string, !Map<string, !Array<DependerInfo>>>}
+     * @private
+     */
+    this.dependersMap_ = new Map();
+  }
+
+  /**
+   * Returns the name of the explicit type bound to the generic type in the
+   * context of the given block, or undefined if the type is not bound.
+   * @param {string} blockId The block id that the generic type is
+   *     possibly bound in.
+   * @param {string} genericType The generic type to find the explicit binding
+   *     of.
+   * @return {undefined|string} The name of the explicit type bound to the
+   *     generic type in the context of the given block, or undefined if the
+   *     type is not bound.
+   */
+  getExplicitType(blockId, genericType) {
+    const priorityMap = this.dependenciesMap_.get(blockId);
+    if (!priorityMap) {
+      return undefined;
+    }
+    const types = priorityMap.getValues(genericType);
+    if (!types) {
+      return undefined;
+    }
+    // In the future we might add logic to figure out what the super type of all
+    // of the types is. But for now there should only be one type bound anyway.
+    return types[0];
+  }
+
+  /**
+   * Binds the given dependerType name to the dependencyType's explicit type in
+   * the context of the dependerId. Should only be called if the dependencyType
+   * is currently bound to an explicit type in the context of its block.
+   * Also associates info about the depender with the dependency so that
+   * dependers can be easily removed if the dependency block ever looses its
+   * explicit type.
+   * @param {string} dependerId The id of the block that is depending on the
+   *     dependency block.
+   * @param {string} dependerType The name of the generic type in the depender
+   *     block that we want to bind to the other type in the dependency block.
+   * @param {string} dependencyId The id of the block that the depender block
+   *     is depending on.
+   * @param {string} dependencyType The name of the generic type in the
+   *     dependency block that we want to bind the dependerType to.
+   * @param {number} priority The priority of the binding. Higher priority
+   *     bindings override lower priority bindings.
+   */
+  bindTypeToGeneric(
+      dependerId, dependerType, dependencyId, dependencyType, priority) {
+    if (!this.workspace_.getBlockById(dependerId)) {
+      throw Error(
+          'The depender id (' + dependerId + ') is not a valid block id');
+    }
+    if (!this.workspace_.getBlockById(dependencyId)) {
+      throw Error(
+          'The dependency id (' + dependencyId + ') is not a valid block id');
+    }
+
+    const explicitType = this.getExplicitType(dependencyId, dependencyType);
+    if (!explicitType) {
+      throw Error('The type ' + dependencyType + ' on block ' + dependencyId +
+          ' is not bound to an explicit type. The generic type must be bound' +
+          ' before another generic type can bind to it.');
+    }
+    this.bindTypeToExplicit(dependerId, dependerType, explicitType, priority);
+
+    let types = this.dependersMap_.get(dependencyId);
+    if (!types) {
+      types = new Map();
+      this.dependersMap_.set(dependencyId, types);
+    }
+    let dependers = types.get(dependencyType);
+    if (!dependers) {
+      dependers = [];
+      types.set(dependencyType, dependers);
+    }
+    dependers.push(new DependerInfo(dependerId, dependerType, dependencyType));
+  }
+
+  /**
+   * Binds the given genericType name to the explicitType name in the context
+   * of the blockId.
+   * @param {string} blockId The id of the block to bind the genericType within.
+   * @param {string} genericType The name of the generic type that we want to
+   *     bind to the explicit type.
+   * @param {string} explicitType The name of the explicit type we want to bind
+   *     the generic type to.
+   * @param {number} priority The priority of the binding. Higher priority
+   *     bindings ovveride lower priority bindings.
+   */
+  bindTypeToExplicit(blockId, genericType, explicitType, priority) {
+    let queueMap = this.dependenciesMap_.get(blockId);
+    if (!queueMap) {
+      queueMap = new PriorityQueueMap();
+      this.dependenciesMap_.set(blockId, queueMap);
+    }
+    queueMap.bind(genericType, explicitType, priority);
+
+    // TODO: Flow through all other connections if necessary.
+    //   Make sure to update them if we get a higher priority binding.
+  }
+
+  /**
+   * Unbinds the given dependerType name from the dependencyType's explicit type
+   * in the context of the dependerId.
+   * Also de-associates info about the depender from the dependency.
+   * @param {string} dependerId The id of the block that is depending on the
+   *     dependency block.
+   * @param {string} dependerType The name of the generic type in the depender
+   *     block that we want to unbind from the other type on in the dependency
+   *     block.
+   * @param {string} dependencyId The id of the block that the depender block
+   *     was depending on.
+   * @param {string} dependencyType The name of the generic type in the
+   *     dependency block that we want to unbind the dependerType from.
+   * @param {number} priority The priority of the binding to remove.
+   */
+  unbindTypeFromGeneric(
+      dependerId, dependerType, dependencyId, dependencyType, priority) {
+    const types = this.dependersMap_.get(dependencyId);
+    if (!types) {
+      return;
+    }
+    const dependers = types.get(dependencyType);
+    if (!dependers) {
+      return;
+    }
+    const index = dependers.findIndex((elem) => {
+      return elem.blockId == dependerId &&
+          elem.dependerType == dependerType &&
+          elem.dependencyType == dependencyType;
+    });
+    if (index != -1) {
+      const explicitType = this.getExplicitType(dependencyId, dependencyType);
+      this.unbindTypeFromExplicit(
+          dependerId, dependerType, explicitType, priority);
+      dependers.splice(index, 1);
+    }
+  }
+
+  /**
+   * Unbinds the given generic type name from the explicit type name in the
+   * context of the blockId.
+   * @param {string} blockId The the block to unbind the types in.
+   * @param {string} genericType The name of the generic type to unbind from the
+   *     explicit type.
+   * @param {string} explicitType The name of the explicit type to unbind from
+   *     the generic type.
+   * @param {number} priority The priority of the binding to remove.
+   */
+  unbindTypeFromExplicit(blockId, genericType, explicitType, priority) {
+    if (this.dependenciesMap_.has(blockId)) {
+      this.dependenciesMap_.get(blockId).unbind(
+          genericType, explicitType, priority);
+    }
+    // TODO: Flow through all other connections.
+  }
+}
+
+/**
+ * A class containing info about a depender. Used to tell the depender to stop
+ * depending on the block it is depending on.
+ */
+class DependerInfo {
+  /**
+   * Constructs a DependerInfo.
+   * @param {string} blockId The id of the depender block.
+   * @param {string} dependerType The type on the depender block that is
+   *     depending on a generic type on another block.
+   * @param {string} dependencyType The generic type on the other block that the
+   *     dependerType is depending on.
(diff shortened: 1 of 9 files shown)


plugins/nominal-connection-checker/src/priority_queue_map.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  /**
+   * Removes the given value with the associated priority from the key's
+   * priority queue of values.
+   * @param {*} key The key to unbind the value from.
+   * @param {*} value The value to unbind from the key.
+   * @param {number} priority The priority of the binding.
+   */
+  unbind(key, value, priority) {
```
rachel-fenichel, line 90: Is it possible to have multiple values with the same priority?

BeksOmega: It is possible to have:
* Multiple different values with different priorities.
* Multiple different values with the same priority.
* Multiple of the same value with different priority.
* Multiple of the same value with the same priority.

So basically: Yes. The PriorityQueueMap doesn't really have any fancy behavior atm, it's just a wrapper for a map where all of the values are priority queues.

rachel-fenichel:
Okay, I'm diving back into this and will have a review for you today. Sorry about the delay!

plugins/nominal-connection-checker/src/index.js
```diff
@@ -42,21 +60,101 @@ export class NominalConnectionChecker extends Blockly.ConnectionChecker {
    */
   init(hierarchyDef) {
     this.typeHierarchy_ = new TypeHierarchy(hierarchyDef);
+    // TODO: Check if we need to release this.
```
rachel-fenichel: Handled in workspace.dispose.

plugins/nominal-connection-checker/src/index.js
```diff
@@ -42,21 +60,101 @@ export class NominalConnectionChecker extends Blockly.ConnectionChecker {
...
+
+    const genericMap = this.getGenericMap();
+    let parentBlock;
+    let parentCon;
+    let explicitFn;
+    let genericFn;
+    if (e.newParentId) {
+      parentBlock = this.workspace_.getBlockById(e.newParentId);
+      parentCon = parentBlock.getInput(e.newInputName).connection;
+      explicitFn = GenericMap.prototype.bindTypeToExplicit.bind(genericMap);
```
rachel-fenichel, line 108: Why are you using a bind instead of just calling the function, given that you also have the object around?

If the goal is only to make it possible to directly call the correct function, I think it would be clearer to set a boolean saying which case it is and then have a helper that checks that boolean and calls the appropriate function.

plugins/nominal-connection-checker/src/generic_map.js
```diff
@@ -0,0 +1,237 @@
...
+   * @return {undefined|string} The name of the explicit type bound to the
+   *     generic type in the context of the given block, or undefined if the
+   *     type is not bound.
+   */
+  getExplicitType(blockId, genericType) {
+    const priorityMap = this.dependenciesMap_.get(blockId);
+    if (!priorityMap) {
+      return undefined;
+    }
+    const types = priorityMap.getValues(genericType);
```
rachel-fenichel, line 92: Is types guaranteed to have at least one entry if it's not null?

rachel-fenichel (approved):
This makes sense overall, though it feels like the interactions are extremely complicated.

With respect to the deep generic nesting, I do wonder if you instead should make it so that you can't do Identity -> Identity -> Identity -> Dog. What if you blocked nested identity blocks entirely, so something has to be bound more immediately? (I have not thought all the way through this for all of the different use cases).

You may want to export a MIN_PRIORITY and MAX_PRIORITY for the generic map.

plugins/nominal-connection-checker/src/index.js
BeksOmega: Yeah this is a really tricky spot, and I still have mixed feelings about it.

The reason I originally aliased these functions instead of just calling them is because I wanted to try and keep the logic at the end of the function as simple as possible. All it cares about is the state of the connections (**E**, **GU**, or **GB**) so I wanted to move the other info about binding vs unbinding somewhere else.

Adding checks for binding vs unbinding at the end of the function would definitely work, but my current feeling is that it would distract from the logic that's really important. I know it's annoying to refer back to the actual function that's getting called, but the idea is that you shouldn't have to do that often because again, that info isn't really important. What's important is when it calls the generic function vs when it calls the explicit function.

That's just my current feeling on it though. As this function expands (especially as I add logic for flowing bindings through connections) something else may be better.

plugins/nominal-connection-checker/src/generic_map.js
BeksOmega: I checked, and the PriorityQueueMap handles that [here](https://github.com/google/blockly-samples/pull/376/files/7d8071b16dd6ee5f8713801a4e4fb84c8d10d085). I'm not sure if I handled it correctly for the dependersMap_ though so I'll have to check. And I'll add some tests as well!

BeksOmega:
> Okay, I'm diving back into this and will have a review for you today. Sorry about the delay!

I started classes last week so this is actually a really good pace for me hehe.

> This makes sense overall, though it feels like the interactions are extremely complicated.

They are quite complicated, which makes me a bit nervous lol. I'm worried that once I'm done, you and I are going to be the only ones that understand this system. Is there anywhere you think inline comments would be useful? Or somewhere I could write dev docs for it?

> With respect to the deep generic nesting, I do wonder if you instead should make it so that you can't do Identity -> Identity -> Identity -> Dog. What if you blocked nested identity blocks entirely, so something has to be bound more immediately? (I have not thought all the way through this for all of the different use cases).

Hmmm are you saying that we should allow:

but disallow (ignoring the fact that this is allowed to connect to the Launch Flying Animal block when it should not be):

Or are you saying that we should make it impossible to chain identity blocks - no matter what order you connect them in?

> You may want to export a MIN_PRIORITY and MAX_PRIORITY for the generic map.

Coolio, what do you think they should be? Right now all values from min int to max int are legal

rachel-fenichel:
I was thinking that we you could make it impossible to chain identity checks regardless of the order.

I suggest 0 for a min priority. Int max for the max is fine.

As far as the complicated code goes--not sure, it's possible it'll be easier to understand when more of the pieces are in place.

rachel-fenichel (approved):

rachel-fenichel:
Approved--go ahead and merge when you're ready.

BeksOmega:
> I was thinking that we you could make it impossible to chain identity checks regardless of the order.

Ok cool! That's what I figured you were talking about, but I wanted to make sure.

I got really excited abou this idea (because it would simplfiy things /a lot/), so as a result I was thinking about it a lot. I think disallowing the connection of multiple generics could work well for simple generics like the ones I'm currently building, but I realized that it might run into problems once we get to parameterized types.

For example if you want to do array processing, you often only specify the type of the array when you define it. Here is some sudo-go that hopefully demonstrates.

```
// Takes in an age and returns a function that takes in an Animal and
// returns true if the Animal is greater than the given age.
func GreaterThanAge(age int) func(Animal) bool { }

// Takes in a weight and returns a function that takes in an Animal and
// returns true if the Animal is greater than the given weight.
func GreaterThanWeight(weight int) func(Animal) bool { }

// Takes in an array of type T and a function that takes in any type
// general-er than T that returns a boolean.
// This function returns an array of all of the elements that make the
// function return true. The type of the array is the original type.
func Filter(type T, U >: T)(array []T, fn func(U) bool) []T { }

// Creates an array of Dogs.
dogArray := []Dog{dog1, dog2, dog3}

// Removes all of the dogs that are less <= 5 years old and <= 10 pounds.
// As you can see the Dog type was only given once above, but filteredDogs
// should still be an array of type Dog.
filteredDogs := Filter(Filter(dogArray, GreaterThanAge(5)), GreaterThanWeight(10))
```

And here is that same thing but in sudo-blocks:
FiltersDemo
We want the Dog types to be able to flow through the generic List and Filter blocks.

So I think that if we /can/ get it to work we should. And I'm pretty sure we can get it there. Currently I've got the connecting all working, so I've (hopefully) just got to clean up some stuff with the disconnecting. We'll just have to wait and see!

BeksOmega:
@moniika It looks like I might have messed something up with the field date tests... The results from travis look weird so here's what I get locally as well:

<details>
<summary>Console output</summary>
Ran at 19:09:17 PT


```
  1) FieldDate
       Constructor
         Empty:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at assertFieldDefault (webpack-internal:///./test/field_date_test.mocha.js:64:5)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:147:7)

  2) FieldDate
       Constructor
         Undefined:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at assertFieldDefault (webpack-internal:///./test/field_date_test.mocha.js:64:5)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:98:7)

  3) FieldDate
       Constructor
         Null:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at assertFieldDefault (webpack-internal:///./test/field_date_test.mocha.js:64:5)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:98:7)

  4) FieldDate
       Constructor
         NaN:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at assertFieldDefault (webpack-internal:///./test/field_date_test.mocha.js:64:5)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:98:7)

  5) FieldDate
       fromJson
         Empty:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at assertFieldDefault (webpack-internal:///./test/field_date_test.mocha.js:64:5)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:186:7)

  6) FieldDate
       fromJson
         Undefined:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at assertFieldDefault (webpack-internal:///./test/field_date_test.mocha.js:64:5)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:98:7)

  7) FieldDate
       fromJson
         Null:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at assertFieldDefault (webpack-internal:///./test/field_date_test.mocha.js:64:5)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:98:7)

  8) FieldDate
       fromJson
         NaN:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at assertFieldDefault (webpack-internal:///./test/field_date_test.mocha.js:64:5)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:98:7)

  9) FieldDate
       setValue
         Empty -> New Value
           Undefined:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:227:7)

  10) FieldDate
       setValue
         Empty -> New Value
           Null:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:227:7)

  11) FieldDate
       setValue
         Empty -> New Value
           NaN:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:227:7)

  12) FieldDate
       setValue
         Empty -> New Value
           Non-Parsable String:

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:227:7)

  13) FieldDate
       setValue
         Empty -> New Value
           Invalid Date - Month(2020-13-20):

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:227:7)

  14) FieldDate
       setValue
         Empty -> New Value
           Invalid Date - Day(2020-02-32):

      Value
      + expected - actual

      -2020-09-30
      +2020-10-01
      
      at assertFieldValue (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:75:47)
      at Context.eval (webpack-internal:///./node_modules/@blockly/dev-tools/src/field_test_helpers.mocha.js:227:7)
```

</details>

I've got no clue what it is.

moniika:
> @moniika It looks like I might have messed something up with the field date tests... The results from travis look weird so here's what I get locally as well:
> 
> Console output
> I've got no clue what it is.

I know what it is. It's on me. I made a mistake in the tests (with how I get the "expected" default value for the field. I'm going to make a PR to fix the test.

moniika:
> > @moniika It looks like I might have messed something up with the field date tests... The results from travis look weird so here's what I get locally as well:
> > Console output
> > I've got no clue what it is.
> 
> I know what it is. It's on me. I made a mistake in the tests (with how I get the "expected" default value for the field. I'm going to make a PR to fix the test.

I think my PR should fix the Travis failures. The issues you have locally (with the date being off by 1 day) I thought were already fixed. Let me know if after merging if it works locally.

BeksOmega:
> I think my PR should fix the Travis failures. The issues you have locally (with the date being off by 1 day) I thought were already fixed. Let me know if after merging if it works locally.

Yep yep it works perfectly! I just forgot to rebase hehe. Thank you so much :D

PR #381, opened 2020-09-24
Fix bugs with field test helpers

moniika:
- Supports passing empty string as expectedText in assertFieldValue
- Fixes creation tests `runConstructorSuiteTests ` and `runFromJsonSuiteTests` for case where optional parameter `assertFieldDefault` is not passed (and invalid test cases are expected to throw)

### DIFF
--- a/plugins/dev-tools/src/field_test_helpers.mocha.js
+++ b/plugins/dev-tools/src/field_test_helpers.mocha.js
@@ -61,7 +61,9 @@ export function assertFieldValue(field, expectedValue,
     expectedText = undefined) {
   const actualValue = field.getValue();
   const actualText = field.getText();
-  expectedText = expectedText || String(expectedValue);
+  if (expectedText === undefined) {
+    expectedText = String(expectedValue);
+  }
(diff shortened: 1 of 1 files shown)


plugins/dev-tools/src/field_test_helpers.mocha.js
```diff
@@ -61,7 +59,9 @@ export function assertFieldValue(field, expectedValue,
     expectedText = undefined) {
   const actualValue = field.getValue();
   const actualText = field.getText();
-  expectedText = expectedText || String(expectedValue);
+  if (expectedText === undefined) {
```
alschmiedt, line 62: would you want to do the same if expectedText is null

alschmiedt (approved):

plugins/dev-tools/src/field_test_helpers.mocha.js
moniika: I'm not sure if we would ever intentionally check that expectedText, but leaving it as this allows us to do so. My intention is just to check whether expectedText was passed.

PR #382, opened 2020-09-28
Set blockly version and make strict connection checker public

rachel-fenichel:


samelhusseini (approved):

plugins/strict-connection-checker/package.json
```diff
@@ -44,7 +43,7 @@
   "devDependencies": {
     "@blockly/dev-scripts": "^1.1.0",
     "@blockly/dev-tools": "^2.0.1",
-    "blockly": "git://github.com/google/blockly.git#develop",
+    "blockly": "^3.20200924.1",
     "chai": "^4.2.0"
   },
   "peerDependencies": {
```
samelhusseini, line 49: Since this won't work on earlier versions of Blockly. You'll want to add a min cap on peer dependencies as well. eg: ">3.20200924.1"

samelhusseini (changes requested):

plugins/strict-connection-checker/package.json
rachel-fenichel: Done.

samelhusseini (approved):

PR #384, opened 2020-09-29
Move test blocks into their own package

samelhusseini:
- Create a new package, @blockly/block-test
- Move all the test blocks into this package
- Add @blockly/block-test as a dependency of @blockly/dev-tools

Tested x-browser in Blockly playground with IE, Edge, and Chrome.


rachel-fenichel:
LGTM. Also add to the README in the plugins directory: https://github.com/google/blockly-samples/tree/master/plugins

PR #386, opened 2020-10-01
Add hierarchy validation checks for generic names - project static

BeksOmega:
### Description

Adds a check that informs the developer if the type hierarchy includes a type that would act like a generic type. Eg if we had a hierarchy like the following:
```json
{
  "a": { }
}
```
The validator would log an error to the console saying "The type a will act like a generic type if used as a connection check, because it is a single character."

### Testing

Added tests to make sure that errors are properly logged if a hierarchy definition containing generic types is passed.

### DIFF
--- a/plugins/nominal-connection-checker/src/index.js
+++ b/plugins/nominal-connection-checker/src/index.js
@@ -12,6 +12,7 @@
 import * as Blockly from 'blockly/core';
 import {TypeHierarchy} from './type_hierarchy';
 import {GenericMap, INPUT_PRIORITY, OUTPUT_PRIORITY} from './generic_map';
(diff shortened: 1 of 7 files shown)


plugins/nominal-connection-checker/src/index.js
```diff
@@ -216,6 +216,7 @@ export class NominalConnectionChecker extends Blockly.ConnectionChecker {
    */
   isGeneric_(connection) {
     const check = this.getCheck_(connection);
+    // If you ever update this, be sure to update hierarchy_validation.js too.
```
rachel-fenichel: It would be better to have these both call the same function. You could make it a static function on GenericMap, or else move it out into a ehlper some other way.

rachel-fenichel (approved):
Looks good.

PR #392, opened 2020-10-06
Editable field date

### LINKED ISSUE #343: Editable date field coexisting with datepicker

I would like to have some sort of a two way binding between the date field input and the date picker, in a way that the text input, today readonly, would be digitable.

The user would be able to type the date, or pick in the dropdown calendar.

My idea is to make the blockly 'field_date' field input digitable, add a mask so the user can type a valid date there, and this date will reflect on the calendar. As the same, calendar date changes would reflect on the field (like today's).

n/a

**Additional context**

I've attached a screenshot of the current date field below, imagine typing in the field and the datepicker reflecting your changes, just like Angular Material's datepicker do =D

[current blockly's datepicker](https://i.imgur.com/d69MNEL.png)
[Angular Material's date picker (two way binding with the input is the perfect example for what I expect)](https://i.imgur.com/jgiH50v.mp4)

moniika:
### Details
Fixes #343

Adding text edit functionality to field-date plugin.
- Changing base class of FieldDate from Field to FieldTextEdit
- Adding option in constructor for enable/disable of text edit
- Showing text editor based on whether text edit is enabled
- Adding logic to handle text editor/date picker interactions
  - Hide text editor when date is selected from picker
  - Show date picker on click of field (to allow for re-openining date picker after it was used)
  - Update date picker on valid change of text input

### Screenshots
dc6f8d35-446c-4f14-9e8e-0699dcd0d66c

### Additional Information
While testing this field, found it was also affected by https://github.com/google/blockly/issues/4345.

### DIFF
--- a/plugins/field-date/src/field_date.js
+++ b/plugins/field-date/src/field_date.js
@@ -34,10 +34,11 @@ goog.require('goog.ui.DatePicker');
  * @param {Function=} opt_validator A function that is called to validate
  *    changes to the field's value. Takes in a date string & returns a
  *    validated date string ('YYYY-MM-DD' format), or null to abort the change.
+ * @param {?(boolean|string)=} opt_textEdit Whether to enable text editor.
  * @extends {Blockly.Field}
  * @constructor
  */
-Blockly.FieldDate = function(opt_value, opt_validator) {
+Blockly.FieldDate = function(opt_value, opt_validator, opt_textEdit = false) {
   /**
    * The default value for this field (current date).
    * @type {*}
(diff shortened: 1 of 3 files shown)


plugins/field-date/src/field_date.js
```diff
@@ -167,24 +175,50 @@ Blockly.FieldDate.prototype.updateEditor_ = function() {
 };
 
 /**
- * Create and show the date field's editor.
+ * Shows the inline free-text editor on top of the text along with the date
+ *    editor.
```
alschmiedt: nit: I don't think this needs to be indented.

alschmiedt (approved):
small nit, but otherwise lgtm

plugins/field-date/src/field_date.js
moniika: Fixed

PR #394, opened 2020-10-08
ContinuousToolbox updates selected category as flyout is scrolled

maribethb:
When flyout is scrolled, the corresponding category in the toolbox is highlighted as selected. Also added some padding to the bottom of the flyout so that when the last category is clicked, you can actually scroll to the top of it. And fixed a bunch of the types that changed in the blockly core flyout code that had changed since last time I worked on this. For now the only type errors are related to the optional fields for FlyoutItemInfo which Abby just fixed in https://github.com/google/blockly/pull/4355 so will error until those changes make it to master.

Also, I ruined everything by merging master into this branch despite knowing I was going to be merging into the toolbox_plugin branch. It's going to make this review messy, but I promise the only files I actually changed are under plugins/continuous-toolbox/src/ (so my two non-merge/publish commits). Sorry :(


plugins/continuous-toolbox/src/ContinuousToolbox.js
```diff
@@ -24,30 +25,41 @@ export class ContinuousToolbox extends Blockly.Toolbox {
...
   }
 
   /**
    * Gets the contents that should be shown in the flyout immediately.
    * This includes all blocks and labels for each category of block.
-   * @return {!Array<!Blockly.utils.toolbox.FlyoutItem>} Flyout contents.
+   * @return {!Blockly.utils.toolbox.FlyoutItemInfoArray} Flyout contents.
    */
   getInitialFlyoutContents() {
+    /** @type {!Array<!Blockly.utils.toolbox.FlyoutItemInfo>} */
```
alschmiedt: I think you can use `!Blockly.utils.toolbox.FlyoutItemInfoArray` here

plugins/continuous-toolbox/src/ContinuousToolbox.js
```diff
@@ -58,13 +70,58 @@ export class ContinuousToolbox extends Blockly.Toolbox {
...
+
+  /**
+   * Gets a category by name.
+   * @param {string} name Name of category to get.
+   * @return {?Blockly.ToolboxCategory} Category, or null if not
+   *    found.
+   */
+  getCategoryByName(name) {
+    const category = this.contents_.find(
+        (item) => item instanceof Blockly.ToolboxCategory &&
```
alschmiedt, line 95: Not for this PR but I think this means we should have the kind stored on the toolbox item.

plugins/continuous-toolbox/src/ContinuousFlyout.js
```diff
@@ -72,6 +86,42 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
...
+    }
+  }
+
+  /**
+   * Scrolls flyout to given position.
+   * @param {number} position The x coordinate to scroll to.
+   */
+  scrollTo(position) {
+    // Set the scroll target to either the scaled position or the lowest
+    // possible scroll point, whichever is smaller
```
alschmiedt: nit: add .

plugins/continuous-toolbox/src/ContinuousFlyout.js
```diff
@@ -41,13 +42,26 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
     this.autoClose = false;
   }
 
+  /**
+   * Gets parent toolbox.
+   * Since we registered the ContinuousToolbox, we know that's its type.
+   * @return {!ContinuousToolbox} Toolbox that owns this flyout.
+   * @private
+   */
+  getParentToolbox() {
```
alschmiedt: Since private add _

plugins/continuous-toolbox/src/ContinuousToolbox.js
```diff
@@ -24,30 +25,41 @@ export class ContinuousToolbox extends Blockly.Toolbox {
...
+    return /** @type {ContinuousFlyout} */ (super.getFlyout());
   }
 
   /**
    * Gets the contents that should be shown in the flyout immediately.
    * This includes all blocks and labels for each category of block.
-   * @return {!Array<!Blockly.utils.toolbox.FlyoutItem>} Flyout contents.
+   * @return {!Blockly.utils.toolbox.FlyoutItemInfoArray} Flyout contents.
    */
   getInitialFlyoutContents() {
```
alschmiedt: Does this need to be public?

plugins/continuous-toolbox/src/ContinuousFlyout.js
```diff
@@ -72,6 +86,42 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
...
+        return;
+      }
+    }
+  }
+
+  /**
+   * Scrolls flyout to given position.
+   * @param {number} position The x coordinate to scroll to.
+   */
+  scrollTo(position) {
```
alschmiedt, line 118: Does this need to be public?

plugins/continuous-toolbox/src/ContinuousFlyout.js
```diff
@@ -72,6 +86,42 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
     return null;
   }
 
+  /**
+   * Selects an item in the toolbox based on the scroll position of the flyout.
+   * @param {number} position Current scroll position of the workspace.
+   */
+  selectCategoryByScrollPosition(position) {
```
alschmiedt: Does this need to be public?

alschmiedt (commented):
lgtm, mostly nits and a few questions about visibility.

plugins/continuous-toolbox/src/ContinuousToolbox.js
maribethb: even if it were provided as a method or something, the type would still have to be refined here because getName is only available on the type ToolboxCategory. And just calling a method like isCategory() (if i'm understanding your comment correctly) wouldn't be enough to tell the compiler that the type of this is indeed ToolboxCategory.

alschmiedt: Ah, I thought since we were already having to cast to `Blockly.ToolboxCategory` below then the check from the compiler didn't matter up here. But that makes sense.

plugins/continuous-toolbox/src/ContinuousFlyout.js
maribethb: Made everything either private (with underscores) or package as needed. Thanks!

alschmiedt (approved):

PR #395, opened 2020-10-08
Adds a codelab about customizing a toolbox

alschmiedt:
Adds a codelab detailing various ways to customize a toolbox.


codelabs/custom_toolbox/custom_toolbox.md
```diff
@@ -0,0 +1,493 @@
...
+After defining your category you need to tell Blockly that it exists.
+Register your category by adding the below code to the end of `custom_category.js`.
+```js
+Blockly.registry.register(
+    Blockly.registry.Type.TOOLBOX_ITEM,
+    Blockly.ToolboxCategory.registrationName,
+    CustomCategory, true);
+```
+By registering our `CustomCategory` with `Blockly.ToolboxCategory.registrationName`
+we are overriding the default category in Blockly. Because we are overriding a
```
maribethb, line 101: This makes me think we should call that flag something like allowOverrides, which might make it easier to explain here.

codelabs/custom_toolbox/custom_toolbox.md
```diff
@@ -0,0 +1,493 @@
...
+
+To start, create a file named `custom_category.js` in the same folder as
+`playground.html`. Include your new file with a script tag:
+```
+<script src="custom_category.js"></script>
+```
+
+Note: you must include your custom code *after* including the Blockly library.
+
+## Define and register a custom category
```
maribethb: Minor detail: I feel like this could go on the same page as the previous step.

codelabs/custom_toolbox/custom_toolbox.md
```diff
@@ -0,0 +1,493 @@
...
+The `colour` passed in is calculated from either the `categorystyle` or the `colour`
+attribute set on the category definition.
+
+For example, the "Logic" category definition looks like: 
+```xml
+<category name="Logic" categorystyle="logic_category">
+...
+</category>
+```
+
```
maribethb, line 144: I feel like this is kind of left hanging where the color definition actually is, so if it's short enough code snippet you could add "and logic_category style is defined..." but then again you did say they should know a little about the toolbox defs. Maybe at least add a link to more info?

codelabs/custom_toolbox/custom_toolbox.md
```diff
@@ -0,0 +1,493 @@
...
+### Change the background of the category
+
+Now that everything is set up, we are going to expand the color of the category
+to cover the entire category div. To do this we are going to override the `addColourBorder_` method.
+
+Add the following code to your `CustomCategory` class.
+```js
+/** @override */
+addColourBorder_(colour){
```
maribethb, line 131: I think it would be nice to explain what this is used for in the default category, so they can get a feel for what options they can change. Just something like, "in the default ToolboxCategory, this method adds the strip of color next to the category name. We can override this method to change this behavior" in the previous paragraph. That helps explain why this method in particular exists and is the one to modify.

then the same for the other overrides like "by default, setSelected changes the background color to white if deselected, or blah blah blah" just to give another feel for how the default stuff works.

codelabs/custom_toolbox/custom_toolbox.md
```diff
@@ -0,0 +1,493 @@
...
+### The result
+If you open your playground, you should see a white gear above your "Logic"
+label, and it should change to blue when the category has been selected.
+
+![](./category_gear.png)
+
+![](./category_gear_selected.png)
+
+## Change the category HTML
+Currently, the element for the icon on the category is a `<span>` element. For a
```
maribethb: This section is a bit abstract right now and it's not clear what differentiates it from the previous section. What about something like
"If you only need to change the css, like we did in the previous section, then using the cssConfig like we did is a great choice. But if you need to change the html, maybe to add text, an image, or anything else, you can override the corresponding method that creates the dom. In this example, we'll use an < i > element for the icon by overriding the createIconDom_ method." and maybe link to either the class itself or if it's explained in the docs.

I think if you wanted to make it even more different then you could change this to make a more substantial change like use an img instead of an icon. The point here isn't that currently it's a span and maybe they might not want it to be, it's that they have control over the html regardless. maybe they still want a span but with other stuff inside it.

I also think it might make sense to move the css stuff first, before the colourBorder stuff, so that you're introducing the easier content first. If you only need to change the css, here's how to do that. if you need to change the html too then here's that. But that might make the setSelected section too complicated to explain in that order so I think this is fine too.

maribethb (commented):
I didn't look at the code yet but here are my initial thoughts. Overall it looks great and a super helpful intro to all the changes you did!

codelabs/custom_toolbox/custom_toolbox.md
alschmiedt: Done

codelabs/custom_toolbox/custom_toolbox.md
alschmiedt: Done.

codelabs/custom_toolbox/custom_toolbox.md
alschmiedt: It's really short, so I added both : )

codelabs/custom_toolbox/custom_toolbox.md
alschmiedt: Ok! Updated this section to be an image instead, but I'm not entirely sure what css part I should move up?

alschmiedt:
I think I addressed most of the PR comments, but I also moved around some of the files so the format is similar to the context menu codelab.

codelabs/custom_toolbox/custom_toolbox.md
alschmiedt: That makes sense to me, I'll make a separate PR to update it in Blockly.

codelabs/custom_toolbox/custom_toolbox.md
maribethb: I was suggesting a reordering of the pages to have
1. how to change css classes (the icon part)
2. how to override methods (color border part and icon img part)
3. how to add custom items (the rest of the codelab stays where you have it)

but I also think the current organization is fine and it makes sense with the way you introduce the changes to setSelected. because i think even in the css part you need to change setSelected for the final product to look right, so you have to introduce that concept anyway, and you've already done that so everything makes sense.

so tldr: you're good, ignore me :)

maribethb (approved):
Looks great! Matching the starter code setup from the codelab Rachel added helps too, good idea!

PR #397, opened 2020-10-12
Pull in code to recycle blocks

alschmiedt:
Summary:

- `clearOldBlocks_` gets called at the beginning of `flyout.show()`. This will add any blocks that can be recycled to the `this.recycledBlocks_` list.

- We use the `recycledBlocks_` list in `createBlock_` to bypass calling `domToBlock` if the block has been recycled. This is the expensive call that we are trying to avoid if possible.
- We empty the `recycledBlocks_` after we have used it.

Note: What makes a block recyclable is probably an application based decision, so this should be something that can be easily changed.

### DIFF
--- a/plugins/continuous-toolbox/src/ContinuousFlyout.js
+++ b/plugins/continuous-toolbox/src/ContinuousFlyout.js
@@ -39,6 +39,21 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
      */
     this.scrollAnimationFraction = 0.3;
 
+    /**
+     * A list of blocks that can be recycled.
+     * @type {!Array.<!Blockly.BlockSvg>}
+     * @private
+     */
+    this.recycleBlocks_ = [];
+
+    /**
+     * Whether to recycle blocks when refreshing the flyout. When false, do not
+     * allow anything to be recycled. The default is to recycle.
+     * @type {boolean}
+     * @private
+     */
+    this.recyclingEnabled_ = true;
+
     this.autoClose = false;
   }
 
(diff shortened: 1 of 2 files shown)


plugins/continuous-toolbox/src/ContinuousFlyout.js
```diff
@@ -185,4 +200,152 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
...
+          if (field.isOptionListDynamic()) {
+            return false;
+          }
+        }
+      }
+      // Check children.
+      if (input.connection) {
+        const targetBlock = /** @type {Blockly.BlockSvg} */
+            (input.connection.targetBlock());
+        if (targetBlock && !this.blockIsRecyclable_(targetBlock)) {
```
rachel-fenichel, line 311: Is `this` bound correctly at this point? Please check that calling `this.blockIsRecyclable_` here actually works as intended if you've also called `setBlockIsRecyclable` to change the function.

plugins/continuous-toolbox/src/ContinuousFlyout.js
```diff
@@ -185,4 +200,152 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
...
+   *     otherwise.
+   * @public
+   */
+  setRecyclingEnabled(isEnabled) {
+    this.recyclingEnabled_ = isEnabled;
+  }
+
+  /**
+   * Puts a previously created block into the recycle bin and moves it to the
+   * top or the workspace. Used during large workspace swaps to limit the number
```
rachel-fenichel: or -> of

plugins/continuous-toolbox/src/ContinuousFlyout.js
```diff
@@ -185,4 +200,152 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
...
+   * of new dom elements we need to create.
```
rachel-fenichel: dom -> DOM

rachel-fenichel (approved):
Mostly looks good, with one thing to double check before merging.

plugins/continuous-toolbox/src/ContinuousFlyout.js
alschmiedt: done

plugins/continuous-toolbox/src/ContinuousFlyout.js
alschmiedt: done

plugins/continuous-toolbox/src/ContinuousFlyout.js
alschmiedt: Just tested re setting this method using `setBlockIsRecyclable` and the `this` looks like it is is bound correctly.

PR #404, opened 2020-10-14
Fix context menu bug in +/- blocks

BeksOmega:
### Description

Fixes the context menu bug brought to my attention by [this post](https://groups.google.com/g/blockly/c/KQS4cTwj4xo).

Also does some general cleanup related to types, conforming to the style guide, and removing TODOs.

[EDIT: Actually I'm going to move this to use for...of constructs]
[EDIT2: Done]

### Testing

1) Right clicked the block in the flyout. Observed how the only option was "help".
2) Dragged the block out, and right clicked it. Observed how there were no "Create 'get'" options.
3) Added two variables to the block, and right clicked it. Observed how there were two "Create 'get'" options.
4) Removed 1 variable from the block, and right clicked it. Observed how there was one "Create 'get'" option.
5) Removed the remaining variable from the block, and right clicked it. Observed how there were no "Create'get'" options.

### DIFF
--- a/plugins/block-plus-minus/src/procedures.js
+++ b/plugins/block-plus-minus/src/procedures.js
@@ -174,7 +174,7 @@ const procedureContextMenu = {
 
     // Add options to create getters for each parameter.
     const varModels = this.getVarModels();
-    for (let i = 0, model; (model = varModels); i++) {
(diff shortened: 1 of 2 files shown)


moniika (approved):
+1

PR #405, opened 2020-10-15
Fix issue with conflicting param names

BeksOmega:
### Description

While fixing #404 (hah) I discovered some bugs with parameter name conflicts.

If the new name became empty, callers would not be updated:
EmptyName

If the new name became conflicting, callers would not be updated:
ConflictingNames

This PR fixes those issues. It also reorganizes the validator a little bit because it was very messy.

### Testing
1. Tested that empty names update callers.
EmptyNameFixed

2. Tested that conflicting names update callers.
ConflictingNamesFixed

1. Tested that creating an empty name without update callers works fine.
EmptyNoErrors

2. Tested that creating a naming conflict without updating callers works fine.
ConflictingNamesNoErrors

### DIFF
--- a/plugins/block-plus-minus/src/procedures.js
+++ b/plugins/block-plus-minus/src/procedures.js
@@ -367,7 +367,8 @@ const procedureDefMutator = {
   addVarInput_: function(name, argId) {
     const nameField = new Blockly.FieldTextInput(name, this.validator_);
     nameField.onFinishEditing_ = this.finishEditing_.bind(nameField);
-    nameField.createdVarIds_ = [];
+    nameField.varIdsToDelete_ = [];
+    nameField.preEditVarModel_ = null;
 
     this.appendDummyInput(argId)
         .setAlign(Blockly.ALIGN_RIGHT)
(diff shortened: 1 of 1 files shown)


plugins/block-plus-minus/src/procedures.js
```diff
@@ -388,65 +389,75 @@ const procedureDefMutator = {
...
+    // Replace all whitespace with normal spaces, then trim.
+    newName = newName.replace(/[\s\xa0]+/g, ' ').trim();
+    const caselessName = newName.toLowerCase();
 
     /**
-     * Checks that all of the args (that aren't this arg) have a different
-     * name than this arg.
-     * @param {{model: Blockly.VariableModel, argId:string}} element
-     * @return {boolean} True if the other name is not a match.
+     * Returns true if the given argDatum is the argDatum associated with this
```
moniika: Would it still be correct if you changed this to:
"Returns true if the given argDatum is associated with this field, or has a different caseless name than the argDatum associated with this field."

plugins/block-plus-minus/src/procedures.js
```diff
@@ -388,65 +389,75 @@ const procedureDefMutator = {
...
+     * field, or has a different caseless name than the argDatum associated
+     * with this field. False otherwise.
+     * @param {{model: Blockly.VariableModel, argId:string}} argDatum The
+     *     argDatum we want to make sure is different from the argDatum
+     *     associated with this field.
```
moniika: "is different from" -> "does not conflict with"

moniika (approved):
Looks good, some minor comments about wording.

moniika:
@BeksOmega looks good. Could you also please resolve conflicts on procedures.js by merging in master.

BeksOmega:
> @BeksOmega looks good. Could you also please resolve conflicts on procedures.js by merging in master.

Sorry! I totally didn't notice, otherwise I would have fixed it.

I decided to resolve the conflict by rebasing, since that's what I'm used to. Do you peoples prefer merges though?

PR #408, opened 2020-10-15
Flesh out priority queue map interface - project static

BeksOmega:
### Description

Fleshes out the priority queue map interface so that it will (hopefully) be really robust, and not require changes in the future.

I think it's a good idea to do a "batch edit" like this because making modifications to lower-level interfaces while working on higher-level problems makes you more likely to create bugs in the lower-level interfaces.

Specifically this adds:
* getAllValues and getAllBindings
* unbindAll, unbindMatching, unbindAllMatching
* has, hasValue, hasMatchingValue
* filter
* forEach, forEachBinding

So it basically adds a robust set of has() and delete() equivalents, and adds some stream processing functions I think are important for all data structures.

It also makes some minor changes to existing functions.

### Testing

Each new function was given a suite of unit tests to document the behavior.

### DIFF
--- a/plugins/nominal-connection-checker/src/priority_queue_map.js
+++ b/plugins/nominal-connection-checker/src/priority_queue_map.js
@@ -72,7 +72,7 @@ export class PriorityQueueMap {
    *     the given key.
    */
   getAllValues(key) {
-    const bindings = this.getAllBindings();
+    const bindings = this.getAllBindings(key);
     return bindings && bindings.map((binding) => binding.value);
   }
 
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
Shoot--did I just drop this one on the floor?

BeksOmega:
Oh nope! In my latest branch I just deleted this class entirely (which really sucks because I like my implementation), so I closed the PR. No sense in making you review something that's going to get deleted later hehe

[Edit: Probably should have left a comment when I closed it. I'll remember to do that next time!]

PR #411, opened 2020-10-20
Redo generic flow - project static

BeksOmega:
### Decription

As stated [here](https://github.com/google/blockly-samples/issues/346#issuecomment-702985955) I ran into a problem with the binding method for handling generics. So I redid the logic and now it does a lookup every time we test a connection.

And trust me this is *so* much better. No more having to do special handling for xml, no more crazy generic map class, no more logic that only exists in Beka's head.

### Testing

Added more tests to cover the added getExplicitType and getExplicitTypeOfConnection functions.
Added tests for generic bindings flowing through connections.


rachel-fenichel (approved):
I agree that this is simpler, albeit at some performance cost. The type hierarchy would have to get pretty crazy for this to be a performance bottleneck.

PR #412, opened 2020-10-21
Make the workspace start at the edge of the flyout

maribethb:
Since the flyout is always open, we make the workspace start at the edge of the flyout instead of the edge of the toolbox. This prevents blocks from going under the always-open flyout (basically, there are the same edge cases there are for the existing toolbox like when the content is too wide, blocks can still be dragged partially under the flyout).

This required overriding two methods, a little heavyhanded for what we ultimately needed to change, but bugs have been or will be filed for improvements.

### DIFF
--- a/plugins/continuous-toolbox/src/ContinuousFlyout.js
+++ b/plugins/continuous-toolbox/src/ContinuousFlyout.js
@@ -185,4 +185,66 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
       this.selectCategoryByScrollPosition_(-this.workspace_.scrollY);
     }
   }
+
+  /**
+   * Overrides the position function solely to change the x coord in rtl mode.
+   * The base function allows the workspace to go "under" the flyout, so
+   * to calculate the left edge of the flyout in rtl you would just subtract
+   * the flyout width from the total viewWidth to get x. However, in our
+   * flyout, the workspace already starts at the left edge of the flyout, so
+   * we don't need to subtract the flyout width again.
+   * Ideally there would be a smaller method for us to override instead,
+   * but for now we copy/paste this method and make our fixes.
+   *  @override
+   */
+  position() {
+    if (!this.isVisible()) {
+      return;
+    }
+    const targetWorkspaceMetrics = this.targetWorkspace.getMetrics();
+    if (!targetWorkspaceMetrics) {
+      // Hidden components will return null.
+      return;
+    }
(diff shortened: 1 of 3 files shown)


maribethb:
Updated after merging, and updated eslint config to use closure mode.

plugins/continuous-toolbox/src/ContinuousFlyout.js
```diff
@@ -201,6 +201,68 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
...
+    const y = 0;
+    let x = 0;
+    // If this flyout is the toolbox flyout.
+    if (this.targetWorkspace.toolboxPosition == this.toolboxPosition_) {
+      // If there is a category toolbox.
+      if (targetWorkspaceMetrics.toolboxWidth) {
+        if (this.toolboxPosition_ == Blockly.TOOLBOX_AT_LEFT) {
+          x = targetWorkspaceMetrics.toolboxWidth;
+        } else {
+          x = targetWorkspaceMetrics.viewWidth;
```
alschmiedt, line 245: Maybe we should create an issue and add a TODO here so we don't forget that we need to refactor, and also so we can easily find what line we need to change

plugins/continuous-toolbox/src/ContinuousToolbox.js
```diff
@@ -28,6 +28,12 @@ export class ContinuousToolbox extends Blockly.Toolbox {
     const flyout = this.getFlyout();
     flyout.show(this.getInitialFlyoutContents_());
     flyout.recordScrollPositions();
+
+    // Replace workspace.getMetrics with a version that measures the flyout
```
alschmiedt: nit: Add .

plugins/continuous-toolbox/src/ContinuousToolbox.js
```diff
@@ -28,6 +28,12 @@ export class ContinuousToolbox extends Blockly.Toolbox {
...
+    // Ideally this would be set using the workspace options struct but that
+    // is not currently possible.
+    this.workspace_.getMetrics =
```
alschmiedt, line 37: Same as before to add a TODO

alschmiedt (approved):

plugins/continuous-toolbox/src/ContinuousFlyout.js
maribethb: Added todo comments!

PR #420, opened 2020-10-24
Remove conflicting css for tables

rachel-fenichel:
- Copied in the slate theme scss file
- Removed the config for using the slate theme
- Directly required the new css
- Removed css for table, tr, etc. from the slate theme

The slate css is fairly aggressive/non-specific. It was applying to the table inside the date picker, and making it look really bad.

This may also be because the order of css files is odd with the blockly page inserted into a section inside the main page, but I am not sure about that.

### DIFF
--- a/gh-pages/_config.yml
+++ b/gh-pages/_config.yml
@@ -1,4 +1,3 @@
-theme: jekyll-theme-slate
 show_downloads: false
 
 defaults:
--- a/gh-pages/_layouts/default.html
+++ b/gh-pages/_layouts/default.html
@@ -5,9 +5,7 @@
   <meta charset='utf-8'>
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width,maximum-scale=2">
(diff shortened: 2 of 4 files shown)


rachel-fenichel:
Friendly ping.

maribethb (approved):
I am concerned that if the current css can affect the blockly plugins, is just removing the table styles enough? Theoretically a plugin could have some other editor UI that interacts poorly so we'd just be playing whackamole. it seems like using a theme at all might not be the best fit, and we could apply more targeted styles to the areas that need them, using classes/ids so we don't interfere with Blockly. this is fine for now to at least unbreak the existing field plugin but might be something to consider.

rachel-fenichel:
@maribethb I agree with all of that.  I think the eventual move is to get rid of the MeyerWeb reset css entirely, and then either delete or make more specific all of the remaining CSS in this file.

The biggest thing this change does is make sure all of our CSS is defined locally rather than in a theme file that we don't have access to. Same with the HTML in the layouts. So now we have full control and can easily customize it.

PR #423, opened 2020-10-26
Update continuous-toolbox readme and package

maribethb:
Updates the README for continuous-toolbox. Also updates some attributes of the package.json such as name, version (to use semver), license string (it was not a valid string), etc.

Also adds a screenshot to the readme (img link won't work until this is merged to master, but you should be able to preview the image itself).


plugins/continuous-toolbox/README.md
```diff
@@ -1,46 +1,38 @@
...
-const plugin = new Plugin(workspace);
-plugin.init();
 ```
 
-## API
-
-<!--
-  - TODO: describe the API.
-  -->
-
```
alschmiedt, line 44: I think information on `setBlockIsRecyclable` can go here and eventually however we decide someone should be able to limit their flyout width.

alschmiedt (approved):

plugins/continuous-toolbox/README.md
maribethb: you're right, but i'll let you add the info for recycling if that's okay? if so you can add the heading back then.

alschmiedt: Yup, sgtm

maribethb:
Verified that installing from scratch works after requiring the latest version of dev-tools. I think after this we should be good to go merging into master.

alschmiedt (approved):

PR #428, opened 2020-10-29
Add support custom create function for field creation tests

moniika:
Allows for passing in custom logic for instantiating fields for creation tests (constructor and fromJson). Also added `call`s  to bind `this` context so that custom creation function can have access to `this` of mocha test (and can access properties like `this.workspace`).
Changes needed for https://github.com/google/blockly/pull/4417.

### DIFF
--- a/plugins/dev-tools/src/field_test_helpers.mocha.js
+++ b/plugins/dev-tools/src/field_test_helpers.mocha.js
@@ -28,7 +28,7 @@ FieldValueTestCase.prototype.expectedValue = undefined;
  */
 FieldValueTestCase.prototype.expectedText = undefined;
 /**
- * @type {RegExp|string|undefined} Optional error message matcher if test case
+ * @type {!RegExp|string|undefined} Optional error message matcher if test case
  *    is expected to throw.
  */
 FieldValueTestCase.prototype.errMsgMatcher = undefined;
@@ -51,7 +51,7 @@ FieldCreationTestCase.prototype.json = undefined;
 
 /**
  * Assert a field's value is the same as the expected value.
- * @param {Blockly.Field} field The field.
+ * @param {!Blockly.Field} field The field.
  * @param {*} expectedValue The expected value.
  * @param {string=} expectedText The expected text.
  */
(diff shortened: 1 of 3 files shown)


plugins/dev-tools/src/field_test_helpers.mocha.js
```diff
@@ -128,19 +128,24 @@ function runCreationTestsAssertThrows_(testCases, creation) {
  * @param {function(Blockly.Field)=} assertFieldDefault Asserts that field has
  *    default values. If undefined, tests will check that field throws when
  *    invalid value is passed rather than asserting default.
+ * @param {function(FieldCreationTestCase=)=} customCreateWithJs Custom creation
```
maribethb: I think FieldCreationTestCase needs a ! (or a ?) and same for most of this file, might be best to fix them all at once, if you want to.

plugins/dev-tools/src/field_test_helpers.mocha.js
```diff
@@ -128,19 +128,24 @@ function runCreationTestsAssertThrows_(testCases, creation) {
...
  */
 export function runConstructorSuiteTests(TestedField, validValueTestCases,
-    invalidValueTestCases, validRunAssertField, assertFieldDefault) {
+    invalidValueTestCases, validRunAssertField, assertFieldDefault,
+    customCreateWithJs) {
   suite('Constructor', function() {
     if (assertFieldDefault) {
       test('Empty', function() {
-        const field = new TestedField();
+        const field = customCreateWithJs ? customCreateWithJs.call(this) :
```
maribethb, line 140: just double checking, there's not supposed to be the fieldCreationTestCase parameter here?

maribethb (approved):

plugins/dev-tools/src/field_test_helpers.mocha.js
moniika: not here, there is no test case to pass.

plugins/dev-tools/src/field_test_helpers.mocha.js
moniika: Done.

PR #431, opened 2020-11-02
Add type unification - project static

BeksOmega:
### Description

Adds the `getNearestCommonParents` function to the type hierarchy. This function finds the nearest common super type of two or more types. This will allow me to add functionality for treating a generic multi-input block with multiple child blocks that share a common super type as that super type. For example, the below "Select Random" block is treated as a "Mammal" because its two child blocks have that supertype in common:
TypeUnification

To achieve this I had to add a pre-processing algorithm to the type hierarchy. The pre-processing runs in O(v*e) where v is the number of vertices in the graph and e is the number of edges. Note that this is longer than the O(v+e) time of the hierarchy validation, so we might want to consider doing validation by default, instead of requiring developers to call it separately.

I included a citation for the original research paper where I found the pre-processing algorithm in the source code, but I also created a [blog post](http://bekawestberg.me/blog/lowest-common-ancestor/) about it which I think explains the algorithm a bit more clearly. Or I hope it does anyway hehe.

Also note: this does **not** have anything to do with union types (eg (bool or string)). At the moment I'm not planning to implement any support for union types specifically.

### Testing
36 unit tests for the `getNearestCommonParents` function covering:
* All the different ways I could think of that types could be related (parents, parsibs, cousins, niblings, etc).
* Tests that require the logic to eliminate duplicates or common parents that are not the *nearest* common parent (see Harder graph unions).
* Tests where there are multiple equally near common parents (see Multiparent unions).

### DIFF
--- a/plugins/nominal-connection-checker/src/type_hierarchy.js
+++ b/plugins/nominal-connection-checker/src/type_hierarchy.js
@@ -29,20 +29,19 @@ export class TypeHierarchy {
 
     /**
      * Map of type names to maps of type names to sets of type names that are
-     * the least common ancestors of the two types. You can think of it like
+     * the nearest common parents of the two types. You can think of it like
      * a two-dimensional array where both axes contain all of the type names.
      *
-     * A least common ancestor of two types u and v is defined as:
+     * A nearest common parent of two types u and v is defined as:
      * A super type of both u and v that has no descendant which is also an
      * ancestor of both u and v.
      * @type {!Map<string, !Map<string, Array<string>>>}
      * @private
      */
-    this.leastCommonAncestors_ = new Map();
+    this.nearestCommonParents_ = new Map();
 
     this.initTypes_(hierarchyDef);
-    this.initLeastCommonAncestors_();
-    console.log(this.leastCommonAncestors_);
+    this.initNearestCommonParents_();
   }
 
   /**
@@ -72,7 +71,7 @@ export class TypeHierarchy {
   }
 
   /**
-   * Initializes the leastCommonAncestors_ graph so the least common ancestors
+   * Initializes the nearestCommonParents_ graph so the least common ancestors
    * of two types can be accessed in constant time.
    *
    * Implements the pre-processing algorithm defined in:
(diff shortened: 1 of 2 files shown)


BeksOmega:
It looks like Node 10 doesn't have the `flatMap` function. Luckily that can easily be changed to a `flat` and a `map`, but I'll wait for feedback before changing anything.

rachel-fenichel:
This looks good, so make that flatmap change and then I'll merge it.

However, a word of warning on algorithmic runtimes: you may be surprised by differences in runtime for different basic operations, based on optimizations in the javascript runtime. 

A specific example: I tried to use a [k-d tree](https://en.wikipedia.org/wiki/K-d_tree) to donearrest neighbor searches in the connection database. Neil benchmarked it and found that (at least at the time) using a k-d tree was slower than the current lookup algorithm because operations on bare arrays are extremely fast.

Should this stop you from using cool algorithms? Nope, but you may want to run benchmarks to check whether you're actually getting runtime improvements. And in some cases you may decide that it doesn't matter either way, because the numbers involved are small or the operation is infrequent. (Avoid premature optimization.)

BeksOmega:
> However, a word of warning on algorithmic runtimes: you may be surprised by differences in runtime for different basic operations, based on optimizations in the javascript runtime.

Thank you that's a really cool story! Benchmarking is something I'll definitely have to keep in mind with this project, because I do like to come up with "clever" solutions to things (which usually end up shooting me in the foot lol). I think doing those sanity checks will hopefully keep me from overcomplicating things (like I tried to do with the previous version of generic flow).

However in this case, this was actually the only algorithm I could find that was semi-decipherable, so I really hope it works out ok haha. I think I found some other algorithms for finding the LCA of a _tree_, but there wasn't a lot of info on DAGs. Maybe all of the good algorithms are hiding in textbooks, and haven't made it to the internet yet :D

But yeah, benchmarking is definitely something I'm going to keep in mind to try and check myself from doing crazy things! I'm trying to train myself that it's better to implement the simple thing _first_ so that you actually know if your optimizations are optimizations or not :D

PR #436, opened 2020-11-05
Tooltip block extension plugin

samelhusseini:
**New Plugin**

Name: @blockly/block-extension-tooltip
Description: A Blockly block extension that adds support for custom tooltip rendering.
Language: TS

Note: Currently requires a monkey patch in Blockly to support custom tooltip rendering.


plugins/block-extension-tooltip/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "name": "@blockly/block-extension-tooltip",
+  "version": "0.0.1",
+  "description": "A Blockly block extension that adds support for custom tooltip rendering.",
+  "scripts": {
+    "audit:fix": "blockly-scripts auditFix",
+    "build": "blockly-scripts build",
+    "clean": "blockly-scripts clean",
+    "dist": "blockly-scripts build prod",
+    "lint": "blockly-scripts lint",
+    "prepublishOnly": "npm run clean && npm run dist",
```
rachel-fenichel, line 12: Also needs a predeploy script. Is that not being added in create-package?

plugins/block-extension-tooltip/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "registry": "https://wombat-dressing-room.appspot.com"
+  },
+  "eslintConfig": {
+    "extends": "@blockly/eslint-config"
+  },
+  "engines": {
+    "node": ">=8.17.0"
+  },
+  "browserslist": [
+    "defaults",
```
rachel-fenichel: Please confirm that this works on IE 11, or else change this.

plugins/block-extension-tooltip/src/tooltip_monkey_patch.ts
```diff
@@ -0,0 +1,94 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview A monkey-patch in Blockly to support custom tooltip rendering.
+ */
+import * as Blockly from 'blockly/core';
+
+
+// TODO: We have no way to do this currently without monkey-patching Blockly.
```
rachel-fenichel: File an issue in blockly core and reference it here.

plugins/block-extension-tooltip/README.md
```diff
@@ -0,0 +1,63 @@
+# @blockly/block-extension-tooltip [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
+
+A [Blockly](https://www.npmjs.com/package/blockly) block extension that adds
```
rachel-fenichel, line 3: Add notes that:
- this is monkeypatching (until we make the correct change in core)
- this uses a block extension (link to the documentation)
- as a result, it has to be added to each block type

plugins/block-extension-tooltip/package.json
samelhusseini: Done, and added to the create-package script.

rachel-fenichel:
I think this is ready to merge, once you confirm that it works on IE 11.

plugins/block-extension-tooltip/package.json
samelhusseini: I can't test on IE because the playground currently doesn't load on there. Removed IE11 from the list in browserslist and filed https://github.com/google/blockly-samples/issues/500

PR #439, opened 2020-11-06
Theme codelab updates

alschmiedt:
Some unnecessary files got merged to the themes codelab.

Remove those files and update some indentation.


mmathrani (approved):
Thanks for making sure this works right!

PR #442, opened 2020-11-09
Support deploying custom demos / examples

samelhusseini:
Support deploying demos / examples to ghpages.

Any examples / demo under `examples/` can be configured with a `demoConfig` in `package.json` that tells the predeploy script what files to publish, and what tabs to include. The schema is as follows:

```
interface DemoConfig {
  title?: string, // Title of the demo / examples, appears in the toolbar and page title.
  description?: string, // Description of the demo, appears in the toolbar as a subtitle.
  files: string[] // The list of files that should be copied over into the publish directory
  pages?: Page[] // Some demos have multiple tabs (eg: README / Playground), or a number of different pages, define pages here.
}
interface Page {
   label: string;
   link: string;
}
```

The index page includes a search feature that runs Fuzzy search on each of the links in the page (index.md) and hides those that don't match.

Removed slate.scss
The styles of the pages match the styles of blocklycodelabs.dev, and are using the same footer which links to the developer reference.

Moved over all the demos (that don't rely on storage), and the blockly-react example.


examples/blockly-react/package.json
```diff
@@ -32,5 +33,13 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "homepage": "./",
+  "demoConfig": {
```
rachel-fenichel, line 38: I think we should use `blocklyDemoConfig` for our name here.

gulpfile.js
```diff
@@ -104,62 +108,122 @@ function publishDryRun(done) {
...
+    return function preDeployPlugin() {
       return preparePlugin(folder);
-    }
+    };
+  }))(done);
+}
+
+
+/**
+ * Copy over the test page (index.html and bundled js) and the readme for
```
rachel-fenichel: Update comment to refer to example instead of plugin.

gh-pages/js/index.js
```diff
@@ -0,0 +1,102 @@
+document.addEventListener('DOMContentLoaded', function() {
```
rachel-fenichel: Break some of this (such as the callbacks or the fuse setup) out into helper functions.

gh-pages/index.md
```diff
@@ -1,3 +1,6 @@
+---
+layout: index
+---
 
 # Plugins
```
rachel-fenichel, line 6: Add the strict-connection-checker.

examples/toolbox-demo/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "blockly-toolbox-demo",
+  "version": "0.0.0",
+  "private": true,
+  "demoConfig": {
```
rachel-fenichel: I'd like to have "see also" links for a lot of these ported demos. Do you think that belongs in the demoConfig, or should just go in the readme?

examples/plane-demo/README.txt
```diff
@@ -0,0 +1,26 @@
+This Blockly demo uses Closure Templates to create a multilingual application.
```
rachel-fenichel, line 1: See https://github.com/google/blockly/issues/2625

I suppose there's no harm in copying the plane demo over (since you've already done the work).

examples/custom-dialogs-demo/package.json
```diff
@@ -0,0 +1,13 @@
+{
```
rachel-fenichel, line 1: This one's interesting. Do we want to port it, given that we now have a modal plugin?

rachel-fenichel (commented):
I looked through everything in gh-pages, as well as looking at all package.json files. Looks good overall.

examples/blockly-react/package.json
samelhusseini: Done.

gulpfile.js
samelhusseini: Done.

gh-pages/index.md
samelhusseini: Done.

examples/toolbox-demo/package.json
samelhusseini: We can standardize anything we want and build it into the template, if we find that we want see also links in all demos, then it should live in the template and move into the blocklyDemoConfig.
I'll refrain from attempting that in this PR though as it's already quite large.

examples/plane-demo/README.txt
samelhusseini: Ah ok. Yeah copied over already, can hide from index.md if we don't want eyes on it.

examples/custom-dialogs-demo/package.json
samelhusseini: This is different to the modal plugin and I would actually not call it custom dialogs and more custom prompts. It's just overriding the built in prompts we use from browser alerts / confirms to dialogs. 
The modal plugin is more like a custom built in "modal UI" piece that can be used for different purposes (including possibly prompts.

I think this points at what we were talking about yesterday, the need to reassess our demos and see where they fit (potentially removing some / renaming some / adding additional descriptions / READMEs).

samelhusseini: For the purposes of this PR, I'll keep as is, however happy to remove from index.md if we want less eyes on it.

gh-pages/js/index.js
samelhusseini: Done.

rachel-fenichel (approved):

PR #458, opened 2020-11-11
Add tutorial

rachel-fenichel:
Adds the Tutorial and TutorialStep classes, and wires up the tutorial to show (with some arbitrary steps) when the tutorial button is pushed from the welcome modal.

Fixes lint in a variety of files, mostly by fixing line length or adding jsdoc.

Right now you go from step to step in the tutorial by pressing the next step button, but I wired it up so that there's a callback that the tutorial step can call as well. Next up is figuring out how to know when to call that. @alschmiedt can I just override the handler for the key press I care about, so that it does the normal action *and* calls out to my tutorial step? Then for each step I could reset the handler back to the original action and move on to the next step.

### DIFF
--- a/examples/keyboard-accessibility/src/help_modal.js
+++ b/examples/keyboard-accessibility/src/help_modal.js
@@ -11,7 +11,16 @@
 
 import {speaker} from './speaker';
 
+/**
+ * Modal to show a help menu to the user.
+ */
 export class HelpModal {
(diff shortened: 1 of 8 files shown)


alschmiedt:
Yes, if you add a keyboard shortcut and just don't return true then it will run both your function and the default function.

PR #495, opened 2020-11-22
Add kicking blocks - project static

BeksOmega:
### Description

Added functionality to kick blocks out of connections when they become invalid because of programmatic binding. Programmatic binding is meant to support developers creating fields or other UI that allow users to explicitly set types.

Also removes the generic map, priority queue map, and related files because they are officially irrelevant!

### Testing
Added tests to make sure that blocks get kicked (or don't!) where appropriate.


plugins/nominal-connection-checker/test/index.js
```diff
@@ -64,6 +64,7 @@ function createWorkspace(blocklyDiv, typeHierarchy, blocks) {
 }
 
 document.addEventListener('DOMContentLoaded', function() {
+  window.Blockly = Blockly;
```
samelhusseini: The playground already does this in https://github.com/google/blockly-samples/blob/master/plugins/dev-tools/src/index.js#L30
Curious why this is necessary here?

samelhusseini (approved):

samelhusseini:
The nominal-connection-checker plugin is still using the latest blockly. I assumed it was necessary as it was depending on changes in blockly/develop that hadn't been released yet. Is this still necessary after the latest release?

plugins/nominal-connection-checker/test/index.js
BeksOmega: You're right, I have no clue why I added this haha. Removed!

BeksOmega:
> The nominal-connection-checker plugin is still using the latest blockly. I assumed it was necessary as it was depending on changes in blockly/develop that hadn't been released yet. Is this still necessary after the latest release?

Thank you for the reminder! I had completely forgotten about fixing that. Should be switched over now =)

samelhusseini (approved):

PR #497, opened 2020-12-01
Fix minimum version of peer dependency to blockly

winterer:
The current peer dependency to blockly requires a version, that does not exist yet. As a result, the plugin fails to install.
Seems to by a typo.


winterer:
@googlebot I signed it!

alschmiedt:
Yup, this looks like a typo! Happy to merge this in once the CLA is signed.

PR #502, opened 2020-12-04
Create a plugin to disable 'disable' option on top-level orphan blocks

### LINKED ISSUE #369: Plugin: Allow only top blocks to be manually disabled
### Problem statement

I like the behaviour of Blockly.Events.disableOrphans, but it only makes sense in conjunction with the workspace disable option - to suppress the 'disable' item from the context menu.

I'd still like the user to be able to disable top blocks though via the menu.

I'm proposing an option to allow the disable menu item to still be available for just these blocks.

Maybe a third value for disable, eg: 'onlyTop'?

Blockly.BlockSvg.prototype.showContextMenu_ would check for this and check whether the block was a 'top' block (in the same way that Blockly.Events.disableOrphans does) to determine whether to add the menu item.

I'd be happy to supply a PR for this if you feel it's something useful enough to add to the core.

maribethb:
https://github.com/google/blockly-samples/issues/369 asks for the ability to allow the 'enable/disable' context menu option only on top-level non-orphan blocks, instead of removing it completely (the current recommendation). I've done that here except you can disable any block that isn't an orphan, even if it's not a top level block. This way you can still disable blocks inside functions, disable top-level function blocks, etc. The only thing you can't do is re-enable a block that has been disabled by the `disableOrphans` event handler, which I believe fits the spirit of the feature request and makes the most sense with the behavior of disableOrphans.

Hopefully I correctly grabbed all the new features of package.json (scripts and stuff)... please double check me on that if you can.

Closes #369.


plugins/disable-top-blocks/GETSTARTED.md
```diff
@@ -0,0 +1,24 @@
+This package was bootstrapped with [@blockly/create-package](https://www.npmjs.com/package/@blockly/create-package).
```
moniika: Should this file be removed? I see we do have it in nominal-connection-checked, but it seems like it is only used to help developers when they first start building their plugin.

plugins/disable-top-blocks/src/index.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+
+/**
+ * A block is an orphan if it doesn't have a parent, but it does have
+ * a previous or next connection (so it expects to be attached to something).
+ * @param {!Blockly.BlockSvg} block Block to check.
+ * @return {boolean} Whether the block is an orphan.
+ */
+function isOrphan(block) {
```
moniika, line 61: Does this also cover case where block is unconnected and has left output? Or is that not an orphan?

rachel-fenichel:
Also add it to the gh-pages index so there will be a link to it.

plugins/disable-top-blocks/src/index.js
maribethb: The code was correct, but not the comment. I updated the comment. Output connections or previous connections, plus no parent, make it an orphan. next connections are actually fine (e.g. a start block)

plugins/disable-top-blocks/GETSTARTED.md
maribethb: I see, I didn't realize this file was meant to be deleted. It could be useful if anyone makes other changes to this plugin in the future, but I guess we have general documentation for that and don't need to repeat it for each plugin.

maribethb (commented):
Added to gh pages index as well.

moniika (approved):

PR #506, opened 2020-12-10
Ui event api updates

moniika:
Updates interpreter and mirror demos to reflect Ui Event Api changes in core.

NOTE: should be merged after Blockly December release


maribethb (approved):
LGTM (for merging after release), though I guess based on our conversation today you should bump the version number in package.json and update the peerDependency blockly version number here in the same PR

moniika:
> LGTM (for merging after release), though I guess based on our conversation today you should bump the version number in package.json and update the peerDependency blockly version number here in the same PR

Since they're not published, I don't think we need to bump the version for theses. As for the dependency, I think these two both use unpkg and don't have peerDependency, so I think they should be ok without update.

PR #510, opened 2020-12-11
Update eslint

rachel-fenichel:
- Updated eslint at root and in dev-scripts.
- Updated peer dependency version of eslint in eslint-config.
- Updated required version of eslint-config to the current released version (2.0.4) for dev-scripts.
- Increased minimum node version from 8 to 10 for eslint-config, which I think we will need to apply more generally.

I can still run lint from the top level and have everything work, so I think this is okay.
Next steps:
- Bump versions for dev-scripts and eslint-config and publish
  - *???* Update dev-scripts to depend on the newly published version of eslint-config, and republish dev-scripts *???*


package.json
```diff
@@ -26,15 +26,15 @@
...
+        "eslint": "^7.15.0",
         "gh-pages": "^3.1.0",
         "gulp": "^4.0.2",
         "gulp-header": "^2.0.9",
         "js-green-licenses": "^1.1.0",
+        "json-to-pretty-yaml": "^1.2.2",
         "lerna": "^3.20.2",
         "rimraf": "^3.0.2",
-        "typescript": "^3.8.3",
-        "json-to-pretty-yaml": "^1.2.2"
```
samelhusseini, line 37: Why are you removing json-to-pretty-yaml. It's used in gulpfile.js

rachel-fenichel: I'm not--they just got alphabetized.

samelhusseini (approved):
LGTM, after reverting json-to-pretty-yaml change

package.json
samelhusseini: Oh true.

PR #517, opened 2020-12-15
Update to new ESLint class from eslint 7

rachel-fenichel:
This is a clumsy attempt at moving to the new async linter and I need some help with promises.

### DIFF
--- a/plugins/dev-scripts/scripts/lint.js
+++ b/plugins/dev-scripts/scripts/lint.js
@@ -26,7 +26,7 @@ console.log(`Running lint for ${packageJson.name}`);
 
 // Create the eslint engine.
 const eslintConfig = require('@blockly/eslint-config');
-const cli = new ESLint({
+const linter = new ESLint({
   extensions: ['.js', '.ts'],
   baseConfig: eslintConfig,
   useEslintrc: false,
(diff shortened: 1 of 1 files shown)


samelhusseini (approved):

samelhusseini:
Your use of promises and async/await looks fine to me. Was there something specific you needed help with?

plugins/dev-scripts/scripts/lint.js
```diff
@@ -25,21 +26,39 @@ console.log(`Running lint for ${packageJson.name}`);
 
 // Create the eslint engine.
 const eslintConfig = require('@blockly/eslint-config');
-const cli = new CLIEngine({
+const cli = new ESLint({
```
alschmiedt: Should be updated to eslint

samelhusseini: What do you mean, the variable above is named `ESLint`?

alschmiedt: I think cli should be eslint.  We are not using CLIEngine anymore so I don't think the name cli makes sense.

samelhusseini: Oh ok 👍

rachel-fenichel:
Re promises, I thought there's a better way to show that the task failed than to return null and parse it in the caller.

But if it's all good, that works too. This was just a step toward getting lint --fix to be available at the top level.

samelhusseini:
Returning null with async/await, is the equivalent of returning Promise.resolve(null) and you can do that if you like for clarity.

As for errors, the async/await way to return errors is to throw an Error. Although I wouldn't call the directory not existing as an error per se.

Issue #492, opened 2020-11-19
Update field-grid-dropdown Blockly dependency after core release


moniika:
**Category**
Plugins

**Component**
field-grid-dropdown

**Describe the issue**

The `peerDependencies` and `devDependencies` for `field-grid-dropdown` should be updated to the latest release of Blockly because it depends on an annotation visibility change made in https://github.com/google/blockly/pull/4453.

PR #518, opened 2020-12-15
Remove private from field-grid-dropdown

moniika:
Remove private from field-grid-dropdown


rachel-fenichel (approved):

moniika:
As per #492. This plugin should also be update to the latest version of blockly after release.

PR #520, opened 2020-12-15
Lint fix

rachel-fenichel:
First commit: Add a top level `lint:fix` script, and pipe the resulting `--fix` all the way into blockly-scripts to output lint fixes.
Second commit: Run the new script, fixing stuff like missing trailing commas.

### DIFF
--- a/plugins/nominal-connection-checker/src/hierarchy_validation.js
+++ b/plugins/nominal-connection-checker/src/hierarchy_validation.js
@@ -95,7 +95,7 @@ function checkSupersDefined(hierarchyDef) {
  * },
  * 'typeB': {
  *   'fulfills': ['typeA']
- * }
+ * }.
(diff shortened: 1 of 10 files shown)


alschmiedt (approved):

plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -95,7 +95,7 @@ function checkSupersDefined(hierarchyDef) {
  * },
  * 'typeB': {
  *   'fulfills': ['typeA']
- * }
+ * }.
```
alschmiedt, line 98: This is a bit weird, but I guess it is just trying to fulfill the rule that comments should be a complete sentence?

rachel-fenichel: Yeah I'm not in love with it but it doesn't seem worth fighting it either.

PR #534, opened 2021-01-02
Specifiy statement connection behavior - project static

BeksOmega:
### Description

This makes it so that statement connections act the same way that other connections do. I decided to make statement connections consistent because:
1) It makes connections work invariantly, which keeps the code simpler. This in turn makes it easier to test, modify, debug, etc.
2) It is less of a mental burden on outside developers.

I also went through and revamped all of the tests so that they test statement connections as well as input connections. This is done using functions that generate mocha suites, but in the future I'd like to change it so that they look more like suites themselves.

There was also a bugfix (really a logical error fix) with how bound types were being handled that got bundled into this PR.

### Testing

As described above, tests now test statement connections and input connections. This is mostly meant as documentation of the specification.

### DIFF
--- a/plugins/nominal-connection-checker/package-lock.json
+++ b/plugins/nominal-connection-checker/package-lock.json
@@ -4,6 +4,1123 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.12.11",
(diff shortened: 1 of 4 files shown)


rachel-fenichel (approved):
+1 to simplifying things/making them more consistent across connection types.

I looked at the tests but have to confess that I did not read every single line :)

PR #536, opened 2021-01-07
Orphan block's childrens are also orphans

maribethb:
Update the logic for whether a block is an orphan.

Test: 
npm run start in disable-top-blocks plugin
Drag an if-then block onto the ws.
Drag a second if-then block connected to the first.
They are both disabled automatically by disableOrphans.

Previously: Second block could be enabled through context menu because it technically had a parent
Now: Neither block can be enabled through context menu because they are both orphans.

### DIFF
--- a/plugins/disable-top-blocks/src/index.js
+++ b/plugins/disable-top-blocks/src/index.js
@@ -53,12 +53,13 @@ export class DisableTopBlocks {
 }
 
 /**
- * A block is an orphan if it doesn't have a parent, but it does have
- * a previous or output connection (so it expects to be attached to something).
+ * A block is an orphan if it doesn't have a parent (or its parent is disabled),
+ * but it does have a previous or output connection (so it expects to be
+ * attached to something).
  * @param {!Blockly.BlockSvg} block Block to check.
  * @return {boolean} Whether the block is an orphan.
  */
 function isOrphan(block) {
-  return !block.getParent() &&
+  return (!block.getParent() || !block.getParent().isEnabled()) &&
       !!(block.outputConnection || block.previousConnection);
 }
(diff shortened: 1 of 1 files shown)


moniika (approved):

BeksOmega:
I think this causes an issue with blocks that are manually disabled :/
ManuallyDisabled

To reproduce:
1) Create a function block with two child if blocks (as shown above)
2) Disable the bottom if block.
3) Disable the top if block.
4) Right click the bottom if block.
5) Observe how you cannot re-enable it.

maribethb:
Good catch, Beka, thanks. I've updated it so all children blocks of an orphan block are also orphans. This catches the original case, but also allows Beka's case to work correctly. I didn't want to have to check all the parents recursively but I don't think there's another way. I tested this with "spaghetti" code and even with deeply nested blocks, the context menu is still generated instantly, so I think it's okay.

plugins/disable-top-blocks/src/index.js
```diff
@@ -53,12 +53,20 @@ export class DisableTopBlocks {
...
  */
 function isOrphan(block) {
+  // If the parent is an orphan block, this block should also be considered
+  // an orphan so it cannot be manually re-enabled.
+  if (block.getParent() &&
+      isOrphan(/** @type {Blockly.BlockSvg} */ (block.getParent()))) {
+    return true;
+  }
   return !block.getParent() &&
       !!(block.outputConnection || block.previousConnection);
```
moniika: Since `block.getParent()` is called multiple times, I think it would benefit to add a variable `parent` or `parentBlock`

```
const parentBlock = block.getParent();
```

moniika (approved):
minor comment

plugins/disable-top-blocks/src/index.js
maribethb: Good call, done.

PR #539, opened 2021-01-21
Reorganize tests - project static

BeksOmega:
### Description & Tests!

Apologies in advance, because this is going to be pretty messy to review.

I went through and reorganized the tests to make sure each one had a single, clear assertion. And as I did I also created tests for any "gaps" in testing I identified.

I also added better logging for some of the less clear errors. And I refactored the test helpers so that there is less duplicate code.

Sorry this isn't the most detailed description, but it's been almost a couple of months since I wrote this hehe. If there's anything confusing just ask and I'll do my best to remember why I did what I did!

### DIFF
--- a/plugins/nominal-connection-checker/test/connection_checker_test_helper.mocha.js
+++ b/plugins/nominal-connection-checker/test/connection_checker_test_helper.mocha.js
@@ -18,7 +18,6 @@
 export function createBlockDefs(types) {
   const blocks = [];
   for (let type of types) {
-    type = type.toLowerCase();
     blocks.push({
       'type': 'static_' + type + '_outer_value',
       'message0': '%1',
@@ -207,7 +206,6 @@ export function runTwoBlockTests() {
    */
   function createGetOuterInput(suffix) {
     return function(type, name) {
-      type = type.toLowerCase();
       const block = this.workspace.newBlock('static_' + type + suffix);
       const inConn = block.nextConnection ||
           block.getInput('INPUT1').connection;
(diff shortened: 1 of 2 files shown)


plugins/nominal-connection-checker/test/connection_checker_test_helper.mocha.js
```diff
@@ -199,58 +198,59 @@ export function clearTwoBlockTests() {
  * called at the end of a suite of two block tests.
  */
 export function runTwoBlockTests() {
+  /**
+   * Creates a function which returns the input of an outer block.
```
rachel-fenichel: Nit: Creates a function which creates a new block and returns its next connection or first input connection.

plugins/nominal-connection-checker/test/connection_checker_test_helper.mocha.js
```diff
@@ -199,58 +198,59 @@ export function clearTwoBlockTests() {
...
+      const block = this.workspace.newBlock('static_' + type + suffix);
+      const inConn = block.nextConnection ||
+          block.getInput('INPUT1').connection;
+      inConn.name = name || type + 'In';
+      return inConn;
+    };
+  }
+
+  /**
+   * Creates a function which returns the output of an inner block.
```
rachel-fenichel: Creates a function which creates a new block and returns its output connection or previous connection.

rachel-fenichel (approved):
Okay, I took a shot at reading through all of the tests and they look reasonable (every one I walked all the way through was doing what it claimed it was doing, and the suites look good).

That said, it's a fairly large pile of tests, so there's always a chance I missed something.

Approve, but we're having Travis issues that are blocking merges. (Same story on core.)

PR #542, opened 2021-01-22
Keyboard nav init

alschmiedt:
Initial commit for keyboard navigation. Files that are not just from running the plugin script: 

- package.json
- flyout_cursor
- line_cursor

The [flyout cursor](https://github.com/google/blockly/blob/master/core/keyboard_nav/flyout_cursor.js) and the [line cursor](https://github.com/google/blockly/blob/52b818ed67c5bfa7fd73edf2b067649b8b1f447b/demos/keyboard_nav/line_cursor.js) were almost entirely copied over from core.  The rest of the files will be modified in following PRs.

### DIFF
--- a/plugins/keyboard-navigation/package.json
+++ b/plugins/keyboard-navigation/package.json
@@ -42,7 +42,7 @@
   "devDependencies": {
     "@blockly/dev-scripts": "^1.2.1",
     "@blockly/dev-tools": "^2.0.6",
-    "blockly": "^3.20200924.3",
+    "blockly": "4.20210325.0-beta.0",
     "chai": "^4.2.0",
     "jsdom": "^16.4.0",
     "jsdom-global": "^3.0.2",
--- a/plugins/keyboard-navigation/package.json
+++ b/plugins/keyboard-navigation/package.json
@@ -1,0 +1,67 @@
+{
+  "name": "blockly-plugin-keyboard-navigation",
+  "version": "0.20201029.0",
+  "description": "A Blockly plugin.",
+  "scripts": {
+    "audit:fix": "blockly-scripts auditFix",
+    "build": "blockly-scripts build",
+    "clean": "blockly-scripts clean",
+    "lint": "blockly-scripts lint",
+    "prepublishOnly": "npm run clean && npm run build",
+    "start": "blockly-scripts start",
+    "test": "blockly-scripts test"
+  },
+  "main": "./dist/index.js",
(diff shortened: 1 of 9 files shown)


plugins/keyboard-navigation/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "blockly-plugin-keyboard-navigation",
+  "version": "0.20201029.0",
```
maribethb: I thought we had fixed this in the script already but it should be 0.1.0.

plugins/keyboard-navigation/package.json
```diff
@@ -0,0 +1,67 @@
...
+  "module": "./src/index.js",
+  "unpkg": "./dist/index.js",
+  "author": "Blockly Team",
+  "private": true,
+  "keywords": [
+    "blockly",
+    "blockly-plugin",
+    "keyboard-navigation"
+  ],
+  "homepage": "https://github.com/alschmiedt/blockly-samples/tree/master/plugins/keyboard-navigation#readme",
```
maribethb: I know you're not publishing yet but you might want to go ahead and update this (and others) to point to google instead of your fork so you don't forget later.

plugins/keyboard-navigation/package.json
```diff
@@ -0,0 +1,67 @@
...
+  "bugs": {
+    "url": "https://github.com/alschmiedt/blockly-samples/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/alschmiedt/blockly-samples.git",
+    "directory": "plugins/keyboard-navigation"
+  },
+  "license": "Apache 2.0",
```
maribethb: This will have to be `Apache-2.0` in order to parse correctly. I swear I fixed this too, so now I'm wondering if you're somehow using an older version of the script or if it didn't get updated correctly?

maribethb (approved):

plugins/keyboard-navigation/package.json
alschmiedt: done

plugins/keyboard-navigation/package.json
alschmiedt: done

plugins/keyboard-navigation/package.json
alschmiedt: I think I copied over the entire package.json from my branch which is pretty old. I updated so should have these changes.

PR #543, opened 2021-01-26
Add constants and gesture overrides

alschmiedt:
Adds constants and gesture overrides needed for keyboard navigation. 

Gesture is needed for when shift clicking on the workspace or a block moves the cursor around.

### DIFF
--- a/plugins/keyboard-navigation/src/gesture_monkey_patch.js
+++ b/plugins/keyboard-navigation/src/gesture_monkey_patch.js
@@ -25,8 +25,8 @@ const oldDoWorkspaceClick = Blockly.Gesture.prototype.doWorkspaceClick_;
  * @override
  */
 Blockly.Gesture.prototype.doWorkspaceClick_ = function(e) {
+  oldDoWorkspaceClick.call(this, e);
   const ws = this.creatorWorkspace_;
-  oldDoWorkspaceClick.bind(this)(e);
   if (e.shiftKey && ws.keyboardAccessibilityMode) {
     const screenCoord = new Blockly.utils.Coordinate(e.clientX, e.clientY);
     const wsCoord = Blockly.utils.screenToWsCoordinates(ws, screenCoord);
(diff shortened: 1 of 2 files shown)


plugins/keyboard-navigation/src/gesture_monkey_patch.js
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Execute a workspace click. When in accessibility mode shift clicking will
+ * move the cursor.
+ * @param {!Event} e A mouse up or touch end event.
+ * @this {Blockly.Gesture}
+ * @override
+ */
+Blockly.Gesture.prototype.doWorkspaceClick_ = function(e) {
+  const ws = this.creatorWorkspace_;
+  oldDoWorkspaceClick.bind(this)(e);
```
moniika: Use `call` instead of `bind` here  and in `doWorkspaceClick`. It's just like bind, but it invokes the function immediately.

plugins/keyboard-navigation/src/gesture_monkey_patch.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Execute a workspace click. When in accessibility mode shift clicking will
+ * move the cursor.
+ * @param {!Event} e A mouse up or touch end event.
+ * @this {Blockly.Gesture}
+ * @override
+ */
+Blockly.Gesture.prototype.doWorkspaceClick_ = function(e) {
+  const ws = this.creatorWorkspace_;
```
moniika, line 29: My preference is that this set is after `oldDoWorkspaceClick` call.

moniika (approved):

plugins/keyboard-navigation/src/gesture_monkey_patch.js
alschmiedt: done

plugins/keyboard-navigation/src/gesture_monkey_patch.js
alschmiedt: done, thanks forgot about call : )

PR #544, opened 2021-01-26
Adds basic shortcut handlers

alschmiedt:
Adds the navigation controller class. This is what the user will use to add the plugin: 

```
const navigationController = new NavigationController();
navigationController.init();
navigationController.addWorkspace(workspace)
```

The init method will also add all necessary shortcuts, however that will be added in a later PR. This PR is just focused on the methods that add shortcut handlers.

### DIFF
--- a/plugins/keyboard-navigation/src/navigation_controller.js
+++ b/plugins/keyboard-navigation/src/navigation_controller.js
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2020 Google LLC
+ * Copyright 2021 Google LLC
  * SPDX-License-Identifier: Apache-2.0
  */
 
@@ -10,7 +10,7 @@
  * @author aschmiedt@google.com (Abby Schmiedt)
  */
 
-import '../src/gesture_monkey_patch';
+import './gesture_monkey_patch';
 
 import * as Blockly from 'blockly/core';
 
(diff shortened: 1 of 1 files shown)


plugins/keyboard-navigation/src/navigation_controller.js
```diff
@@ -0,0 +1,208 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
rachel-fenichel: 2021 :)

plugins/keyboard-navigation/src/navigation_controller.js
```diff
@@ -0,0 +1,208 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Registers all of the keyboard shortcuts that are necessary for
+ * navigating blockly using the keyboard.
+ * @author aschmiedt@google.com (Abby Schmiedt)
+ */
+
+import '../src/gesture_monkey_patch';
```
rachel-fenichel: Why not just `./gesture_monkey_patch`?

plugins/keyboard-navigation/src/navigation_controller.js
```diff
@@ -0,0 +1,208 @@
...
+  disable(workspace) {
+    this.navigation.disableKeyboardAccessibility(workspace);
+  }
+
+
+  /**
+   * Removes all the keyboard navigation shortcuts.
+   * @public
+   */
+  dispose() {
```
rachel-fenichel, line 220: This removes the shortcuts but leaves the handlers in place, right?

rachel-fenichel (approved):

plugins/keyboard-navigation/src/navigation_controller.js
alschmiedt: Yup, good point. Added a `removeShortcutHanlders` method.

plugins/keyboard-navigation/src/navigation_controller.js
alschmiedt: Fixed.

plugins/keyboard-navigation/src/navigation_controller.js
alschmiedt: fixed.

PR #546, opened 2021-01-28
Add basic navigation

alschmiedt:
These are the methods needed to set up navigation on a workspace and on a flyout. Most of these methods are new since in core, I would directly add them where they needed to be called, instead of listening for events.

### DIFF
--- a/plugins/keyboard-navigation/src/navigation.js
+++ b/plugins/keyboard-navigation/src/navigation.js
@@ -1,0 +1,377 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Holds all methods necessary to use Blockly through the
+ * keyboard.
+ * @author aschmiedt@google.com (Abby Schmiedt)
+ */
+
+import * as Blockly from 'blockly/core';
+import * as Constants from './constants';
+import {
+  registrationName as cursorRegistrationName,
+  registrationType as cursorRegistrationType} from './flyout_cursor';
+
+/**
+ * Class that holds all methods necessary for keyboard navigation to work.
+ */
+export class Navigation {
+  /**
+   * Constructor for keyboard navigation.
+   */
+  constructor() {
+    /**
+     * Object holding the location of the cursor for each workspace.
+     * Possible locations of the cursor are: workspace, flyout or toolbox.
+     * @type {Object<string,Constants.STATE>}
+     * @protected
+     */
+    this.workspaceStates = {};
+
+    /**
+     * An optional method that allows a developer to customize how to handle
+     * logs, warnings, and errors. The first argument is one of 'log', 'warn',
+     * or 'error'. The second argument is the message.
+     * @type {?function(string, string)}
+     * @public
+     */
+    this.loggingCallback = null;
+
+    /**
+     * The distance to move the cursor when the cursor is on the workspace.
+     * @type {number}
+     * @public
+     */
+    this.WS_MOVE_DISTANCE = 40;
+
+    /**
+     * The name of the marker to use for keyboard navigation.
+     * @type {string}
+     * @public
+     */
+    this.MARKER_NAME = 'local_marker_1';
+
+    /**
+     * The default coordinate to use when focusing on the workspace and no
+     * blocks are present.
+     * @type {!Blockly.utils.Coordinate}
+     * @public
+     */
+    this.DEFAULT_WS_COORDINATE = new Blockly.utils.Coordinate(100, 100);
+
+    /**
+     * The default coordinate to use when moving the cursor to the workspace
+     * after a block has been deleted.
+     * @type {!Blockly.utils.Coordinate}
+     * @public
+     */
+    this.WS_COORDINATE_ON_DELETE = new Blockly.utils.Coordinate(100, 100);
+
+    /**
+     * Wrapper for method that deals with workspace changes.
+     * Used for removing change listener.
+     * @type {Function}
+     * @protected
+     */
+    this.wsChangeWrapper = this.workspaceChangeListener.bind(this);
+
+    /**
+     * Wrapper for method that deals with flyout changes.
+     * Used for removing change listener.
+     * @type {Function}
+     * @protected
+     */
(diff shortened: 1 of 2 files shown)


plugins/keyboard-navigation/src/navigation.js
```diff
@@ -0,0 +1,377 @@
...
+     * Object holding the location of the cursor for each workspace.
+     * Possible locations of the cursor are: workspace, flyout or toolbox.
+     * @type {Object<string,Constants.STATE>}
+     * @protected
+     */
+    this.workspaceStates = {};
+
+    /**
+     * An optional method that allows a developer to customize how to handle
+     * logs, warnings, and errors. The first argument is one of 'log', 'warn',
```
maribethb, line 37: This could possibly be an enum?

plugins/keyboard-navigation/src/navigation.js
```diff
@@ -0,0 +1,377 @@
...
+     * or 'error'. The second argument is the message.
+     * @type {?function(string, string)}
+     * @public
+     */
+    this.loggingCallback = null;
+
+    /**
+     * The distance to move the cursor when the cursor is on the workspace.
```
maribethb, line 45: The units for this would be helpful here since we're all thinking about metrics right now.

plugins/keyboard-navigation/src/navigation.js
```diff
@@ -0,0 +1,377 @@
...
+     * Used for removing change listener.
+     * @type {Function}
+     * @protected
+     */
+    this.flyoutChangeWrapper = this.flyoutChangeListener.bind(this);
+
+    /**
+     * The list of registered workspaces.
+     * Used when removing change listeners in dispose.
+     * @protected
```
maribethb, line 96: `@type {!Array<!Blockly.WorkspaceSvg>}`

plugins/keyboard-navigation/src/navigation.js
```diff
@@ -0,0 +1,377 @@
...
+   * @protected
+   */
+  setState(workspace, state) {
+    this.workspaceStates[workspace.id] = state;
+  }
+
+  /**
+   * Gets the navigation state of the current workspace.
+   * @param {!Blockly.WorkspaceSvg} workspace The workspace to get the state of.
+   * @return {Constants.STATE} The sate of the given workspace.
```
maribethb: typo: state

plugins/keyboard-navigation/src/navigation.js
```diff
@@ -0,0 +1,377 @@
...
+
+    if (flyout) {
+      this.removeFlyout(flyout);
+    }
+  }
+
+  /**
+   * Sets the state for the given workspace.
+   * @param {!Blockly.WorkspaceSvg} workspace The workspace to set the state on.
+   * @param {Constants.STATE} state The navigation state.
```
maribethb: Here and elsewhere, you should use ? if you're not using ! to explicitly acknowledge it can be nullable

plugins/keyboard-navigation/src/navigation.js
```diff
@@ -0,0 +1,377 @@
...
+   * @param {!Blockly.Events.Abstract} e The Blockly event to process.
+   * @protected
+   */
+  flyoutChangeListener(e) {
+    const flyoutWorkspace = Blockly.Workspace.getById(e.workspaceId);
+    const mainWorkspace = flyoutWorkspace.targetWorkspace;
+    const flyout = mainWorkspace.getFlyout();
+
+    if (mainWorkspace && mainWorkspace.keyboardAccessibilityMode &&
+        !flyout.autoClose) {
```
maribethb, line 255: This somewhat confuses me. do we not need to do anything if the flyout autocloses? or on the other hand, it looks like this only works on "simple" flyouts which i didn't think could close. So either way I'm a bit confused on this condition (no change needed, just asking for my own understanding)

plugins/keyboard-navigation/src/navigation.js
```diff
@@ -0,0 +1,377 @@
...
+        this.handleBlockClickInFlyout(mainWorkspace, block);
+      }
+    }
+  }
+
+  /**
+   * Moves the cursor to the workspace if a block has been dragged from a simple
+   * toolbox. For a category toolbox this is handled in
+   * handleToolboxCategoryClick_.
+   * @param {Blockly.WorkspaceSvg} workspace The workspace the cursor belongs
```
maribethb: Should be ! I think (and below)

maribethb (approved):
mostly type nits :)

plugins/keyboard-navigation/src/navigation.js
alschmiedt: done

plugins/keyboard-navigation/src/navigation.js
alschmiedt: done

plugins/keyboard-navigation/src/navigation.js
alschmiedt: done

plugins/keyboard-navigation/src/navigation.js
alschmiedt: done

plugins/keyboard-navigation/src/navigation.js
alschmiedt: So I think this is for the case when the toolbox has a flyout and autoclose = false. `workspace.getFlyout()` returns either the flyout on the toolbox, or the simple flyout on the workspace.

I'll try to add a comment to help explain.

plugins/keyboard-navigation/src/navigation.js
alschmiedt: done

plugins/keyboard-navigation/src/navigation.js
alschmiedt: Ok, I added. It is a bit weird since they should be in workspace coordinates I think since we are positioning something that is on the workspace and not above it. However, we don't know the workspace  scale here so they have to be converted to workspace coordinates when they are used.

plugins/keyboard-navigation/src/navigation.js
alschmiedt: Ok, added a comment. Lmk if that helps!

PR #557, opened 2021-02-04
Copy over navigation methods from core

alschmiedt:
These methods were copied over from core, the only part I changed was some of the comments and the access specifiers.

### DIFF
--- a/plugins/keyboard-navigation/src/navigation.js
+++ b/plugins/keyboard-navigation/src/navigation.js
@@ -366,7 +366,7 @@ export class Navigation {
    * to that stack of blocks and setting the state of navigation to the flyout.
    * @param {!Blockly.WorkspaceSvg} mainWorkspace The workspace the user clicked
    *     on.
-   * @param {?Blockly.BlockSvg} block The block the user clicked on.
+   * @param {!Blockly.BlockSvg} block The block the user clicked on.
    * @protected
    */
   handleBlockClickInFlyout(mainWorkspace, block) {
@@ -503,7 +503,7 @@ export class Navigation {
     const flyout = workspace.getFlyout();
     const cursor = flyout ? flyout.getWorkspace().getCursor() : null;
 
-    return /** @type {Blockly.FlyoutCursor} */ (cursor);
+    return /** @type {?Blockly.FlyoutCursor} */ (cursor);
   }
 
   /**
@@ -530,8 +530,8 @@ export class Navigation {
    *     wrong.
    * @protected
    */
-  modify(workspace, markerNode, cursorNode) {
-    if (!this.modifyWarn(markerNode, cursorNode)) {
+  tryToConnectMarkerAndCursor(workspace, markerNode, cursorNode) {
+    if (!this.logConnectionWarning(markerNode, cursorNode)) {
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


plugins/keyboard-navigation/src/navigation.js
```diff
@@ -380,4 +380,505 @@ export class Navigation {
...
+          'Connection failed with error: ' +
+          checker.getErrorMessage(reason, movingConnection, destConnection));
+      return false;
+    }
+  }
+
+  /**
+   * Finds the inferior connection on the source block if the given connection
+   * is superior.
+   * @param {?Blockly.RenderedConnection} connection The connection trying to be
```
maribethb, line 713: This parameter is marked nullable, but in the body of the function you assume it's non-null. Ideally Closure should tell you if you've done this but maybe that's part of strict checking we don't have on yet. Most of the other functions that have nullable parameters handle it correctly but some of them don't (the next function also does this and maybe more).

plugins/keyboard-navigation/src/navigation.js
```diff
@@ -380,4 +380,505 @@ export class Navigation {
...
+  /**
+   * Tries to connect the given marker and cursor node.
+   * @param {!Blockly.WorkspaceSvg} workspace The main workspace.
+   * @param {!Blockly.ASTNode} markerNode The node to try to connect to.
+   * @param {!Blockly.ASTNode} cursorNode The node to connect to the markerNode.
+   * @return {boolean} True if the key was handled; false if something went
+   *     wrong.
+   * @protected
+   */
+  modify(workspace, markerNode, cursorNode) {
```
maribethb: I feel this name could be more descriptive (I'm not actually sure what's being modified here) and moving it might be a good time to change it, but also fine if you don't want to change things.

plugins/keyboard-navigation/src/navigation.js
```diff
@@ -380,4 +380,505 @@ export class Navigation {
...
+      const cursorBlock = /** @type {!Blockly.BlockSvg} */ (cursorLoc);
+      const markerConnection =
+          /** @type {!Blockly.RenderedConnection} */ (markerLoc);
+      return this.insertBlock(cursorBlock, markerConnection);
+    } else if (markerType == Blockly.ASTNode.types.WORKSPACE) {
+      const block = cursorNode ? cursorNode.getSourceBlock() : null;
+      return this.moveBlockToWorkspace(
+          /** @type {Blockly.BlockSvg} */ (block), markerNode);
+    }
+    this.warn('Unexpected state in setWorkspaceState.');
```
maribethb: this comment did not help my naming confusion because there is no function here called setWorkspaceState. either that or I am more confused than I realized.

maribethb (approved):

plugins/keyboard-navigation/src/navigation.js
alschmiedt: Yeah this must have been a copy and paste error. Fixed.

plugins/keyboard-navigation/src/navigation.js
alschmiedt: Changed to tryToConnectMarkerAndCursor. Let me know what you think.

plugins/keyboard-navigation/src/navigation.js
alschmiedt: Ok, went through and I think I got all of them.

plugins/keyboard-navigation/src/navigation.js
maribethb: Makes sense :) thanks!

PR #558, opened 2021-02-04
Add test page

alschmiedt:
Add a test page similar to the old keyboard navigation demo page.

### DIFF
--- a/plugins/keyboard-navigation/test/index.html
+++ b/plugins/keyboard-navigation/test/index.html
@@ -12,6 +12,30 @@
 </head>
 
 <body>
+  <p>Keyboard Navigation is our first step towards an accessible Blockly.<br />
+  For more information on how the default keyboard navigation works please see
+  the <a href="https://developers.google.com/blockly/guides/configure/web/keyboard-nav">documentation</a>.
+  <br />
+  <br />
+
+  <b>Cursors</b><br />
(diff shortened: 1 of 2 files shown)


plugins/keyboard-navigation/test/index.html
```diff
@@ -12,6 +12,30 @@
...
+  <p>Keyboard Navigation is our first step towards an accessible Blockly.<br />
+  For more information on how the default keyboard navigation works please see
+  the <a href="https://developers.google.com/blockly/guides/configure/web/keyboard-nav">documentation</a>.
+  <br />
+  <br />
+
+  <b>Cursors</b><br />
+  The cursor controls how the user navigates the blocks, inputs, fields and connections on a workspace.
+  This demo shows three different cursors:<br />
+  <b>Default Cursor:</b> Explained in <a href="https://developers.google.com/blockly/guides/configure/web/keyboard-nav">documentation</a>.<br />
```
maribethb: I think it'd be useful to have a short description here as well in addition to pointing to the full docs. Like "This cursor uses previous, next, in, and out commands to navigate through levels of the code. See the full explanation at..."

You might also want to explain what the buttons are for each of those to make it easier to test, for someone new?

maribethb (approved):

plugins/keyboard-navigation/test/index.html
alschmiedt: done!

PR #559, opened 2021-02-04
Add navigation test

alschmiedt:
Copy over navigation tests from core.

### DIFF
--- a/plugins/keyboard-navigation/test/navigation_modify_test.mocha.js
+++ b/plugins/keyboard-navigation/test/navigation_modify_test.mocha.js
@@ -40,7 +40,8 @@ suite('Insert/Modify', function() {
   function assertModifyFails(navigation, workspace, markerNode, cursorNode) {
     let modifyResult;
     const warnings = captureWarnings(function() {
-      modifyResult = navigation.modify(workspace, markerNode, cursorNode);
+      modifyResult = navigation.tryToConnectMarkerAndCursor(
+          workspace, markerNode, cursorNode);
     });
     assert.isFalse(modifyResult);
     assert.equal(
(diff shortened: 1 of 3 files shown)


plugins/keyboard-navigation/test/navigation_modify_test.mocha.js
```diff
@@ -0,0 +1,476 @@
...
+    try {
+      console.warn = function(msg) {
+        msgs.push(msg);
+      };
+      innerFunc();
+    } finally {
+      console.warn = nativeConsoleWarn;
+    }
+    return msgs;
+  }
```
moniika, line 31: I think we should consider providing this helper in dev-tools plugin. Maybe not in this PR, but file a bug.

plugins/keyboard-navigation/test/navigation_modify_test.mocha.js
```diff
@@ -0,0 +1,476 @@
...
+   * @param {!Blockly.ASTNode} cursorNode The node to connect to the markerNode.
+   */
+  function assertModifyFails(navigation, workspace, markerNode, cursorNode) {
+    let modifyResult;
+    const warnings = captureWarnings(function() {
+      modifyResult = navigation.modify(workspace, markerNode, cursorNode);
+    });
+    assert.isFalse(modifyResult);
+    assert.equal(
+        warnings.length, 1, 'Expecting 1 warnings for why modify failed.');
```
moniika, line 48: I would recommend also adding an assertion checking the warning text.
You could use `chai.assert.match` to check part of the message, so we can reasonably conclude that it's the message we were expecting.

moniika (commented):
I'd like for you to also try to fix the eslint errors so the build passes.

plugins/keyboard-navigation/test/navigation_modify_test.mocha.js
alschmiedt: So I did go through and start to update this, but I think because of how experimental it is I'm not sure I want to add this check. There wasn't a lot of reasoning behind the wording we chose for the error messages and I expect us to change them in the future. I think it would be more helpful to add them back in once we settle on the wording we want.

plugins/keyboard-navigation/test/navigation_modify_test.mocha.js
alschmiedt: done. #567

PR #560, opened 2021-02-04
Relaxed requirements on blocks with multiple generic inputs - project static

BeksOmega:
### Description

Second to last PR for simple generic checks! And the last one is just tests for round-tripping though XML :D

This PR implements [Proposal A](https://docs.google.com/document/d/1QKYkmWjkle1JWCi3O8jXr8-7Toazh1pW_4EaVVgB_OI/edit#heading=h.z2m9hs1ghrwp) for multiple generic checks from the design document. The reason I decided to go with Proposal A is it means that the logic can be consistent when loading from XML (which connects blocks in order from leaves to roots).

Note that you should still be able to simulate Proposal B using the `bindType()` function provided by the connection checker =)

### Tests

* Connecting two blocks with sibling types to a generic parent.
* Explicitly binding the generic parent  to a supertype of the sibling types, then doing the same thing.
* Not allowing two unrelated types to connect to a generic parent.
* Explicit binding is given higher priority than parameter types.
* etc

I was also able to remove some of the `bindConnection` calls because of this change =)

### DIFF
--- a/plugins/nominal-connection-checker/package-lock.json
+++ b/plugins/nominal-connection-checker/package-lock.json
@@ -4,6 +4,1111 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.12.13",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.13.tgz",
+      "integrity": "sha512-HV1Cm0Q3ZrpCR93tkWOYiuYIgLxZXZFVG2VgK+MBWjUqZTundupbfx2aXarXuw5Ko5aMcjtJgbSs4vUGBS5v6g==",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "^7.12.13"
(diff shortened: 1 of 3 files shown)


BeksOmega:
I think this PR may have gotten lost hehe. No rush if you guys are busy, just thought I'd bump it =)

rachel-fenichel (approved):
Whoops. Yes, this got lost in the rush. Approved, and I'll merge it.

BeksOmega:
I don't know how github works so this might be a bit of a weird question. But like... would it be helpful if I requested reviews on things? I know you guys do it, but I don't even know where that information shows up haha.

I can't request reviews in the core repo, but I do have permissions here, so if that'd be helpful I'm happy to. Otherwise I'll just keep doing what I'm doing :D

rachel-fenichel:
Our oncall person is generally responsible for triaging incoming pull requests and distributing them to people to review. Honestly, what happened here is just that I'm the only person with context to do this review, and I've been doing a lot of context-switching.

PR #565, opened 2021-02-16
Update toolbox metrics

alschmiedt:
Fixes issue google/blockly#4377.

Overrides the new metrics manager to exclude the flyout from the viewport.

### DIFF
--- a/plugins/continuous-toolbox/package-lock.json
+++ b/plugins/continuous-toolbox/package-lock.json
@@ -4,11 +4,1451 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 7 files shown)


maribethb (approved):
We need to make sure we update the Blockly dependency... hard to do when we don't have a date but I guess we can do that when we merge this branch into master

alschmiedt:
If everything goes according to plan the next release is scheduled for March 25th. I added 4.20210325.0 as the peer dependency and updated the dev version to be 4.202010325.0-beta.3 so if someone wants to pull this down it will at least work. 

I will also add a note in the release doc to check this on all the packages.

PR #571, opened 2021-02-23
Register tests

alschmiedt:
Tests functionality around copying, deleting, and pasting a block using keyboard shortcuts specific to keyboard navigation. These tests are new, and not copied over directly from core.

### DIFF
--- a/plugins/keyboard-navigation/test/shortcuts_test.mocha.js
+++ b/plugins/keyboard-navigation/test/shortcuts_test.mocha.js
@@ -16,12 +16,11 @@ suite('Shortcut Tests', function() {
   /**
    * Creates a test for not running keyDown events when the workspace is in read
    * only mode.
+   * @param {string} testCaseName The name of the test case.
    * @param {Object} keyEvent Mocked key down event. Use createKeyDownEvent.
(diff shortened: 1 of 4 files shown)


plugins/keyboard-navigation/test/shortcuts_test.mocha.js
```diff
@@ -0,0 +1,416 @@
...
+    require('./test_helper');
+
+suite('Shortcut Tests', function() {
+  /**
+   * Creates a test for not running keyDown events when the workspace is in read
+   * only mode.
+   * @param {Object} keyEvent Mocked key down event. Use createKeyDownEvent.
+   * @param {string=} optName An optional name for the test case.
+   */
+  function runReadOnlyTest(keyEvent, optName) {
```
moniika: Could you swap these parameters around so they're the same as the other helper functions?
Also, since `optName` is always passed, you could drop the optional and just make it required. Up to you.

moniika (approved):
lgtm

plugins/keyboard-navigation/test/shortcuts_test.mocha.js
alschmiedt: done

PR #577, opened 2021-03-06
Add zoom-to-fit plugin.

moniika:
Adds zoom-to-fit plugin.
image

### DIFF
--- a/plugins/zoom-to-fit/README.md
+++ b/plugins/zoom-to-fit/README.md
@@ -1,8 +1,5 @@
 # @blockly/zoom-to-fit [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
 
-<!--
-  - TODO: Edit plugin description.
-  -->
 A [Blockly](https://www.npmjs.com/package/blockly) plugin that adds a
 zoom-to-fit control to the workspace.
 
--- a/plugins/zoom-to-fit/README.md
+++ b/plugins/zoom-to-fit/README.md
@@ -22,7 +22,7 @@ npm install @blockly/zoom-to-fit --save
 
 ```js
 import * as Blockly from 'blockly';
-import {Plugin} from '@blockly/zoom-to-fit';
+import {ZoomToFitControl} from '@blockly/zoom-to-fit';
 
 // Inject Blockly.
 const workspace = Blockly.inject('blocklyDiv', {
--- a/plugins/zoom-to-fit/README.md
+++ b/plugins/zoom-to-fit/README.md
@@ -1,0 +1,55 @@
+# @blockly/zoom-to-fit [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
+
+<!--
+  - TODO: Edit plugin description.
+  -->
+A [Blockly](https://www.npmjs.com/package/blockly) plugin that adds a
+zoom-to-fit control to the workspace.
+
+## Installation
+
+### Yarn
+```
+yarn add @blockly/zoom-to-fit
+```
+
+### npm
+```
+npm install @blockly/zoom-to-fit --save
+```
+
+## Usage
+
+```js
+import * as Blockly from 'blockly';
+import {Plugin} from '@blockly/zoom-to-fit';
+
+// Inject Blockly.
+const workspace = Blockly.inject('blocklyDiv', {
+  toolbox: toolboxCategories,
+});
+
+// Initialize plugin.
+const zoomToFit = new ZoomToFitControl(workspace);
+zoomToFit.init();
+workspace.getPluginManager().addPlugin({
+  id: 'zoomToFit',
+  plugin: zoomToFit,
+  weight: 1,
+  types: [Blockly.PluginManager.Type.POSITIONABLE],
+});
+
+// resize Blockly to position ui elements.
+workspace.resize();
+```
+
+## API
+
+- `init`: Initializes the zoom-to-fit control.
+- `dispose`: Disposes of the zoom-to-fit control.
+- `getBoundingRectangle`: Returns the bounding rectangle of the UI element in
+pixel units relative to the Blockly injection div.
+- `position`: Positions the zoom-to-fit control.
+
+## License
+Apache 2.0
--- a/plugins/zoom-to-fit/README.md
+++ b/plugins/zoom-to-fit/README.md
@@ -36,7 +36,7 @@ workspace.getPluginManager().addPlugin({
   types: [Blockly.PluginManager.Type.POSITIONABLE],
 });
 
-// resize Blockly to position ui elements.
+// Resize Blockly to position ui elements.
 workspace.resize();
 ```
 
--- a/plugins/zoom-to-fit/README.md
+++ b/plugins/zoom-to-fit/README.md
@@ -29,15 +29,6 @@ const workspace = Blockly.inject('blocklyDiv', {
 // Initialize plugin.
 const zoomToFit = new ZoomToFitControl(workspace);
 zoomToFit.init();
-workspace.getPluginManager().addPlugin({
-  id: 'zoomToFit',
-  plugin: zoomToFit,
-  weight: 1,
-  types: [Blockly.PluginManager.Type.POSITIONABLE],
-});
-
-// Resize Blockly to position ui elements.
-workspace.resize();
 ```
 
 ## API
--- a/plugins/zoom-to-fit/package.json
+++ b/plugins/zoom-to-fit/package.json
@@ -40,7 +40,7 @@
   "devDependencies": {
     "@blockly/dev-scripts": "^1.2.4",
     "@blockly/dev-tools": "^2.0.10",
-    "blockly": "^4.20201217.0"
+    "blockly": "^4.20210325.0-beta.4"
   },
   "peerDependencies": {
     "blockly": ">=4"
--- a/plugins/zoom-to-fit/package.json
+++ b/plugins/zoom-to-fit/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@blockly/zoom-to-fit",
-  "version": "0.0.0",
+  "version": "1.0.0",
   "description": "A Blockly plugin.",
   "scripts": {
     "audit:fix": "blockly-scripts auditFix",
--- a/plugins/zoom-to-fit/package.json
+++ b/plugins/zoom-to-fit/package.json
@@ -43,7 +43,7 @@
     "blockly": "^4.20201217.0"
   },
   "peerDependencies": {
-    "blockly": ">=3"
+    "blockly": ">=4"
   },
   "publishConfig": {},
   "eslintConfig": {
--- a/plugins/zoom-to-fit/package.json
+++ b/plugins/zoom-to-fit/package.json
@@ -19,13 +19,13 @@
     "blockly-plugin",
     "zoom-to-fit"
   ],
-  "homepage": "https://github.com/moniika/blockly-samples/tree/master/plugins/zoom-to-fit#readme",
+  "homepage": "https://github.com/google/blockly-samples/tree/master/plugins/zoom-to-fit#readme",
   "bugs": {
-    "url": "https://github.com/moniika/blockly-samples/issues"
+    "url": "https://github.com/google/blockly-samples/issues"
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/moniika/blockly-samples.git",
+    "url": "https://github.com/google/blockly-samples.git",
     "directory": "plugins/zoom-to-fit"
   },
   "license": "Apache-2.0",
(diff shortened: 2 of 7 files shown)


plugins/zoom-to-fit/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "main": "./dist/index.js",
+  "module": "./src/index.js",
+  "unpkg": "./dist/index.js",
+  "author": "",
+  "keywords": [
+    "blockly",
+    "blockly-plugin",
+    "zoom-to-fit"
+  ],
+  "homepage": "https://github.com/moniika/blockly-samples/tree/master/plugins/zoom-to-fit#readme",
```
rachel-fenichel: Update `moniika` -> `google`, here and below.

plugins/zoom-to-fit/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "dist",
+    "src"
+  ],
+  "devDependencies": {
+    "@blockly/dev-scripts": "^1.2.4",
+    "@blockly/dev-tools": "^2.0.10",
+    "blockly": "^4.20201217.0"
+  },
+  "peerDependencies": {
+    "blockly": ">=3"
```
rachel-fenichel: Should this match the dev dependency?

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,230 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
rachel-fenichel: Nit: Copyright 2021.

I bet that means we have to update the tool @samelhusseini

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,230 @@
...
+    this.workspace_ = workspace;
+
+    /**
+     * The SVG group containing the zoom controls.
+     * @type {SVGElement}
+     * @private
+     */
+    this.svgGroup_ = null;
+
+    /**
```
rachel-fenichel, line 40: The next four properties define a bounding box, which can be defined with `Blockly.utils.Rect`. That also gives us `intersects` and `contains` functions.

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,230 @@
...
+   * to prevent memory leaks.
+   */
+  dispose() {
+    for (const event of this.boundEvents_) {
+      Blockly.unbindEvent_(event);
+    }
+    this.boundEvents_ = null;
+    if (this.htmlDiv_) {
+      this.htmlDiv_.remove();
+      this.htmlDiv_ = null;
```
rachel-fenichel: Don't need to set these to null--garbage collection will handle it.

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,230 @@
...
+    this.actionDiv_ = null;
+    this.inputElement_ = null;
+  }
+
+  /**
+   * Creates DOM for ui element.
+   * @private
+   */
+  createDom_() {
+    // const parentSvg = this.workspace_.getParentSvg();
```
rachel-fenichel: Delete commented out code.

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,230 @@
...
+        this.top_, this.top_ + this.HEIGHT_,
+        this.left_, this.left_ + this.WIDTH_);
+  }
+
+  /**
+   * Positions the zoom-to-fit control.
+   * It is positioned in the opposite corner to the corner the
+   * categories/toolbox starts at.
+   * @param {!Blockly.MetricsManager.UiMetrics} metrics The workspace metrics.
+   * @param {!Array<Blockly.utils.Rect>} savedPositions List of rectangles that
```
rachel-fenichel: `!Array.<!Blockly.utils.Rect>`

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,230 @@
...
+   *     are already on the workspace.
+   */
+  position(metrics, savedPositions) {
+    // Not yet initialized.
+    if (!this.verticalSpacing_) {
```
rachel-fenichel: Keep an `initialized` property instead of using `verticalSpacing_` as an indicator.

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,230 @@
...
+     */
+    this.MARGIN_SIDE_ = 20;
+
+    /**
+     * The starting vertical distance between the workspace edge and the
+     * control. The value is initialized during `init`.
+     * @type {?number}
+     * @private
+     */
+    this.verticalSpacing_ = null;
```
rachel-fenichel: This can be 0 until initialized. Then the types are simplified, since it's always valid/never null.

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,230 @@
...
+    }
+    // Clamp top value within valid range.
+    this.top_ = Blockly.utils.math.clamp(minTop, this.top_, maxTop);
+
+    this.svgGroup_.setAttribute('transform',
+        'translate(' + this.left_ + ',' + this.top_ + ')');
+  }
+}
+
+// Blockly.Css.register([
```
rachel-fenichel: Do you need this?

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,230 @@
...
+    } else {
+      // Left corner placement.
+      this.left_ = this.MARGIN_SIDE_ + Blockly.Scrollbar.scrollbarThickness;
+    }
+
+    // Upper corner placement
+    const minTop = this.top_ =
+        metrics.absoluteMetrics.top + this.verticalSpacing_;
+    // Bottom corner placement
+    const maxTop = metrics.absoluteMetrics.top + metrics.viewMetrics.height -
```
rachel-fenichel, line 206: Why is the scrollbar thickness considered for the horizontal placement but not vertical placement?

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,230 @@
...
+        this.HEIGHT_ - this.verticalSpacing_;
+    const placeBottom =
+        metrics.toolboxMetrics.position !== Blockly.TOOLBOX_AT_BOTTOM;
+    this.top_ = placeBottom ? maxTop : minTop;
+
+    // Check for collision and bump if needed.
+    let boundingRect = this.getBoundingRectangle();
+    for (let i = 0, otherEl; (otherEl = savedPositions[i]); i++) {
```
rachel-fenichel, line 217: Can this ever loop infinitely, in a small enough workspace?

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,230 @@
...
+        } else {
+          this.top_ = otherEl.bottom + this.MARGIN_BOTTOM_;
+        }
+        // Recheck other savedPositions
+        boundingRect = this.getBoundingRectangle();
+        i = -1;
+      }
+    }
+    // Clamp top value within valid range.
+    this.top_ = Blockly.utils.math.clamp(minTop, this.top_, maxTop);
```
rachel-fenichel, line 231: Having this after means clamping is more important than collisions, right?

plugins/zoom-to-fit/test/index.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <title>Blockly Plugin Test</title>
```
rachel-fenichel: Update title

plugins/zoom-to-fit/src/index.js
samelhusseini: https://github.com/google/blockly-samples/pull/590

plugins/zoom-to-fit/src/index.js
moniika: No. It is either always bumped up or always bumped down.

plugins/zoom-to-fit/src/index.js
moniika: Yes, that's how I've decided to deal with placement here.

plugins/zoom-to-fit/package.json
moniika: Updated

plugins/zoom-to-fit/src/index.js
moniika: I didn't before, but now it's being used.

plugins/zoom-to-fit/src/index.js
moniika: I've removed `this.verticalSpacing_` to reduce confusion surrounding the math and added check for whether the scrollbar padding is necessary.

plugins/zoom-to-fit/src/index.js
moniika: I think storing the top/left/width/height properties separately is more beneficial than storing them all on a rectangle for the following reasons:
1. When returning the bounding box for the object, there's no concern that the caller could modify the properties on the bounding box (which could mess up other computation during positioning) since it is a new object every time.
2. I'd have to re-compute the width/height every time `position` is called to use for computing the new bounding box.
2. (minor plus) It allows me to conveniently call `getBoundingBox` during `position` logic to quickly recompute the new potential bounding box

plugins/zoom-to-fit/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+// Inject Blockly.
+const workspace = Blockly.inject('blocklyDiv', {
+  toolbox: toolboxCategories,
+});
+
+// Initialize plugin.
+const zoomToFit = new ZoomToFitControl(workspace);
+zoomToFit.init();
+workspace.getPluginManager().addPlugin({
```
rachel-fenichel: Let's have a chat as a team about how we want to add plugins to the workspace (like this vs in the config struct).

plugins/zoom-to-fit/src/index.js
rachel-fenichel: For the first one: definitely would want to return a copy, but that's easy.

For the second and third ones: okay, fine to leave it this way since it has these benefits.

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,251 @@
...
+    this.createDom_();
+    this.initialized_ = true;
+  }
+  /**
+   * Disposes of workspace search.
+   * Unlink from all DOM elements and remove all event listeners
+   * to prevent memory leaks.
+   */
+  dispose() {
+    if (this.onZoomOutWrapper_) {
```
rachel-fenichel: Will you ever want to dispose of the zoom control other than when the whole workspace is being disposed?

If so, this should remove the svg group from the DOM (although it doesn't need to set `this.svgGroup_` to `null`).

plugins/zoom-to-fit/src/index.js
rachel-fenichel: Right, but after a bump you reset the index to -1 to restart the loop. So if on the first run it has to bump up, and then on the second try it overlaps the other direction and bumps down, it will keep looping.

plugins/zoom-to-fit/src/index.js
```diff
@@ -0,0 +1,251 @@
...
+    this.svgGroup_.setAttribute('transform',
+        'translate(' + this.left_ + ',' + this.top_ + ')');
+  }
+}
+
+/**
+ * Base64 encoded data uri for zoom to fit  icon.
+ * @type {string}
+ */
+const ZOOM_TO_FIT_SVG_DATAURI =
```
rachel-fenichel, line 242: Why do you have both a datauri and an svg file?

rachel-fenichel (commented):
I'm still concerned about an infinite loop. Other than that, I think this is good to go as a version 0. Let's discuss the remaining questions as a team, since they're more about how we want to have these positionables work.

plugins/zoom-to-fit/src/index.js
moniika: The svg file is committed so that we have a copy of the original svg that the datauri is generated from.
This is the same thing we have done in WorkspaceSearch

plugins/zoom-to-fit/src/index.js
moniika: done

plugins/zoom-to-fit/src/index.js
moniika: What I meant by always bumped up or down is that the `bumpUp` boolean is never changed between loops.
It's set before the loop only once.

plugins/zoom-to-fit/README.md
moniika: Discussed and `addPlugin` call was moved inside of `init` since we may be changing the api in the near future.

rachel-fenichel (approved):

PR #578, opened 2021-03-06
Fixed edges plugin

moniika:
Adds fixed edges plugin that provides a custom `Blockly.MetricsManager` that allows for more easily configurable fixed edges.

### DIFF
--- a/plugins/fixed-edges/package.json
+++ b/plugins/fixed-edges/package.json
@@ -1,0 +1,60 @@
+{
+  "name": "@blockly/fixed-edges",
+  "version": "0.0.0",
+  "description": "A Blockly plugin.",
+  "scripts": {
+    "audit:fix": "blockly-scripts auditFix",
+    "build": "blockly-scripts build",
+    "clean": "blockly-scripts clean",
+    "lint": "blockly-scripts lint",
+    "prepublishOnly": "npm run clean && npm run build",
+    "start": "blockly-scripts start"
+  },
+  "main": "./dist/index.js",
+  "module": "./src/index.js",
+  "unpkg": "./dist/index.js",
+  "author": "",
+  "keywords": [
+    "blockly",
+    "blockly-plugin",
+    "fixed-edges"
+  ],
+  "homepage": "https://github.com/moniika/blockly-samples/tree/master/fixed-edges#readme",
+  "bugs": {
+    "url": "https://github.com/moniika/blockly-samples/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/moniika/blockly-samples.git",
+    "directory": "fixed-edges"
+  },
+  "license": "Apache-2.0",
+  "directories": {
+    "dist": "dist",
+    "src": "src"
+  },
+  "files": [
+    "dist",
+    "src"
+  ],
+  "devDependencies": {
+    "@blockly/dev-scripts": "^1.2.4",
+    "@blockly/dev-tools": "^2.0.9",
+    "blockly": "^4.20201217.0"
+  },
+  "peerDependencies": {
+    "blockly": ">=3"
+  },
+  "publishConfig": {},
+  "eslintConfig": {
+    "extends": "@blockly/eslint-config"
+  },
+  "engines": {
+    "node": ">=8.17.0"
+  },
+  "browserslist": [
+    "defaults",
+    "IE 11",
+    "IE_Mob 11"
+  ]
+}
--- a/plugins/fixed-edges/package.json
+++ b/plugins/fixed-edges/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@blockly/fixed-edges",
-  "version": "0.0.0",
+  "version": "1.0.0",
   "description": "A Blockly MetricsManager for configuring fixed sides.",
   "scripts": {
     "audit:fix": "blockly-scripts auditFix",
--- a/plugins/fixed-edges/package.json
+++ b/plugins/fixed-edges/package.json
@@ -40,7 +40,7 @@
   "devDependencies": {
     "@blockly/dev-scripts": "^1.2.4",
     "@blockly/dev-tools": "^2.0.9",
-    "blockly": ">=4"
+    "blockly": "^4.20210325.0-beta.4"
   },
   "peerDependencies": {
     "blockly": ">=4"
--- a/plugins/fixed-edges/package.json
+++ b/plugins/fixed-edges/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@blockly/fixed-edges",
   "version": "0.0.0",
-  "description": "A Blockly plugin.",
+  "description": "A Blockly MetricsManager for configuring fixed sides.",
   "scripts": {
     "audit:fix": "blockly-scripts auditFix",
     "build": "blockly-scripts build",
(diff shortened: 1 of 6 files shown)


plugins/fixed-edges/src/index.js
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+
+/**
+ * @fileoverview Plugin overview.
+ */
+
+import Blockly from 'blockly/core';
+
+let fixedEdges = {
```
moniika, line 14: I'm not sure about this type.

plugins/fixed-edges/src/index.js
```diff
@@ -0,0 +1,102 @@
...
+};
+
+/**
+ * The manager for a workspace with fixed edges.
+ * @param {!Blockly.WorkspaceSvg} workspace The workspace to calculate metrics
+ *     for.
+ * @implements {Blockly.IMetricsManager}
+ * @constructor
+ */
+export class FixedEdgesMetricsManager extends Blockly.MetricsManager {
```
moniika, line 26: How does `ConstantEdgesMetricsManger` sound as an alternative?

plugins/fixed-edges/src/index.js
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Sets which edges are fixed.
+   * @param {{
+   *   top: (boolean|undefined),
+   *   bottom: (boolean|undefined),
+   *   left: (boolean|undefined),
+   *   right: (boolean|undefined)
+   * }} newFixedEdges The edges to set as fixed.
+   */
+  static setFixedEdges(newFixedEdges) {
```
moniika: I'd like feedback on this method.
I'm not confident with this choice of param type.
My intention is that you can set top or bottom or neither as "fixed"/"contant" AND left or right or neither as "fixed"/"contant".

plugins/fixed-edges/src/index.js
moniika, line 40: The terminology of "fixed' may not be clear enough as to what this does. You can  also achieve fixed edges by disabling scrollbars.

plugins/fixed-edges/test/index.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
alschmiedt: 2021

plugins/fixed-edges/test/index.js
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Plugin test.
```
alschmiedt: Update this.

plugins/fixed-edges/src/index.js
alschmiedt: So, if I am understanding correct if you disable horizontal or vertical scrollbars you can: 
- Fix Top and bottom 
- Fix left and right

With this plugin we add the ability to fix the following combinations: 
- Top and left 
- Bottom and right

Does this sound right?

plugins/fixed-edges/src/index.js
moniika: updated wording throughout

plugins/fixed-edges/src/index.js
moniika: updated wording

plugins/fixed-edges/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+ *   bottom: (boolean|undefined),
+ *   left: (boolean|undefined),
+ *   right: (boolean|undefined)
+ * }}
+ */
+let FixedEdgesConfig;
+
+/**
+ * The current configuration for fixed edges.
+ * @type {!FixedEdgesConfig}
```
alschmiedt, line 27: @private?

plugins/fixed-edges/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+    if (fixedEdges.left && fixedEdges.right) {
+      right = viewMetrics.width;
+    }
+
+    const edges = {
+      top: top,
+      bottom: bottom,
+      left: left,
+      right: right,
+    };
```
alschmiedt: Could you move this to the top? Something like: 
const edges = {
   top: fixedEdges.top ? 0 : undefined,
   ...
}
Just thinking then we wouldn't have two top, right, left, bottom variables to keep track of.

plugins/fixed-edges/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+    if (!vScrollEnabled) {
+      if (top !== undefined) {
+        edges.bottom = top + viewMetrics.height;
```
alschmiedt: Use edges.top for consistency.

plugins/fixed-edges/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "main": "./dist/index.js",
+  "module": "./src/index.js",
+  "unpkg": "./dist/index.js",
+  "author": "",
+  "keywords": [
+    "blockly",
+    "blockly-plugin",
+    "fixed-edges"
+  ],
+  "homepage": "https://github.com/moniika/blockly-samples/tree/master/fixed-edges#readme",
```
alschmiedt: Urls here and below should point to google instead of your fork.

plugins/fixed-edges/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@blockly/fixed-edges",
+  "version": "0.0.0",
+  "description": "A Blockly plugin.",
```
alschmiedt: Update description

plugins/fixed-edges/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "src": "src"
+  },
+  "files": [
+    "dist",
+    "src"
+  ],
+  "devDependencies": {
+    "@blockly/dev-scripts": "^1.2.4",
+    "@blockly/dev-tools": "^2.0.9",
+    "blockly": "^4.20201217.0"
```
alschmiedt: I think we want this to be the beta release. My reasoning is if someone wants to pull this and test it, then it will build properly.

moniika: I've updated the version

plugins/fixed-edges/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+  hasFixedEdges() {
+    return true;
+  }
+
+  /**
+   * Computes the fixed edges of the scroll area.
+   * @param {!Blockly.MetricsManager.ContainerRegion=} opt_viewMetrics The view
+   *     metrics if they have been previously computed. Passing in null may
+   *     cause the view metrics to be computed again, if it is needed.
+   * @return {Blockly.MetricsManager.FixedEdges} The fixed edges of the scroll
```
alschmiedt: Update to `!Blockly.MetricsManager.FixedEdges`

plugins/fixed-edges/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@blockly/fixed-edges",
+  "version": "0.0.0",
```
alschmiedt: I think this can be updated to version 1.0.0 since it does not monkey patch core.

alschmiedt (approved):

plugins/fixed-edges/package.json
moniika: Updated

plugins/fixed-edges/src/index.js
moniika: Updated. I'm also going to make a PR in core, since I shouldn't be changing the type for an overridden method.

PR #582, opened 2021-03-09
Add tests for round-tripping through XML

BeksOmega:
### Description

Adds tests to ensure that blocks properly round-trip through XML. These tests include:
  - Simple subtyping tests
  - Simple generic tests
  - Multiple inputs with generics
    These cases are what would have caused problems if we went with [Proposal B](https://docs.google.com/document/d/1QKYkmWjkle1JWCi3O8jXr8-7Toazh1pW_4EaVVgB_OI/edit#heading=h.mriw7sjc5ine) For example this set of steps:
    bad-serialization
    Would not deserialize properly. Dog would be attached first, causing the SelectRandom block to bind to Dog, so Cat would not be able to connect.
  - Generics with types that have multiple nearest common parents
  - Blocks that use the programmatic binding system, and round-trip that information using mutators
    If outside developers wanted to provide things like type dropdowns, they'd do something similar to this.

I think that should cover any weird cases!

### DIFF
--- a/plugins/nominal-connection-checker/test/serialization_test.mocha.js
+++ b/plugins/nominal-connection-checker/test/serialization_test.mocha.js
@@ -1,0 +1,399 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Unit tests for round-tripping though serialization.
+ */
+
+const chai = require('chai');
+const Blockly = require('blockly/node');
+
+const {pluginInfo} = require('../src/index.js');
+const {getCheck} = require('../src/utils');
+const {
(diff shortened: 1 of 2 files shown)


plugins/nominal-connection-checker/test/serialization_test.mocha.js
```diff
@@ -0,0 +1,399 @@
...
+      const actualXmlText = Blockly.Xml.domToText(targetXml);
+
+      chai.assert.equal(actualXmlText, expectedXmlText);
+    };
+  });
+
+  teardown(function() {
+    for (const block of this.blocks) {
+      delete Blockly.Blocks[block.type];
+    }
```
rachel-fenichel, line 84: Also dispose of the workspace, I think.

rachel-fenichel (approved):

PR #589, opened 2021-03-12
Add block dynamic connection plugin

ajpal:
Right now just has versions of `text_join`, `create_list_with`, and `controls_if`.
In the future, I could see this being useful for functions and possibly other blocks as well.

Mar-12-2021 11-57-57
Mar-12-2021 11-58-04
Apr-01-2021 13-36-13

### DIFF
--- a/plugins/block-dynamic-connection/README.md
+++ b/plugins/block-dynamic-connection/README.md
@@ -1,46 +1,31 @@
-# blockly-plugin-block-dynamic-connection [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
+# @blockly/block-dynamic-connection [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
 
-<!--
-  - TODO: Edit plugin description.
-  -->
-A [Blockly](https://www.npmjs.com/package/blockly) plugin that ...
+A group of [Blockly](https://www.npmjs.com/package/blockly) blocks that
+add and remove connections dynamically.
 
 ## Installation
-
-### Yarn
 ```
-yarn add blockly-plugin-block-dynamic-connection
-```
-
-### npm
-```
-npm install blockly-plugin-block-dynamic-connection --save
+npm install @blockly/block-dynamic-connection --save
 ```
 
 ## Usage
 
-<!--
-  - TODO: Update usage.
-  -->
 ```js
 import * as Blockly from 'blockly';
-import {Plugin} from 'blockly-plugin-block-dynamic-connection';
-
-// Inject Blockly.
-const workspace = Blockly.inject('blocklyDiv', {
-  toolbox: toolboxCategories,
-});
-
-// Initialize plugin.
-const plugin = new Plugin(workspace);
-plugin.init();
+import * as BlockDynamicConnection from '@blockly/block-dynamic-connection';
 ```
 
 ## API
+- `overrideOldBlockDefinitions`: Replaces the Blockly default blocks with the
+  dynamic connection blocks. This enables projects to use the dynamic block
+  plugin without changing existing XML.
 
-<!--
-  - TODO: describe the API.
-  -->
+## XML
+```xml
+<block type="dynamic_text_join"></block>
+<block type="dynamic_list_create"></block>
+<block type="dynamic_if"></block>
+```
 
 ## License
 Apache 2.0
--- a/plugins/block-dynamic-connection/README.md
+++ b/plugins/block-dynamic-connection/README.md
@@ -1,0 +1,46 @@
+# blockly-plugin-block-dynamic-connection [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
+
+<!--
+  - TODO: Edit plugin description.
+  -->
+A [Blockly](https://www.npmjs.com/package/blockly) plugin that ...
+
+## Installation
+
+### Yarn
+```
+yarn add blockly-plugin-block-dynamic-connection
+```
+
+### npm
+```
+npm install blockly-plugin-block-dynamic-connection --save
+```
+
+## Usage
+
+<!--
+  - TODO: Update usage.
+  -->
+```js
+import * as Blockly from 'blockly';
+import {Plugin} from 'blockly-plugin-block-dynamic-connection';
+
+// Inject Blockly.
+const workspace = Blockly.inject('blocklyDiv', {
+  toolbox: toolboxCategories,
+});
+
+// Initialize plugin.
+const plugin = new Plugin(workspace);
+plugin.init();
+```
+
+## API
+
+<!--
+  - TODO: describe the API.
+  -->
+
+## License
+Apache 2.0
--- a/plugins/block-dynamic-connection/package.json
+++ b/plugins/block-dynamic-connection/package.json
@@ -8,6 +8,7 @@
     "clean": "blockly-scripts clean",
     "lint": "blockly-scripts lint",
     "prepublishOnly": "npm run clean && npm run build",
+    "predeploy": "npm run build && blockly-scripts predeploy",
     "start": "blockly-scripts start",
     "test": "blockly-scripts test"
   },
@@ -48,7 +49,10 @@
   "peerDependencies": {
     "blockly": ">=3"
   },
-  "publishConfig": {},
+  "publishConfig": {
+    "access": "public",
+    "registry": "https://wombat-dressing-room.appspot.com"
+  },
   "eslintConfig": {
     "extends": "@blockly/eslint-config"
   },
--- a/plugins/block-dynamic-connection/package.json
+++ b/plugins/block-dynamic-connection/package.json
@@ -41,7 +41,9 @@
   "devDependencies": {
     "@blockly/dev-scripts": "^1.2.4",
     "@blockly/dev-tools": "^2.0.10",
-    "blockly": "^4.20201217.0"
+    "blockly": "^4.20201217.0",
+    "chai": "^4.2.0",
+    "mocha": "^7.1.0"
   },
   "peerDependencies": {
     "blockly": ">=3"
--- a/plugins/block-dynamic-connection/package.json
+++ b/plugins/block-dynamic-connection/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "blockly-plugin-block-dynamic-connection",
+  "name": "@blockly/block-dynamic-connection",
   "version": "0.0.0",
   "description": "A group of blocks that add connections dynamically.",
   "scripts": {
--- a/plugins/block-dynamic-connection/package.json
+++ b/plugins/block-dynamic-connection/package.json
@@ -1,0 +1,60 @@
+{
+  "name": "blockly-plugin-block-dynamic-connection",
+  "version": "0.0.0",
+  "description": "A Blockly plugin.",
+  "scripts": {
+    "audit:fix": "blockly-scripts auditFix",
+    "build": "blockly-scripts build",
+    "clean": "blockly-scripts clean",
+    "lint": "blockly-scripts lint",
+    "prepublishOnly": "npm run clean && npm run build",
+    "start": "blockly-scripts start"
+  },
+  "main": "./dist/index.js",
+  "module": "./src/index.js",
+  "unpkg": "./dist/index.js",
+  "author": "",
+  "keywords": [
+    "blockly",
+    "blockly-plugin",
+    "block-dynamic-connection"
+  ],
+  "homepage": "https://github.com/ajpal/blockly-samples/tree/master/plugins/block-dynamic-connection#readme",
+  "bugs": {
+    "url": "https://github.com/ajpal/blockly-samples/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ajpal/blockly-samples.git",
+    "directory": "plugins/block-dynamic-connection"
+  },
+  "license": "Apache-2.0",
+  "directories": {
+    "dist": "dist",
+    "src": "src"
+  },
+  "files": [
+    "dist",
+    "src"
+  ],
+  "devDependencies": {
+    "@blockly/dev-scripts": "^1.2.4",
+    "@blockly/dev-tools": "^2.0.10",
+    "blockly": "^4.20201217.0"
+  },
+  "peerDependencies": {
+    "blockly": ">=3"
+  },
+  "publishConfig": {},
+  "eslintConfig": {
+    "extends": "@blockly/eslint-config"
+  },
+  "engines": {
+    "node": ">=8.17.0"
+  },
+  "browserslist": [
+    "defaults",
+    "IE 11",
+    "IE_Mob 11"
+  ]
+}
--- a/plugins/block-dynamic-connection/package.json
+++ b/plugins/block-dynamic-connection/package.json
@@ -8,7 +8,8 @@
     "clean": "blockly-scripts clean",
     "lint": "blockly-scripts lint",
     "prepublishOnly": "npm run clean && npm run build",
-    "start": "blockly-scripts start"
+    "start": "blockly-scripts start",
+    "test": "blockly-scripts test"
   },
   "main": "./dist/index.js",
   "module": "./src/index.js",
--- a/plugins/block-dynamic-connection/package.json
+++ b/plugins/block-dynamic-connection/package.json
@@ -1,7 +1,7 @@
 {
   "name": "blockly-plugin-block-dynamic-connection",
   "version": "0.0.0",
-  "description": "A Blockly plugin.",
+  "description": "A group of blocks that add connections dynamically.",
   "scripts": {
     "audit:fix": "blockly-scripts auditFix",
     "build": "blockly-scripts build",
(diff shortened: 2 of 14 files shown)


ajpal:
@googlebot I fixed it.

plugins/block-dynamic-connection/src/dynamic_list_create.js
```diff
@@ -0,0 +1,124 @@
+Blockly.Blocks['dynamic_list_create'] = {
```
alschmiedt, line 14: Add license header and fileoverview.

plugins/block-dynamic-connection/src/dynamic_text_join.js
```diff
@@ -0,0 +1,124 @@
+Blockly.Blocks['dynamic_text_join'] = {
```
alschmiedt, line 14: Add header.

plugins/block-dynamic-connection/src/index.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
alschmiedt: Update to 2021

plugins/block-dynamic-connection/src/index.js
```diff
@@ -0,0 +1,11 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import './insertion_marker_manager_monkey_patch';
+
+import './dynamic_text_join.js';
+
+import './dynamic_list_create.js';
```
alschmiedt: Remove extra space between the imports.

plugins/block-dynamic-connection/test/index.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Plugin test.
```
alschmiedt: Update the fileoverview description.

plugins/block-dynamic-connection/src/dynamic_list_create.js
```diff
@@ -0,0 +1,124 @@
...
+    }
+    const next = parseInt(xmlElement.getAttribute('next'));
+    this.inputCounter = next;
+  },
+
+  /**
+   * Check whether a new input should be added and determine where it should go.
+   * @param {Blockly.Connection} connection - The connection that has a
+   * pending connection.
+   * @return {Number} The index before which to insert a new input,
```
alschmiedt: Should be lowercase `number` because it is a primitive. If you are curious, more information on types can be found on their [github wiki](https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System#the-javascript-type-language).

plugins/block-dynamic-connection/README.md
```diff
@@ -0,0 +1,46 @@
+# blockly-plugin-block-dynamic-connection [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
```
alschmiedt: Just a reminder to go through and update any TODOs.

plugins/block-dynamic-connection/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "unpkg": "./dist/index.js",
+  "author": "",
+  "keywords": [
+    "blockly",
+    "blockly-plugin",
+    "block-dynamic-connection"
+  ],
+  "homepage": "https://github.com/ajpal/blockly-samples/tree/master/plugins/block-dynamic-connection#readme",
+  "bugs": {
+    "url": "https://github.com/ajpal/blockly-samples/issues"
```
alschmiedt: Update urls to be "google/blockly-samples" instead of "ajpal/blockly-samples".

plugins/block-dynamic-connection/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "blockly-plugin-block-dynamic-connection",
+  "version": "0.0.0",
+  "description": "A Blockly plugin.",
```
alschmiedt: Update description.

plugins/block-dynamic-connection/src/dynamic_list_create.js
```diff
@@ -0,0 +1,124 @@
...
+   * or null if no input should be added.
```
alschmiedt: Our linter should have caught this, but any comments that go onto a second line (like here and the @param above) should be indented.

plugins/block-dynamic-connection/src/dynamic_list_create.js
```diff
@@ -0,0 +1,124 @@
...
+      this.inputList[0]
+          .appendField(Blockly.Msg['LISTS_CREATE_WITH_INPUT_WITH']);
+    }
+    const next = parseInt(xmlElement.getAttribute('next'));
+    this.inputCounter = next;
+  },
+
+  /**
+   * Check whether a new input should be added and determine where it should go.
+   * @param {Blockly.Connection} connection - The connection that has a
```
alschmiedt: Remove '-'

plugins/block-dynamic-connection/src/dynamic_text_join.js
```diff
@@ -0,0 +1,124 @@
...
+    this.setTooltip(Blockly.Msg['TEXT_JOIN_TOOLTIP']);
+  },
+  /**
+   * Create XML to represent number of text inputs.
+   * @return {!Element} XML storage element.
+   * @this {Blockly.Block}
+   */
+  mutationToDom: function() {
+    const container = Blockly.utils.xml.createElement('mutation');
+    const inputNames = this.inputList.map((input) => input.name).join(',');
```
alschmiedt, line 48: Do we need to store the input names? Or could we get away with just storing the number of inputs?

plugins/block-dynamic-connection/src/dynamic_list_create.js
```diff
@@ -0,0 +1,124 @@
...
+      return connectionIndex + 1;
+    }
+
+    // Don't add new connection
+    return null;
+  },
+
+  /**
+   * Called when a block is dragged over one of the connections on this block.
+   * @param {Blockly.Connection} connection - The connection on this block that
```
alschmiedt: Here and throughout the file, if a type can't be null then there should be a ! in front of it. ex `{!Blockly.Connection}`.

plugins/block-dynamic-connection/src/dynamic_list_create.js
```diff
@@ -0,0 +1,124 @@
...
+    this.moveNumberedInputBefore(this.inputList.length - 1, insertIndex);
+  },
+
+  /**
+   * Called when a block is dropped into one of the connections on this block,
+   * or when a block drags over connections on this block and then is dragged
+   * away.
+   */
+  finalizeConnections: function() {
+    if (this.inputList.length > 2) {
```
alschmiedt: I think the 2 here and below is `minInputs`? I think making this into a variable would make that a bit more clear.

alschmiedt (commented):
Overall this looks good! Most of my comments are just nits about style and types.

plugins/block-dynamic-connection/src/dynamic_text_join.js
ajpal: Yeah we do, because the inputs won't always be ordered from 0. For example, if you make a block with four inputs, then remove the top two, the inputs will be named `ADD2` and `ADD3`. If we just stored the number, the next time we create the block we would name the inputs `ADD0` and `ADD1`, and the inputs wouldn't match.

BeksOmega:
Hello! I absolutely love this idea for a plugin, and I think it looks great so far :D But one thing I noticed is that all of the blocks are under new names, which means that applications with existing user projects saved as XML won't be able to upgrade to use this plugin. Should the new dynamic blocks override the built-in definitions instead? It might be better to leave as-is, I'm not sure.

alschmiedt:
Hello!
That is a great point. So we recommended that Anjali use different names because we were concerned about possible overlap in plugins and the blocks that they override. (Which is I think different than what we recommended for your plugin). We figured that if a user wanted to use these new blocks to override their original blocks they could set that themselves. Something like `Blockly.Blocks['old_name'] = Blockly.Blocks['new_name']`. However, Rachel brought up that it might be worth it to export a function that does these renames for the user. Something like `overrideOldBlockDefinitions`, which a user can call if they want it to work with old xml. @BeksOmega and @ajpal let me know what you think! Anjali, I think this just means we would need to export that extra function and add information in the readme about how to use it.

plugins/block-dynamic-connection/src/dynamic_text_join.js
```diff
@@ -0,0 +1,136 @@
...
+ */
+
+/**
+ * @fileoverview Defines a version of the text_join block with dyanmic
+ *    inputs that appear when a block is dragged over inputs on the block.
+ */
+
+Blockly.Blocks['dynamic_text_join'] = {
+  inputCounter: 2,
+  minInputs: 2,
```
alschmiedt: Add jsdocs.

plugins/block-dynamic-connection/src/dynamic_if.js
```diff
@@ -0,0 +1,130 @@
+Blockly.Blocks['dynamic_if'] = {
+  inputCounter: 1,
+  minInputs: 1,
```
alschmiedt: Add a header and jsdocs to this file.

plugins/block-dynamic-connection/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "@blockly/dev-scripts": "^1.2.4",
+    "@blockly/dev-tools": "^2.0.10",
+    "blockly": "^4.20201217.0",
+    "chai": "^4.2.0",
+    "mocha": "^7.1.0"
+  },
+  "peerDependencies": {
+    "blockly": ">=3"
+  },
+  "publishConfig": {},
```
alschmiedt: This is another field out create script should fill out, however we are still working on that : ) in the meantime can you add: 
```
  "publishConfig": {
    "access": "public",
  },
```

plugins/block-dynamic-connection/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "blockly-plugin-block-dynamic-connection",
+  "version": "0.0.0",
+  "description": "A group of blocks that add connections dynamically.",
+  "scripts": {
+    "audit:fix": "blockly-scripts auditFix",
+    "build": "blockly-scripts build",
+    "clean": "blockly-scripts clean",
+    "lint": "blockly-scripts lint",
+    "prepublishOnly": "npm run clean && npm run build",
```
alschmiedt, line 10: Add `"predeploy": "npm run build && blockly-scripts predeploy",`

plugins/block-dynamic-connection/src/dynamic_if.js
```diff
@@ -0,0 +1,130 @@
+Blockly.Blocks['dynamic_if'] = {
```
alschmiedt, line 14: Any file where we use Blockly, we need ```import Blockly from 'blockly/core';```

plugins/block-dynamic-connection/src/dynamic_if.js
```diff
@@ -0,0 +1,130 @@
...
+    this.setTooltip(Blockly.Msg['LISTS_CREATE_WITH_TOOLTIP']);
+  },
+
+  mutationToDom: function() {
+    const container = Blockly.utils.xml.createElement('mutation');
+    const inputNames = this.inputList
+        .filter((input) => input.name.includes('IF'))
+        .map((input) => input.name.slice(2)).join(',');
+    container.setAttribute('inputs', inputNames);
+    const hasElse = this.inputList[this.inputList.length - 1].name == 'ELSE';
```
alschmiedt: I think this can be: 
```
const hasElse = !!this.getInput('ELSE');
```

plugins/block-dynamic-connection/src/dynamic_if.js
```diff
@@ -0,0 +1,130 @@
...
+    this.appendStatementInput('DO0');
+
+    this.setTooltip(Blockly.Msg['LISTS_CREATE_WITH_TOOLTIP']);
+  },
+
+  mutationToDom: function() {
+    const container = Blockly.utils.xml.createElement('mutation');
+    const inputNames = this.inputList
+        .filter((input) => input.name.includes('IF'))
+        .map((input) => input.name.slice(2)).join(',');
```
alschmiedt: I think I would prefer a input.name.replace('IF') here. Mostly because I think it is a bit more clear as to what is happening.

plugins/block-dynamic-connection/src/dynamic_if.js
```diff
@@ -0,0 +1,130 @@
...
+    container.setAttribute('else', hasElse);
+    container.setAttribute('next', this.inputCounter);
+    return container;
+  },
+
+  domToMutation: function(xmlElement) {
+    const inputs = xmlElement.getAttribute('inputs');
+    if (inputs) {
+      const inputNumbers = inputs.split(',');
+      this.inputList = [];
```
alschmiedt: For removing the inputs, I think I would prefer if you did something [like this](https://github.com/google/blockly/blob/4be82c0b30c9f45bf2cb617f68e5011ba482ce05/blocks/logic.js#L452). Calling `this.removeInput` updates extra fields on the block (like statementInputCount). I don't think it makes a difference for your block, but I think it is a bit safer.

plugins/block-dynamic-connection/src/dynamic_if.js
```diff
@@ -0,0 +1,130 @@
...
+          this.insertElseIf(inputIndex + 2);
+        }
+      }
+    }
+  },
+
+  finalizeConnections: function() {
+    const inputInfo = this.inputList.map((input) => {
+      return {name: input.name, target: input.connection.targetConnection};
+    });
```
alschmiedt: Do we need this? Could we just use the input object instead?

plugins/block-dynamic-connection/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "blockly-plugin-block-dynamic-connection",
```
alschmiedt: Sould be updated to be @blockly/block-dynamic-connection

plugins/block-dynamic-connection/src/dynamic_if.js
```diff
@@ -0,0 +1,166 @@
...
+   * @type {number}
+   */
+  inputCounter: 1,
+
+  /**
+   * Minimum number of inputs for this block.
+   * @type {number}
+   */
+  minInputs: 1,
+  /**
```
alschmiedt: There should be a line break between the fields and the next comment. I think this happens a few times in this file and in `dynamic_list_create`.

plugins/block-dynamic-connection/src/dynamic_if.js
```diff
@@ -0,0 +1,166 @@
...
+
+    this.appendValueInput('IF0')
+        .setCheck('Boolean')
+        .appendField(Blockly.Msg['CONTROLS_IF_MSG_IF'], 'if');
+    this.appendStatementInput('DO0');
+
+    this.setTooltip(Blockly.Msg['LISTS_CREATE_WITH_TOOLTIP']);
+  },
+
+  mutationToDom: function() {
```
alschmiedt, line 52: This file needs jsdocs added to the functions.

plugins/block-dynamic-connection/src/dynamic_if.js
```diff
@@ -0,0 +1,166 @@
...
+   * @this {Blockly.Block}
+   */
+  init: function() {
+    this.setHelpUrl(Blockly.Msg['CONTROLS_IF_HELPURL']);
+    this.setStyle('logic_blocks');
+
+    this.appendValueInput('IF0')
+        .setCheck('Boolean')
+        .appendField(Blockly.Msg['CONTROLS_IF_MSG_IF'], 'if');
+    this.appendStatementInput('DO0');
```
alschmiedt: We need to add the text field for the "do" that sits in front of the statement inputs. The message string is `CONTROLS_IF_MSG_THEN`.

We also need to add the next and previous statement to the block.
```
this.setNextStatement(true);
this.setPreviousStatement(true);
```

plugins/block-dynamic-connection/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "src": "src"
+  },
+  "files": [
+    "dist",
+    "src"
+  ],
+  "devDependencies": {
+    "@blockly/dev-scripts": "^1.2.4",
+    "@blockly/dev-tools": "^2.0.10",
+    "blockly": "^4.20201217.0",
```
alschmiedt, line 45: This can be updated to our latest release.

alschmiedt (commented):
A few final comments, then lgtm : )

plugins/block-dynamic-connection/package.json
ajpal: When I bump the version, some of the serialization tests fail with a similar error to what I was running into before, which ended up being due to not importing `Blockly` in the src files.
image

I also get these warnings when I run `npm run test`:
image

Again, only happens when I bump the blockly version to `5.20210325.0`. Does this look familiar to you?

alschmiedt: Yeah, I think it has to do with #4573. I tried with other packages, and it looks like we are getting a similar error when we update. I don't think this is a problem related to this plugin, so it won't hold up this PR.

plugins/block-dynamic-connection/src/dynamic_if.js
```diff
@@ -0,0 +1,203 @@
...
+   * has a pending connection.
+   */
+  onPendingConnection: function(connection) {
+    if (connection.type === Blockly.NEXT_STATEMENT && !this.getInput('ELSE')) {
+      this.appendStatementInput('ELSE')
+          .appendField(Blockly.Msg['CONTROLS_IF_MSG_ELSE'], 'else');
+    }
+    const inputIndex = this.findInputIndexForConnection(connection);
+    const input = this.inputList[inputIndex];
+    if (connection.targetConnection && input.name.includes('IF')) {
```
alschmiedt, line 151: I was playing with the plugin, and ran into the problem where `input` was null here. I am having difficulty reproducing the error, however I think it makes sense to handle the case where `inputIndex` is `undefined`. Or, I think it would be better if we returned `null` from `firstInputIndexForConnection` and update the type to be `{?number}`. (The ? means that the type can be null.)

alschmiedt (approved):
One final comment, then LGTM.

alschmiedt:
Thanks @ajpal for all your work on this! 

This will be published with the rest of our blockly samples this Thursday.

PR #595, opened 2021-03-16
Make playground createWorkspace optional

samelhusseini:
Use Blockly.inject as the default workspace creator resulting in:
- createWorkspace
- defaultOptions

as optional parameters to `createPlayground`.

### DIFF
--- a/plugins/dev-tools/src/playground/index.js
+++ b/plugins/dev-tools/src/playground/index.js
@@ -72,16 +72,17 @@ let PlaygroundAPI;
 /**
  * Create the Blockly playground.
  * @param {!HTMLElement} container Container element.
- * @param {CreateWorkspaceFn} createWorkspace A workspace creation method called
- *     every time the toolbox is re-configured.
- * @param {Blockly.BlocklyOptions} defaultOptions The default workspace options
+ * @param {CreateWorkspaceFn=} createWorkspace A workspace creation method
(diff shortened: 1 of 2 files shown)


plugins/dev-tools/src/playground/index.js
```diff
@@ -72,16 +72,17 @@ let PlaygroundAPI;
...
+ * @param {Blockly.BlocklyOptions=} defaultOptions The default workspace options
  *     to use.
  * @param {PlaygroundConfig=} config Optional Playground config.
  * @param {string=} vsEditorPath Optional editor path.
  * @return {Promise<PlaygroundAPI>} A promise to the playground API.
  */
-export function createPlayground(container, createWorkspace,
-    defaultOptions, config = {}, vsEditorPath) {
+export function createPlayground(container, createWorkspace =
+    (blocklyDiv, options) => Blockly.inject(blocklyDiv, options),
```
alschmiedt: Just curious, where is `blocklyDiv` and `options` getting set?

alschmiedt (approved):

plugins/dev-tools/src/playground/index.js
samelhusseini: Called when the playground is ready to generate the workspace: https://github.com/google/blockly-samples/blob/master/plugins/dev-tools/src/playground/index.js#L280

PR #600, opened 2021-03-18
Add support for parameterized types to type hierarch (only) - project static

BeksOmega:
### Description

The first step towards Lists, Dictionaries, Functions, and Sets!

This PR adds support for parameters to the TypeDef class, and adds logic in the TypeHierarchy to initialize those values.

Every TypeDef now has a property called `params` which is an array of ParamDefs. ParamDefs have a `name` and a `variance`, which is specified using json like:
```javascript
{
  'List': {
    'params': [
      {
        'name': 'A',
        'variance': 'co'
      }
    ]
  }
}
```
For information about what variance is and how it works see the [glossary](https://docs.google.com/document/d/1QKYkmWjkle1JWCi3O8jXr8-7Toazh1pW_4EaVVgB_OI/edit#heading=h.bviy0z22j7ic) section of the design document.

Because types can now have parameters, we need a way to subtype those types and fulfill those parameters. This is done using a string (like the below) which can be parsed into a TypeStructure. TypeStructures are like the instantiation of a TypeDef. They have a `name` and a `params` array, which contains other TypeStructures. As more support for parameterized types is added, many functions will switch from using strings to TypeStructures.
```javascript
{
  'ProducerList': {
    'params': [
      {
        'name': 'A',
        'variance': 'inv'
      }
    ]
  },
  'List': {
    'fulfills': ['ProducerList[A]'],  // Here!
    'params': [
      {
        'name': 'A',
        'variance': 'inv'
      }
    ]
  }
}
```
One important function to look at is the `getParamsForAncestor` method. This method puts the params of the TypeDef the method is called on into a state so that they can easily be compared to a type of an ancestor. This is important for future type checking. For example say we had the following hierarchy definition:
```javascript
{
  "List": {
    "params": [
      {
        "name": "T",
        "variance": "inv"
      }
    ]
  },
  "Tuple": {
    "params": [
      {
        "name": "A",
        "variance": "co"
      },
      {
        "name": "B",
        "variance": "co"
      }
    ]
  },
  "Dictionary": {
    // Dictionary fulfills all functionality of a list of tuples
    //  (most don't but just for example's sake)
    "fulfills": [
      "List[Tuple[K, V]]"
    ],
    "params": [
      {
        "name": "K",
        "variance": "inv"
      },
      {
        "name": "V",
        "variance": "inv"
      }
    ]
  }
}
```
And that we wanted to see if a Dictionary[String, Dog] can be used as a List[Tuple[String, Dog]]. To do this we need to "map" [String, Dog] to Tuple[String, Dog], which this system handles.

# Testing

* Adds a bunch of tests for getParamsForAncestor that should cover all edge cases.
* Adds some tests for TypeStructure parsing, but this will get filled out more in the next PR when I move the parsing logic.

### DIFF
--- a/plugins/nominal-connection-checker/src/type_hierarchy.js
+++ b/plugins/nominal-connection-checker/src/type_hierarchy.js
@@ -250,8 +250,13 @@ export class TypeHierarchy {
     }, [types[0]]);
   }
 
-  // TODO: Might be useful to look into how lexers do this.
   // TODO: Move all the TypeStructure stuff into its own file in the next PR.
+  /**
+   * Parses a string into a TypeStructure.
+   * @param {string} str The string to parse into a TypeStructure.
+   * @return {!TypeStructure} The created TypeStructure.
+   * @private
+   */
   parseType_(str) {
     const typeStruct = {};
     const bracketIndex = str.indexOf('[');
@@ -267,6 +272,13 @@ export class TypeHierarchy {
     return typeStruct;
   }
 
+  /**
+   * Parses a string into an array of type structures.
+   * @param {string} str The string to parse into an array of type structures.
+   *     The string is expected to not be surrounded by brackets.
+   * @return {!Array<!TypeStructure>} The created TypeStructure array.
+   * @private
+   */
   parseParamsArray_(str) {
     const params = [];
     let latestIndex = 0;
@@ -362,7 +374,7 @@ class TypeDef {
 
     /**
      * The caseless names of the parameters of this type.
-     * @type {!Array<Param>}
+     * @type {!Array<ParamDef>}
      * @private
      */
     this.params_ = [];
@@ -574,10 +586,21 @@ class TypeDef {
     return this.paramsMap_.get(ancestorName);
   }
 
+  /**
+   * Returns the index of the parameter with the given name.
+   * @param {string} paramName The name of the parameter.
+   * @return {number} The index of hte parameter.
+   */
   getIndexOfParam(paramName) {
     return this.params_.findIndex((param) => param.name == paramName);
   }
 
+  /**
+   * Returns the parameter definition for the parameter at the given index.
+   * @param {number} index The index to get the parameter definition of.
+   * @return {!ParamDef} The parameter definition for the parameter at the
+   *     given index.
+   */
   getParamForIndex(index) {
     return this.params_[index];
   }
--- a/plugins/nominal-connection-checker/src/type_hierarchy.js
+++ b/plugins/nominal-connection-checker/src/type_hierarchy.js
@@ -54,7 +54,7 @@ export class TypeHierarchy {
     // cyclic type hierarchy (eg Dog <: Mammal <: Dog). They are expected to
     // not do that.
 
-    // Init types, and direct supers.
+    // Init types, direct supers, and parameters.
     for (const typeName of Object.keys(hierarchyDef)) {
       const lowerCaseName = typeName.toLowerCase();
       const type = new TypeDef(lowerCaseName);
@@ -62,6 +62,12 @@ export class TypeHierarchy {
       if (info.fulfills && info.fulfills.length) {
         info.fulfills.forEach(
             (superType) => type.addSuper(superType.toLowerCase()));
+      }
+      if (info.params && info.params.length) {
+        info.params.forEach((param) => {
+          type.addParam(
+              param.name.toLowerCase(), stringToVariance(param.variance));
+        });
       }
       this.types_.set(lowerCaseName, type);
     }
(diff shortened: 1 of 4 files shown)


plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -54,14 +54,20 @@ export class TypeHierarchy {
     // cyclic type hierarchy (eg Dog <: Mammal <: Dog). They are expected to
     // not do that.
 
-    // Init types, and direct supers.
+    // Init types, direct supers, and parameters.
```
rachel-fenichel, line 57: Meta note: each change has only added a few lines to `initTypes`, but it's getting long and unwieldy. Consider breaking it apart in a future PR.

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -199,15 +206,17 @@ export class TypeHierarchy {
...
+   * @param {string} subType The stringified possible subtype.
+   * @param {string} superType The stringified possible supertype.
    * @return {boolean} True if the types are identical, or if the first type
    *     fulfills the second type (directly or via its supertypes) as specified
    *     in the type hierarchy definition. False otherwise.
    */
-  typeFulfillsType(subName, superName) {
-    return this.types_.get(subName.toLowerCase())
-        .hasAncestor(superName.toLowerCase());
+  typeFulfillsType(subType, superType) {
```
rachel-fenichel, line 215: This is reparsing both types every time it's called. Is it called enough to cache that information?

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -240,8 +249,82 @@ export class TypeHierarchy {
           });
     }, [types[0]]);
   }
+
+  // TODO: Move all the TypeStructure stuff into its own file in the next PR.
```
rachel-fenichel, line 253: :+1:

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -326,6 +434,22 @@ class TypeDef {
...
+   * Adds the given parameter info to the list of parameters of this type.
+   * @param {string} paramName The caseless name of the parameter.
+   * @param {!Variance} variance The variance of the parameter.
+   * @param {number=} index The index to insert the parameter at. If undefined,
+   *     the parameter will be added at the end.
+   */
+  addParam(paramName, variance, index = undefined) {
+    const param = new ParamDef(paramName, variance);
+    if (index != undefined) {
+      this.params_.splice(index, param);
```
rachel-fenichel: This is missing the first optional parameter, `deleteCount`. Should be `this.params_.splice(index, 0, param)`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice#syntax

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -439,4 +563,140 @@ class TypeDef {
...
+  getParamsForAncestor(ancestorName, actualTypes = undefined) {
+    if (ancestorName == this.name && !this.paramsMap_.has(this.name)) {
+      // Convert this type's params to a type structure.
+      this.paramsMap_.set(
+          this.name,
+          this.params_.map((param) => {
+            return {name: param.name, params: []};
+          }));
+    }
+    const params = this.paramsMap_.get(ancestorName);
```
rachel-fenichel: Is this getting a copy, or are you directly modifying the `paramsMap` with the following `replaceFn`?

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -439,4 +563,140 @@ class TypeDef {
...
+   */
+  hasParameter(paramName) {
+    return this.params_.some((param) => param.name == paramName);
+  }
+
+  /**
+   * Returns the index of the parameter with the given name, or -1 if the
+   * parameter does not exist..
+   * @param {string} paramName The name of the parameter.
+   * @return {number} The index of hte parameter.
```
rachel-fenichel: `hte` -> `the`

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -439,4 +563,140 @@ class TypeDef {
...
+   *     given index.
+   */
+  getParamForIndex(index) {
+    return this.params_[index];
+  }
+
+  /**
+   * Returns the ParamDef with the given name, or undefined if not found.
+   * @param {string} paramName The name of the parameter to find the
+   *     ParamDef of.
```
rachel-fenichel: We're trying to reduce lint so that we can turn on warnings as errors for pull requests--please fix this one by moving `the` down a line.

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Hmm quite possibly! But in a future PR I move the parsing into the connection checker, so I don't want to touch it right now.

I've created an item in my task manager to look into caching :D

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Thanks for the catch!

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Yeah this should be doing a copy, fixed :D In the next PR (where I move and expand the type structure stuff) I'll change it to do a deep copy as well.

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: After all of the parameterized type stuff, there's a branch where I cleanup some todos. I've added breaking up initTypes to that branch ([relevant commit](https://github.com/BeksOmega/blockly-samples/commit/766c0feb7275200eba5afce7ee016a29caac0d13))

PR #617, opened 2021-03-26
Add predeploy script to plugins

moniika:


samelhusseini (approved):

maribethb (approved):
I realized the template is also missing `"test": "blockly-scripts test"` if you want to add that one as well while you're here, or I can do it separately

PR #618, opened 2021-03-26
Add missing dev scripts to plugin template.

### LINKED ISSUE #629: Add predeploy to package.json templates in dev-create
**Category**
Plugins

**Component**
dev-create

When we create a new plugin we need predeploy defined in the package.json, otherwise it will cause issues when the publish script is run.

I'd like for predeploy to be included in the template so it's not forgotten.

No change. We manually add predeploy to scripts in package.json when we create a plugin.

**Additional context**

maribethb:
When creating a new plugin from `dev-create` script, the resulting package.json is missing the `predeploy` and `test` scripts. This PR adds them to the template so that new plugins are created with the proper scripts. 

I verified this has the expected output with `node  ./dev-create/bin/create.js test-plugin` and the resulting plugin has the expected scripts.

One question I have is if this also needs to be added to `bin/create.js` as there is also a scripts section there. However, I'm not sure when that would be triggered over the file here since it would always have a template version available.

Fixes #629


plugins/dev-create/templates/plugin/template.json
```diff
@@ -5,8 +5,10 @@
     "build": "blockly-scripts build",
     "clean": "blockly-scripts clean",
     "lint": "blockly-scripts lint",
+    "predeploy": "blockly-scripts predeploy",
```
samelhusseini, line 8: These should be updated for all template.jsons (field, block, etc)

samelhusseini:
https://github.com/google/blockly-samples/blob/master/plugins/dev-create/bin/create.js#L125 this is the fallback if a template doesn't have a scripts section. Which I think they all do right now. No harm in adding it there too.

plugins/dev-create/templates/plugin/template.json
maribethb: Okay, added it to all types and the create script. Also added auditFix to the create script.

samelhusseini (approved):

PR #646, opened 2021-04-01
Move TypeStructure into new file - project static

BeksOmega:
### Description

* Changes the TypeStructure type def into an actual class.
* Adds an `.equals()` function for deep equality testing.
* Moves all of the old parsing logic into the new file. 
* Adds a structureToString function to convert the structure back to a human-readable string.
* Adds clear errors for missing brackets/misplaced brackets and other common problems.

### Testing

* Includes all of the previous unit tests for parsing.
* Adds tests for structureToString.
* Adds tests for errors.


BeksOmega:
I tested caching on the fully-functional parameters branch, and it does seem to make things a bit faster (as you'd expect). I recorded the performance of dragging the below set of  `Select Random` blocks toward the `Uses GetterList of Mammal` block, with caching and without.

TestBlocks

(All of those blocks have to be traversed)

I did 3 trials for each, and recorded the results in [this google doc](https://docs.google.com/spreadsheets/d/1Jecd1rvE5CaUuZhiCpUQwYG9cLosjdYPUc9E4N9KKmw/edit?usp=sharing). I don't think there's enough data there to give a statistically significant result (idk I haven't taken statistics) but it does seem to support the idea that caching is faster :D

plugins/nominal-connection-checker/src/type_structure.js
```diff
@@ -0,0 +1,292 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
rachel-fenichel, line 3: 2021

rachel-fenichel (commented):
One nit and a trailing comma to fix, and then I'll merge it.

PR #649, opened 2021-04-02
Gh pages script

alschmiedt:
This is a script to update all plugins and examples to the latest beta version of Blockly.

In order for the examples to work I would have to go through and update all of the index files to rely on the `blockly_compressed` file it finds in the node_modules, instead of getting the file from unpkg. 

I did this for turtle and single direction scrollbars to make sure we were ok with this approach, before I did it for the rest of the examples.


gulpfile.js
```diff
@@ -265,6 +265,32 @@ function deployToGhPages(repo) {
   };
 }
 
+/**
+ * Prepares plugins to be tested locally.
+ * @param {Function} done Completed callback.
+ */
+function prepareToTestPlugins(done) {
```
samelhusseini: preparePluginsForBeta? might be more descriptive of what this is doing.

gulpfile.js
```diff
@@ -265,6 +265,32 @@ function deployToGhPages(repo) {
...
+  execSync(`npm run boot`, {stdio: 'inherit'});
+  execSync(`npm run deploy:prepare:plugins`, {stdio: 'inherit'});
+  done();
+}
+
+/**
+ * Prepares examples to be tested locally.
+ * @param {Function} done Completed callback.
+ */
+function prepareToTestExamples(done) {
```
samelhusseini: prepareExamplesForBeta

package.json
```diff
@@ -24,7 +24,8 @@
         "publish:check": "lerna changed",
         "publish:dryrun": "gulp publishDryRun",
         "publish:force": "gulp forcePublish",
-        "publish:_internal": "lerna publish"
+        "publish:_internal": "lerna publish",
+        "test:ghPages": "gulp testGhPagesLocally"
```
samelhusseini: rather not use capitalization in npm scripts, change too `test:ghpages`

samelhusseini (approved):

PR #653, opened 2021-04-06
Add params checks to hierarchy validation - project static

BeksOmega:
### Description

Adds 10 new checks to hierarchy validation :D They include:
* Checking that syntactic characters are not used as names (comma, space, left bracket, right bracket)
* Checking that the 'params' property is an array.
* Checking that the names of the params are generic.
* Checking that the names of the params do not conflict.
* Checking that the variances of the params are valid.
* Checking that the 'fulfills' property is an array.
* Checking that supers in the 'fulfills' array parse properly.
* Checking that all of the params passed to supers are defined.
* Checking that the number of params passed to the supers are correct.
* Checking that the variances of the params passed to the supers are compatible with the supers' variances.

### Testing

Unit tests have been added for all new checks.

### DIFF
--- a/plugins/nominal-connection-checker/src/hierarchy_validation.js
+++ b/plugins/nominal-connection-checker/src/hierarchy_validation.js
@@ -10,6 +10,7 @@
 'use strict';
 
 import {isGeneric} from './utils';
+import {Variance, stringToVariance} from './type_hierarchy';
 import {TypeStructure, parseType,
   structureToString, TypeParseError} from './type_structure';
 
@@ -38,6 +39,7 @@ export function validateHierarchy(hierarchyDef) {
   checkSupersDefined(hierarchyDef);
   checkSuperParamsDefined(hierarchyDef);
   checkSuperNumParamsCorrect(hierarchyDef);
+  checkSuperParamVariancesCompatible(hierarchyDef);
   checkCircularDependencies(hierarchyDef);
 }
 
@@ -181,7 +183,7 @@ function checkSuperParamsDefined(hierarchyDef) {
 
 /**
  * Checks the hierarchy def for any parameterized super types that are not
- * provied the correct number of types. Also checks for cases where generic
+ * provided the correct number of types. Also checks for cases where generic
  * parameters are given type lists, which is not allowed.
  * @param {!Object} hierarchyDef The definition of the type hierarchy.
  */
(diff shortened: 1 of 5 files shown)


plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -24,18 +27,34 @@ export function validateHierarchy(hierarchyDef) {
...
 }
 
 /**
  * Checks the hierarchyDef for conflicting types.
  * @param {!Object} hierarchyDef The definition of the type hierarchy.
  */
 function checkConflictingTypes(hierarchyDef) {
-  const conflictMsg = 'The type name \'%s\' conflicts with the type name(s) %s';
+  const conflictMsg =
+      'The type name \'%s\' conflicts with the type name(s) [%s]';
```
alschmiedt: You can use a template literal here if you want to avoid escaping the single quotes.

plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -10,6 +10,9 @@
 'use strict';
 
 import {isGeneric} from './utils';
+import {Variance, stringToVariance, VarianceError} from './type_hierarchy';
+import {TypeStructure, parseType,
+  structureToString, TypeParseError} from './type_structure';
 
 /**
  * Validates the given hierarchy definition. Does checks for duplicate types,
```
alschmiedt, line 18: Doesn't need to be in this PR, but I think it would be nice if we could eventually give `hierarchyDef` a more specific type.

plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -77,12 +279,244 @@ function checkSupersDefined(hierarchyDef) {
...
+          throw e;
+        } // Otherwise it will have been handled by our specific check.
+      }
+    }
+  }
+}
+
+/**
+ * Checks the hierarchy def for any parameterized super types that are not
+ * provided the correct number of types. Also checks for cases where generic
```
alschmiedt, line 300: "providing"?

plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -77,12 +279,244 @@ function checkSupersDefined(hierarchyDef) {
...
+    }
+  }
+}
+
+/**
+ * Checks the hierarchy def for any parameterized super types that are not
+ * provided the correct variance of parameter.
+ * @param {!Object} hierarchyDef The definition of the type hierarchy.
+ */
+function checkSuperParamVariancesCompatible(hierarchyDef) {
```
alschmiedt, line 378: This is mostly my personal opinion as there is not a lot in the style guide to back me up here, so I am fine with whatever you decide to do : ) 

This function is a bit difficult to read imo. The multiple layers of nested functions makes it difficult for me to tell what code is actually being run in the parent function. I think this could be fixed by either: 
- Moving these to a new file, if you feel like you want to keep all the functions grouped together
- There are quite a few nested functions that look like helper functions for variances, I wonder if these could be grouped together possibly closer to the variance definition?
- Or by simply refactoring it so that the functions are defined at the same level. 

Even if you wanted to just remove one of the layers of nesting, I think that would help as well.

plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -102,37 +536,38 @@ function checkCircularDependencies(hierarchyDef) {
...
-      currentTraversal.push(typeName);
-      logCircularDependency(currentTraversal);
-      currentTraversal.pop();
-      return;
-    }
+  function searchForCyclesRec(typeName, currentTraversal) {
+    try {
+      const typeStructure = parseType(typeName, false);
+      const caselessName = typeStructure.name.toLowerCase();
+      const typeInfo = types.get(caselessName);
```
alschmiedt, line 551: It looks like this function relies on being called in a place where `types` has already been defined with the correct type? If this is the case it seems a bit fragile to me. For a few reasons: 
- It is non obvious that this function relies on types. 
- It is non obvious what type `types` needs to be for this function to work properly.

If possible, I think it would be a bit more clear if we just pass it in as a parameter.

plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -57,18 +76,201 @@ function checkConflictingTypes(hierarchyDef) {
...
+
+  for (const type of Object.keys(hierarchyDef)) {
+    const typeDef = hierarchyDef[type];
+    if (!typeDef.params || !Array.isArray(typeDef.params)) {
+      continue;
+    }
+
+    // Map of caseless names to cased names.
+    const definedParamNames = new Map();
+    // Map of original names to arrays of conflicting names.
```
alschmiedt: Update to "to array of [position, conflicting name]" or something that mentions we are storing the position as well.

plugins/nominal-connection-checker/src/hierarchy_validation.js
BeksOmega: Provided is correct in this case =) Parameterized types are sort of like function calls. You provide them parameters.

So if you have a type hierarchy like this:
```
{
  'typeA': {
    'params': [
      {
         'name': 'A',
         'variance': 'co',
      },
    ],
  },
  'typeB': {
    'fulfills': ['typeA']
  }
}
```
The supertype typeA of typeB is not being provided enough parameters. It expects 1 but is getting 0.

plugins/nominal-connection-checker/src/hierarchy_validation.js
BeksOmega: Yeah I definitely see what you're saying. I'll refactor them to all be at the same level!

alschmiedt (approved):
lgtm!

plugins/nominal-connection-checker/src/hierarchy_validation.js
BeksOmega: Yeah I would definitely love to do that for some of the other files! I'm not sure if you can create some sort of JSON definition in the same way that core has that XML definition, but if it were possible that would be super cool :D 

I'm not sure if type info should be added to this file though... Because this one is sort of meant to take in whatever random object, and then help devs figure out what's broken. Basically I'm worried that adding typing will lead me (or future devs) to make assumptions about the passed object which I shouldn't be making - hence causing uninformitive errors. Do you get where I'm coming from? or am I thinking about it the wrong way?

alschmiedt: I was thinking of just adding a typedef similar to how we do for the [toolbox definition](https://github.com/google/blockly/blob/3fe56b083c62549cb2bdb17270846bf17ced5d7a/core/utils/toolbox.js#L25).

I definitely understand where you are coming from, however I think that the clarity of knowing what the object is suppose to look like outweighs the confusion that might be caused from a developer assuming that certain fields are on the object. 

The way we generally treat types in Blockly is that they are a contract between the method and the developer. If we say a method needs a string, we generally don’t go out of our way to make sure that the value they give us is in fact a string. We just carry on as if it were a string and any errors are up to the developer to solve.

I think you are trying to do some extra validation here that provides information that a compiler would normally provide to a developer. One way to accomplish this while still using a more specific type would be to: 
- Validate the shape of the object using your functions like `checkFulfillsIsArray` and `checkParamsIsArray`. 
- Throw an error if something is incorrect
- Safely cast the object to your type now that you know it is the correct shape.

Let me know if this seems reasonable or if you have any other questions/concerns : )

BeksOmega: > I was thinking of just adding a typedef similar to how we do for the toolbox definition.

Ah that makes sense!

> Let me know if this seems reasonable or if you have any other questions

Yeah that seems totally reasonable! And I'm definitely onboard with adding some more type info. But I'm also still a bit confused.

I think the point that's getting me mixed up is that the rest of the plugin doesn't actually rely on this code. It's not getting run anywhere except by unit tests. So other places would expect the hierarchy definition to have the correct shape - but they're completely disconnected from this code.

You mention that it seems like the extra validation is kind of like the information a compiler would provide, and that was exactly the intention. The idea was that developers would run their hierarchy definition through this code while they were developing it, or as part of a test suite. But validateHierarchy() would never be run in production.

So I'm still not understand the advantage of putting more strict type info on this file specifically, when it doesn't actually rely on the type having that strict structure. It seems like should be able to take in whatever random Object, and throw informative errors accordingly.

Again, I'm totally cool with adding more type info. I'd just like to be unconfuzzled haha.

alschmiedt: Ah, this is my bad then! I didn't realize this file was only meant as a helper file for developers. I was assuming this was going to be eventually run on the hierarchyDef passed to TypeHierarchy. If that is not the case, then yes adding a typedef in other files, but not this one makes sense : )

Issue #635, opened 2021-03-30
Only run linter once on GH Actions


maribethb:
**Is your feature request related to a problem? Please describe.**

GitHub Actions runs our test script multiple times, for different versions of node. The test script includes linting. Lint errors are presumably going to be identical across Node versions. The result of this is that every lint error appears twice (three times in core) which contributes to notification fatigue. 

**Describe the solution you'd like**

1. Separate out the lint from the test script
2. Run linting as a separate step in the Actions pipeline, and only once
3. Lint errors should only appear once

**Describe alternatives you've considered**

We could just pick a version of node and have it skip linting for the other versions (would still require doing step 1). However, I like the idea of it being a separate action, that way you get more descriptive errors/status messages (e.g. the build succeeded but linting failed) whereas right now it is all under the same check.

**Additional context**

This isn't directly related to the release process situation we had last week, but it's in keeping with the theme of reducing false notifications so I'm adding it to the bash.

PR #662, opened 2021-04-08
Update build and test scripts to accept a skip-lint option

maribethb:
This PR adds a `--skip-lint` option to the `test` and `build` scripts. For example:

```
npm run test -- --skip-lint
```

This is required for #635. 

Before change: The linter is always run on build or test. This causes problems because the GitHub Action runs all of build, test, and lint, so all lint errors show up 3-9 times (for multiple versions of node, etc).
After change: If the flag is passed, webpack doesn't lint when it runs. If the flag is not passed, there is no change from before and linter is run.

Tested:
from `blockly-samples/plugins/block-plus-minus` ran

```
node ../dev-scripts/scripts/build -- --skip-lint
node ../dev-scripts/scripts/build 
node ../dev-scripts/scripts/test -- --skip-lint
node ../dev-scripts/scripts/test
```

 and saw correct behavior in all cases.

### DIFF
--- a/plugins/dev-scripts/config/webpack.config.js
+++ b/plugins/dev-scripts/config/webpack.config.js
@@ -67,28 +67,11 @@ module.exports = (env) => {
     blocklyAliasSuffix = '/dist';
   }
 
-  return {
-    target,
-    mode: isProduction ? 'production' : 'development',
-    entry: entry,
-    devtool: isProduction ? 'source-map' : 'eval-source-map',
-    output: {
-      path: isProduction ? resolveApp('dist') : resolveApp('build'),
-      publicPath: isProduction ? '/dist/' : '/build/',
-      filename: outputFile,
-      libraryTarget: 'umd',
-      globalObject: 'this',
-    },
-    resolve: {
-      alias: {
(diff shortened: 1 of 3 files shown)


plugins/dev-scripts/config/webpack.config.js
```diff
@@ -109,29 +92,55 @@ module.exports = (env) => {
...
+  // The following rules are always in effect.
+  rules.push(
+      // Load Blockly source maps.
+      {
+        test: /(blockly\/.*\.js)$/,
+        use: [require.resolve('source-map-loader')],
+        enforce: 'pre',
+      },
+      // Run babel to compile both JS and TS.
+      {
```
samelhusseini: Instead of forking the script with a conditional, you can follow a similar pattern to how we determine whether or not to run the TS checker plugin.

eg: 

```
[
 !env.skipLint && { ...}
].filter(Boolean)
```

maribethb: Genius, done, and tested the same way and it all still works.

samelhusseini (approved):

PR #668, opened 2021-04-15
Use the new positionable api in zoom-to-fit control

moniika:
Update zoom-to-fit control plugin to use new positionable apis from core provided in https://github.com/google/blockly/pull/4782.


moniika:
Decided to not include positionable APIs directly in core.

PR #670, opened 2021-04-15
Add support for parameters to getNearestCommonParents - project static

BeksOmega:
### Description

So this is kind of a lot. I'm going to go through some examples to explain all the stuff this code has to handle. There are also more examples in the tests.

So when we're trying to find the nearest common parent of two types, we not only need to find the nearest common parent of their "outer" types, but also the nearest common parents of their parameters. But the parameters don't always match up in convenient ways.

For example:
```
{
  'Collection': {
    'params': [
      {
        'name': 'A',
        'variance': 'inv', // for simplicity
      }
    ]
  }
},
{
  'List': {
    'fulfills': ['Collection[A]'],
    'params': [
      {
        'name': 'A',
        'variance': 'inv', // for simplicity
      }
    ]
  }
},
{
  'Dict': {
    'fulfills': ['Collection[B]'],
    'params': [
      {
        'name': 'A',
        'variance': 'inv', // for simplicity
      },
      {
        'name': 'B',
        'variance': 'inv', // for simplicity
      }
    ]
  }
}
```
In this case both List and Dict fulfill the type Collection, but List does so with its first param, while Dict does so with its second param. This means to see if List[String] and Dict[Number, String] have a common ancestor Collection[Something] we need to compare String and String. typeFulfillsType does this too so it should be pretty familiar.

Where it gets really tricky is when we have to deal with variance. Say we have 6 types.
  * GetterList[A] - covariant
  * AdderList[A] - contravariant
  * List[A] - invariant, and a subtype of GetterList and AdderList
  * Mammal
  * Dog - subtype of mammal
  * Bat - subtype of mammal
  * DogBat - subtype of Dog and bat

It is easy to find common parents of GetterList types. Because it is covariant with respect to its parameter, we just have to make sure that its parameters also share a common parent. Eg the common parent of GetterList[Dog] and GetterList[Bat] is GetterList[Mammal]

AdderList is harder. Because it is contravariant with respect to its parameter, we have to make sure that its parameters share a common *subtype*. And because we can have *nested* parameterized types (eg AdderList[AdderList[Dog]]) we can't just do a simple check, we need to *recursively* evaluate any parameters to ensure they are compatible, hence the addition of `getNearestCommonDescendants`. For example AdderList[Dog] and AdderList[Bat] have the common parent AdderList[DogBat] (even though DogBat is *subtype* of Dog and Bat).

List is once again pretty easy. Because it is invariant with respect ot its parameter, we just have to make sure they are entirely identical. For example List[Dog] and List[Dog] have the common parents GetterList[Dog], AdderList[Dog] and List[Dog] (List[Dog] being the *nearest* common parent).

Finally there are complications caused by allowing developers to specify concrete types in 'fulfills' statements in the hierarchy definition.

For example if we have the following type hierarchy definition:
```
            {
              'typeA': {
                'params': [
                  {
                    'name': 'A',
                    'variance': 'covariant',
                  },
                ],
              },
              'typeB': {
                'params': [
                  {
                    'name': 'A',
                    'variance': 'covariant',
                  },
                ],
              },
              'typeC': {
                'fulfills': ['typeA[A]', 'typeB[A]'],
                'params': [
                  {
                    'name': 'A',
                    'variance': 'covariant',
                  },
                  {
                    'name': 'B',
                    'variance': 'covariant',
                  },
                ],
              },
              'typeD': { },
              'typeE': { },
              'typeF': {
                'fulfills': ['typeD', 'typeE'],
              },
            }
```
And we want to find the nearest common **descendants** of typeA[typeD] and typeB[typeE] it should obviously be typeC[typeF, something] But what that "something" should be is unclear, and currently it just gets lost by the code. To fix this we need to be able to specify a generic param for "something", which means we need to be able to evaluate generic params. I actually just got this working yesterday :D But that's like 6 PRs from now.

Hopefully that gives enough context that the code will make sense. But if you have any questions definitely ask! And please tell me where I can add more comments, or break things up in to subprocedures.

### Testing

So many tests. Mostly tests for `getParamsForDescendant_`,  `getNearestCommonParent` and `getNearestCommonDescendant`. But I'm not going to list all of them haha.


plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -138,51 +178,64 @@ export class TypeHierarchy {
...
+  initNearest_(nearestCommonMap, relevantRelatives, isNearest) {
     const unvisitedTypes = new Set(this.types_.keys());
     while (unvisitedTypes.size) {
       for (const typeName of unvisitedTypes) {
         const type = this.types_.get(typeName);
-        const unvisitedSupers = type.supers().filter(
-            unvisitedTypes.has, unvisitedTypes);
-        if (unvisitedSupers.length) {
+        const hasUnvisited = !!relevantRelatives(type).filter(
+            unvisitedTypes.has, unvisitedTypes).length;
```
moniika: Since you use it multiple times, `relativeRelatives(type)` would be better as a local variable so you can reference it instead of calling the method each time.

You can use `some` here (this would also make it more efficient because `some` can exit early while filter would have to loop through all the array elements):
```
const hasUnvisited = relevantRelatives(type).some(unvisitedTypes.has, unvisitedTypes);
```

Although it takes more characters to write, I think it would be easier to read if you renamed `hasUnvisited` to `hasUnvisitedRelatives`

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -138,51 +178,64 @@ export class TypeHierarchy {
...
+          if (isNearest(type, otherTypeName)) {
+            nearestCommon.push(typeName);
           } else {
-            // Get all the least common ancestors this type's direct
-            // ancestors have with the otherType.
-            type.supers().forEach((superTypeName) => {
-              leastCommonAncestors.push(
-                  ...this.nearestCommonParents_.get(superTypeName)
+            // Get all the nearest common types this type's relevant relatives
+            // have with the otherType.
```
moniika: Get all the nearest common types **that** this type's relevant relatives **share** with otherType.

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -138,51 +178,64 @@ export class TypeHierarchy {
...
+              nearestCommon.push(
+                  ...nearestCommonMap.get(relTypeName)
                       .get(otherTypeName));
             });
-            // Only include types that have no descendants in the array.
-            leastCommonAncestors = leastCommonAncestors.filter(
+            // Remove types that have a nearer relative in the array.
+            nearestCommon = nearestCommon.filter(
                 (typeName, i, array) => {
                   return !array.some((otherTypeName) => {
```
moniika: For readability I would change 
`!array.some` -> `array.every` (and flip the returns)

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -138,51 +178,64 @@ export class TypeHierarchy {
...
         unvisitedTypes.delete(typeName);
 
         const map = new Map();
-        this.nearestCommonParents_.set(typeName, map);
+        nearestCommonMap.set(typeName, map);
         for (const [otherTypeName] of this.types_) {
-          let leastCommonAncestors = [];
-          if (type.hasDescendant(otherTypeName)) {
-            leastCommonAncestors.push(typeName);
+          let nearestCommon = [];
```
moniika, line 220: Why isn't this a set to prevent duplicates?

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -138,51 +178,64 @@ export class TypeHierarchy {
...
                       .get(otherTypeName));
             });
-            // Only include types that have no descendants in the array.
-            leastCommonAncestors = leastCommonAncestors.filter(
+            // Remove types that have a nearer relative in the array.
+            nearestCommon = nearestCommon.filter(
                 (typeName, i, array) => {
                   return !array.some((otherTypeName) => {
                     // Don't match the type against itself, but do match against
                     // duplicates.
```
moniika: Could you clarify this comment?

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -138,51 +178,64 @@ export class TypeHierarchy {
...
    * edges.
+   *
+   * @param {!Map<string, !Map<string, Array<string>>>} nearestCommonMap The
+   *      map of nearest common types (either ancestors or descendants) that we
+   *     are initializing.
+   * @param {function(TypeDef):!Array<string>} relevantRelatives Returns the
+   *     relatives that are relevant to this procedure. In the case of
+   *     ancestors, returns supertypes, and in the case of descendants, returns
+   *     subtypes.
+   * @param {function(TypeDef, string):boolean} isNearest Returns true if the
```
moniika: Instead of nearest, what about `isDirectRelative`. Is that an accurate name?

Also, the description "nearest common X" is not clear. Assuming direct relative is correct), here's an alternative description:
"Returns whether the type of the second, typeName, parameter is the direct relative of the type passed in as the first parameter."

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -138,51 +178,64 @@ export class TypeHierarchy {
...
+   * But the above has been slightly modified to work for both ancestors and
+   * descendants.
+   *
    * Operates in O(nm) where n is the number of nodes and m is the number of
    * edges.
+   *
+   * @param {!Map<string, !Map<string, Array<string>>>} nearestCommonMap The
+   *      map of nearest common types (either ancestors or descendants) that we
+   *     are initializing.
+   * @param {function(TypeDef):!Array<string>} relevantRelatives Returns the
```
moniika: To easier distinguish this as a function, I'd recommend update to `getRelevantRelatives` (otherwise without memorizing the type, it looks like it could be a list of relatives)

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -35,20 +36,34 @@ export class TypeHierarchy {
...
+     * the nearest common descendants of the two types. You can think of it like
+     * a two-dimensional array where both axes contain all of the type names.
+     *
+     * A nearest common descendant of two types u and v is defined as:
+     * A sub type of both U and v that has no ancestor which is also a
+     * descendant of both u and v.
+     * @type {!Map<string, !Map<string, Array<string>>>}
+     * @private
+     */
+    this.nearestCommonDescendants_ = new Map();
```
moniika: This and the `nearestCommonParents_` would benefit from starting with a description that focuses on the purpose, rather than the types, similar to how you document the init methods for this graph that explains how this is a directed graph that allows you to access the nearest common ancestor/descendant in constant time.

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -138,51 +178,64 @@ export class TypeHierarchy {
...
                 (typeName, i, array) => {
                   return !array.some((otherTypeName) => {
                     // Don't match the type against itself, but do match against
                     // duplicates.
                     if (array.indexOf(otherTypeName) == i) {
                       return false;
                     }
-                    return this.types_.get(typeName)
-                        .hasDescendant(otherTypeName);
+                    return isNearest(this.types_.get(typeName), otherTypeName);
```
moniika: For my understanding, this is how I interpret this:
We are filtering out any types that are direct relatives (direct ancestor or descendent, depending on which version) of other types in the `nearestCommon` array because if they are a direct relative of something else then they are a farther common relative.

You could probably add a comment here.

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -138,51 +178,64 @@ export class TypeHierarchy {
...
+            relevantRelatives(type).forEach((relTypeName) => {
+              nearestCommon.push(
+                  ...nearestCommonMap.get(relTypeName)
                       .get(otherTypeName));
             });
-            // Only include types that have no descendants in the array.
-            leastCommonAncestors = leastCommonAncestors.filter(
+            // Remove types that have a nearer relative in the array.
+            nearestCommon = nearestCommon.filter(
                 (typeName, i, array) => {
```
moniika: Can you use a different name than `typeName` here?
Since it is the same variable name as the what is used in the outer loop, it caused me confusion with the isNearest call inside of this filter function.

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -138,51 +178,64 @@ export class TypeHierarchy {
...
                   return !array.some((otherTypeName) => {
                     // Don't match the type against itself, but do match against
                     // duplicates.
                     if (array.indexOf(otherTypeName) == i) {
                       return false;
                     }
```
moniika: Using `indexOf` would iterate through the whole array, which seems like it could be unecessary.
Could we compare the strings `typeName` and `otherTypeName` instead?

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -261,34 +308,126 @@ export class TypeHierarchy {
...
-   * A nearest common parent of a set of types A is defined as:
+   * Returns an array of all the nearest common ancestors of the given types.
+   * A nearest common ancestor of a set of types A is defined as:
    * A super type of all types in A that has no descendant which is also an
    * ancestor of all types in A.
-   * @param {...string} types A variable number of types that we want to find
-   *     the nearest common parents of.
-   * @return {!Array<string>} An array of all the nearest common parents of the
-   *     given types.
+   * @param {...TypeStructure} types A variable number of types that we want to
```
moniika, line 317: Is `TypeStructure` related to `TypeDef`? What's the difference?

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -261,34 +308,126 @@ export class TypeHierarchy {
...
+      paramsLists = this.unifyParamsLists(
+          paramsLists,
+          commonDef,
+          this.getNearestCommonParents.bind(this),
+          this.getNearestCommonDescendants.bind(this));
+      paramsLists[0] = paramsLists[0].map((val) => [val]);
+      const combinations = this.combine_(paramsLists);
+      if (!combinations.length) {
+        // If this commonType didn't unify, maybe the parents of that type will.
+        commonTypes.push(...commonDef.supers());
```
moniika, line 354: could you continue here? (and use `push` instead of indexing to add to `mappedTypes`)

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -261,34 +308,126 @@ export class TypeHierarchy {
...
+        // Remove duplicates.
+        .filter((typeStructs, i, arrays) => {
+          const typeName = typeStructs[0].name;
+          return arrays.findIndex(
+              (typeStructs2) => typeStructs2[0].name == typeName) == i;
+        })
+        // Remove non-nearest types.
+        .filter((typeStructs, i, arrays) => {
+          const typeName = typeStructs[0].name;
+          return !arrays.some((typeStructs2, j) => {
```
moniika: Similar to previous comment. I think `arrays.every` would be clearer.

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -309,6 +448,186 @@ export class TypeHierarchy {
...
+    return typeNames.reduce((accumulator, currType) => {
+      const commonMap = nearestCommon.get(currType);
+      // Note: neither flatMap() nor flat() work on Node 10. See #431.
+      return accumulator
+          .map((type) => {
+            return [...commonMap.get(type)]; // Create copy to avoid corruption.
+          })
+          .reduce((flat, toFlatten) => {
+            return [...flat, ...toFlatten];
+          }, [])
```
moniika: Could you further explain what's happening here, maybe add a comment in the reduce function.

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -309,6 +448,186 @@ export class TypeHierarchy {
...
+      });
+    });
+    return paramsLists;
+  }
+
+  /**
+   * Returns an array of arrays of common types for each of the lists of actual
+   * types for a given parameter of the commonDef. Parameter lists are unified
+   * based on the variance of the formal parameter of the common type definition
+   * at that index.
```
moniika, line 517: What does "unified" mean?

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -309,6 +448,186 @@ export class TypeHierarchy {
...
+   *     of the second array onto. Should be an array of arrays for proper
+   *     combinating.
+   * @param {!Array<*>} secondArray An array of elements used to create
+   *     combinations.
+   * @param {!Array<!Array<*>>} rest The rest of the arrays of elements.
+   * @return {!Array<!Array<*>>} All combinations of elements in all of the
+   *     subarrays.
+   * @private
+   */
+  combine_([firstArray, ...[secondArray, ...rest]]) {
```
moniika, line 568: Is this used anywhere but with paramLists?

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: So originally this wasn't a set because the set API is kinda garbage haha. It doesn't support any comprehensions like filter, map, or reduce. This means that when I go to to filter out the types that have a nearer relative in the set, I have to convert it to an array first. Originally I avoided that so I wouldn't have to deal with the costs of conversion.

However the arrays/sets we're dealing with are super tiny in the *vast* majority of cases. Getting even as high as 10 would be really hard - I can't think of a case where you'd need 2 types to share 10 direct super types unless you were going a bit wild and created one type for each method on the class.

So if it's going to make readability easier I'll switch it over to a set. Please see my top level comment wrt the readability discussion.

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Done, but see my top level comment about readability.

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: This comment has been removed, but see my top level comment about readability.

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: We can, but we also have to do some checks on the indices (which is not a bad thing).

For example if we just do:
```javascript
['a', 'b', 'a'].filter((type, i, array) =>
    array.every((otherType, j) => type != otherType));
```
We'll end up with an empty array because every item will match against itself.

Instead you have to do a check like this:
```javascript
['a', 'b', 'a'].filter((type, i, array) =>
    array.every((otherType, j) => i <= j || type != otherType));
```

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Yeah I really don't like `isNearest` :/ But I'm not sure if `isDirectRelative` works either. Let me lay out the example I'm concerned about.

So if we have the a type hierarchy like below, and we are creating the nearestCommonDescendants map, `isNearest(c, a)` would return true.
```
a   x
 \ /
  b
  |
  c
```
Do you think that `isDirectRelative(c, a)` returning true for the above case feels right?

Right now I'm leaning towards going with `isDirectRelative` because it's obviously better than `isNearest`, but I wanted to see if you have any other thoughts given that example.

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Coolio! I added the local variable for relevantRelatives and I switched to `some`. Switching to `some` made the line short enough that I could just embed it in the if, so I got rid of `hasUnvisited` altogether.

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Switched to `nearestTypeName` and `otherNearestTypeName` for now. Do you think that works? Maybe it would actually be more readable to switch to abbreviated names like `type1`, `type2`, etc...

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: TypeDef stores information about the definition of a types. This includes how it relates to other types (its ancestors and descendants) as well as info about its parameters (their names, and variances).

A TypeStructure is kind of like an instances of a type. This plugin converts the strings on connection checks into TypeStructures to make the recursive type checking logic simpler. TypeStructures don't know anything about how they relate to anything, or what the variances of their params are. They just know the structure of the instance.

Hopefully that helps! But you can also checkout the definitions for [TypeDef](https://github.com/google/blockly-samples/blob/d4d114754ba0a12e5eb47ee98ffa7341ee557758/plugins/nominal-connection-checker/src/type_hierarchy.js#L317) and [TypeStructure](https://github.com/google/blockly-samples/blob/d4d114754ba0a12e5eb47ee98ffa7341ee557758/plugins/nominal-connection-checker/src/type_structure.js#L14). I might want to add some better inline docs for TypeDef haha.

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Definitely will do after we figure out what to name this function!

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Yes I can! Great suggestion :D

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Hmmm this is a pattern I use for for flattening an array of arrays (because Node 10 is stupid and doesn't think flattening is important I guess). I try to address this in the `Note:` comment at the top of the return. Do you think there's more detail that needs to be added?

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Unified means we get the array of common types. In this case the "direction" of the unification/common types is based on the variance of the formal parameter.

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: Not at the moment - but in a few PRs it gets moved into a utils file because it starts being used elsewhere.

The reason I separated it out here is b/c I've gotten comments on some past PRs about nested functions being hard to read.

BeksOmega (commented):
First of all, sorry about leaving a review instead of responding individually. I know it makes the conversations hard to read, but there was just a lot to respond to.

What I'd like to address here is some thoughts wrt readability and efficiency of removing duplicate elements from an array.

The pattern I usually use for this is filtering the array like so:
```javascript
array.filter((elem, i, arr) => arr.indexOf(elem) == i);
```
indexOf always returns the first instance of an item in an array. So if you're looking at the first instance, indexOf() will return the index of the first, and you'll know that item's ok. On the other hand if you're looking at the second instance, indexOf() will return the index of the first, and you'll know you can remove that item.

Once you're familiar with the pattern I think it's easy enough to read, however it's not very efficient.

As I address below, there are other ways that you can do this by adding some additional checks based on the indices:
```javascript
array.filter((elem, i, arr) =>
    arr.every((elem2, j) => i <= j || elem != elem2));
```

But both of these have the problem of being n^2. Another option is to use a Set() when initialy creating the collection, so that duplicates are impossible. But I don't really like this solution. (1) It requires the set to be converted to an array if we want to do any filtering, which I don't think is very readable. (2) It doesn't actually prevent duplicates in the case that we're creating a set of objects (it only works for primitive types). So I'd rather go with one of the other two solutions.

Which solution do you prefer?

And one other thing, I think that for readability it might be a good idea to separate out any filters that remove duplicates from other filters. That way we can keep them consistent and recognizable, and we won't have to use as many comments. Similar to how I try to keep the flattening `reduce()` consistent. Do you think that's a good idea?

plugins/nominal-connection-checker/src/type_hierarchy.js
moniika: I really like the new wording.

plugins/nominal-connection-checker/src/type_hierarchy.js
moniika: The `type1` and `type2` is nice since it's shorter (and I don't think we're getting much by adding `nearestType` in the front).
However, if we don't to that I think should sub `nearest` with `near` (i.e., `nearTypeName` and `otherNearTypeName`) since technically one of the types might not be "nearest".

plugins/nominal-connection-checker/src/type_hierarchy.js
moniika: Ok. If it wasn't, then I was going to suggest more specific jsDoc for this function.

plugins/nominal-connection-checker/src/type_hierarchy.js
moniika: Ok. I'm fine with leaving it as an array.

plugins/nominal-connection-checker/src/type_hierarchy.js
moniika: I see. I originally thought that `subs()` was all subtypes including indirect and `descendants` was only direct subtypes. But after your explanation and looking at the jsDoc I see that it is the opposite.

So I don't think this function should be `isDirectRelative` now because that isn't the case (I wouldn't consider a to be a direct relative of c, I would only consider b to be a direct relative).

What's the motivation for calling it "nearest" in the first place?
When I read the description, I assumed that nearest == direct supertype/subtype (depending on which case).

I'm still thinking if I have a specific suggestion for a name.

plugins/nominal-connection-checker/src/type_hierarchy.js
moniika: I had to look up `flat()` and `flatMap()` since it's not a common function I use. 
I think the comment is helpful, but a comment by the reduce function would make comprehension easier, though not strictly necessary.
Something like  "Flatten the array of arrays" on L474 (or add Arr or Array to the variables of the reduce function as a reminder of the types)

moniika:
> First of all, sorry about leaving a review instead of responding individually. I know it makes the conversations hard to read, but there was just a lot to respond to.
> 
> What I'd like to address here is some thoughts wrt readability and efficiency of removing duplicate elements from an array.
> 
> The pattern I usually use for this is filtering the array like so:
> 
> ```js
> array.filter((elem, i, arr) => arr.indexOf(elem) == i);
> ```
> 
> indexOf always returns the first instance of an item in an array. So if you're looking at the first instance, indexOf() will return the index of the first, and you'll know that item's ok. On the other hand if you're looking at the second instance, indexOf() will return the index of the first, and you'll know you can remove that item.
> 
> Once you're familiar with the pattern I think it's easy enough to read, however it's not very efficient.
> 
> As I address below, there are other ways that you can do this by adding some additional checks based on the indices:
> 
> ```js
> array.filter((elem, i, arr) =>
>     arr.every((elem2, j) => i <= j || elem != elem2));
> ```
> 
> But both of these have the problem of being n^2. Another option is to use a Set() when initialy creating the collection, so that duplicates are impossible. But I don't really like this solution. (1) It requires the set to be converted to an array if we want to do any filtering, which I don't think is very readable. (2) It doesn't actually prevent duplicates in the case that we're creating a set of objects (it only works for primitive types). So I'd rather go with one of the other two solutions.
> 
> Which solution do you prefer?
> 
> And one other thing, I think that for readability it might be a good idea to separate out any filters that remove duplicates from other filters. That way we can keep them consistent and recognizable, and we won't have to use as many comments. Similar to how I try to keep the flattening `reduce()` consistent. Do you think that's a good idea?

If you're only removing duplicates, `indexOf` is a bit easier to read, so I would lean towards that approach (if there's not going to be a comment).
However, if you've already got an `every` (or `some`) inside of the `filter` (eg., [L227](https://github.com/google/blockly-samples/pull/670/files#diff-c0324ad5941634030054630b09ce52ef3bc3c48214517c2cc53caf185f92007eR227)), then I would go with the second approach (rather than nesting an `indexOf`) because it would otherwise be O(n^3).
I see that in `getNearestCommon` there is a filter to remove duplicates. I'd be fine with that one using `indexOf`, but there's already a comment "Get rid of duplicates" so to me it makes no difference which approach because the comment makes it clear what's happening.

--

As for the set, I would evaluate on a case by case basis on whether it adds a performance benefit. If the only processing you're doing on the list is removing duplicates, then it could be worth it, but otherwise it may not make a difference (and thus the array approach is more convenient). I initially suggested a Set in `initNearest_`, but now I think that it wouldn't actually make a difference since you need to process the list later anyways.

--

**For readability**, you could create a helper function that does the filter operation (if you're doing that a lot):
```js
array.filter((elem, i, arr) => arr.indexOf(elem) == i);
```
Naming it something like `filterDuplicates`. This would both make it easier to read and would reuse code (assuming you do this more than once).

And one other thing, I think that for readability it might be a good idea to separate out any filters that remove duplicates from other filters. That way we can keep them consistent and recognizable, and we won't have to use as many comments. Similar to how I try to keep the flattening `reduce()` consistent. Do you think that's a good idea?

I'm not as sure about the idea of "separating" the filter. I think it's technically more efficient to combine it some cases and I don't see any harm with comments. If you're copying the same comments, then there might be an opportunity to create a helper function.

plugins/nominal-connection-checker/src/type_hierarchy.js
BeksOmega: First: I reorganized the params/internal logic isNearest a bit so that it makes more sense in context (hopefully).

I originally called it nearest because this function returns true if the first type is the nearest common ancestor/descendant of itself and the second type. I think this makes more sense if you look at how it is used on L214. If it returns true, we can add the first type to the list of nearest common types between the two types.

Also note that by definition the second type can only be the nearest common ancestor/descendant of itsefl and the first type if the second type is the ancestor/descendant of the first (respectively). Which is why the anonymous functions passed to this function work.

BeksOmega:
> If you're only removing duplicates, indexOf is a bit easier to read, so I would lean towards that approach (if there's not going to be a comment).
> However, if you've already got an every (or some) inside of the filter (eg., L227), then I would go with the second approach (rather than nesting an indexOf) because it would otherwise be O(n^3).
> I see that in getNearestCommon there is a filter to remove duplicates. I'd be fine with that one using indexOf, but there's already a comment "Get rid of duplicates" so to me it makes no difference which approach because the comment makes it clear what's happening.

Makes sense! What I decided to do is change all of them to use the findIndex method for now, and additionally I've added comments to all of the filters so that it's clear what they're doing (even if people don't understand the logic behind using getIndex()).

> As for the set, I would evaluate on a case by case basis on whether it adds a performance benefit. If the only processing you're doing on the list is removing duplicates, then it could be worth it, but otherwise it may not make a difference (and thus the array approach is more convenient). I initially suggested a Set in initNearest_, but now I think that it wouldn't actually make a difference since you need to process the list later anyways.

Makes sense! I'll be sure to watch out for opportunities to use it :D

> For readability, you could create a helper function that does the filter operation (if you're doing that a lot):
>```
>array.filter((elem, i, arr) => arr.indexOf(elem) == i);
>```
> Naming it something like filterDuplicates. This would both make it easier to read and would reuse code (assuming you do this more than once).

I love this idea :D But right now I'm leaning toward leaving it out of this PR. My struggle is that there are really 2 different ways of removing duplicates. (1) Remove shallow duplicates, ie identical primitive data or different pointers to the same object. (2) Remove deep duplicates, eg different objects which look the same. (1) is easy to handle with a helper function like you've shown, but for (2) each you always need a different method for telling if objects are deeply equialent.

I'm thinking that in the future (2) might "condence" in that we're really only filtering for deep duplicates on a couple of different types (eg TypeStructures). And then it would be really easy to add helper functions for all of them at once. But for now I think I'd rather leave things slightly more explicit.

---

Please feel free to tell me if there are any changes in the latest commit you don't like :D Thank you for the work you've done on reviewing this! I think this is really helping clean up the logic and make it (hopefully) maintainable :D

plugins/nominal-connection-checker/src/type_hierarchy.js
moniika: Thanks for the explanation and I think your reworded jsDoc for the `isNearest` param helps a lot.

moniika (commented):
Thanks for all the updates.

(Side note: I just remembered that we can add inline JSDoc to the params of the arrow function, which I think may be sometimes helpful, for types. I'm not suggesting that for this PR though, but it might be worth keeping in mind for future changes.)

My only nit with the changes is that I'd like you to add curly braces to the arrow functions that are now not declared on the same line.

I'd just like you to add the curly braces to the arrow functions before I approve/merge. There's just three places I think (L155-156, L168-169, and L228-230).

PR #676, opened 2021-04-20
Add a new scroll-options plugin

maribethb:
I updated the values in the package-json but not in the readme or anywhere else so I left those todos there.

### DIFF
--- a/plugins/scroll-options/package.json
+++ b/plugins/scroll-options/package.json
@@ -1,0 +1,62 @@
+{
+  "name": "@blockly/plugin-scroll-options",
+  "version": "0.1.0",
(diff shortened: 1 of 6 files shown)


plugins/scroll-options/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@blockly/plugin-scroll-options",
+  "version": "0.1.0",
+  "description": "A Blockly plugin that adds advanced scroll options such as scroll-on-drag.",
```
alschmiedt: ... and scroll while holding a block.

alschmiedt (approved):

PR #695, opened 2021-05-08
Add checking for duplicates in fulfills

BeksOmega:
### Description

Adds a check to type hierarchy validation to ensure that types are not being fulfilled/subtyped multiple times by the same type.

The reason I added this check is because supporting multiple inheritance of the same type adds too much ambiguity (in my opinion).

For example, say we want to have a `Dict[K, V]` that acts as both a `List[K]` and a `List[V]`. If we then pass the dictionary to a `reverse(List[T]) : List[T]` block, which list should be reversed? Do we get a reversed `List[K]` or a reversed `List[V]`? I'm not sure how this case should be handled it, so I'm just banning it haha.

Also note that this can be trivially worked around by adding a block to your blocks language that "converts" one type to one of the other types. For example a `getListOfKeys(Dict[K, V]) : List[K]` block and a `getListOfValues(Dict[K, V]) : List[V]` block would solve the above scenario.

### Tests

Adds unit tests for properly catching multiple fulfills with:
  * Idential type names
  * Differently cased names
  * Different parameters passed to the fulfilled type

### DIFF
--- a/plugins/nominal-connection-checker/src/hierarchy_validation.js
+++ b/plugins/nominal-connection-checker/src/hierarchy_validation.js
@@ -42,6 +42,7 @@ export function validateHierarchy(hierarchyDef) {
   checkFulfillsIsArray(hierarchyDef);
   checkSupersParsing(hierarchyDef);
   checkSupersDefined(hierarchyDef);
+  checkDupeSupers(hierarchyDef);
   checkSuperParamsDefined(hierarchyDef);
   checkSuperNumParamsCorrect(hierarchyDef);
   checkSuperParamVariancesCompatible(hierarchyDef);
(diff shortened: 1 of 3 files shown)


plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -228,6 +229,48 @@ function checkSupersDefined(hierarchyDef) {
   }
 }
 
+/**
+ * Checks the hierarchy def for any types which include the same type multiple
+ * times in their fulfills array.
```
moniika: I think you could add that this logs an error for any types which includes the same type multiple times. 
That way it's clear from the start how this method handles finding the same type.

plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -228,6 +229,48 @@ function checkSupersDefined(hierarchyDef) {
...
+function checkDupeSupers(hierarchyDef) {
+  const errorMsg = 'The type %s fulfills the type %s multiple times.';
+
+  for (const type of Object.keys(hierarchyDef)) {
+    const typeInfo = hierarchyDef[type];
+    if (!typeInfo.fulfills || !Array.isArray(typeInfo.fulfills)) {
+      continue;
+    }
+    const superOccurrences = new Map();
+    const superNames = new Map();
```
moniika: You could combine these maps by having the value for the map be an object like:
```
{
  count: {number}
  caseSensitiveName: {string}
}
```
You would only have to check for existence in the map once inside of the loop.

plugins/nominal-connection-checker/src/hierarchy_validation.js
```diff
@@ -228,6 +229,48 @@ function checkSupersDefined(hierarchyDef) {
...
+        }
+      } catch (e) {
+        if (!(e instanceof TypeParseError)) {
+          throw e;
+        } // Otherwise it will have been handled by our specific check.
+      }
+    }
+    for (const [lowerCaseName, occurences] of superOccurrences) {
+      if (occurences > 1) {
+        console.error(errorMsg, type, superNames.get(lowerCaseName));
```
moniika: Idea (feel free to leave as is):
If you aren't actually using the count of occurrences in the error message, you could change your algorithm from a map to using sets.
Instead of the map `superOccurences`, you use the first set to store the supers that you encounter in fulfills and a second set for any multiples (if you find a super that's already in the first set) you encounter. Then the last loop will just go through that second set.

moniika (commented):
Just a few comments, but looks good.

plugins/nominal-connection-checker/src/hierarchy_validation.js
BeksOmega: This is a great idea :D I think that makes the code a lot simpler. Done!

moniika (approved):

PR #698, opened 2021-05-11
Cleanup TODOs after adding parameterized types

BeksOmega:
### Description

Each commit cleans up a little TODO issue, like reorganizing code or making function calls more consistent.

### Testing
I think there are some tests to make sure errors are thrown properly, but it's mostly just a bit of reorganization.

### DIFF
--- a/plugins/nominal-connection-checker/src/index.js
+++ b/plugins/nominal-connection-checker/src/index.js
@@ -66,9 +66,31 @@ export class NominalConnectionChecker extends Blockly.ConnectionChecker {
    * @override
    */
   doTypeChecks(a, b) {
+    try {
+      return this.doTypeChecksInternal_(a, b);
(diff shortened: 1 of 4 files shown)


moniika (approved):
lgtm
One thing I want to suggest is you could consider making use of template strings (backticks: \`\`) for strings that contain a lot of variables.

BeksOmega:
> lgtm
> One thing I want to suggest is you could consider making use of template strings (backticks: ``) for strings that contain a lot of variables.

Huh for some reason I thought templates still had to follow the indentation rules, which made them kind of useless. But I just checked and [templates are exempt](https://google.github.io/styleguide/jsguide.html#features-strings-template-strings) :D Thank you for pointing this out for me! I'll definiteliy use them in the future.

PR #705, opened 2021-05-13
Fix insertion

alschmiedt:
Stores the drag and scroll delta in the ScrollBlockDragger. We use these values to find the total distance that the block has moved from the original location that it was picked up from. This totalDelta can be used to update the insertion markers and to move the block off of the drag surface.

I am still trying to figure out a better way to add my changes to the `endBlockDrag` method so that we don't have to copy the entire method.

Uses the APIs in google/blockly#4828

### DIFF
--- a/plugins/scroll-options/src/ScrollBlockDragger.js
+++ b/plugins/scroll-options/src/ScrollBlockDragger.js
@@ -16,81 +16,86 @@ import * as Blockly from 'blockly/core';
  * someone is dragging it.
  */
 export class ScrollBlockDragger extends Blockly.BlockDragger {
+  /** @override */
   constructor(block, workspace) {
     super(block, workspace);
 
-    this.scrollDelta = new Blockly.utils.Coordinate(0, 0);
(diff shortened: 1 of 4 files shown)


plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -16,30 +16,80 @@ import * as Blockly from 'blockly/core';
...
 export class ScrollBlockDragger extends Blockly.BlockDragger {
+  /** @override */
+  constructor(block, workspace) {
+    super(block, workspace);
+
+    /**
+     * How much the block has been moved due to scrolling.
+     * @type {!Blockly.utils.Coordinate}
+     * @protected
+     * TODO: Do we still do underscore in Blockly samples?
```
BeksOmega, line 27: I do :D But according to the [styleguide](https://google.github.io/styleguide/jsguide.html#features-classes-fields) it only applies to private fields, not protected ones.

PR #712, opened 2021-05-14
Adding flyout support for backpack.

moniika:
### Description
- Darkens backpack when stack of blocks is dragged over it (previously button did not change if mouse was dragging a stack of blocks)
- Adds flyout to Backpack that opens when the backpack button is clicked
- Adds block stack to backpack flyout if dropped on button 
  - returns position of stack back to the start of the drag
  - ignores duplicate blocks

#### Screenshots
61624897-429c-4b7d-941e-66ce3cf83d47 (2)

### Additional Information

### DIFF
--- a/plugins/workspace-backpack/src/backpack_helpers.js
+++ b/plugins/workspace-backpack/src/backpack_helpers.js
@@ -1,0 +1,105 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Helper and utility methods for the Backpack plugin.
+ * @author kozbial@google.com (Monica Kozbial)
+ */
+
+import * as Blockly from 'blockly/core';
+
+/**
+ * Creates a backpack flyout.
+ * @param {!Blockly.WorkspaceSvg} targetWorkspace The target workspace for the
+ *    backpack.
+ * @return {!Blockly.IFlyout} The backpack flyout.
+ */
+export function createBackpackFlyout(targetWorkspace) {
+  // Create flyout options.
+  const flyoutWorkspaceOptions = new Blockly.Options(
+      /** @type {!Blockly.BlocklyOptions} */
+      ({
+        'scrollbars': true,
+        'parentWorkspace': targetWorkspace,
+        'rtl': targetWorkspace.RTL,
+        'oneBasedIndex': targetWorkspace.options.oneBasedIndex,
+        'renderer': targetWorkspace.options.renderer,
+        'rendererOverrides': targetWorkspace.options.rendererOverrides,
+        'move': {
+          'scrollbars': true,
+        },
+      }));
+  // Create vertical or horizontal flyout.
+  if (targetWorkspace.horizontalLayout) {
+    flyoutWorkspaceOptions.toolboxPosition =
+        (targetWorkspace.toolboxPosition ===
+            Blockly.utils.toolbox.Position.TOP) ?
+            Blockly.utils.toolbox.Position.BOTTOM :
+            Blockly.utils.toolbox.Position.TOP;
+    const HorizontalFlyout = Blockly.registry.getClassFromOptions(
+        Blockly.registry.Type.FLYOUTS_HORIZONTAL_TOOLBOX,
+        targetWorkspace.options, true);
+    return new HorizontalFlyout(flyoutWorkspaceOptions);
+  }
+  flyoutWorkspaceOptions.toolboxPosition =
+      (targetWorkspace.toolboxPosition ===
+          Blockly.utils.toolbox.Position.RIGHT) ?
+          Blockly.utils.toolbox.Position.LEFT :
+          Blockly.utils.toolbox.Position.RIGHT;
+  const VerticalFlyout = Blockly.registry.getClassFromOptions(
+      Blockly.registry.Type.FLYOUTS_VERTICAL_TOOLBOX,
+      targetWorkspace.options, true);
+  return new VerticalFlyout(flyoutWorkspaceOptions);
+}
+
+/**
+ * Converts XML representing a block into text that can be stored in the
+ *    content array.
+ * @param {!Element} xml An XML tree defining the block and any
+ *    connected child blocks.
+ * @return {string} Text representing the XML tree, cleaned of all unnecessary
+ * attributes.
+ * @private
+ */
+export function cleanBlockXML(xml) {
+  const xmlBlock = xml.cloneNode(true);
+  let node = xmlBlock;
+  while (node) {
+    // Things like text inside tags are still treated as nodes, but they
+    // don't have attributes (or the removeAttribute function) so we can
+    // skip removing attributes from them.
+    if (node.removeAttribute) {
+      node.removeAttribute('x');
+      node.removeAttribute('y');
+      node.removeAttribute('id');
+      node.removeAttribute('disabled');
+      if (node.nodeName == 'comment') { // Future proof just in case.
+        node.removeAttribute('h');
+        node.removeAttribute('w');
+        node.removeAttribute('pinned');
+      }
+    }
+
+    // Try to go down the tree
+    let nextNode = node.firstChild || node.nextSibling;
+    // If we can't go down, try to go back up the tree.
+    if (!nextNode) {
+      nextNode = node.parentNode;
+      while (nextNode) {
+        // We are valid again!
+        if (nextNode.nextSibling) {
+          nextNode = nextNode.nextSibling;
+          break;
+        }
+        // Try going up again. If parentNode is null that means we have
+        // reached the top, and we will break out of both loops.
+        nextNode = nextNode.parentNode;
+      }
+    }
+    node = nextNode;
+  }
+  return Blockly.Xml.domToText(xmlBlock);
+}
--- a/plugins/workspace-backpack/src/backpack_helpers.js
+++ b/plugins/workspace-backpack/src/backpack_helpers.js
@@ -31,7 +31,7 @@ export function cleanBlockXML(xml) {
       node.removeAttribute('y');
       node.removeAttribute('id');
       node.removeAttribute('disabled');
-      if (node.nodeName == 'comment') { // Future proof just in case.
+      if (node.nodeName == 'comment') {
         node.removeAttribute('h');
         node.removeAttribute('w');
         node.removeAttribute('pinned');
--- a/plugins/workspace-backpack/src/backpack_helpers.js
+++ b/plugins/workspace-backpack/src/backpack_helpers.js
@@ -13,12 +13,11 @@ import * as Blockly from 'blockly/core';
 
 /**
  * Converts XML representing a block into text that can be stored in the
- *    content array.
+ * content array.
  * @param {!Element} xml An XML tree defining the block and any
  *    connected child blocks.
  * @return {string} Text representing the XML tree, cleaned of all unnecessary
  * attributes.
- * @private
  */
 export function cleanBlockXML(xml) {
   const xmlBlock = xml.cloneNode(true);
--- a/plugins/workspace-backpack/src/backpack_helpers.js
+++ b/plugins/workspace-backpack/src/backpack_helpers.js
@@ -12,50 +12,6 @@
 import * as Blockly from 'blockly/core';
 
 /**
- * Creates a backpack flyout.
- * @param {!Blockly.WorkspaceSvg} targetWorkspace The target workspace for the
- *    backpack.
- * @return {!Blockly.IFlyout} The backpack flyout.
- */
-export function createBackpackFlyout(targetWorkspace) {
-  // Create flyout options.
-  const flyoutWorkspaceOptions = new Blockly.Options(
-      /** @type {!Blockly.BlocklyOptions} */
-      ({
-        'scrollbars': true,
-        'parentWorkspace': targetWorkspace,
-        'rtl': targetWorkspace.RTL,
-        'oneBasedIndex': targetWorkspace.options.oneBasedIndex,
-        'renderer': targetWorkspace.options.renderer,
-        'rendererOverrides': targetWorkspace.options.rendererOverrides,
-        'move': {
-          'scrollbars': true,
-        },
-      }));
-  // Create vertical or horizontal flyout.
-  if (targetWorkspace.horizontalLayout) {
-    flyoutWorkspaceOptions.toolboxPosition =
-        (targetWorkspace.toolboxPosition ===
-            Blockly.utils.toolbox.Position.TOP) ?
-            Blockly.utils.toolbox.Position.BOTTOM :
-            Blockly.utils.toolbox.Position.TOP;
-    const HorizontalFlyout = Blockly.registry.getClassFromOptions(
-        Blockly.registry.Type.FLYOUTS_HORIZONTAL_TOOLBOX,
-        targetWorkspace.options, true);
-    return new HorizontalFlyout(flyoutWorkspaceOptions);
-  }
-  flyoutWorkspaceOptions.toolboxPosition =
-      (targetWorkspace.toolboxPosition ===
-          Blockly.utils.toolbox.Position.RIGHT) ?
-          Blockly.utils.toolbox.Position.LEFT :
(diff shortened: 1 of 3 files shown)


plugins/workspace-backpack/src/backpack_monkey_patch.js
```diff
@@ -0,0 +1,175 @@
...
+      }
+      const toolbox = workspace.getToolbox();
+      if (toolbox &&
+          toolbox.getFlyout() &&
+          toolbox.getFlyout().autoClose) {
+        toolbox.clearSelection();
+      }
+      if (workspace.backpack) {
+        workspace.backpack.close();
+      }
```
alschmiedt, line 41: Since this isn't on a class that we can easily register, it seems like removing this monkey patch might be more involved. Do we have a plan/idea on how we might be able to take this out?

plugins/workspace-backpack/src/backpack_monkey_patch.js
```diff
@@ -0,0 +1,175 @@
...
+      toolbox.removeStyle(style);
+    }
+    this.workspace_.setResizesEnabled(true);
+
+    toolbox = this.workspace_.getToolbox();
+    if (toolbox && typeof toolbox.removeStyle == 'function') {
+      style = this.draggingBlock_.isDeletable() ? 'blocklyToolboxDelete' :
+          'blocklyToolboxGrab';
+      toolbox.removeStyle(style);
+    }
```
alschmiedt: I think this is a copy and paste error? We should only need to call `setResizeEnaled` and change the toolbox style once.

plugins/workspace-backpack/src/backpack_helpers.js
```diff
@@ -0,0 +1,105 @@
...
+          Blockly.utils.toolbox.Position.RIGHT;
+  const VerticalFlyout = Blockly.registry.getClassFromOptions(
+      Blockly.registry.Type.FLYOUTS_VERTICAL_TOOLBOX,
+      targetWorkspace.options, true);
+  return new VerticalFlyout(flyoutWorkspaceOptions);
+}
+
+/**
+ * Converts XML representing a block into text that can be stored in the
+ *    content array.
```
alschmiedt: nit: unnecessary spacing.

plugins/workspace-backpack/src/backpack_helpers.js
```diff
@@ -0,0 +1,105 @@
...
+
+import * as Blockly from 'blockly/core';
+
+/**
+ * Creates a backpack flyout.
+ * @param {!Blockly.WorkspaceSvg} targetWorkspace The target workspace for the
+ *    backpack.
+ * @return {!Blockly.IFlyout} The backpack flyout.
+ */
+export function createBackpackFlyout(targetWorkspace) {
```
alschmiedt: My concern in putting this function in a separate file is that developers wouldn't be able to override it to place the flyout wherever they want. If you don't have any objections, I think putting this back with the plugin might make it easier to extend.

plugins/workspace-backpack/src/backpack_monkey_patch.js
```diff
@@ -0,0 +1,175 @@
...
+      this.draggingBlock_.setDeleteStyle(true);
+      if (this.deleteArea_ == Blockly.DELETE_AREA_TRASH && trashcan) {
+        trashcan.setLidOpen(true);
+      }
+    } else {
+      this.draggingBlock_.setDeleteStyle(false);
+      if (trashcan) {
+        trashcan.setLidOpen(false);
+      }
+    }
```
alschmiedt: Instead of having this code here, could you save the old method above and call it here. It is not ideal, but it mimics calling super and would mean that we wouldn't have to worry about this getting out of sync with Blockly as much. It is also might help us figure out where the block dragger could be split up to allow for calling super.

plugins/workspace-backpack/src/backpack_monkey_patch.js
```diff
@@ -0,0 +1,175 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
alschmiedt: nit: 2021

plugins/workspace-backpack/src/index.js
```diff
@@ -10,6 +10,8 @@
  */
 
 import * as Blockly from 'blockly/core';
+import {createBackpackFlyout, cleanBlockXML} from './backpack_helpers';
+import './backpack_monkey_patch';
 
 /**
  * Class for backpack.
```
alschmiedt: I think adding more information on what the backpack actually does would be helpful. Saves blocks, opens a flyout, etc.

plugins/workspace-backpack/src/index.js
```diff
@@ -19,16 +21,20 @@ import * as Blockly from 'blockly/core';
...
      * The workspace.
      * @type {!Blockly.WorkspaceSvg}
      * @protected
      */
-    this.workspace_ = workspace;
+    this.workspace_ = targetWorkspace;
+
+    // This set is part of the monkeypatch, so that a reference to the backpack
+    // can be accessed on a given workspace.
+    this.workspace_.backpack = this;
```
alschmiedt, line 38: I think similar question as above, is there a plan/idea for removing this monkey patch?

plugins/workspace-backpack/src/index.js
```diff
@@ -119,6 +131,25 @@ export class Backpack {
...
+     * A list of XML (stored as strings) representing blocks in the backpack.
+     * @type {!Array<string>}
+     * @private
+     */
+    this.contents_ = [];
+
+    /**
+     * The maximum items that can be stored on the backpack.
+     */
+    this.maxItems_ = 32;
```
alschmiedt: Add access modifier. 

Is there a way for someone to easily change this?

plugins/workspace-backpack/src/index.js
```diff
@@ -289,19 +353,123 @@ export class Backpack {
...
+   * @param {!Array<string>} contents The backpack contents to merge.
+   */
+  mergeContents(contents) {
+    contents.forEach((item) => {
+      this.addItem(item);
+    });
+  }
+
+  /**
+   * Whether the backpack is open-able.
```
alschmiedt, line 475: This confuses me a bit, why is it only openable if it is already open? Openable to me sounds like it is able to be opened? So it should return true even if the flyout is closed.

Also, if this is private it should have the "_".

plugins/workspace-backpack/src/index.js
```diff
@@ -289,19 +353,123 @@ export class Backpack {
         'translate(' + this.left_ + ',' + this.top_ + ')');
   }
 
+  /**
+   * Returns the count of items in the backpack.
+   * @return {number} The count.
```
alschmiedt: nit: Elaborate on this a bit.

plugins/workspace-backpack/src/index.js
```diff
@@ -289,19 +353,123 @@ export class Backpack {
...
+   */
+  handleBlockDrop(block) {
+    const blockXml = Blockly.Xml.blockToDomWithXY(block);
+    this.addItem(cleanBlockXML(blockXml));
+  }
+
+  /**
+   * Adds item to backpack.
+   * @param {string} item Text representing the XML tree of a block to add,
+   * cleaned of all unnecessary attributes.
```
alschmiedt: nit: second line should be indented.

plugins/workspace-backpack/src/index.js
```diff
@@ -289,19 +353,123 @@ export class Backpack {
...
+    }
+    // TODO: Fire UI event for Backpack content change.
+  }
+
+  /**
+   * Sets backpack contents XML.
+   * @param {!Array<string>} contents The new backpack contents.
+   */
+  setContents(contents) {
+    this.contents_.length = 0;
```
alschmiedt: Just curious, why do this over `this.contents_ = []`?

plugins/workspace-backpack/src/index.js
```diff
@@ -289,19 +353,123 @@ export class Backpack {
...
+    // TODO: Fire UI event for Backpack content change.
+    this.close();
+  }
+
+  /**
+   * Handles a block drop on this backpack.
+   * @param {!Blockly.BlockSvg} block The block being dropped on the backpack.
+   */
+  handleBlockDrop(block) {
+    const blockXml = Blockly.Xml.blockToDomWithXY(block);
```
alschmiedt: It looks like you are stripping the x and y attribute in the `cleanBlockXML` method, could you use `blockToDom` instead?

plugins/workspace-backpack/src/backpack_helpers.js
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/**
+ * Converts XML representing a block into text that can be stored in the
+ *    content array.
+ * @param {!Element} xml An XML tree defining the block and any
+ *    connected child blocks.
+ * @return {string} Text representing the XML tree, cleaned of all unnecessary
+ * attributes.
+ * @private
```
alschmiedt: Remove private.

plugins/workspace-backpack/src/backpack_helpers.js
```diff
@@ -0,0 +1,105 @@
...
+  while (node) {
+    // Things like text inside tags are still treated as nodes, but they
+    // don't have attributes (or the removeAttribute function) so we can
+    // skip removing attributes from them.
+    if (node.removeAttribute) {
+      node.removeAttribute('x');
+      node.removeAttribute('y');
+      node.removeAttribute('id');
+      node.removeAttribute('disabled');
+      if (node.nodeName == 'comment') { // Future proof just in case.
```
alschmiedt: This seems a bit suspicious to me. Is this future proofing for workspace comments?

plugins/workspace-backpack/src/backpack_monkey_patch.js
moniika: I see this as being a change that should happen in the Base `BlockDragger`. If it was extended as a custom Dragger, then multiple plugins that use this feature would not be compatible and I think that it would be common for plugins to want to know when a cursor enters/exits over it (when holding block).

The plan is to implement an API in core to support this. This may looks something like adding an additional registratable type for plugins for `IDragTarget` and making changes in the `BlockDragger` to support it.

plugins/workspace-backpack/src/backpack_monkey_patch.js
moniika: good catch, yes it looks like it is a copy/paste error.

plugins/workspace-backpack/src/index.js
moniika: Added some more info

plugins/workspace-backpack/src/index.js
moniika: The plan is to design and implement a solution in core in the next couple weeks.

plugins/workspace-backpack/src/index.js
moniika: Not yet, but I plan to add a way in future PR.

plugins/workspace-backpack/src/index.js
moniika: I've updated the return description, but I'm not sure if was exactly what you were asking.

plugins/workspace-backpack/src/index.js
moniika: done

plugins/workspace-backpack/src/index.js
moniika: done

plugins/workspace-backpack/src/index.js
moniika: Both approaches work. The benefit to setting the `length` to 0 is that if there was any references to `this.contents_`, this set would affect all references to the array, while setting this variable to an empty array would be reassigning the reference to a new array.

Not related to that question, I this this clear of the array is actually unnecessary since I'm setting `this.contents_` to a new array right afterwards.

plugins/workspace-backpack/src/index.js
moniika: That's a good question. I am also confused why this works now.

moniika: I was missing some `!` in a few places. I've updated and also removed the check for the count because, for now I want to support opening an empty backpack (and the count check prevents an empty backpack from opening).

plugins/workspace-backpack/src/backpack_helpers.js
moniika: I've refactored this logic into `initFlyout_`

plugins/workspace-backpack/src/backpack_helpers.js
moniika: done

plugins/workspace-backpack/src/backpack_helpers.js
moniika: done

plugins/workspace-backpack/src/backpack_helpers.js
moniika: I'm not sure. It's copied from the trashcan implementation.

plugins/workspace-backpack/src/backpack_monkey_patch.js
moniika: done

plugins/workspace-backpack/src/backpack_monkey_patch.js
moniika: I don't have a plan yet, but I was going to look into it next week. Possibly there could be another type for plugins that we could register for those that have flyouts that need closed here.

plugins/workspace-backpack/src/index.js
alschmiedt: Sounds good, it still needs @type and @private

plugins/workspace-backpack/src/index.js
```diff
@@ -289,19 +391,122 @@ export class Backpack {
...
+    this.flyout.show(xml);
     // TODO: Fire UI event for Backpack open.
   }
 
   /**
    * Closes the backpack flyout.
    */
   close() {
-    // TODO: Implement flyout close.
+    if (!this.isOpen()) {
```
alschmiedt, line 507: Should this also use isOpenable_ for consistency?

plugins/workspace-backpack/src/index.js
```diff
@@ -119,6 +132,25 @@ export class Backpack {
...
+     * The maximum items that can be stored on the backpack.
+     */
+    this.maxItems_ = 32;
+
+    /**
+     * The backpack flyout. Initialized during init.
+     * @type {Blockly.IFlyout|undefined}
+     * @package
+     */
+    this.flyout = undefined;
```
alschmiedt, line 155: Why not null here?

alschmiedt (approved):
Some small nits, but otherwise lgtm

plugins/workspace-backpack/src/backpack_helpers.js
BeksOmega: The original intention was to fix this: https://github.com/google/blockly/issues/2574. I think the reason it says "future proof" is incase there are ever other nodes that have 'h', 'w', and 'pinned' attributes? But honestly I'm not sure haha

plugins/workspace-backpack/src/index.js
moniika: I think `isOpen` is the better check here. When closing the flyout, it can be closed if it was open and whether it is currently openable doesn't matter. 
Changing this to `isOpenable_` could cause problem with edge cases (depending on how `isOpenable_` is overridden) if the state changes such that is isn't openable while the flyout is open and then it can't be closed.

plugins/workspace-backpack/src/index.js
moniika: I set it to `undefined` because I don't want to assign a value. Setting it to undefined communicates that the variable has been declared but not defined.
That way I'm saying that this object is never null after it is initialized.

plugins/workspace-backpack/src/index.js
moniika: done

plugins/workspace-backpack/src/backpack_helpers.js
moniika: Thanks Beka. I'm going to remove the comment from here since there's confusion.

PR #714, opened 2021-05-15
Solve Custom fields in React demo are marked as EDITABLE but not also SERIALIZABLE warning

### LINKED ISSUE #607: Custom fields in React demo are marked as EDITABLE but not also SERIALIZABLE
**Category**

Examples

**Component**

blockly-react

**Describe the bug**

The custom fields in the react demo: `BlocklyReactField` and `ReactDateField` are editable but not serializable, resulting in the warning:
```
blockly_compressed.js:464 Detected an editable field that was not serializable. Please define SERIALIZABLE property as true on all editable custom fields. Proceeding with serialization.
```

**To Reproduce**

Run the react demo and look at console warnings.

**Expected behavior**

These fields should probably be marked as `SERIALIZABLE = true` on the prototype.

Harshita-Kanal:
Added `SERIALIZABLE = true` to the class definition

Fixes #607

### DIFF
--- a/examples/blockly-react/src/fields/BlocklyReactField.jsx
+++ b/examples/blockly-react/src/fields/BlocklyReactField.jsx
@@ -31,6 +31,8 @@ import * as Blockly from 'blockly/core';
 
 class BlocklyReactField extends Blockly.Field {
 
+  SERIALIZABLE = true
+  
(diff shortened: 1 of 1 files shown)


moniika (approved):
Thanks for the PR.

PR #716, opened 2021-05-17
Add ScrollMetricsManager

maribethb:
Add ScrollMetricsManager for scroll-options plugin that can use cached content metrics.

### DIFF
--- a/plugins/scroll-options/src/ScrollMetricsManager.js
+++ b/plugins/scroll-options/src/ScrollMetricsManager.js
@@ -14,8 +14,8 @@ import * as Blockly from 'blockly/core';
  * if you hadn't yet picked up the block.
  *
  * To use the cached value of the content metrics instead of calculating new
- * ones, set `stopCalculating` to true before calling `getMetrics` and false
- * when the metrics can be recalculated again.
+ * ones, set `useCachedContentMetrics` to true before calling `getMetrics`
+ * false when the metrics can be recalculated again.
  */
 export class ScrollMetricsManager extends Blockly.MetricsManager {
   /** @override */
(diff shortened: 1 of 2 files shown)


plugins/scroll-options/src/ScrollMetricsManager.js
```diff
@@ -0,0 +1,77 @@
...
+
+      toolboxWidth: toolboxMetrics.width,
+      toolboxHeight: toolboxMetrics.height,
+      toolboxPosition: toolboxMetrics.position,
+
+      flyoutWidth: flyoutMetrics.width,
+      flyoutHeight: flyoutMetrics.height
+    };
+  }
+};
```
alschmiedt: nit: add newline.

plugins/scroll-options/src/ScrollMetricsManager.js
```diff
@@ -0,0 +1,77 @@
...
+     * Whether to stop recalculating content metrics and used the cached value
+     * instead. Note that if there are no cached metrics, they will be
+     * recalculated even if `stopCalculating` is true.
+     * @type {boolean}
+     */
+    this.stopCalculating = false;
+  }
+
+  /** @override */
+  getMetrics() {
```
alschmiedt: Unfortunately jus thought of this :/ but what if instead of only stopping updating for the content metrics we stopped updating all the metrics? It would be something like: 

```
if we are calculating
    this.metrics = super.getMetrics();
return this.metrics
```
My motivation behind this is that anytime we call super we decrease the amount of maintenance the function requires.

plugins/scroll-options/src/ScrollMetricsManager.js
maribethb: Done.

plugins/scroll-options/src/ScrollMetricsManager.js
maribethb: As discussed offline, we can't cache all the metrics as some of them still need to update (such as view and scroll). However, changed to only override getContentMetrics method, which is much simpler.

Also, changed the name of the property that controls this behavior to be more clear, ptal.

alschmiedt (approved):

PR #717, opened 2021-05-17
Backpack events

moniika:
### Description
- Adds and emits UI events: `BackpackChange` and `BackpackOpen`
- Refactors `setContents` and `mergeContents` (changing `mergeContents to `addItems`)
- Changes access modifier from `private` to `protected`

### Additional Information

### DIFF
--- a/plugins/workspace-backpack/src/index.js
+++ b/plugins/workspace-backpack/src/index.js
@@ -11,6 +11,7 @@
 
 import * as Blockly from 'blockly/core';
 import {cleanBlockXML} from './backpack_helpers';
+import {BackpackChange, BackpackOpen} from './ui_events';
 import './backpack_monkey_patch';
 
 /**
@@ -40,7 +41,7 @@ export class Backpack {
     /**
      * The SVG group containing the backpack.
      * @type {SVGElement}
-     * @private
+     * @protected
      */
     this.svgGroup_ = null;
 
@@ -122,28 +123,28 @@ export class Backpack {
      * Array holding info needed to unbind events.
      * Used for disposing.
      * @type {!Array<!Blockly.browserEvents.Data>}
-     * @private
+     * @protected
      */
     this.boundEvents_ = [];
 
     /**
      * Whether this has been initialized.
      * @type {boolean}
-     * @private
+     * @protected
      */
     this.initialized_ = false;
 
     /**
      * A list of XML (stored as strings) representing blocks in the backpack.
      * @type {!Array<string>}
-     * @private
+     * @protected
      */
     this.contents_ = [];
 
     /**
      * The maximum items that can be stored on the backpack.
      * @type {number}
-     * @private
+     * @protected
      */
     this.maxItems_ = 32;
 
@@ -419,7 +420,7 @@ export class Backpack {
       return;
     }
     this.contents_.length = 0;
-    // TODO: Fire UI event for Backpack content change.
+    Blockly.Events.fire(new BackpackChange(this.workspace_.id));
     this.close();
   }
 
@@ -438,15 +439,20 @@ export class Backpack {
    *     cleaned of all unnecessary attributes.
    */
   addItem(item) {
-    if (this.contents_.indexOf(item) !== -1) {
-      return;
-    }
+    this.addItems([item]);
+  }
 
-    this.contents_.unshift(item);
-    while (this.contents_.length > this.maxItems_) {
-      this.contents_.pop();
-    }
-    // TODO: Fire UI event for Backpack content change.
+  /**
+   * Adds multiple items to the backpack.
+   * @param {!Array<string>} items The backpack contents to add.
+   */
+  addItems(items) {
+    const filteredItems = items.filter((item) => {
+      return this.contents_.indexOf(item) === -1;
+    });
+    this.contents_.unshift(...filteredItems);
+    this.trimContents_();
+    Blockly.Events.fire(new BackpackChange(this.workspace_.id));
   }
 
   /**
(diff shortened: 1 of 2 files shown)


plugins/workspace-backpack/src/index.js
```diff
@@ -438,43 +439,48 @@ export class Backpack {
...
    * Sets backpack contents XML.
    * @param {!Array<string>} contents The new backpack contents.
    */
   setContents(contents) {
-    this.contents_ = [...contents];
-    while (this.contents_.length > this.maxItems_) {
-      this.contents_.pop();
-    }
-    // TODO: Fire UI event for Backpack content change.
+    this.contents_ = contents.filter((item) => {
```
maribethb: This is confusing to me. It looks like you're overwriting the contents, not adding to them, so why do you need to check if it was in the old contents?

plugins/workspace-backpack/src/index.js
```diff
@@ -438,43 +439,48 @@ export class Backpack {
...
-      this.contents_.pop();
-    }
-    // TODO: Fire UI event for Backpack content change.
+  /**
+   * Adds multiple items to the backpack.
+   * @param {!Array<string>} items The backpack contents to add.
+   */
+  addItems(items) {
+    const filteredItems = items.filter((item) => {
+      return this.contents_.indexOf(item) === -1;
```
maribethb: You do this in a couple places (though I have a comment about the one below) so it might be useful to have this as a separate `contains` function, or similar? I could see it being useful as a public method, e.g. a context menu item to add to backpack could be disabled if the backpack already contains the item.

plugins/workspace-backpack/src/index.js
```diff
@@ -40,7 +41,7 @@ export class Backpack {
     /**
      * The SVG group containing the backpack.
      * @type {SVGElement}
```
maribethb: This type needs a `?`

plugins/workspace-backpack/src/ui_events.js
```diff
@@ -0,0 +1,101 @@
...
+export class BackpackOpen extends Blockly.Events.UiBase {
+  /**
+   * Type of this event.
+   * @type {string}
+   */
+  type = BACKPACK_OPEN;
+
+  /**
+   * Class for a backpack open event.
+   * @param {boolean=} isOpen Whether the backpack flyout is opening (false
```
maribethb, line 32: Here it says false if opening, but in `this.isOpen` it says false if closing. 

Maybe `isOpening` would be more clear if the backpack is opening during this event vs was already open?

plugins/workspace-backpack/src/ui_events.js
```diff
@@ -0,0 +1,101 @@
...
+ * @author kozbial@google.com (Monica Kozbial)
+ */
+
+import * as Blockly from 'blockly/core';
+
+/**
+ * Name of event that records a backpack open.
+ * @const
+ */
+export const BACKPACK_OPEN = 'backpack_open';
```
maribethb, line 18: Do these have to be exported separately? I'm curious why you made them separate instead of a property on the event class or something. It doesn't look like you use this outside of this file so I'm not sure why you need to export it like this, but I'm not 100% sure on the requirements of registering the event

plugins/workspace-backpack/src/index.js
```diff
@@ -438,43 +439,48 @@ export class Backpack {
...
-   * Merges backpack contents XML.
-   * @param {!Array<string>} contents The backpack contents to merge.
+   * Trims the internal contents array so that it does not exceed max size.
+   * @private
    */
-  mergeContents(contents) {
-    contents.forEach((item) => {
-      this.addItem(item);
-    });
+  trimContents_() {
```
maribethb: This feels like something that should be an error rather than silently dropping items, but I'm not sure what UX you have in mind for adding and managing items so I could be convinced otherwise

plugins/workspace-backpack/src/index.js
moniika: done

plugins/workspace-backpack/src/index.js
moniika: Thanks for pointing it out. This is a bug. My intention was to filter the contents array passed in so that I could guarantee that all the contents in `this.contents_` were always unique (this check doesn't do that right).

plugins/workspace-backpack/src/ui_events.js
moniika: It's a typo that came from copying it from `Blockly Events.TrashcanOpen`. It should be "false if closing".
I'd refer to leave it as `isOpen` for consistency with the `TrashcanOpen` event.

plugins/workspace-backpack/src/index.js
moniika: Also, looking further. I seem to have made a similar mistake in how I implemented `addItems`.

plugins/workspace-backpack/src/index.js
moniika: I've refactored, but I think the `contains` method is an interesting idea. 
Since this PR is supposed to be about Events, I don't want to update it here, but I can address in the context menu PR.
https://github.com/google/blockly-samples/pull/728

plugins/workspace-backpack/src/index.js
moniika: This is how we handle items in the trashcan (silently dropping them), so I followed the same pattern here.
There are other ways we could handle this, we can discuss offline and possibly just remove the maxItem feature for now (and add back later).

plugins/workspace-backpack/src/ui_events.js
moniika: For other events in Blockly we also export the type constant separately 
It's needed to make checks like:
```
if (event.type === BACKPACK_OPEN) {
}
```
This property (`type`) is defined on the prototype, so it wouldn't be "right" to do something like:
```
if (event.type === BackpackOpen.protoype.type) {
}
```

plugins/workspace-backpack/src/index.js
moniika: As discussed offline. I've removed maxItems in this PR and we can address how handle max items in future.

plugins/workspace-backpack/src/index.js
moniika: I've refactored and added a private helper `filterDuplicates_` that handles the duplicate items correctly.

plugins/workspace-backpack/src/ui_events.js
maribethb: I see, thanks for the explanation!

plugins/workspace-backpack/src/index.js
```diff
@@ -438,43 +432,44 @@ export class Backpack {
...
    * Sets backpack contents XML.
    * @param {!Array<string>} contents The new backpack contents.
    */
   setContents(contents) {
-    this.contents_ = [...contents];
-    while (this.contents_.length > this.maxItems_) {
-      this.contents_.pop();
-    }
-    // TODO: Fire UI event for Backpack content change.
+    this.contents_ = this.filterDuplicates_(contents);
```
maribethb, line 453: I think this one might still have a problem.
Block A is already in the backpack.
Call setContents(A, B, C)
results in the backpack containing B, C

(These would be good unit test candidates if you're planning to add them)

moniika: That is a good one. I have a branch for tests but I was having trouble getting them to run because I need to mock more things to make it work.

I've added a setting this.contents_ to empty array before this line, which I think fixes the issue.

maribethb (approved):

PR #720, opened 2021-05-18
replaced var with const

### LINKED ISSUE #719: Use `const` instead of `var` in workspace search
**Category**
 - Plugins

**Component**

workspace-search

**Describe the bug**

One of the source files uses `var`. In blockly-samples, this is a lint error as we follow the [Google style guide](https://google.github.io/styleguide/jsguide.html#features-use-const-and-let), which requires that we use `let` or `const` instead of `var`.

**To Reproduce**

Run `npm run lint` in blockly-samples. Two of the lint errors say:  
```
plugins/workspace-search/src/WorkspaceSearch.js#L325
Unexpected var, use let or const instead
```

You can also see this in the lint report of any PR opened in blockly-samples.

**Expected behavior**

This lint error does not exist.

**Additional context**

This would be a good first issue for new contributors. To help you get started, here are the steps you should follow to resolve this issue.

1. In your fork of blockly-samples, open the affected file: https://github.com/google/blockly-samples/blob/master/plugins/workspace-search/src/WorkspaceSearch.js#L325
2. These variables are never reassigned, so they are `const`s. Replace `var` with `const`.
3. Run `npm run lint` to ensure this lint error has been resolved and you have not introduced any others. (There are other lint errors and warnings in blockly-samples, but there shouldn't be any new ones nor any in this particular file).
4. Open a PR and address any comments.

If your PR is accepted and merged, we'll be able to see results right away as there will be one fewer lint error on every PR opened in blockly-samples!

himanshu007-creator:
resolves #719

### DIFF
--- a/plugins/workspace-search/src/WorkspaceSearch.js
+++ b/plugins/workspace-search/src/WorkspaceSearch.js
@@ -322,8 +322,8 @@ export class WorkspaceSearch {
    * @return {!Blockly.utils.Rect} The plugin’s bounding box.
    */
   getBoundingRectangle() {
-    var top = this.htmlDiv_.style.top;
-    var left = this.htmlDiv_.style.left;
+    const top = this.htmlDiv_.style.top;
+    const left = this.htmlDiv_.style.left;
     return new Blockly.utils.Rect(
         top, top + this.htmlDiv_.style.height,
         left, left + this.htmlDiv_.style.width);


alschmiedt (approved):

alschmiedt:
It looks like GitHub Actions is having some issues, so I will merge this when that is fixed and the build finishes running. 

Thanks for your contribution!

PR #721, opened 2021-05-18
Backpack example

moniika:
### Description
- Adds example of Backpack that demonstrates sharing blocks between two workspaces.
- Adds example and plugin to index on gh-pages
- Adds bugfix to check when registering context menu whether option is already registered.

#### Screenshots
54337695-52f9-4791-b110-3fc2dc301009

### Additional Information

### DIFF
--- a/examples/backpack-demo/index.html
+++ b/examples/backpack-demo/index.html
@@ -122,10 +122,10 @@
       });
 
   // Add backpacks
-  const backpack1 = new Backpack(primaryWorkspace);
-  backpack1.init();
-  const backpack2 = new Backpack(secondaryWorkspace);
-  backpack2.init();
+  const primaryBackpack = new Backpack(primaryWorkspace);
+  primaryBackpack.init();
+  const secondaryBackpack = new Backpack(secondaryWorkspace);
+  secondaryBackpack.init();
 
   // Listen to events on both workspace.
   primaryWorkspace.addChangeListener(updateBackpack);
@@ -139,12 +139,12 @@
     let contents;
     let targetBackpack;
     if (primaryWorkspace.id === event.workspaceId) {
-      targetBackpack = secondaryWorkspace.backpack;
-      contents = primaryWorkspace.backpack.getContents();
+      targetBackpack = secondaryBackpack;
+      contents = primaryBackpack.getContents();
       console.log('second workspace backpack updated');
     } else { // secondaryWorkspace.id === event.workspaceId
-      targetBackpack = primaryWorkspace.backpack;
-      contents = secondaryWorkspace.backpack.getContents();
+      targetBackpack = primaryBackpack;
+      contents = secondaryBackpack.getContents();
       console.log('first workspace backpack updated');
     }
     targetBackpack.setContents(contents);
--- a/examples/backpack-demo/index.html
+++ b/examples/backpack-demo/index.html
@@ -1,0 +1,98 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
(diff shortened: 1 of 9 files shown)


examples/backpack-demo/index.html
```diff
@@ -0,0 +1,156 @@
...
+
+  function updateBackpack(event) {
+    if (event.type !== 'backpack_change') {
+      return;
+    }
+    Blockly.Events.disable();
+    let contents;
+    let targetBackpack;
+    if (primaryWorkspace.id === event.workspaceId) {
+      targetBackpack = secondaryWorkspace.backpack;
```
alschmiedt: If this is a monkey patch that we expect to remove at some point, could we instead use `backpack1` and `backpack2` here instead of getting it from the workspace?

examples/backpack-demo/index.html
```diff
@@ -0,0 +1,156 @@
...
+        media: 'https://unpkg.com/blockly/media/',
+        toolbox: document.getElementById('toolbox'),
+        trashcan: true
+      });
+
+  // Add backpacks
+  const backpack1 = new Backpack(primaryWorkspace);
+  backpack1.init();
+  const backpack2 = new Backpack(secondaryWorkspace);
+  backpack2.init();
```
alschmiedt: Super nit: `primaryBackpack` and `secondaryBackpack` to match the workspace names?

examples/backpack-demo/index.html
```diff
@@ -0,0 +1,156 @@
...
+    if (primaryWorkspace.id === event.workspaceId) {
+      targetBackpack = secondaryWorkspace.backpack;
+      contents = primaryWorkspace.backpack.getContents();
+      console.log('second workspace backpack updated');
+    } else { // secondaryWorkspace.id === event.workspaceId
+      targetBackpack = primaryWorkspace.backpack;
+      contents = secondaryWorkspace.backpack.getContents();
+      console.log('first workspace backpack updated');
+    }
+    targetBackpack.setContents(contents);
```
alschmiedt, line 150: I'm thinking about if someone wanted to create a backpack that only lets the owner of the workspace delete items in the backpack, but lets the other users add items. I think this is possible now, but would it be easier if we added information on the event as to what changed? I'm not sure it is worth it, just curious : )

alschmiedt (approved):

examples/backpack-demo/index.html
moniika: I think adding more information to the change event could be useful.
I'll add it to the design doc and if it doesn't get sorted out before it is published, then I'll file a bug to keep track.
Since I'm still playing around with it, I'm not sure what kind of information is useful. So, I want to wait to add properties so there's minimal changing of properties on the event (only adding).

alschmiedt: Yup, sounds great : )

plugins/workspace-backpack/src/backpack_helpers.js
```diff
@@ -172,10 +191,10 @@ export function registerAllContextMenus(workspace) {
   registerEmptyBackpack(workspace);
```
alschmiedt, line 191: Out of scope for this PR, but we might want to think about how someone would want to only add a subset of these context menu options.

alschmiedt (approved):

plugins/workspace-backpack/src/backpack_helpers.js
moniika: Agreed. I have a TODO comment referencing this. I think it is something that someone would want to be able to customize.

PR #728, opened 2021-05-21
Add context menus for backpack.

moniika:
### Description
Updates the README documentation and
adds the following context menu options for the backpack:
- "Copy all Blocks" to Backpack in main Workspace
- "Paste all Blocks" from Backpack in main Workspace
- "Remove from Backpack" on Block stack in Backpack Flyout
- "Add to Backpack" on Block stack in main Workspace
- "Empty" option when right clicking the Backpack that is disabled if the Backpack is empty

#### Screenshots
c28008d3-d4b1-4d48-832a-8b42e549d5a2

### Additional Information

### DIFF
--- a/plugins/workspace-backpack/README.md
+++ b/plugins/workspace-backpack/README.md
@@ -37,8 +37,8 @@ can be translated by assigning the following properties of Blockly.Msg
 - `EMPTY_BACKPACK` (Default: "Empty") context menu - Empty the backpack.
 - `REMOVE_FROM_BACKPACK` (Default: "Remove from Backpack") context menu - Remove
 the selected Block from the backpack.
-- `COPY_TO_BACKPACK` (Default: "Add to Backpack") context menu - Add the
-selected block to the backpack.
+- `COPY_TO_BACKPACK` (Default: "Copy to Backpack") context menu - Copy the
+selected Block to the backpack.
 - `COPY_ALL_TO_BACKPACK` (Default: "Copy All Blocks to Backpack") Context menu -
 copy all Blocks on the workspace to the backpack.
 - `PASTE_ALL_FROM_BACKPACK` (Default: "Paste All Blocks from Backpack") context
--- a/plugins/workspace-backpack/README.md
+++ b/plugins/workspace-backpack/README.md
@@ -30,15 +30,49 @@ const backpack = new Backpack(workspace);
 backpack.init();
 ```
 
+### Blockly Languages
+We do not currently support translating the text in this plugin to different
+languages. However, if you would like to support multiple languages the messages
+can be translated by assigning the following properties of Blockly.Msg
+- `EMPTY_BACKPACK` (Default: "Empty") context menu - Empty the backpack.
+- `REMOVE_FROM_BACKPACK` (Default: "Remove from Backpack") context menu - Remove
+the selected Block from the backpack.
+- `COPY_TO_BACKPACK` (Default: "Add to Backpack") context menu - Add the
+selected block to the backpack.
+- `COPY_ALL_TO_BACKPACK` (Default: "Copy All Blocks to Backpack") Context menu -
+copy all Blocks on the workspace to the backpack.
+- `PASTE_ALL_FROM_BACKPACK` (Default: "Paste All Blocks from Backpack") context
+menu - Paste all Blocks from the backpack to the workspace.
+
+```javascript
+Blockly.Msg['EMPTY_BACKPACK'] = 'Opróżnij plecak';  // Polish 
+// Inject workspace, etc...
+```
+
 ## API
 
 - `init`: Initializes the backpack.
 - `dispose`: Disposes of backpack.
+
+- `isOpen`: Returns whether the backpack is open.
+- `open`: Opens the backpack flyout.
+- `close`: Closes the backpack flyout.
+
+- `getCount`: Returns the count of items in the backpack.
+- `getContents` Returns backpack contents.
+- `empty`: Empties the backpack's contents. If the contents-flyout is currently
+open it will be closed.
+- `addItem`: Adds item to backpack.
+- `deleteItem`: Deletes item from the backpack.
+- `setContents`: Sets backpack contents.
+
+- `handleBlockDrop`: Handles a block drop on this backpack.
+- `onDragEnter`: Handle mouse over.
+- `onDragExit`: Handle mouse exit.
+
 - `getBoundingRectangle`: Returns the bounding rectangle of the UI element in
 pixel units relative to the Blockly injection div.
 - `position`: Positions the backpack UI element.
-- `open`: Opens the backpack flyout.
-- `close`: Closes the backpack flyout.
 
 ## License
 Apache 2.0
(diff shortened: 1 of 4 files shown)


plugins/workspace-backpack/src/backpack_helpers.js
```diff
@@ -10,6 +10,171 @@
...
 
 import * as Blockly from 'blockly/core';
+import './msg';
+
+/**
+ * Registers a context menu option to empty the backpack when right-clicked.
+ * @param {!Blockly.WorkspaceSvg} workspace The workspace to register the
+ *   context menu option on.
+ */
+function registerEmptyBackpack(workspace) {
```
maribethb, line 21: This doesn't really play well with customizing the context menu options since it removes everything else indiscriminately. But I haven't thought of a way around this either - maybe we could pass the event in with the scope, but then to remove the other options you'd have to have all of them check if you're in the right area. It would be cool if we could register additional `scopeType`s so you could register a function on the BackpackIcon scope, but that would require some changes to how we decide which context menu to show, which this shouldn't block on obviously.

I am somewhat concerned because the workspace can only have one `configureContextMenu` function defined, and it seems like a fairly common thing people have used before we added the context menu registry. So this plugin wouldn't be compatible with anyone who has already used the `configureContextMenu` API, if I'm understanding correctly. Or you'd get different behavior depending on who implements this method last. Sam [gets around this in dev-tools](https://github.com/google/blockly-samples/blob/master/plugins/dev-tools/src/playground/index.js#L317) by calling the previously set `configureContextMenu`. In your case, you'd probably only need to do that in the section where this click was not on the backpack. That could still be broken if the developer adds their own implementation after this code is run.

Perhaps you could define and export the function itself separately? That way you can register it by default, but the developer could call the function from their own configureContextMenu if they've created their own.

plugins/workspace-backpack/src/backpack_helpers.js
```diff
@@ -10,6 +10,171 @@
...
+/**
+ * Registers a context menu option to remove a block from a backpack flyout.
+ */
+function registerRemoveFromBackpack() {
+  const removeFromBackpack = {
+    displayText: Blockly.Msg['REMOVE_FROM_BACKPACK'],
+    preconditionFn: function(
+        /** @type {!Blockly.ContextMenuRegistry.Scope} */ scope) {
+      const ws = scope.block.workspace;
+      if (ws.isFlyout && ws.targetWorkspace && !!ws.targetWorkspace.backpack) {
```
maribethb: Is this enabled in the regular (toolbox) flyout of a workspace that has a backpack?

plugins/workspace-backpack/src/msg.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * @fileoverview Translatable messages used in backpack.
+ * @author kozbial@google.com (Monica Kozbial)
+ */
+
+import * as Blockly from 'blockly/core';
+
+/** @type {string} */
+// context menu - Copy all Blocks on the workspace to the backpack.
+Blockly.Msg['COPY_ALL_TO_BACKPACK'] = 'Copy All Blocks to Backpack';
```
maribethb, line 16: This is a minor note, but to be more like core blockly it would say "Copy 42 blocks to backpack" (e.g. the delete all blocks option). That would also give more flexibility to language translations by using the built-in tools we have for message interpolation (e.g. if the number shouldn't go at the end, in some language)

Also, I think it should be consistent between "add" and "copy" to backpack

plugins/workspace-backpack/src/backpack_helpers.js
moniika: I've added the call to the previous context menu implementation if the backpack is not clicked. 

I added a section to the README on "Compatibility" to let the developer know that they need to call the old `workspace.configureContextMenu`. However, I agree that it would be better if we had a hook in core, like an additional `scopeType`.

plugins/workspace-backpack/src/backpack_helpers.js
moniika: Yes and it's not supposed to. Good catch. I'll try and fix that.

plugins/workspace-backpack/src/msg.js
moniika: I'm not sure that the number of Blocks is as useful of information for a Backpack. The number of Blocks and the number of "items" added to the Backpack is not equivalent.

I have updated "Add" to "Copy" for consistency now.

plugins/workspace-backpack/src/backpack_helpers.js
moniika: Fixed.

plugins/workspace-backpack/src/msg.js
moniika: Also, to clarify, the number that appears to the right of the context menu option in the gif is the count of the current number of items in the Backpack.

plugins/workspace-backpack/README.md
```diff
@@ -74,5 +74,23 @@ open it will be closed.
 pixel units relative to the Blockly injection div.
 - `position`: Positions the backpack UI element.
 
+## Compatibility
```
maribethb, line 80: This is a great idea! More plugins probably need this section.

While you're here, I might also mention that you use the registered Flyout classes. I'm curious how compatible this will work out to be with the continuous-toolbox which registers a custom flyout.

plugins/workspace-backpack/src/msg.js
maribethb: Sounds good.

plugins/workspace-backpack/src/msg.js
```diff
@@ -15,14 +15,14 @@ import * as Blockly from 'blockly/core';
 // context menu - Copy all Blocks on the workspace to the backpack.
 Blockly.Msg['COPY_ALL_TO_BACKPACK'] = 'Copy All Blocks to Backpack';
 /** @type {string} */
-// ontext menu - Add the selected block to the backpack.
-Blockly.Msg['COPY_TO_BACKPACK'] = 'Add to Backpack';
+// ontext menu - Copy the selected Block to the backpack.
```
maribethb: super nit: typo here `ontext`

maribethb (approved):

plugins/workspace-backpack/README.md
moniika: It looks like it is compatible with the continuous toolbox after testing.
I've added a section about how it uses the registered Flyout classes.

zhaoyaning30059089: In my project, I am using both the @blockly/workspace-backpack and @blockly/continuous-toolbox plugins simultaneously. The @blockly/workspace-backpack appears to be malfunctioning, with issues such as DOM overlapping and residual DOM elements.

maribethb: @zhaoyaning30059089 can you please file an issue with this information? it's going to get lost if it's only in a comment on a closed PR.

Please include which versions of blockly and each plugin you're using, and include a picture if possible. Thanks!

PR #732, opened 2021-05-24
Update continuous

alschmiedt:
- Fixes calculateBottomPadding to work with our new metrics manager. 
- Updates our README to include the ContinuousCategory.

### DIFF
--- a/plugins/continuous-toolbox/README.md
+++ b/plugins/continuous-toolbox/README.md
@@ -24,7 +24,6 @@ Note that this plugin uses APIs introduced in the `3.20200924.3` release of Bloc
 
 ```js
 import * as Blockly from 'blockly';
-import '../src/ContinuousCategory';
 import {ContinuousToolbox, ContinuousFlyout, ContinuousMetrics} from '@blockly/continuous-toolbox';
 
 // Inject Blockly.
--- a/plugins/continuous-toolbox/README.md
+++ b/plugins/continuous-toolbox/README.md
@@ -24,6 +24,7 @@ Note that this plugin uses APIs introduced in the `3.20200924.3` release of Bloc
 
 ```js
 import * as Blockly from 'blockly';
+import '../src/ContinuousCategory';
 import {ContinuousToolbox, ContinuousFlyout, ContinuousMetrics} from '@blockly/continuous-toolbox';
 
 // Inject Blockly.
--- a/plugins/continuous-toolbox/src/ContinuousFlyout.js
+++ b/plugins/continuous-toolbox/src/ContinuousFlyout.js
@@ -79,6 +79,7 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
    * @package
    */
   recordScrollPositions() {
+    this.scrollPositions = [];
     const categoryLabels = this.buttons_.filter((button) => button.isLabel() &&
         this.getParentToolbox_().getCategoryByName(button.getButtonText()));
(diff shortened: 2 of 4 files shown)


plugins/continuous-toolbox/README.md
```diff
@@ -24,6 +24,7 @@ Note that this plugin uses APIs introduced in the `3.20200924.3` release of Bloc
 
 ```js
+import '../src/ContinuousCategory';
```
maribethb: So I'm pretty sure we only have this here because the ContinuousCategory registers itself (in the Blockly registry). But we shouldn't make the end user deal with that. I think it would work if we added this line in ContinuousToolbox instead, so that way when they include the toolbox, it will automatically include the category as well?

I'm pretty sure that should work, we can do it in the test directory and find out. (Feel free to ping me to do it if you want)

plugins/continuous-toolbox/src/ContinuousFlyout.js
```diff
@@ -166,18 +170,20 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
...
       const lastPosition = lastCategory.position.y * this.workspace_.scale;
-      const lastCategoryHeight = metrics.contentHeight - lastPosition;
-      if (lastCategoryHeight < metrics.viewHeight) {
-        return metrics.viewHeight - lastCategoryHeight;
+      const lastCategoryHeight = contentMetrics.height - lastPosition;
+      if (lastCategoryHeight < viewMetrics.height) {
+        return viewMetrics.height - lastCategoryHeight;
       }
     }
     return 0;
```
maribethb, line 190: I think getMetrics_ (below this) is obsolete now?

plugins/continuous-toolbox/src/ContinuousMetricsFlyout.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Overrides metrics to add extra padding, if needed to the
```
maribethb: Random note: `@fileoverview`s aren't required by the style guide so I've started leaving them out if they are redundant with the class docs. I got tired of trying to figure out how to same the same sentence twice and they aren't adding value so feel free to do the same (unless this is a Blockly-specific rule I've missed, then, oops)

plugins/continuous-toolbox/README.md
alschmiedt: Ok, so I think the problem is in `test/index.js` we are using the source files directly instead of using them through the `src/index.js` file. The `ContinuousCategory` is exported in there so if we use the `src/index.js` file we shouldn't have to include the `ContinuousCategory` anywhere else.

plugins/continuous-toolbox/src/ContinuousFlyout.js
alschmiedt: Yup! Forgot to remove, thanks!

plugins/continuous-toolbox/README.md
maribethb: Ok, good to know. Do we need to export the new class `ContinuousMetricsFlyout` in `src/index.js` as well? Right now it's only used internally, but someone might need to access it.

maribethb (approved):

maribethb (approved):

PR #742, opened 2021-05-27
plugins/block-test package-lock update after removing a dependency

rachel-fenichel:
Follow-on to #740. The package-lock file does actually need to change because of the deletion--it's not just different versions of the same packages.


alschmiedt (approved):
Whoops, forgot to add. Sorry about that!

PR #746, opened 2021-06-03
Add basic version of autoscroll on drag.

maribethb:
Adds functionality for scrolling the workspace as you drag a block near the edge.

Currently, this is a rather basic autoscroll with a constant value. It can scroll in multiple directions at once (e.g. diagonally) but it always has a constant speed.

This works for different zoom levels, connection manager, bubbles, etc. and all are in the correct places. See video recording for demo.
However, there is still work remaining:

- Add smarter logic for when and how fast to autoscroll. For example, scroll when the block is near the edge, not just past it. Or scroll when the mouse is near the edge rather than just the block location (especially noticeable when dragging large block stacks because the top left corner needs to be at the edge)
- Make this behavior (and the pan on scroll move behavior) configurable by the plugin user. Including adjusting speed values.
- Clean up the plugin (e.g. some duplicated methods, plugin name in index.js, etc)

https://user-images.githubusercontent.com/8573958/120569513-e2c05e00-c3ca-11eb-84ae-c35f52dcbf65.mov


plugins/scroll-options/src/AutoScroll.js
```diff
@@ -0,0 +1,169 @@
...
+     * @type {number}
+     * @private
+     */
+    this.lastTime_ = Date.now();
+
+    /**
+     * Whether the scroll animation should continue. If this is false, the next
+     * animation frame will not be requested.
+     * @type {boolean}
+     * @private
```
alschmiedt: Do you see any of these being needed by someone who wanted to subclass this? Should any of these be @protected?

plugins/scroll-options/src/AutoScroll.js
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Ticks scrolling behavior and triggers another
+   * frame request.
+   * @param {number} now Current time in ms. This is usually passed
+   *     automatically by `requestAnimationFrame`.
+   * @private
+   */
+  nextAnimationStep_(now) {
+    if (this.shouldAnimate_) {
+      const dt = now - this.lastTime_;
```
alschmiedt: nit: prefer delta over dt.

plugins/scroll-options/src/AutoScroll.js
```diff
@@ -0,0 +1,169 @@
...
+      this.lastTime_ = now;
+      // This method could be called multiple times per ms, and we only want to
+      // scroll if we should actually move.
+      if (dt > 0) {
+        this.scrollTick_(dt);
+      }
+
+      this.animationFrameId_ =
+          requestAnimationFrame(this.nextAnimationStep_.bind(this));
```
alschmiedt: I think you can use an arrow function here instead of bind.

plugins/scroll-options/src/AutoScroll.js
```diff
@@ -0,0 +1,169 @@
...
+    const deltaX = newLocation.x - oldLocation.x;
+    const deltaY = newLocation.y - oldLocation.y;
+
+    // If we did scroll, we need to let the dragger know by how much.
+    // The dragger will update its values so that things like connection
+    // markers will stay consistent.
+    if (deltaX || deltaY) {
+      this.workspace_.currentGesture_.getCurrentDragger()
+          .moveBlockWhileDragging(deltaX, deltaY);
+    }
```
alschmiedt, line 140: We can pull this into a new method that can be reused between both of our parts of the plugin. But I think you mentioned this in your PR comment, so totally fine if you want to do that in a different PR.

plugins/scroll-options/src/AutoScroll.js
```diff
@@ -0,0 +1,169 @@
...
+    return new Blockly.utils.Coordinate(workspaceOffset.x, workspaceOffset.y);
+  }
+
+  /**
+   * Updates the scroll vector for the current autoscroll and begins the
+   * animation.
+   * @param {!Blockly.utils.Coordinate} scrollVector New scroll velocity vector
+   *     in pixels per ms.
+   */
+  updateProperties(scrollVector) {
```
alschmiedt, line 162: What about `startAnimation`? `updateProperties` makes it seem like it should be getting called constantly, but I think that is not the case? It seems like it just gets called to kick off the animation.

plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -121,5 +128,96 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
...
+    if (newLoc.y > metrics.top + metrics.height) {
+      const scrollVector =
+          SCROLL_DIRECTION_VECTORS['bottom'].scale(SCROLL_SPEED);
+      candidateScrolls.push(scrollVector);
+    }
+    if (newLoc.x < metrics.left) {
+      const scrollVector = SCROLL_DIRECTION_VECTORS['left'].scale(SCROLL_SPEED);
+      candidateScrolls.push(scrollVector);
+    }
+    if (newLoc.x > metrics.left + metrics.width) {
```
alschmiedt, line 184: This can be for another PR, but we might want to make this use `newLoc.x + this.draggingBlock_.width`. And `newLoc.y + this.draggingBlock_.height` when we are deciding to scroll past the bottom. Looking at code.org's implementation I think it starts to scroll when the bottom edge of the block is over the bottom of the viewport.

plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -121,5 +128,96 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
...
+      const scrollVector =
+          SCROLL_DIRECTION_VECTORS['right'].scale(SCROLL_SPEED);
+      candidateScrolls.push(scrollVector);
+    }
+
+    // Get the overall scroll direction vector (could scroll diagonally).
+    // CDO reduces down to just one vector per direction from all the possible
+    // ones they generate. Currently we just have one per direction so we don't
+    // need to do anything else.
```
alschmiedt: I would write out code.org it took me a second to realize that is what CDO meant haha. 

I'm a bit confused here. Do you know why they generate more than one per direction?

plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -84,6 +85,12 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
         Blockly.utils.Coordinate.sum(this.scrollDelta_, currentDragDeltaXY);
     super.drag(e, totalDelta);
     this.dragDelta_ = currentDragDeltaXY;
+
+    // Calculate the location the block is being dragged to, in ws units.
+    // This same calculation is done in super.drag().
+    const deltaPx = this.pixelsToWorkspaceUnits_(totalDelta);
+    const newLoc = Blockly.utils.Coordinate.sum(this.startXY_, deltaPx);
```
alschmiedt, line 92: I think this might point to us needing some kind of `getNewLocation` method in core.

alschmiedt (approved):
lgtm!
There are a few TODOs that I am ignoring because you mentioned you wanted to address in a different PR.

plugins/scroll-options/src/AutoScroll.js
maribethb: Yeah, you mentioned possibly having `getWsTranslation` just returning a copy in core, so then we might not even need this at all, so I didn't bother deduplicating it yet until we figured that out.

plugins/scroll-options/src/AutoScroll.js
maribethb: This could get called in the middle of an animation. For example, you are holding the block at the top-left corner, and while scrolling, you move the block to just the top of the screen. In that case, you're not really starting the animation from scratch, at least from the perspective of the block dragger, where this is called.

In the original version, it was `updateProperties` because it also passed the mouseX and mouseY values. I'm not yet sure whether I'll need them in the next phase of smarter auto-dragging because I haven't tried to implement them yet. If not, I could possibly rename this then, but I'm not sure `startAnimation` is quite accurate.

I think I need to look into not requesting a new animation frame if we're already in the middle of one, but I'll do that separately.

plugins/scroll-options/src/ScrollBlockDragger.js
maribethb: Cool, I'll do that next in core.

plugins/scroll-options/src/ScrollBlockDragger.js
maribethb: Yeah, they have several different ways to detect if/when the block should start scrolling, and how fast. I want to get the initial one in, and then refine it, and I do plan for it to work when the block goes over the edge at all, not just the top left corner.

plugins/scroll-options/src/AutoScroll.js
maribethb: Good call, I just made them all protected.

plugins/scroll-options/src/AutoScroll.js
maribethb: Done

plugins/scroll-options/src/AutoScroll.js
maribethb: Done

plugins/scroll-options/src/ScrollBlockDragger.js
maribethb: This was more of a todo/note to myself :) I don't think it'll be in here after the extra dragging logic is added.

They have more than one per direction because they have several conditions that could be fulfilled simultaneously. For example, they could have "scroll fast if the block is 30 px past the edge of the ws" *and* "scroll slow if the mouse is within 10px of the edge". Both of those conditions could be met simultaneously, and you don't want to just add them together to scroll even faster. You just want the fastest one to apply. So you'd have to select the fastest vector in each direction, and then add the resulting one vector from each direction together. Here, I'm only adding the directions together, hence the reminder to condense down to one per direction before adding.

plugins/scroll-options/src/AutoScroll.js
maribethb: Okay, I changed it so it doesn't start a new animation frame unless we haven't had one yet. So now it truly is just updating the properties, usually, as opposed to starting a new one.

It's also way smoother now that it's working properly instead of basically starting a new animation with every drag step. Yay!

plugins/scroll-options/src/ScrollBlockDragger.js
maribethb: Circling back, we decided offline not to do this in core since it needs both `delta` and `newLoc` so we'd just be wrapping the one line into a new function, which isn't super valuable here.

PR #749, opened 2021-06-07
Add additional logic for auto-scrolling workspace

maribethb:
Adds additional logic/conditions for when the workspace should auto-scroll. This version takes into account the mouse/cursor location, as well as the full block dimensions rather than just the corner of the block. The speed changes depending on how far out of bounds you are dragging.

I used the same options that code.org is using in their implementation. It seems a little fast to me, but I'm no expert. 

In the next PR, I'll add the ability for a developer to supply any/all of these options, as well as the ability to disable the feature completely (and/or disable the mousewheel handler as well). In addition, everything in this class is overridable so a developer could also subclass the ScrollBlockDragger in order to provide different logic.

https://user-images.githubusercontent.com/8573958/121089228-b0807900-c79b-11eb-91e5-c25a8c41425a.mov

### DIFF
--- a/plugins/scroll-options/src/ScrollBlockDragger.js
+++ b/plugins/scroll-options/src/ScrollBlockDragger.js
@@ -13,6 +13,11 @@ import * as Blockly from 'blockly/core';
 import {AutoScroll} from './AutoScroll';
 
 /**
+ * @typedef {!Object<string, !Array<!Blockly.utils.Coordinate>>}
+ */
+let CandidateScrolls;
+
+/**
  * A block dragger that adds the functionality for a block to be moved while
  * someone is dragging it.
  */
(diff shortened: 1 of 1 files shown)


plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -34,7 +39,51 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
      * @protected
      */
     this.dragDelta_ = new Blockly.utils.Coordinate(0, 0);
+
+    // TODO(maribethb): See if we can actually check this
```
alschmiedt: Can this be removed?

plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -140,61 +185,51 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
...
 
-    // Get the overall scroll direction vector (could scroll diagonally).
-    // Note: code.org reduces down to just one vector per direction from
-    // all the possible ones they generate. Currently we just have one per
-    // direction so we don't need to do anything else.
-    candidateScrolls.forEach(function(scroll) {
-      overallScrollVector =
-          Blockly.utils.Coordinate.sum(overallScrollVector, scroll);
-    });
+    this.getBlockCandidateScrolls_(candidateScrolls, metrics, mouse);
```
alschmiedt: Since this doesn't return anything, should this instead be something like `computeBlockCandidateScrolls_`.

plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -204,11 +239,187 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
...
+   * left.
+   * @param {!CandidateScrolls} candidateScrolls Existing lists of candidate
+   *     scrolls. Will be modified in place.
+   * @return {!Blockly.utils.Coordinate} Overall scroll vector.
+   * @protected
+   */
+  getOverallScrollVector_(candidateScrolls) {
+    let overallScrollVector = new Blockly.utils.Coordinate(0, 0);
+    for (const direction of this.scrollDirections_) {
+      const scroll = candidateScrolls[direction].reduce((fastest, current) => {
```
alschmiedt: nit: scroll -> fastestScroll or maybe just add a comment "// Find the fastest scroll in the given direction".

plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -204,11 +239,187 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
...
+   * value will be negative.
+   *
+   * @param {!Blockly.MetricsManager.ContainerRegion} metrics View metrics for
+   *     the workspace.
+   * @param {!Blockly.utils.Coordinate} mouse Mouse coordinates.
+   * @return {!Object<string, number>} An object describing the amount of
+   *     overflow in each direction.
+   * @protected
+   */
+  getMouseOverflows_(metrics, mouse) {
```
alschmiedt: nit: metrics -> viewMetrics.

alschmiedt (approved):
This is so cool! lgtm few small nits

plugins/scroll-options/src/ScrollBlockDragger.js
maribethb: We have `isMovable` etc that I want to use to get these dynamically (no need to calculate if we can't even scroll horizontally) but I want to do more thorough testing, so in another PR. But I improved the todo.

plugins/scroll-options/src/ScrollBlockDragger.js
maribethb: Good call, done

PR #753, opened 2021-06-10
Add config options to backpack for contextmenu.

moniika:
### This PR introduces the following changes:
- Adds configuration options to the Backpack constructor for context menus
- Adds information about context menus and configuration to README

### The following gifs have also been added to the README:
(which wont show until this branch merges into master)
#### An animated picture of all the context menu options in use
context-menu

#### An animated picture of the "Copy to Backpack" context menu
context-menu-precondition


plugins/workspace-backpack/src/options.js
```diff
@@ -0,0 +1,78 @@
...
+      copyToBackpack: true,
+      copyAllToBackpack: false,
+      pasteAllToBackpack: false,
+      disablePreconditionChecks: false,
+    },
+  };
+  if (!options) {
+    return parsedOptions;
+  }
+  if (options.contextMenu !== undefined) {
```
maribethb: I'm still working on my implementation so I haven't fully tested it and make sure the types work out, but I am using the spread operator to basically do
`this.options_ = {...defaultOptions, ...options}`

which will use the new values for any provided, and the default values for those not specified. so the developer can't pass "undefined" as an option explicitly, just leave it out, which seems reasonable.

plugins/workspace-backpack/src/backpack.js
```diff
@@ -10,8 +10,11 @@
  */
 
 import * as Blockly from 'blockly/core';
-import {cleanBlockXML, registerAllContextMenus} from './backpack_helpers';
+import {cleanBlockXML, registerContextMenus} from './backpack_helpers';
 import {BackpackChange, BackpackOpen} from './ui_events';
+import {
+  BackpackContextMenuOptions, BackpackOptions, parseOptions,
```
maribethb, line 16: it looks like `parseOptions` isn't being called anywhere

moniika: I think somehow I lost some logic merging. I'll fix it.

moniika: Updated.

plugins/workspace-backpack/src/options.js
moniika: I can do something similar to this (and I have just tested it works). Take a look.

plugins/workspace-backpack/README.md
```diff
@@ -30,6 +41,35 @@ const backpack = new Backpack(workspace);
 backpack.init();
 ```
 
+### Configuration
+This plugin takes an optional configuration object.
```
maribethb, line 45: I think we should explain that you can leave any of these out of the object to use the default

maribethb (approved):

PR #754, opened 2021-06-10
Backpack cleanup

moniika:
No change in behavior.
This PR does the following:
- Rearranges properties in Backpack constructor and updates some private properties to protected visibilty
- Refactors attaching listeners into a separate function
- Updates wording and capitalization in JsDoc.

### DIFF
--- a/plugins/workspace-backpack/src/backpack.js
+++ b/plugins/workspace-backpack/src/backpack.js
@@ -39,12 +39,6 @@ export class Backpack {
     this.workspace_.backpack = this;
 
     /**
-     * The backpack options.
-     * @type {!BackpackOptions}
-     */
-    this.options_ = parseOptions(backpackOptions);
-
-    /**
      * The backpack flyout. Initialized during init.
      * @type {?Blockly.IFlyout}
      * @protected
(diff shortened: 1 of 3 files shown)


plugins/workspace-backpack/src/backpack.js
```diff
@@ -39,114 +39,108 @@ export class Backpack {
...
-    this.initialized_ = false;
+    this.svgGroup_ = null;
 
     /**
-     * A list of XML (stored as strings) representing blocks in the backpack.
-     * @type {!Array<string>}
-     * @protected
+     * Top offset for backpack in svg.
+     * @type {number}
+     * @private
```
maribethb, line 126: Should these be protected as well? `getTargetArea` uses them, so if someone overrides `createDom_` to use a different image, they'd either need to use override these, or else the entire `getTargetArea` function, just to change these values.

I could be missing something though

otherwise, lgtm

maribethb (approved):

plugins/workspace-backpack/src/backpack.js
moniika: The constants starting with `SPRITE_*` are specific to the Backpack svg currently used to handle the whitespace that exists around the svg.
There isn't currently a way to change the button asset without subclassing. They would need to override both `createDom_` and `getTargetArea`. I'll make a separate PR addressing adding hints to the API to help let developers know that they should be overriding both.

PR #756, opened 2021-06-11
allow users searching block by id

duzc2:
add block.id into searching text


moniika:
Thanks for the PR.
I don't think we want this feature added to the base version of workspace search because the block id isn't something that front-end users typically have access to.
If this is something that you want for your own use, then I would recommend extending `WorkspaceSearch` in your own code after importing the plugin. I see that your edit is inside of the private function `isBlockMatch_`, which makes me think that it would be better if that function was protected (so subclasses could access it). I'll file an issue to update the visibility of that function.

PR #759, opened 2021-06-14
Add high contrast theme plugin

samelhusseini:
Add the high contrast theme plugin, replacing Blockly.Themes.HighContrast


plugins/theme-highcontrast/README.md
```diff
@@ -0,0 +1,30 @@
+# @blockly/theme-highcontrast [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
+
+A [Blockly](https://www.npmjs.com/package/blockly) high contrast theme.
```
alschmiedt, line 3: "A Blockly theme that uses darker colors for the blocks to create contrast between the block color and the white text."?

Also, I think adding a photo would be helpful. Doesn't have to be all the blocks, but maybe a stack of three or four to give people an idea.

alschmiedt (approved):

PR #763, opened 2021-06-14
Add tritanopia theme as a plugin

samelhusseini:


alschmiedt (approved):
Same comment as other themes. I think some kind of photo in the README would be nice.

PR #782, opened 2021-06-24
Add nested suites to TestSuite

BeksOmega:
### Description

I'm going to do some Test Driven Development for Project Cereal. This includes adding a test helper which makes sure that serializers round-trip the data correctly. I want the tests provided with the test helper to have a nested structure (eg Serializer.Fields.Number.Decimal) so I needed to add support for this to the TestSuite.

### Testing
In the first commit you can see a trial of this logic which I added in the blocks-plus-minus package. It output a test structure that looked like this:
```
outer suite
  inner suite 1
    - test 1
    - test 2
  inner suite 2
    - test 3
```

Which was exactly what I wanted =)

### DIFF
--- a/plugins/dev-tools/src/common_test_helpers.mocha.js
+++ b/plugins/dev-tools/src/common_test_helpers.mocha.js
@@ -28,6 +28,7 @@ TestCase.prototype.only = false;
  * Test suite configuration information.
  * @record
  * @template {TestCase} T
+ * @template {TestSuite} U
  */
 export function TestSuite() {}
 /**
@@ -48,6 +49,10 @@ TestSuite.prototype.only = false;
  * @type {!Array<T>} The associated test cases.
  */
 TestSuite.prototype.testCases = [];
+/**
+ * @type {!Array<U>} Optional inner test suites.
+ */
+TestSuite.prototype.testSuites = [];
 
 /**
  * Runs provided test cases.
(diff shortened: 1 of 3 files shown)


plugins/dev-tools/src/common_test_helpers.mocha.js
```diff
@@ -77,7 +82,12 @@ export function runTestSuites(testSuites, createTestCaseCallback) {
     let suiteCall = (testSuite.skip ? suite.skip : suite);
     suiteCall = (testSuite.only ? suite.only : suiteCall);
     suiteCall(testSuite.title, function() {
-      runTestCases(testSuite.testCases, createTestCaseCallback(testSuite));
+      if (testSuite.testSuites && testSuite.testSuites.length) {
+        runTestSuites(testSuite.testSuites, createTestCaseCallback);
+      }
+      if (testSuite.testCases && testSuite.testCases.length) {
+        runTestCases(testSuite.testCases, createTestCaseCallback(testSuite));
+      }
```
moniika, line 90: The check for whether the properties exist is a little confusion given that the record structure indicates them as non-nullable.
I think that this check is reasonable though, given that with the new addition of the testSuites property, I think it would be convenient to not necessarily define both.

So, I'd like the type for `testCases` and `testSuites` to be updated to be nullable (ex: `?Array<T>` instead of `!Array<T>`).

moniika (approved):
Just one comment, otherwise looks good.

PR #795, opened 2021-07-01
Fix deprecation warning in vue example

BeksOmega:
### Description

In the latest Blockly release variableDB_ was renamed to nameDB_ and this was causing a deprecation warning.

Also fixed the use of the deprecated Blockly.Variable.NAME_TYPE.

### Testing

1) Linked this sample to the latest core dist.
2) Tested generating the code for the price finding block.

Before I would get a deprecation warning in the console, and now I do not.

### DIFF
--- a/examples/blockly-vue/package-lock.json
+++ b/examples/blockly-vue/package-lock.json
@@ -2361,6 +2361,12 @@
         "@types/node": "*"
       }
     },
+    "node_modules/@types/color-name": {
+      "version": "1.1.1",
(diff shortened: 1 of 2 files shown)


alschmiedt:
It doesn't matter as much now since the release is out, but we generally try to keep all of our release related changes in a separate branch. This quarter the branch is `q2_2021_release`, so it might be better if you could merge it into there instead of master.

NeilFraser (approved):

PR #806, opened 2021-07-02
Updates name of content-highlight plugin

alschmiedt:
Updates the name of the content highlight plugin.


moniika (approved):
Looks right. package-lock.json probably should be updated because package.json was edited.

alschmiedt:
Yup, good call! Just added.

PR #817, opened 2021-07-09
Remove peer dependencies from templates and inject the latest version…

rachel-fenichel:
… of blockly through create.js

Clang-format formatted parts of create.js and I didn't want to fight it.

The goal of this change is to make sure that peer dependencies are always recent when using the create script *without* needing to update the templates during our release step.

Note that the create script does something similar with dev dependencies: for each named dependency it finds the most recent version and sets the dependency to ^latest.


moniika (approved):
lgtm.
It would be nice to do something similar for the dev-tools devDependency (in a future PR).

PR #819, opened 2021-07-15
Fix typing and update declaration of @record types to match style guide

moniika:
Update typing for test types:
- Use syntax for declaring `@record` type as recommended in [style guide](https://google.github.io/styleguide/jsguide.html#features-classes-interfaces)
- Correct typing in `runTestSuites`
- Remove second template type from `TestSuite`

### Additional Information
The eslint warnings that this fixes are:
```
/usr/local/google/home/kozbial/github/blockly/samples-moniika/plugins/dev-tools/src/common_test_helpers.mocha.js
  49:0  warning  The type 'T' is undefined  jsdoc/no-undefined-types
  53:0  warning  The type 'U' is undefined  jsdoc/no-undefined-types
```

### DIFF
--- a/plugins/dev-tools/src/block_test_helpers.mocha.js
+++ b/plugins/dev-tools/src/block_test_helpers.mocha.js
@@ -17,66 +17,81 @@ const {
 } = commonTestHelpers;
 
 /**
- * Code generation test case.
- * @extends {TestCase}
+ * Code generation test case configuration.
+ * @implements {TestCase}
  * @record
  */
-export function CodeGenerationTestCase() {}
-CodeGenerationTestCase.prototype = new TestCase();
-/**
- * @type {string} The expected code.
- */
-CodeGenerationTestCase.prototype.expectedCode = '';
-/**
- * @type {number|undefined} The expected inner order.
- */
-CodeGenerationTestCase.prototype.expectedInnerOrder = undefined;
-/**
- * @type {boolean|undefined} Whether to use workspaceToCode instead of
- * blockToCode for test.
- */
-CodeGenerationTestCase.prototype.useWorkspaceToCode = false;
-/**
- * A function that creates the block for the test.
- * @param {!Blockly.Workspace} workspace The workspace context for this test.
- * @return {!Blockly.Block}
- */
-CodeGenerationTestCase.prototype.createBlock = undefined;
+export class CodeGenerationTestCase {
+  /**
+   * Class for a code generation test case.
+   */
+  constructor() {
+    /**
+     * Creates the block to use for this test case.
+     * @param {!Blockly.Workspace} workspace The workspace context for this
+     *    test.
+     * @return {!Blockly.Block}
+     */
+    this.createBlock;
+    /**
+     * @type {string} The expected code.
+     */
+    this.expectedCode;
+    /**
+     * @type {boolean|undefined} Whether to use workspaceToCode instead of
+     * blockToCode for test.
+     */
+    this.useWorkspaceToCode;
+    /**
+     * @type {number|undefined} The expected inner order.
+     */
+    this.expectedInnerOrder;
+  }
+}
 
 /**
  * Code generation test suite.
  * @extends {TestSuite<CodeGenerationTestCase>}
  * @record
  */
-export function CodeGenerationTestSuite() {}
-CodeGenerationTestSuite.prototype = new TestSuite();
-/**
- * @type {!Blockly.Generator} The generator to use for running test cases.
- */
-CodeGenerationTestSuite.prototype.generator = undefined;
+export class CodeGenerationTestSuite {
+  /**
+   * Class for a code generation test suite.
+   */
+  constructor() {
+    /**
+     * @type {!Blockly.Generator} The generator to use for running test cases.
+     */
+    this.generator;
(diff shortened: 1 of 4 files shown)


plugins/dev-tools/src/block_test_helpers.mocha.js
```diff
@@ -17,66 +17,81 @@ const {
...
+   * Class for a code generation test case.
+   */
+  constructor() {
+    /**
+     * Creates the block to use for this test case.
+     * @param {!Blockly.Workspace} workspace The workspace context for this
+     *    test.
+     * @return {!Blockly.Block}
+     */
+    this.createBlock;
```
BeksOmega: I think methods are supposed to be declared in the body of the class, based on the linked guide. I think the only other place a function is declared is for `assertBlockStructure`.

plugins/dev-tools/src/common_test_helpers.mocha.js
```diff
@@ -5,54 +5,65 @@
...
+    /**
+     * @type {string} The title for the test case.
+     */
+    this.title;
+    /**
+     * @type {?Array<T>} The associated test cases.
+     */
+    this.testCases;
+    /**
+     * @type {?Array<TestCase<T>>} Optional inner test suites.
```
BeksOmega: I think the previous annotation is correct, because test suites can now be parameterized with their inner test suite type. In Java this would look like:
```java
public interface TestSuite<U extends TestSuite, T extends TestCase> { }

// Sort of a recursive type definition, but totally legal.
public class GeneratorTestSuite implements TestSuite<GeneratorTestSuite, GeneratorTestCase> { }
```

If we declare it as `?Array<!TestSuite<T>>` I don't that's not correct because the subclass of `TestSuite` (eg `GeneratorTestSuite`) could have properties that don't exist on `TestSuite` and this type annotation would make those properties illegal to access, because it would be legal to store any type of `TestSuite` including the base one, in this array.

BeksOmega (commented):
Yay this is much nicer :D Originally I didn't like class syntax for JS, but it's really grown on me =)

One style fix, and then I think the previous annotation for `testSuites` was correct.

plugins/dev-tools/src/common_test_helpers.mocha.js
moniika: That's a good point.

The problem I'm trying to solve is that `runTestSuites` currently makes the assumption that the `testSuites` stored accept the same `TestCase` type (`T`) as the parent `TestSuite`, so I'm trying to more formally capture that relationship.

I'm going to think a little about this.

BeksOmega: Hmmm I wonder if it's possible to declare something like:
```java
// Added <U, T> to the TestSuite type parameter
public interface TestSuite<U extends TestSuite<U, T>, T extends TestCase> { }
```

I think that notates the fact that the inner `TestSuite` type must recursively contain itself, and the same `TestCase` type as the outer `TestSuite`.

But I'm not sure if the closure compiler supports type constraints like that.

moniika: I've reverted my changes to the class annotation and updated the method `runTestSuites`

plugins/eslint-config/index.js
```diff
@@ -53,6 +53,9 @@ module.exports = {
     'no-warning-comments': 'warn',
     'new-cap': ['error', {'capIsNewExceptionPattern': '^.*Error'}],
     'no-invalid-this': 'off',
+    // valid-jsdoc does not work properly for interface methods.
+    // https://github.com/eslint/eslint/issues/9978
+    'valid-jsdoc': 'off',
```
BeksOmega, line 58: Do we want to turn this off globally?

plugins/dev-tools/src/field_test_helpers.mocha.js
```diff
@@ -9,45 +9,55 @@ import {runTestCases, TestCase} from './common_test_helpers.mocha';
...
-FieldValueTestCase.prototype.expectedText = undefined;
-/**
- * @type {!RegExp|string|undefined} Optional error message matcher if test case
- *    is expected to throw.
- */
-FieldValueTestCase.prototype.errMsgMatcher = undefined;
+export class FieldValueTestCase {
+  /**
+   * Class for a a field value test case.
+   */
```
BeksOmega, line 18: No need to change this, but I learned the other day that [descriptions are optional for constructors](https://google.github.io/styleguide/jsguide.html#jsdoc-class-comments) (you still need to provide parameter annotations though).

BeksOmega (approved):
LGTM!

plugins/dev-tools/src/common_test_helpers.mocha.js
BeksOmega: Looks correct to me :D

plugins/eslint-config/index.js
moniika: This was intentional. This rule is deprecated according to: https://eslint.org/blog/2018/11/jsdoc-end-of-life (which is also linked in the github issue).
The `eslint-plugin-jsdoc` is separate and will still be checking our jsdoc (so we don't lose coverage there).

Issue #768, opened 2021-06-16
block-plus-minus triggers warnings about overwriting block definitions


moniika:
**Category**

Plugins

**Component**
block-plus-minus

**Describe the bug**
Warning in console about overwriting existing json definitions.
```
Block definition #0 in JSON array overwrites prior definition of "lists_create_with".
Block definition #0 in JSON array overwrites prior definition of "procedures_defnoreturn".
Block definition #0 in JSON array overwrites prior definition of "procedures_defreturn".
```

**To Reproduce**

1. Run plugin using `npm run start`
2. Observe warnings in console.

**Expected behavior**

No warnings in console.

**Additional context**

This could be fixed by adding APIs in core to allow for overriding exising block definitions.

BeksOmega:
Related to https://github.com/google/blockly/issues/2788

johnnyoshika:
@BeksOmega Is this happening because keys are not unregistering before overriding the default block behaviors? I noticed that for `controls_if_mutator` and `text_join_mutator`, their keys gets unregistered before registering, so a warning is not logged to the console for those:

https://github.com/google/blockly-samples/blob/master/plugins/block-plus-minus/src/if.js#L178-L180
https://github.com/google/blockly-samples/blob/master/plugins/block-plus-minus/src/text_join.js

`lists_create_with`, `procedures_defnoreturn`, and `procedures_defreturn` on the other hand do not unregister their keys.

BeksOmega:
> @BeksOmega Is this happening because keys are not unregistering before overriding the default block behaviors? I noticed that for `controls_if_mutator` and `text_join_mutator`, their keys gets unregistered before registering, so a warning is not logged to the console for those:
> 
> https://github.com/google/blockly-samples/blob/master/plugins/block-plus-minus/src/if.js#L178-L180
> https://github.com/google/blockly-samples/blob/master/plugins/block-plus-minus/src/text_join.js
> 
> `lists_create_with`, `procedures_defnoreturn`, and `procedures_defreturn` on the other hand do not unregister their keys.

Yeah definitely! So mutators/extensions have the ability to be unregistered, but blocks do not (currently) because they're applied directly to the `Blockly.Blocks` object, rather than being provided via the registry.

[...]

PR #829, opened 2021-07-26
Plus / Minus blocks - delete original blocks before registering new ones

johnnyoshika:
Delete original blocks to prevent the following warnings in the console:

```
Block definition #0 in JSON array overwrites prior definition of "lists_create_with".
Block definition #0 in JSON array overwrites prior definition of "procedures_defnoreturn".
Block definition #1 in JSON array overwrites prior definition of "procedures_defreturn".
```

Relates to google/blockly-samples#768

### DIFF
--- a/plugins/block-plus-minus/src/list_create.js
+++ b/plugins/block-plus-minus/src/list_create.js
@@ -12,7 +12,8 @@ import Blockly from 'blockly/core';
 import {createPlusField} from './field_plus';
 import {createMinusField} from './field_minus';
 
-// Delete original block because there's no way to unregister it: https://github.com/google/blockly-samples/issues/768#issuecomment-885663394
+// Delete original block because there's no way to unregister it:
+// https://github.com/google/blockly-samples/issues/768#issuecomment-885663394
 delete Blockly.Blocks['lists_create_with'];
 
 /* eslint-disable quotes */
--- a/plugins/block-plus-minus/src/list_create.js
+++ b/plugins/block-plus-minus/src/list_create.js
@@ -12,6 +12,9 @@ import Blockly from 'blockly/core';
 import {createPlusField} from './field_plus';
 import {createMinusField} from './field_minus';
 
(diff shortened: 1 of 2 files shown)


johnnyoshika:
@googlebot I signed it!

plugins/block-plus-minus/src/list_create.js
```diff
@@ -12,6 +12,9 @@ import Blockly from 'blockly/core';
 import {createPlusField} from './field_plus';
 import {createMinusField} from './field_minus';
 
+// Delete original block because there's no way to unregister it: https://github.com/google/blockly-samples/issues/768#issuecomment-885663394
```
BeksOmega:
```suggestion
// Delete original block because there's no way to unregister it:
// https://github.com/google/blockly-samples/issues/768#issuecomment-885663394
```

plugins/block-plus-minus/src/procedures.js
```diff
@@ -14,6 +14,10 @@ import {createPlusField} from './field_plus';
 
 Blockly.Msg['PROCEDURE_VARIABLE'] = 'variable:';
 
+// Delete original blocks because there's no way to unregister them: https://github.com/google/blockly-samples/issues/768#issuecomment-885663394
```
BeksOmega:
```suggestion
// Delete original blocks because there's no way to unregister them:
// https://github.com/google/blockly-samples/issues/768#issuecomment-885663394
```

BeksOmega (approved):
Hi! Thanks so much for the PR and the write up of how you tested it :D The logic looks perfect, it just needs a couple of format changes. You should be able to accept and commit them right from the github UI =)

Thank you again, once those changes are in I should be able to merge this!

johnnyoshika:
@BeksOmega accepted all your suggestions, thank you!

BeksOmega:
Wooo all checks passed! Thank you again for putting up this PR :D

These changes will be published to npm on Thursday (07/29), so after that you should be able to upgrade, and you'll no longer have those gross warnings in your tests :D

PR #834, opened 2021-07-29
Updates the version of Blockly in devDependencies

alschmiedt:
The modal plugin was not working. This is because the modal plugin used an old version of Blockly as the devDependency. This is fine if running by itself, however we use dev-tools when creating the test pages and this relies on a later version of Blockly.

I think this is further support that the plugins devDependencies need to be updated every release to use the latest version of Blockly.


gonfunko (approved):
Looks good, do you think we should just update the version for all the plugins in this PR?

alschmiedt:
Since the other ones aren't broken I am going to leave for right now just so I can finish releasing. And I will create a PR after releasing to update all of them.

PR #839, opened 2021-08-04
Add the ability to configure scroll-options and gets it ready for publishing

maribethb:
**This removes `private: true` from package.json so the plugin will be published during next release. Please review for readiness**

This PR adds the ability to configure the behavior in the scroll-options plugin. The API is documented in the README.

It also takes care of details for publishing like correcting the publishConfig and updating the version number.

I've tested all of the various option configurations by hand. It would be cool to add sliders to the advanced playground to play around with it but that required more time than I had :)

Things to note:
- The package is named `@blockly/plugin-scroll-options`. We aren't currently consistent with adding `plugin` eg `@blockly/continuous-toolbox` vs `@blockly/plugin-modal`. [The docs](https://developers.google.com/blockly/guides/plugins/naming) say to add the `plugin` so I did.
- I'm using a JsDoc-style typedef instead of a Closure-style typedef for EdgeScrollOptions. The reason for this is basically because the Closure-style typedef converts badly to TypeScript, so anyone trying to compile this to TypeScript or using ts language services (basically enabled by default in VS Code) gets a very misleading error message since this is directly related to a public API. And if they follow the error message's suggested resolution, it would actually break the plugin. Since we don't actually use the Closure compiler, using the JsDoc style should be fine. We do tell eslint to use closure style, but eslint appears to have no problems with this code.

More details about this below in case anyone's curious:
Compare the most recent 2 commits for the typedef style change.
`EdgeScrollOptions` typedef has various properties that are optional. in closure, I don't think you can do the `=` syntax to mark properties as optional, you have to say `| undefined`. However, TypeScript seems to then interpret this as the property *must* be specified, though you can pass the value as undefined. However if you actually do pass eg `{slowBlockSpeed: undefined}` that will break the functionality. Because of the property merging we are doing, you just have to leave out properties so that they won't be overwritten. If you explicitly pass undefined, the default value will be overwritten with `undefined` as the value, which will break the plugin. But if you leave out any properties, you get the following typescript error
 
 ```js
  scrollOptionsPlugin.updateEdgeScrollOptions({slowMouseSpeed: 2});
// Argument of type '{ slowMouseSpeed: number; }' is not assignable to parameter of type 'EdgeScrollOptions'.
// Type '{ slowMouseSpeed: number; }' is missing the following properties from type 'EdgeScrollOptions': slowBlockSpeed, fastBlockSpeed, slowBlockStartDistance, fastBlockStartDistance, and 5 more.
```

So TypeScript is instructing them to add properties they really shouldn't add. 

I also tried using a `@record` type instead of a typedef, and the same TS error occurs (record types also don't support the `=` syntax for optional properties so I assume it's the same problem).

Using the JsDoc style typedef leads to the correct TS typings, plus it also is better commented/documented than what is possible to do with Closure.

### DIFF
--- a/plugins/scroll-options/README.md
+++ b/plugins/scroll-options/README.md
@@ -82,36 +82,46 @@ on `ScrollOptions`:
 -   `enableEdgeScroll`/`disableEdgeScroll`
 -   `updateEdgeScrollOptions`
 
-The edge scroll options are provided in an object with the following properties:
+**IMPORTANT**: Currently, the options to configure the edge scroll behavior are
+provided statically due to the way a BlockDragger is initialized. Thus, each
+instance of the plugin will share the same settings, even if you call
+`updateEdgeScrollOptions` on different instances of the plugin (i.e., on
+different workspaces). You can have some workspaces using the plugin and some
+not, but any that are using the plugin will share the settings. If this is a
+blocking issue for you, please file an issue on blockly-samples.
 
--   `slowBlockSpeed`, `fastBlockSpeed`: Pixels per ms to scroll based on how far
-    the block is from the edge of the viewport.
--   `slowBlockStartDistance`: Distance in workspace units that the edge of the
-    block is from the edge of the viewport before the corresponding scroll speed
-    will be used. Can be negative to start scrolling before the block extends
-    over the edge.
--   `fastBlockStartDistance`: Same as above, for fast speed. Must be larger than
-    `slowBlockStartDistance`.
--   `oversizeBlockThreshold`: If a block takes up this percentage of the
+The edge scroll options are provided in an object with the following properties
+(default values in parentheses):
+
+-   `slowBlockSpeed` (0.28), `fastBlockSpeed` (1.4): Pixels per ms to scroll
+    based on how far the block is from the edge of the viewport.
+-   `slowBlockStartDistance` (0): Distance in workspace units that the edge of
+    the block is from the edge of the viewport before the corresponding scroll
+    speed will be used. Can be negative to start scrolling before the block
+    extends over the edge.
+-   `fastBlockStartDistance` (50): Same as above, for fast speed. Must be larger
+    than `slowBlockStartDistance`.
+-   `oversizeBlockThreshold` (0.85): If a block takes up this percentage of the
     viewport or more, it will be considered oversized. Rather than using the
     block edge, we use the mouse cursor plus the given margin size to activate
     block-based scrolling.
--   `oversizeBlockMargin`: Cursor margin to use for oversized blocks. A bigger
-    value will cause the workspace to scroll sooner, i.e., the mouse can be
-    further inward from the edge when scrolling begins.
--   `slowMouseSpeed`, `fastMouseSpeed`: Pixels per ms to scroll based on how far
-    the mouse is from the edge of the viewport.
--   `slowMouseStartDistance`: Distance in workspace units that the mouse is from
-    the edge of the viewport before the corresponding scroll speed will be used.
-    Can be negative to start scrolling before the mouse extends over the edge.
--   `fastMouseStartDistance`: Same as above, for fast speed. Must be larger than
-    `slowMouseStartDistance`.
+-   `oversizeBlockMargin` (15): Cursor margin to use for oversized blocks. A
+    bigger value will cause the workspace to scroll sooner, i.e., the mouse can
+    be further inward from the edge when scrolling begins.
+-   `slowMouseSpeed` (0.5), `fastMouseSpeed` (1.6): Pixels per ms to scroll
+    based on how far the mouse is from the edge of the viewport.
+-   `slowMouseStartDistance` (0): Distance in workspace units that the mouse is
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: plugins/scroll-options/README.md [hunk 1]
```diff
@@ -1,46 +1,133 @@
...
-  -->
+The plugin is configured during initialization for convenience. It can also be
+configured, or reconfigured, at any time afterwards using the following methods
+on `ScrollOptions`:
+
+-   `enableWheelScroll`/`disableWheelScroll`
+-   `enableEdgeScroll`/`disableEdgeScroll`
+-   `updateEdgeScrollOptions`
+
+The edge scroll options are provided in an object with the following properties:
```
alschmiedt: Could we add what the default value of these are in the description?

plugins/scroll-options/src/index.js
```diff
@@ -4,18 +4,21 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-// TODO: Edit plugin overview.
-/**
- * @fileoverview Plugin overview.
- */
+import Blockly from 'blockly/blockly';
```
alschmiedt: Why not 'blockly/core'?

plugins/scroll-options/src/index.js
```diff
@@ -4,18 +4,21 @@
...
+
```
alschmiedt: nit: remove extra line

plugins/scroll-options/src/index.js
```diff
@@ -29,16 +32,94 @@ export class Plugin {
...
 
   /**
-   * Moves the currently dragged block as the user scrolls the workspace.
+   * Disables scrolling with mousewheel during block drag.
+   */
+  disableWheelScroll() {
+    if (!this.wheelEvent_) {
+      // Already disabled.
+      return;
+    }
```
alschmiedt, line 90: I think this condition will only work if the wheel scroll has never been enabled. When we call `unbind` it clears the array, but `this.wheelEvent` will still equal `[]`. I think we need to set `this.wheelEvent_` to null after we unbind if we want this to work.

plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -352,21 +415,26 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
...
-          blockBounds.left, mouse.x - this.options_.oversizeBlockMargin);
+          blockBounds.left,
+          mouse.x - ScrollBlockDragger.options_.oversizeBlockMargin);
       blockBounds.right = Math.min(
-          blockBounds.right, mouse.x + this.options_.oversizeBlockMargin);
+          blockBounds.right,
+          mouse.x + ScrollBlockDragger.options_.oversizeBlockMargin);
     }
 
     // The coordinate system is negative in the top and left directions, and
```
alschmiedt, line 440: We aren't currently consistent with how we do this in samples, but I think it is probably worth it to register this class so a user can use the name or the class depending on their needs.

alschmiedt (approved):
Overall lgtm. 

Talked offline about what this would look like for multiple workspaces. Currently, we do not allow having different options for separate workspaces. If a user calls `scrollOptionsPlugin.init(options)` it will update the options for any scroll option plugins that were initialized previously. If we can't find a good solution for this then we will settle for updating documentation and fixing it if someone creates a feature request for it later on.

plugins/scroll-options/src/ScrollBlockDragger.js
maribethb: Done

plugins/scroll-options/src/index.js
maribethb: Good catch! I fixed this and tested that calling disable after enable works

plugins/scroll-options/src/index.js
maribethb: Fixed

plugins/scroll-options/README.md
maribethb: Done

maribethb:
> Overall lgtm.
> 
> Talked offline about what this would look like for multiple workspaces. Currently, we do not allow having different options for separate workspaces. If a user calls `scrollOptionsPlugin.init(options)` it will update the options for any scroll option plugins that were initialized previously. If we can't find a good solution for this then we will settle for updating documentation and fixing it if someone creates a feature request for it later on.

To close the loop on this, we decided this is good for v1 and I've updated the instructions to address this issue. If this is blocking your adoption of this plugin, please file an issue so we can prioritize appropriately. In the meantime, this version will go in as the public API of the plugin does not need to change as we can probably change the implementation in ScrollBlockDragger. One option Abby suggested is storing the options in the plugin instead of the block dragger, and listening for a block drag event and updating the options from the ones stored in the plugin.

The reason this isn't straightforward is because a new BlockDragger is created every time a new gesture is started, so we can't just init the class with options when the plugin is initialized, and it would be a mess to try to change the initialization of a block dragger since it is done deep in core. Listening for the block drag event seems very hacky to me but would probably work.

PR #846, opened 2021-08-10
feat: add JSON options to serialization test helpers

BeksOmega:
### Description

Adds an option to the serialization test helpers to test JSON serialization instead of XML serialization. This should not be merged until the JSON serialization system is in core.

I've also been unable to test this. I tried adding tests to the blocks-plus-minus package, but that required linking that package to dev-tools, and then linking dev-tools to my local checkout of core, which I couldn't get to work :/

### DIFF
--- a/plugins/dev-tools/src/block_test_helpers.mocha.js
+++ b/plugins/dev-tools/src/block_test_helpers.mocha.js
@@ -94,7 +94,7 @@ export class SerializationTestCase {
     this.json;
     /**
      * @type {string|undefined} The expected json after round trip. Provided if
-     *    it different from json that was passed in.
+     *    it is different from json that was passed in.
      */
     this.expectedJson;
   }
--- a/plugins/dev-tools/src/block_test_helpers.mocha.js
+++ b/plugins/dev-tools/src/block_test_helpers.mocha.js
@@ -168,7 +168,7 @@ export const runSerializationTestSuite = (testCases) => {
    * @param {!SerializationTestCase} testCase The test case information.
    * @return {!Function} The test callback.
    */
-  const createXmlToBlockTestCallback = (testCase) => {
+  const createSerializedDataToBlockTestCallback = (testCase) => {
     return function() {
       let block;
       if (testCase.json) {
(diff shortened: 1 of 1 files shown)


plugins/dev-tools/src/block_test_helpers.mocha.js
```diff
@@ -143,8 +158,14 @@ export const runSerializationTestSuite = (testCases) => {
    */
   const createXmlToBlockTestCallback = (testCase) => {
     return function() {
-      const block = Blockly.Xml.domToBlock(Blockly.Xml.textToDom(
-          testCase.xml), this.workspace);
+      let block;
+      if (testCase.json) {
+        block = Blockly.serialization.blocks.load(
```
BeksOmega, line 163: Should be switched to `blocks.append`

plugins/dev-tools/src/block_test_helpers.mocha.js
```diff
@@ -60,23 +60,38 @@ CodeGenerationTestSuite.prototype.generator = undefined;
...
+     */
+    this.expectedXml;
+    /**
+     * @type {string} The block json to use for test. Do not provide if xml is
+     *     provided.
+     */
+    this.json;
+    /**
+     * @type {string|undefined} The expected json after round trip. Provided if
+     *    it different from json that was passed in.
```
alschmiedt: if it **is** different

plugins/dev-tools/src/block_test_helpers.mocha.js
```diff
@@ -155,13 +176,21 @@ export const runSerializationTestSuite = (testCases) => {
    */
   const createXmlRoundTripTestCallback = (testCase) => {
```
alschmiedt: Could we update the name of this and the above? Either by removing `Xml` or by adding `createXmlOrJson...`?

alschmiedt (approved):
lgtm

Just a remind to update the base branch to be master.

plugins/dev-tools/src/block_test_helpers.mocha.js
BeksOmega: ~~We can, but it would be a breaking change :/ Just confirming that's ok before I change it?~~

Never mind, I wasn't looking at this closely, will change!

alschmiedt:
Thanks @BeksOmega!

PR #858, opened 2021-08-12
feat: add JSO serialization to advanced playground

BeksOmega:
### Description

This should be merged after the JSO serializer is released.

Adds support for the new JSO serializer to the advanced playground. This will make it easy to use/test the JSO serializer.

### Testing
Manually tested that the tab and context menu items are properly added. But the actual serialization can't be tested until the JSO serializer is released

### DIFF
--- a/plugins/dev-tools/src/playground/index.js
+++ b/plugins/dev-tools/src/playground/index.js
@@ -118,6 +118,8 @@ export function createPlayground(
         const errorModel = window.monaco.editor.createModel('');
         const editorXmlContextKey =
             editor.createContextKey('isEditorXml', true);
+        const editorJsonContextKey =
+            editor.createContextKey('isEditorJson', true);
 
         // Load / Save playground state.
         const playgroundState = new LocalStorageState(`playgroundState_${id}`, {
@@ -191,6 +193,7 @@ export function createPlayground(
           currentTab = tab;
           currentGenerate = tab.generate;
           const isXml = tab.state.name == 'XML';
+          const isJson = tab.state.name == 'JSON';
           editor.setModel(currentTab.state.model);
           editor.updateOptions({
             readOnly: !isXml,
@@ -203,6 +206,7 @@ export function createPlayground(
                   (t.tabElement == tab.tabElement) ? '#1E1E1E' : '#2D2D2D');
           // Update editor state.
           editorXmlContextKey.set(isXml);
+          editorJsonContextKey.set(isJson);
           playgroundState.set('activeTab', tab.state.name);
           playgroundState.save();
         };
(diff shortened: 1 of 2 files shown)


plugins/dev-tools/src/playground/index.js
```diff
@@ -485,13 +507,23 @@ function registerTabButtons(editor, playground, tabButtons, updateEditor) {
...
-    if (playground.getCurrentTab().state.name !== 'XML') {
-      return;
-    }
+  const loadXml = () => {
     const xml = editor.getModel().getValue();
     const workspace = playground.getWorkspace();
-    Blockly.Xml.domToWorkspace(Blockly.Xml.textToDom(xml), workspace);
+    try {
+      Blockly.Xml.domToWorkspace(Blockly.Xml.textToDom(xml), workspace);
+    } catch(e) {
```
alschmiedt: Add a space after catch to fix the lint error.

plugins/dev-tools/src/playground/index.js
```diff
@@ -521,6 +553,7 @@ function registerEditorCommands(editor, playground) {
     contextMenuOrder: 1,
     run: save,
   });
+  // Add Clean XML action (only available on thee JSON tab).
```
alschmiedt: thee --> the

alschmiedt: Also should this be only available on the XML tab and the one below be for the JSON tab?

plugins/dev-tools/src/playground/index.js
```diff
@@ -550,12 +595,11 @@ function registerEditorCommands(editor, playground) {
...
     const ctrlCmd = e.metaKey || e.ctrlKey;
     if (ctrlCmd && e.keyCode === Blockly.utils.KeyCodes.S) {
-      // Save.
       save();
       e.preventDefault();
     } else if (ctrlCmd && e.keyCode === Blockly.utils.KeyCodes.ENTER) {
-      // Load.
-      load();
+      loadXml();
+      loadJson();
```
alschmiedt: Wouldn't this load more than the desired amount of blocks on the workspace?

alschmiedt (approved):

plugins/dev-tools/src/playground/index.js
BeksOmega: Lol yes, ty for the catch!

plugins/dev-tools/src/playground/index.js
BeksOmega: I don't think so b/c they're both looking at the current value in the editor tab, and only one of them will work correctly based on that (the other will error and return). However, this is confusing, so I added return values to make it more explicit.

BeksOmega:
Also added a fix for #1049

plugins/dev-tools/src/playground/index.js
```diff
@@ -191,6 +193,7 @@ export function createPlayground(
           currentTab = tab;
           currentGenerate = tab.generate;
           const isXml = tab.state.name == 'XML';
+          const isJson = tab.state.name == 'JSON';
```
cpcallen, line 196: Since `isXml` and `isJson` are mutually exclusive (but not exhaustive), might it be better for the two booleans to be replaced by an enum?

plugins/dev-tools/src/playground/index.js
```diff
@@ -191,6 +193,7 @@ export function createPlayground(
...
           editor.setModel(currentTab.state.model);
           editor.updateOptions({
             readOnly: !isXml,
```
cpcallen: Is the JSON tab supposed to be readonly?

plugins/dev-tools/src/playground/index.js
```diff
@@ -242,29 +246,47 @@ export function createPlayground(
 
         // Register default tabs.
         const tabs = {
+          'JSON': registerGenerator(
+              'JSON',
+              'json',
+              (ws) => {
```
cpcallen, line 252: Will clang-format preserve this more compact formatting?:
```suggestion
          'JSON': registerGenerator( 'JSON', 'json', (ws) => {
```
(Here and below.)

cpcallen (approved):

plugins/dev-tools/src/playground/index.js
BeksOmega: I think they need to be booleans to set the context keys below.

If they weren't just local to this function I would definitely change it, but since they're only used in this one spot I think keeping them as booleans makes more sense.

plugins/dev-tools/src/playground/index.js
BeksOmega: Nope! Thanks for the catch!

plugins/dev-tools/src/playground/index.js
BeksOmega: I think it will, but doesn't that break the [recommendation](https://google.github.io/styleguide/jsguide.html#formatting-where-to-break) of wrapping at a higher syntactic level? I know this takes up more space, but personally I find it more readable.

PR #859, opened 2021-08-12
Update package-locks in preparation for release

maribethb:
It seems every plugin had an inconsistent version in the package-lock, e.g.

```json
"name": "@blockly/dev-tools",
	"version": "2.5.4",
	"lockfileVersion": 2,
	"requires": true,
	"packages": {
		"": {
			"name": "@blockly/dev-tools",
			"version": "2.5.3",
```

Where the version under `packages` was one version behind the actual plugin's version. Seems weird and I wonder if this is a new problem introduced by package-lock version 2 and the version not being updated here by lerna.


package-lock.json
```diff
@@ -6070,7 +6070,72 @@
...
+                "signal-exit",
+                "string_decoder",
+                "string-width",
+                "strip-ansi",
+                "strip-json-comments",
+                "tar",
+                "util-deprecate",
+                "wide-align",
+                "wrappy",
+                "yallist"
```
BeksOmega, line 6138: Did we expect this to grow so much? I'm not sure what bundled dependencies are tbh.

BeksOmega (approved):
LGTM, just one question about bundled dependencies. (I hope you didn't have to do this manually lol)

package-lock.json
maribethb: Nah, this is purely autogenerated. I'm not sure why it changed so much either because none of the versions of this package changed... but generally I don't understand what npm is doing to the package locks and I just let it do its thing... I reran install just to be sure and npm is sure, apparently.

PR #860, opened 2021-08-12
Fix static properties in ScrollBlockDragger.

maribethb:
Fix static properties in ScrollBlockDragger because `static` is apparently experimental and babel didn't decide to tell me until now...

I had to make `options` public because it didn't like it being protected now that it's not on the class itself.

### DIFF
--- a/plugins/scroll-options/src/ScrollBlockDragger.js
+++ b/plugins/scroll-options/src/ScrollBlockDragger.js
@@ -97,53 +97,6 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
   }
 
   /**
-   * Whether the behavior to scroll the workspace when a block is dragged near
-   * the edge is enabled.
-   * @type {boolean}
-   * @public
-   */
-  static edgeScrollEnabled = true;
-
-  /**
-   * Configuration options for the scroll-options settings.
-   * @type {!EdgeScrollOptions}
-   * @protected
-   */
-  static options_ = defaultOptions;
-
-  /**
-   * Update the scroll options. Only the properties actually included in the
-   * `options` parameter will be set. Any unspecified options will use the
-   * previously set value (where the initial value is from `defaultOptions`).
-   * Therefore, do not pass in any options with explicit `undefined` or `null`
-   * values. The plugin will break. Just leave them out of the object if you
-   * don't want to change the default value.
-   *
-   * This method is safe to call multiple times. Subsequent calls will add onto
-   * previous calls, not completely overwrite them. That is, if you call this
-   * with:
-   *
-   *     `updateOptions({fastMouseSpeed: 5});
-   *     updateOptions({slowMouseSpeed: 2});`.
-   *
-   * Then the final options used will include both `fastMouseSpeed: 5` and
-   * `slowMouseSpeed: 2` with all other options being the default values.
-   * @param {!EdgeScrollOptions} options Object containing any or all of
-   *     the available options. Any properties not present will use the existing
-   *     value.
-   */
-  static updateOptions(options) {
-    ScrollBlockDragger.options_ = {...ScrollBlockDragger.options_, ...options};
-  }
-
-  /**
-   * Resets the options object to the default options.
-   */
-  static resetOptions() {
-    ScrollBlockDragger.options_ = defaultOptions;
-  }
-
-  /**
    * Updates the location of the block that is being dragged.
    * @param {number} deltaX Horizontal offset in pixel units.
    * @param {number} deltaY Vertical offset in pixel units.
@@ -345,11 +298,11 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
     const blockOverflows = this.getBlockBoundsOverflows_(viewMetrics, mouse);
     for (const direction of this.scrollDirections_) {
       const overflow = blockOverflows[direction];
-      if (overflow > ScrollBlockDragger.options_.slowBlockStartDistance) {
+      if (overflow > ScrollBlockDragger.options.slowBlockStartDistance) {
         const speed =
-            overflow > ScrollBlockDragger.options_.fastBlockStartDistance ?
-            ScrollBlockDragger.options_.fastBlockSpeed :
-            ScrollBlockDragger.options_.slowBlockSpeed;
+            overflow > ScrollBlockDragger.options.fastBlockStartDistance ?
+            ScrollBlockDragger.options.fastBlockSpeed :
+            ScrollBlockDragger.options.slowBlockSpeed;
         const scrollVector =
             this.SCROLL_DIRECTION_VECTORS_[direction].clone().scale(speed);
         candidateScrolls[direction].push(scrollVector);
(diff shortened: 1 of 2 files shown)


moniika (approved):

plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -96,53 +96,6 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
...
-   *
-   * Then the final options used will include both `fastMouseSpeed: 5` and
-   * `slowMouseSpeed: 2` with all other options being the default values.
-   * @param {!EdgeScrollOptions} options Object containing any or all of
-   *     the available options. Any properties not present will use the existing
-   *     value.
-   */
-  static updateOptions(options) {
-    ScrollBlockDragger.options_ = {...ScrollBlockDragger.options_, ...options};
-  }
```
cpcallen, line 137: Static _methods_ really ought to be supported by Babel, since they've been around since ES6 (2015).  Did you move this one (and `resetOptions`, below) out of the class body only because it references the `.options_` property and the linter was complaining about use-before-declaration?

plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -489,5 +442,50 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
...
+ * the edge is enabled.
+ * @type {boolean}
+ */
+ScrollBlockDragger.edgeScrollEnabled = true;
+
+/**
+ * Configuration options for the scroll-options settings.
+ * @type {!EdgeScrollOptions}
+ */
+ScrollBlockDragger.options = defaultOptions;
```
cpcallen, line 456: 1. I'm a little perplexed about your comment:

   > I had to make `options` public because it didn't like it being protected now that it's not on the class itself.

   Which piece of tooling complained about that?

0. If these properties were `@protected` previously then presumably we are expecting subclasses to make use of it—so is it OK that `options_` has been renamed?

0. Might it be better to make `options` an unexported module-global variable, and have `updateOptions` copy any properties that exist on both the `options` global and it's `options` argument from the latter to the former?

cpcallen (commented):
This has already gone in, but I thought I'd offer a couple comments anyway since you'd asked, @maribethb.

plugins/scroll-options/src/ScrollBlockDragger.js
BeksOmega: > If these properties were @protected previously then presumably we are expecting subclasses to make use of it—so is it OK that options_ has been renamed?

Since this plugin was just released, it's unlikely anyone had time to subclass it, so I don't think it's a problem.

plugins/scroll-options/src/ScrollBlockDragger.js
maribethb: I didn't try moving them separately. Babel just quit when it found the first error (e.g. it only errored on the first property instead of giving me every error in the file) so I didn't try it to see if the static methods would have been fine because I didn't know there was a difference. 

I didn't see any use-before-declaration errors / lint messages, and I am using `options` before declaring it even now. We have a different eslint config for samples though and it may just not be turned on here.

plugins/scroll-options/src/ScrollBlockDragger.js
maribethb: 1. It's the TypeScript language service. Technically, we don't really care because we don't publish the generated TypeScript typings. But there are some projects that use these plugins from TypeScript and I don't want them to have issues. And if we get to the point of converting plugins to TypeScript, I don't want to have issues. So I try to listen to the TS language service when possible. The error it gives me on the variable when I declare it and mark it `@protected` is `Property 'options' is protected and only accessible within class 'ScrollBlockDragger' and its subclasses.`

and when I use the property from within a class method I get 
`Property 'options' is protected and only accessible through an instance of class 'ScrollBlockDragger'. This is an instance of class 'typeof ScrollBlockDragger'.` That doesn't really make any sense to me and I can't find out why TypeScript would be upset with this, so it might be a case of just badly generated TypeScript typings (similar to the typedef issue I was having and is described in the PR) but again I didn't really want to ruin things for anyone who is using TypeScript.

2. Like Beka said, this is the first time we're publishing this plugin so the rename is okay.
3. That makes sense if we don't think `options` should be public, but then wouldn't subclasses not have access to it?

plugins/scroll-options/src/ScrollBlockDragger.js
cpcallen: > I didn't see any use-before-declaration errors / lint messages

Great.  In core it was being (in my view) excessively zealous about that.

plugins/scroll-options/src/ScrollBlockDragger.js
cpcallen: > 1\. The error it gives me on the variable when I declare it and mark it `@protected` is `Property 'options' is protected and only accessible within class 'ScrollBlockDragger' and its subclasses.`

It seems it's idea of `@protected` is a bit stricter than Closure Compiler's.  CC _has_ to allow this for ES5-style classes (since there is no `class` declaration) and—I'd thought—allows it for ES6-style ones too.  (What we'd really like here would be a combination `@package @protected`, but even CC doesn't know how to combine those two, sadly.)

> When I use the property from within a class method I get
> `Property 'options' is protected and only accessible through an instance of class 'ScrollBlockDragger'. This is an instance of class 'typeof ScrollBlockDragger'.` That doesn't really make any sense to me and I can't find out why TypeScript would be upset with this

Well, "this is an instance of class 'typeof ScrollBlockDragger'" is a long-winded way of saying "this is the ScrollBlockDragger constructor, not a ScrollBlockDragger instance", but why it would it would claim "Property 'options' is protected and only accessible through an instance of class 'ScrollBlockDragger'" I can't say: that claim seems patently false.

> 3\. That makes sense if we don't think `options` should be public, but then wouldn't subclasses not have access to it?

I think there are three possible approaches here:

1. Have a public (in this case `static`) `.options` property which can be directly set by external code.  Don't bother having a `setOptions` method at all.  This is the simplest approach.
0. Have an actually-`@protected` `.options_` property that subclasses can read from, and have a `setOptions` method on the base class only.  I think this is what you were aiming for.
0. Have options stored in unexported, module-local variables, and have the base class and each subclass provide a `setOptions` implementation that extracts the relevant part from its `options` argument and stores it said variables (after having called `super.setOptions(options)` so the base class can do the same).  This is pretty complex but is the 'safest', most-encapsulated approach.

I think that what I don't like about the code you've ended up with (and this is no critique of your design skills, since you obviously intend it to be this way) is that there is _both_ a public `.options` property _and_ a `.setOptions` method.  You lose the potential benefits of encapsulation (since why would I bother calling `.setOptions` if I can write directly to the `.options` object?) while still keeping much of the complexity.

maribethb: Yeah, I wanted option 2. In TypeScript it seems like you can have static properties defined directly on the class (so something close to the original syntax I had, but where `protected` is an actual keyword in TS instead of in the JsDoc) so maybe one day we can have that again.

I get what you're saying, so maybe we could do Option 1 and remove the `setOptions` method, but at the same time it would just be moving that logic up to the plugin level because I wanted the main entry point to configuring this to be at the plugin level rather than necessarily having to reach down into the `ScrollBlockDragger`. So I don't think it really removes that much complexity to remove the function, it just moves it to a different place.

I don't really like Option 3 though because I don't want to have to force that complexity onto the subclassers. If I understand correctly though, I guess they wouldn't have to deal with that unless they were specifically changing something about the auto-scroll functionality that uses the options, which I expect to be a minority of subclassers, so maybe it would be fine.

PR #861, opened 2021-08-12
Update package-locks for release

maribethb:
Update all the package-locks again, similar to https://github.com/google/blockly-samples/pull/859

This is going to be an ongoing issue with package-lock v2 because it doesn't get updated after lerna bumps the versions :/


BeksOmega (approved):
LGTM

PR #867, opened 2021-08-18
fix!: remove recycling logic from continous flyout

BeksOmega:
### Description

This is dependent on changes made by the new JSO serialization system.

Removes the recycling logic from the continuous toolbox since it has been moved in to core.

### Testing

~~Was not able to test whether this works correctly or not because it is dependent on changes in core, and linking is being uncooperative.~~

Manually tested this by editing one of the shadows of the color blend block, and then creating a variable to refresh the flyout. The edits to the shadow were maintained, signaling that recycling was occuring.

### API Change

Along with https://github.com/google/blockly/pull/5392 this removes the protected `createBlock_` and `clearOldBlocks_` functions from the public API

### DIFF
--- a/plugins/continuous-toolbox/package.json
+++ b/plugins/continuous-toolbox/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@blockly/continuous-toolbox",
-  "version": "1.0.20",
+  "version": "2.0.20",
   "description": "A Blockly plugin that adds a continous-scrolling style toolbox and flyout",
   "scripts": {
     "build": "blockly-scripts build",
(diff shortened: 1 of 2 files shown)


alschmiedt:
@BeksOmega is this good to be converted from a draft?

BeksOmega:
> @BeksOmega is this good to be converted from a draft?

Yep! All of my drafts should be gtg now.

alschmiedt (approved):
The change LGTM! Just a few things in the package need to be updated.

- The peerDependency should be updated to `^7.20211209.0` since recycling will no longer work when this package is used with any version below v7. 
- The major version should also be updated so that developers don't automatically get this change pulled in.

Also, if you could change the target branch to be master that would be great. Thanks!

PR #876, opened 2021-09-08
Use new location for `genUid` on `Blockly.utils.idGenerator`

cpcallen:
`genUid` is being moved from `Blockly.utils` to `Blockly.utils.idGenerator`.  Probe for the new location and, if found, stub it there instead of the old location.

See google/blockly#5441.


alschmiedt (approved):
Lgtm once the lint error is fixed.

PR #877, opened 2021-09-08
Update package locks

rachel-fenichel:
Similar to https://github.com/google/blockly-samples/pull/861

### DIFF
--- a/plugins/dev-tools/package-lock.json
+++ b/plugins/dev-tools/package-lock.json
@@ -6,7 +6,7 @@
 	"packages": {
 		"": {
 			"name": "@blockly/dev-tools",
-			"version": "2.5.5",
+			"version": "2.5.6",
 			"license": "Apache-2.0",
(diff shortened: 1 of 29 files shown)


plugins/dev-tools/package-lock.json
```diff
@@ -6,7 +6,7 @@
 	"packages": {
 		"": {
 			"name": "@blockly/dev-tools",
-			"version": "2.5.5",
```
cpcallen, line 9: Shouldn't this be a minor (rather than patch) version bump?

cpcallen (changes requested):

plugins/dev-tools/package-lock.json
cpcallen: OK, per discussion (see #861) this is just recording the current version of the package, as bumped by the previous release.

cpcallen (approved):

PR #879, opened 2021-09-10
fix: update webpack dev server version to fix loading

BeksOmega:
### Description

Fixes `npm run start` loading an initially blank page.

### Testing

Linked to the field-grid-dropdown package and ran `npm run start`. Observed that while it does kind of load a blank page, there are clear messages in the console telling you that the server is waiting for the bundle to finish. The page also continues loading while the bundling is happening, and automatically refreshes when it finishes.

Overall, I think this is a much better experience.


plugins/dev-scripts/config/webpackDevServer.config.js
```diff
@@ -9,25 +9,18 @@
...
 
-const fs = require('fs');
-const path = require('path');
-
-const appDirectory = fs.realpathSync(process.cwd());
-const resolveApp = (relativePath) => path.resolve(appDirectory, relativePath);
-
 module.exports = () => {
   return {
     port: 3000,
```
alschmiedt, line 14: I'm not entirely sure why this fix changes this, but I am getting an overlay when there are warnings instead of only when there are errors. 

I believe this can be fixed by adding the below to this config.
```
    client: {
      overlay: {
        errors: true,
        warnings: false,
      },
    },
```

alschmiedt (approved):

plugins/dev-scripts/config/webpackDevServer.config.js
BeksOmega: I added the things, but I wasn't able to get the linking working (can't remember how I was doing it before) so I wasn't able to test :/

alschmiedt: No worries, I can test on my end before merging : ) Thanks for adding that! 

But also in case you are curious, you should be able to test by running `npm run boot` from the root directory.

alschmiedt:
Ok I tested and it looks like it is working as expected. I think the only thing that is left is fixing the conflict in the dev-scripts/package-lock.json. Would you mind taking a look? If you don't have time I'm happy to fix it up : )

BeksOmega:
Sorry I should have noticed that and fixed it sooner! I think this is gtg :crossed_fingers:

alschmiedt:
No worries! Thanks for the fix!

plugins/dev-scripts/scripts/test.js
```diff
@@ -30,7 +30,7 @@ console.log(`Building tests for ${packageJson.name}`);
 const args = process.argv.slice(2);
 const skipLint = args.includes('--skip-lint');
 const config = webpackConfig({
-  mode: 'test',
+  mode: 'development',
```
maribethb, line 33: I think this change broke all the tests. When running `npm run test` you now get "0 tests passed" for all plugins. I think because of this line https://github.com/BeksOmega/blockly-samples/blob/04004b282c59ed65db0a14e8312303cd61c5d275/plugins/dev-scripts/config/webpack.config.js#L27

without `isTest` being true the entry points for the mocha files never get created.

However just changing this line back to `mode: 'test'` gives me additional errors like

```
ValidationError: Invalid options object. Ignore Plugin has been initialized using an options object that does not match the API schema.
```

that I don't fully understand. I am looking into this issue but I'm wondering if either Beka you could help look into this to fix it more quickly or if we should revert this PR until it's fixed? Because right now all PRs will report that all the tests pass, but in reality it's just not running any tests.

BeksOmega: Hmmm so I changed this because it seemed like `test` was no longer a valid mode, I thought `development` was the new name (see [the docs](https://webpack.js.org/configuration/mode/) and [the upgrade guide](https://webpack.js.org/migrate/5/#make-sure-to-use-mode)). But honestly it's a bit unclear.

It looks like what Abby is doing is working. But another idea is based on [this old github issue](https://github.com/webpack/webpack/issues/7659), maybe we actually need to be setting [optimization.nodeEnv](https://webpack.js.org/configuration/optimization/#optimizationnodeenv)?

alschmiedt: I'll have to double check with Sam, but AFAICT 'test' was never a webpack option (even if you go to [webpack 4](https://v4.webpack.js.org/configuration/mode/) that we were using before it does not mention 'test' as an option) , but instead something we set and [use internally](https://github.com/google/blockly-samples/blob/47d85d1e5c53a716692d3091ff3ab39422b020e7/plugins/dev-scripts/config/webpack.config.js#L27).

Again, I'll have to double check. Not pretending to be super confident in my webpack config knowledge here haha.

plugins/dev-scripts/config/webpack.config.js
```diff
@@ -85,6 +85,9 @@ module.exports = (env) => {
       },
       extensions: ['.ts', '.js']
           .filter((ext) => isTypescript || !ext.includes('ts')),
+      fallback: {
+        'util': false,
+      },
```
cpcallen, line 90: What was the purpose of this change, @BeksOmega?

BeksOmega: Guess this was causing problems when building for node, cuz this isn't included as a polyfill anymore. https://webpack.js.org/configuration/resolve/#resolvefallback

I was probably getting some error and this was the suggested solution on stack overflow, cuz I don't remember details.

Issue #727, opened 2021-05-21
how to configure the field slider options


pelikhan:
The field-slider README should contain an example on setting custom min/max (other options) values.

alschmiedt:
Thanks for filing! 

I talked with Monica and she suggested adding something like below under the Usage header.

"This field is an extension of the Blockly.FieldNumber field. See the [Blockly.FieldNumber documentation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/number#creation) on what parameters and configurations this field supports."

pelikhan:
Ohh..... so does field_slider replace the field_number field in place?

BeksOmega:
> Ohh..... so does field_slider replace the field_number field in place?

Not by default no. It's a subclass of field_number, so it has all of the same functionality, but it still has a separate constructor and json keyword.

If you do want it to act as a replacement you can unregister the field_number json keyword, and register the field_slider constructor in its place. This means that any blocks defined using json which currently use the field_number field will now use a field_slider field. But any blocks that have a javascript definition where they construct a `Blockly.FieldNumber` will still use the number definition.
```javascript
Blockly.fieldRegistry.unregister('field_number');
Blockly.fieldRegistry.register('field_number', FieldSlider);
```

```javascript
{
  "type": "math_number"
  "message0": "%1",
  "args0": [
    {
      "type": "field_number"  // This would now add a FieldSlider field to the block.
    }
  ]
}
```

```javascript
this.appendDummyInput()
     .appendField(new Blockly.FieldNumber()) // Still adds a FieldNumber to the block.
```

pelikhan:
This is very helpful info! Thanks.


Sent from Outlook<http://aka.ms/weboutlook>

________________________________

PR #900, opened 2021-10-01
add usage description

derwehr:
## Description
Adds a short usage description to the field-slider plugin's readme as suggested in #727


moniika (approved):
Thanks for the PR!

PR #921, opened 2021-10-19
fix: change 'Blocky' to 'Blockly'

Oddant1:


Oddant1:
@googlebot I signed it!

PR #930, opened 2021-11-18
Make FieldGridDropdown primaryColour and borderColour configurable

ajpal:
Allow users to customize the colors of the FieldGridDropdown, but still default to matching the color of the block if no colors are provided in the config

With custom colors provided in the config:
image

With no colors provided:
image


gonfunko:
Thanks for the PR! Could you add a quick note to the README documenting the new options?

gonfunko:
Thanks for the README update, looks good!

PR #931, opened 2021-11-22
fix: update to use non deprecated function

alschmiedt:

This involved: 

*Additional Information* 
I think CI isn't passing because the blockly version here does not mesh with the blockly version in dev-tools which is...fun.


plugins/keyboard-navigation/src/navigation_controller.js
```diff
@@ -851,7 +851,7 @@ export class NavigationController {
           return false;
         }
         this.navigation.moveCursorOnBlockDelete(workspace, sourceBlock);
-        Blockly.deleteBlock(sourceBlock);
+        this.navigation.deleteBlock(sourceBlock);
```
maribethb: instead of copying the deleteBlock function , you should be able to use `sourceBlock.checkAndDelete()`

well, that's the intention anyway, if that isn't working we should look into it

maribethb (approved):
not sure what's going on with CI. I think since you're using the beta version of dev-tools, that one is using the beta version of blockly, so that should be fine (and the CI would tell you if it weren't). the error is in a dependency and i don't understand the call stack so not sure what's going on there

plugins/keyboard-navigation/src/navigation_controller.js
alschmiedt: Thanks! This works perfectly.

alschmiedt:
I had to release a new beta version of Blockly because I needed a bug fix in it, so I don't think the version in dev-tools is compatible anymore. So locally at least it complains about the version number and I have to do `npm install --force` to get it to work.

alschmiedt:
Closing in favor of  #938

PR #932, opened 2021-11-23
feat: Configurable backpack open

Chappie74:
**Description**

Added functionality for configuring whether the backpack should open if empty. 

**Testing**

When instantiating the backpack, add  `allowEmptyBackpackOpen: false` as an option. This defaults to true, which keeps the default blockly behaviour.

### DIFF
--- a/plugins/workspace-backpack/src/backpack.js
+++ b/plugins/workspace-backpack/src/backpack.js
@@ -578,7 +578,8 @@ export class Backpack extends Blockly.DragTarget {
    * @protected
    */
   isOpenable_() {
-    return !this.isOpen();
+    return !this.isOpen() &&
+    this.options_.allowEmptyBackpackOpen ? true : this.getCount() > 0;
   }
 
   /**
--- a/plugins/workspace-backpack/src/options.js
+++ b/plugins/workspace-backpack/src/options.js
@@ -24,6 +24,7 @@ export let BackpackContextMenuOptions;
 
 /**
  * @typedef {{
+ *    allowEmptyBackpackOpen: (boolean|undefined),
  *    contextMenu:(!BackpackContextMenuOptions|undefined),
  * }}
  */
@@ -37,6 +38,7 @@ export let BackpackOptions;
  */
 export function parseOptions(options) {
   const defaultOptions = {
+    allowEmptyBackpackOpen: true,
     contextMenu: {
       emptyBackpack: true,
       removeFromBackpack: true,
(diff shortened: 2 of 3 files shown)


plugins/workspace-backpack/src/options.js
```diff
@@ -44,6 +45,7 @@ export function parseOptions(options) {
       copyAllToBackpack: false,
       pasteAllToBackpack: false,
       disablePreconditionChecks: false,
+      allowEmptyBackpackOpen: true,
```
maribethb: This should be a new top-level option instead of being nested under contextMenu options. The context menu options configures which actions are available in the context menu, and this option is unrelated.

plugins/nominal-connection-checker/src/type_hierarchy.js
```diff
@@ -497,16 +497,17 @@ export class TypeHierarchy {
...
+   * the paramters of to the common type.
    * @param {string} commonType The name of the common type we are getting the
-   *     params in the order of.
+   * params in the order of.
+   * @param getParamsForCommon
    * @param {function(!TypeDef, string, !Array<!TypeStructure>)
-   * :!Array<!TypeStructure>}
-   *     getParamsForCommon A function that takes in the definition of a type
-   *     and maps the given actual params to the parameters of the common type.
+:!Array<!TypeStructure>}
```
maribethb: Could you revert all the changes that aren't in the backpack plugin? They aren't related and some of the comments are incorrectly formatted - e.g. not starting the comment line with the asterisk can cause problems with jsdoc parsing.

plugins/workspace-backpack/src/backpack.js
```diff
@@ -578,7 +578,8 @@ export class Backpack extends Blockly.DragTarget {
    * @protected
    */
   isOpenable_() {
-    return !this.isOpen();
+    return !this.isOpen() &&
+    this.options_.contextMenu.allowEmptyBackpackOpen? true: this.getCount() > 0;
```
maribethb: Surprised the linter didn't complain about this, but could you add a space before the `?` and `:` please: https://google.github.io/styleguide/jsguide.html#:~:text=On%20both%20sides%20of%20any%20binary%20or%20ternary%20operator.

and then this line will probably need to be wrapped to stay under 80 chars

maribethb (commented):
Thank you for the PR and for contributing to Blockly! I think this just needs a bit of cleanup and then we should be good to go. Let me know if you have any questions!

Could you also add information about the new option in the README for this plugin?

Chappie74:
Thanks for the feedback. Will take a look and fix the issues.

plugins/nominal-connection-checker/src/type_hierarchy.js
Chappie74: Yeah I was wondering about all those changes. I couldn't get the project to run due to linting errors. And I ran the lint:fix command, which changed a lot of files.

alschmiedt (approved):

alschmiedt:
Just looked it over and the changes look good to me.

Thank you for your contribution!

PR #933, opened 2021-11-23
fix: don't scrub values of 0 in the "Custom Generator" codelab

GregoryCannon:
Cast `math_number` blocks as a string to avoid them being scrubbed later on in the code generation process.


GregoryCannon:
Accidentally was branched off another PR and `git rebase` was being difficult so I'm gonna just redo it.

PR #934, opened 2021-11-23
fix: handle value of zero in the custom generator codelab

GregoryCannon:
Cast numbers as strings so they're not scrubbed later in the code generation process.


maribethb (approved):

alschmiedt:
It looks like just updating the title doesn't update the actual commit so it is still failing. I am just going to do it before merging in. I think in the future this would be fixed by changing the commit locally by doing
```
git commit --amend -m "New message"
```

And then git force pushing to this branch...which is never my favorite thing.

PR #938, opened 2021-12-01
fix: keyboard navigation to use new Blockly APIs

alschmiedt:

This involved: 

*Additional Information* 
I think CI isn't passing because the blockly version here does not mesh with the blockly version in dev-tools which is...fun.


alschmiedt:
Just a heads up @maribethb this is a dupe of #931 so you have already reviewed this.

PR #942, opened 2021-12-02
fix: updates keyboard nav to work with the 2021 dec release

alschmiedt:

This involved: 

*Additional Information*


rachel-fenichel:
`checkAndDelete` is marked package in core, so we'll also need to make it public if we want to use it here. Either change it in core or file an issue to discuss if you're not sure if it should be public.

rachel-fenichel (approved):

PR #947, opened 2021-12-03
chore: update major versions for packages that rely on dec 2021 release

alschmiedt:
These should have been done in previous PRs, but I forgot so here we are. 

- Updates major version for block-plus-minus since it relies on the most recent release of Blockly
- Updates the minor version of keyboard navigation. This plugin is still in beta which is why it has a 0 as the major version. I think updating the minor version here is the right call, but not entirely sure.


rachel-fenichel (approved):
+1 to updating the minor version of keyboard nav. Eventually we'll get these updates automated based on commit messages.

Issue #917, opened 2021-10-15
Update eslint-related packages


maribethb:
**Category**

- Tooling

**Component**

eslint-config

**Describe the bug**

* blockly-samples/plugins/eslint-config uses eslint v7
* It also uses `@typescript-eslint/parser`, `@typescript-eslint/eslint-plugin`, and `jsdoc-eslint-plugin` versions that are several years old and are not compatible with eslint v7
* This problem is exposed if you have to use the `--legacy-peer-deps` flag for any reason (e.g. install blockly beta with a plugin that hasn't been updated yet) where you get an error message about eslint not being installed.

**To Reproduce**

Use the `--legacy-peer-deps` flag and eslint will not be installed. 

**Additional context**

See https://groups.google.com/g/blockly/c/MpMkF-hgloA for initial context.

Currently, I've published beta versions of eslint-config (2.1.5-beta.0), dev-scripts (1.2.11-beta.0), and dev-tools (2.6.2-beta.0) that use the newest version of each of the dependencies listed above. However there is more additional work. Over time some of the rules have changed, been added/removed/renamed/etc so there are new lint errors and warnings.

- [ ] Investigate and fix missing rule names
- [ ] Resolve the additional lint errors/warnings (either by fixing actual style problems, or updating/disabling the new rules)
- [x] Release non-beta versions of eslint-config, dev-scripts, and dev-tools that use the new versions

While dev-tools has some changes that should remain part of the beta release (e.g. all changes related to the q4 release of Blockly) this problem is unrelated to the core beta release and can be fixed ahead of time.

maribethb:
Current status: New versions of the eslint-related packages were used in order to solve some peerDependency issues, but we didn't sort out the newly introduced warnings. So this still needs some work to decide what to do with those warnings. I would have fixed errors but it looks like after the update there were only warnings and no errors.

Run `npm run lint` in samples to see warnings. I believe other than in nominal-connection-checker, previously the only warnings were about TODO comments and now there are more. Ideally we would have no lint messages

BeksOmega:
Nominal connection checker can be deleted whenever. I talked w/ Rachel and we decided it would be better to pull it into a separate repo. I can put up a PR for removing it if you want =)

maribethb:
Remaining lint errors/warnings were fixed in q1 2022.

PR #950, opened 2021-12-10
release: update package versions in prep for release

maribethb:
Update plugin versions to be compatible with latest blockly release.

Review checklist (please help me make sure I've followed this and updated each plugin):

- [ ] Each plugin (unless listed below) should list devDependencies: `blockly: ^7.20211209.0` and `@blockly/dev-tools: ^3.0.0`
- [ ] Each plugin should have a peerDependency that includes blockly `7.20211209.0` as part of the range. If no changes were required to make the plugin compatible with the new release, it gets added to the end of the range. If changes were required, a new range is started so v7 is at the bottom (range can be specified with `^` but **must not** be specified with `>`)
- [ ] If the plugin is now **only** compatible with v7 and higher (i.e. second scenario above i.e. changes were required in response to api changes in blockly, so now v7 is at the bottom of the peerDependency range) then the plugin must have a new **major** version listed
- [ ] No plugin depends on a beta version of either blockly or dev-tools in any type of dependency

Exceptions:
block-extension-tooltip is not compatible with the new version of blockly because of how it monkey patches blockly. This plugin needs some additional work that we haven't done yet. Therefore it is explicitly not listed as compatible with the newest version of dev-tools or blockly. I also tried to "pin" their versions (specified an exact version instead of a range) hoping that lerna won't update it, but i don't know if that will work or not. I also added a warning to its README saying it's not compatible with the new version of blockly.

Haven't touched demos or examples yet.

Preview these plugins at https://maribethb.github.io/blockly-samples/

------

Additional changes in this PR:

- Merged master into release branch (needed to pick up some changes to mocha tests)
- Updated GH Actions workflow to install npm@latest which is needed to avoid this issue: https://github.com/npm/cli/issues/2633
- Updated eslint versions to fix this issue: https://github.com/google/blockly-samples/issues/917
- Updates typed-variable-modal and tests to use `Blockly.dialog.alert` instead of `Blockly.alert`. This was blocking due to the way it was mocked causing tests to fail. I wanted to do this in a separate PR but I couldn't because it depends on the new version of blockly (which is done in this PR) and I couldn't submit this without fixing the tests. Had to do it all at once.


plugins/typed-variable-modal/package.json
```diff
@@ -41,15 +41,15 @@
...
+    "@blockly/dev-tools": "^3.0.0",
+    "blockly": "^7.20211209.0",
     "jsdom": "^19.0.0",
     "jsdom-global": "3.0.2",
     "mocha": "^7.1.0",
     "sinon": "7.5.0"
   },
   "peerDependencies": {
-    "blockly": "3.20200625.0 - 6"
+    "blockly": "3.20200625.0 - 7"
```
alschmiedt: Should this just be `7.20211209.0` since we had to update to use `Blockly.dialog.alert`?

plugins/keyboard-navigation/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@blockly/keyboard-navigation",
-  "version": "0.2.0",
+  "version": "0.3.0",
```
alschmiedt: I think this was already updated in a previous PR from 0.1 to 0.2, so I think we can just keep it at 0.2.0

alschmiedt (approved):

plugins/keyboard-navigation/package.json
maribethb: Done

plugins/typed-variable-modal/package.json
maribethb: Good catch, therefore gave this one a new major version as well. And updated the dependency on plugin-modal to be the newest version so that the peerDependency on blockly is satisfied on both.

PR #963, opened 2021-12-23
chore: remove unregisters from themes

alschmiedt:
This fixes the warnings related to themes we are seeing in the advanced playground when it is run with the most recent release of Blockly.

### DIFF
--- a/plugins/theme-dark/package.json
+++ b/plugins/theme-dark/package.json
@@ -45,7 +45,7 @@
     "blockly": "^7.20211209.0"
   },
   "peerDependencies": {
-    "blockly": "3.20200123.0 - 7"
+    "blockly": "^7.20211209.0"
   },
   "publishConfig": {
     "access": "public",
--- a/plugins/theme-dark/package.json
+++ b/plugins/theme-dark/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@blockly/theme-dark",
-  "version": "2.0.7",
+  "version": "3.0.0",
   "description": "A Blockly dark theme.",
   "scripts": {
     "audit:fix": "blockly-scripts auditFix",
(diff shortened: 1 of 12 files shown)


plugins/theme-dark/package.json
```diff
@@ -45,7 +45,7 @@
     "blockly": "^7.20211209.0"
   },
   "peerDependencies": {
-    "blockly": "3.20200123.0 - 7"
+    "blockly": "^7.20211209.0"
```
maribethb, line 48: I think whenever we make a plugin only compatible with the new version, it needs to get a major version bump... I can't find where we've written this down but I think that's the convention we follow, so that people using plugins don't automatically get upgraded to a version they can't use.

Alternatively I thought we added a mechanism to check if something is registered? If so then we could call that and only unregister if needed, which might be safer anyway? i.e. to protect against someone registering it twice or registering two custom themes with the same name.

It would be nice if `unregister` actually threw an error instead of logging to the console, that way we could catch and ignore it here, but it doesn't...

Anyway, I'm fine with leaving it the way you have it but I think this would be a breaking change if you make it non-compatible with previous versions.

maribethb (approved):

plugins/theme-dark/package.json
alschmiedt: I went back and forth with adding a check to see if it was registered. I decided against it since I figured we would want to remove it once we updated the theme to a new major version - since it would no longer be necessary. However, if we are ok with leaving it in long term then I am happy to add it back.

maribethb: Yeah, I could see it going either way. I guess if you check if it's registered and then unregister it, then in the cases I listed above e.g. registering two custom themes with the same name, we're actually hiding something that should be an error. So then your approach is good and we just need to do a major version bump on the theme itself.

PR #973, opened 2022-01-19
chore: fix lint errors in blockly-samples

### LINKED ISSUE #630: Fix eslint errors in plugins
We'd like to be able to include eslint as part of checks before merging, but first we need to clean up pre-existing eslint errors.

### LINKED ISSUE #650: Update eslint to allow long `import` lines.
The eslint config doesn't allow `import` lines to be longer than 80. It does allow `goog.require` statements to be longer. The Google style guide doesn't address this issue, but clang-format formats `import` lines to stay on one line, which can exceed 80 characters.

We should update the eslint config to ignore the max-length rule on `import` lines. See "ignore pattern" here: https://eslint.org/docs/rules/max-len

maribethb:
After I updated the eslint and eslint-plugin-jsdoc packages, there were some new lint errors and some caused by rule names having changed, etc. My end goal is for there to be no lint errors in blockly-samples except for in nominal-connection-checker which will be removed soon.

Fixes #630 
Fixes #650 

- Update some rules which changed name or format (e.g. the camel-case rules changed formats slightly)
- Update to an even newer version of eslint-plugin-jsdoc which is supposed to have a fix for `@suppress` annotations. However despite asking for the new version in the eslint config, and deleting all the package locks and node_modules and reinstalling, I'm still getting the old version :/ I think this might be because of the way the packages are being installed/linked via lerna and I'm hoping that once eslint-config is published then the other packages will know what to do and use the right version. So there may or may not still be one error that says "visibility is not defined" but hopefully that will go away on its own when the versions are sorted out.
- Turned off the rule that warns about TODO comments since the style guide allows them and we use them. we could decide to add some phrases it will warn about such as "DO NOT SUBMIT" if we wanted to but I haven't
- Add an exception to the line length rules that won't warn about long import statements or URLs
- Turned off the rule that requires periods at the end of sentences. We may or may not actually want to keep that, but right now it's broken for long record type params so I just disabled it, and will file a bug with eslint-plugin-jsdoc
- Adjusted the typescript-specific rules for ts code. The style guide says if the types are specified in ts then we shouldn't also put them into JsDoc so I adjusted those rules and removed the jsdoc types.
- Removed trailing commas from record types, which have always been incorrect but now cause an eslint error
- Fix some syntax errors related to specifying arrays in jsdoc
- Add some disable-next-line comments for cases where we are already breaking the rule but where we don't want to generally allow that going forward (ts/no-explicit-any)
- Added `monaco` and `dat` to a global comment so that eslint believes they have been defined elsewhere. I looked into doing this "properly" but because we load the packages using custom loaders I don't know how else to tell eslint/jsdoc that the types really are defined so settled for this method to silence the warnings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,9 +15,15 @@ build/
 gh-pages/_site
 gh-pages/plugins
 gh-pages/examples
+.DS_Store
 
 # TypeScript cache
 *.tsbuildinfo
 
 # Optional npm cache directory
 .npm
+
+# IDEs and editors
+.idea
+*.sublime-workspace
+.vscode/*
--- a/codelabs/context_menu_option/context_menu_option.md
+++ b/codelabs/context_menu_option/context_menu_option.md
@@ -262,14 +262,12 @@ As an example, update the help item's `callback` to add a block to the workspace
 
 ```js
 callback: function(scope) {
-  const domText = Blockly.Xml.textToDom(`
-  <xml xmlns="https://developers.google.com/blockly/xml">
-    <block type="text">
-      <field name="TEXT">Now there is a block</field>
-    </block>
-    </xml>
-  `);
-  Blockly.Xml.domToWorkspace(domText, scope.workspace);
+  Blockly.serialization.blocks.append({
+    'type': 'text',
+    'fields': {
+      'TEXT': 'Now there is a block'
+    }
+  });
 }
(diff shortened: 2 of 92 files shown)


maribethb:
Re-running CI to make sure tests pass now that #974 is merged.

PR #974, opened 2022-01-20
fix: the test configuration

alschmiedt:
Addresses the first half of [this comment](https://github.com/google/blockly-samples/pull/879/commits/04004b282c59ed65db0a14e8312303cd61c5d275#r788125510).

The second half I believe can be addressed with an npm install. We updated webpack versions so if we don't install then we are using v5 config with v4 of webpack.

### DIFF
--- a/plugins/dev-scripts/scripts/test.js
+++ b/plugins/dev-scripts/scripts/test.js
@@ -30,7 +30,7 @@ console.log(`Building tests for ${packageJson.name}`);
 const args = process.argv.slice(2);
 const skipLint = args.includes('--skip-lint');
 const config = webpackConfig({
-  mode: 'development',
+  mode: 'test',
   skipLint: skipLint,
 });
 if (!config.entry) {
--- a/plugins/dev-scripts/config/webpack.config.js
+++ b/plugins/dev-scripts/config/webpack.config.js
@@ -158,7 +158,9 @@ module.exports = (env) => {
       }),
       // canvas should only be required by jsdom if the 'canvas' package is
       // installed in package.json. Ignoring canvas require errors.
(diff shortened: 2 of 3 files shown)


plugins/dev-scripts/scripts/test.js
```diff
@@ -30,7 +30,7 @@ console.log(`Building tests for ${packageJson.name}`);
 const args = process.argv.slice(2);
 const skipLint = args.includes('--skip-lint');
 const config = webpackConfig({
-  mode: 'development',
+  mode: 'test',
   skipLint: skipLint,
 });
```
BeksOmega, line 35: Sorry you had to dig into this :/ I should have taken the time to look into it more.

But now I kind of understand, so webpackConfig doesn't actually take in a webpack config. It takes in a custom config, that we use to configure the webpack config. Is there somewhere that info could be put? Eg a JSDoc comment on the webpackConfig definition. Obviously I should have dug deeper into this before changing it willy-nilly, but it would also be nice if it didn't require deeper digging.

alschmiedt: No worries at all. You looked into a lot more webpack config stuff so that I didn't have to so super appreciative : )  

And yeah I completely agree. I'll take a look and see if I can add more helpful documentation.

alschmiedt:
Ok so I believe this one is ready to review and I will create another PR for updating documentation. I might also see if there is any way we can fail when we do not run tests when there are tests so we don't end up in the same state where all of our tests are just silently failing. But not entirely sure how to do that just yet.

maribethb (approved):
Thanks, Abby!

PR #981, opened 2022-01-21
Revert "Publish"

alschmiedt:
This reverts commit 7649b47cbb366115604d5f3c8e4aa8b578d7121d.

The release script failed before it published to npm but not before it added a bunch of tags to github and added this commit. Reverting this will hopefully mean that npm and git are now in sync.

### DIFF
--- a/plugins/block-dynamic-connection/package-lock.json
+++ b/plugins/block-dynamic-connection/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "@blockly/block-dynamic-connection",
-	"version": "0.1.18",
+	"version": "0.1.17",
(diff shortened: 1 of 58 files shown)


maribethb (approved):
Will this revert the tags or do you have to delete those separately? Alternatively, you could do

lerna publish from-git 

Which would use the tags it just uploaded to GitHub instead of making new ones. But up to you, not sure which is less headache tbh.

alschmiedt:
Yeah I already removed all the tags.

I wasn't aware of the `lerna publish from-git` method so I will have to try that next time because deleting that many tags was really not fun.

Issue #899, opened 2021-10-01
block-extension-tooltip not rendering tooltip correctly


moniika:
**Category**
Plugins

**Component**
 block-extension-tooltip

**Describe the bug**
Tooltip not rendering correctly when depending on Blockly beta (post `goog.module` migration).

Expected:
![image](https://user-images.githubusercontent.com/6621618/135675255-2cb2bb43-6234-4656-86b7-8989f9d6e106.png)

Actual:
![image](https://user-images.githubusercontent.com/6621618/135675265-9db335fc-c9e2-4111-8c03-6d44872129c4.png)

**To Reproduce**

1. `npm install blockly@beta`
2. Run `blocky-extension-tooltip` with `npm run start`

BeksOmega:
The monkey patch is touching some [private properties](https://github.com/google/blockly-samples/blob/21e3e227a16a171f95f93b6cc1947514b9764287/plugins/block-extension-tooltip/src/tooltip_monkey_patch.ts#L45) :/

maribethb:
I made this plugin private in #983 so that it doesn't get published again until we sort this out. So when we do fix this, we need to remember to remove that from the package.json. Also, in `src/index.ts` we need to import from `blockly/core` instead of `blockly`.

alschmiedt:
@maribethb can this be closed, or are you waiting until you implement your plan for deprecating?

maribethb:
Remaining work: deprecate and remove the plugin

PR #982, opened 2022-01-22
fix!: Return a non-zero error code on errors in build, lint

### LINKED ISSUE #816: Lint errors should cause github actions to fail
**Category**

Plugins
**Component**
dev-scripts

In our last release of blockly samples we found out that lint errors were causing our packages to be published without a dist folder. 

When someone submits a PR with lint errors the errors are displayed in the file diff, but ci does not fail. We would rather ci fail so we can keep linting errors out of master.

**Additional context**
See #818 for more information on why this is necessary.

samelhusseini:
Return a non-zero error code on errors in the build and lint scripts.

Fixes #816

Additions by maribethb:
- Remove the "example" directory under workspace-search plugin. The reason I did that during this PR was because its build script was being run by lerna leading to a lot of extraneous output, and because if those examples ever fail to build then it could block PR submission/publishing, which would be unfortunate. These examples are severely out of date and generally not useful since the plugin's test directory serves as an example itself.
- Make lerna show all lint messages even if there is an error
    - First, by waiting for the results of both src and test before exiting (if there are errors in src, you will still see messages in test)
    - Second, by adding "--no-bail" argument to top-level lint script in package.json. This mimics the previous behavior (which never errored therefore never exited early) where CI will show all messages at once but you may have to scroll through some output in order to see the messages. Now lerna will at least tell you there was an error but unfortunately it will not tell you which package it came from (https://github.com/lerna/lerna/issues/1997)
- Silence webpack warnings from source-map-loader so the build, test, and start output is not littered with spurious warnings about the closure sourcemaps not being found. This also removes those same warnings from the console when running a plugin.
- Fixed a line continuation lint error
- Reverted the change that bumped the blockly version used in block-extension-tooltip


plugins/block-extension-tooltip/package.json
```diff
@@ -42,7 +42,7 @@
   "devDependencies": {
     "@blockly/dev-scripts": "^1.2.12",
     "@blockly/dev-tools": "2.6.1",
-    "blockly": "5.20210325.1",
+    "blockly": "^7.20211209.0",
```
alschmiedt: I don't think this can be updated yet because of [this issue](https://github.com/google/blockly-samples/issues/899).

plugins/dev-scripts/scripts/lint.js
```diff
@@ -52,22 +52,33 @@ async function lintDir(dir, linter) {
     if (shouldFix) {
       await ESLint.outputFixes(results);
     }
-    return results;
+    return [results, ESLint.getErrorResults(results)];
```
alschmiedt, line 57: I think the return type for this needs to be updated.

alschmiedt (approved):
lgtm, I think the build and lint files might benefit from some more comments/more information in the fileoverview on what is exactly happening. However, all of dev-tools needs a bit more documentation so I am fine if we want to work on that in a follow up PR.

maribethb:
Added a few additional comments but yes agreed that dev-tools as a whole could use some attention and probably best done separately

PR #983, opened 2022-01-25
fix: Make block-extension-tooltip build successfully

### LINKED ISSUE #805: block-extension-tooltip build error
**Category**
Plugin

**Component**
block-extension-tooltip

**Describe the bug**
There is an error when building the production version of block-extension-tooltip.

**To Reproduce**

1. In the `block-extension-tooltip` directory, run `npm run build`
```
ERROR in /usr/local/google/home/kozbial/github/blockly/samples-moniika/plugins/block-extension-tooltip/test/index.ts
ERROR in /usr/local/google/home/kozbial/github/blockly/samples-moniika/plugins/block-extension-tooltip/test/index.ts(79,53):
TS2345: Argument of type '(blocklyDiv: HTMLElement, options: BlocklyOptions) => WorkspaceSvg' is not assignable to parameter of type '(blocklyDiv: HTMLElement, options: BlocklyOptions) => Workspace'.
  Types of parameters 'options' and 'options' are incompatible.
    Type 'import("/usr/local/google/home/kozbial/github/blockly/samples-moniika/plugins/dev-tools/node_modules/blockly/blockly").BlocklyOptions' is not assignable to type 'import("/usr/local/google/home/kozbial/github/blockly/samples-moniika/plugins/block-extension-tooltip/node_modules/blockly/blockly").BlocklyOptions'.
      Types of property 'parentWorkspace' are incompatible.
        Type 'WorkspaceSvg' is missing the following properties from type 'WorkspaceSvg': getMetricsManager, setMetricsManager, getCachedParentSvgSize, setCachedParentSvgSize, and 2 more.
```

**Expected behavior**
No build error.

**Additional context**
May be related (or possibly the cause of) the issue with the released version of the plugin missing it's compressed files (tracked in https://github.com/google/blockly-samples/issues/793)

maribethb:
Fixes #805 

Fixes two different build errors; see issue for more details.

Before: `npm run build` failed in `block-extension-tooltip`
Now: `npm run build` succeeds (with warnings, sorry)

### DIFF
--- a/plugins/block-extension-tooltip/src/index.ts
+++ b/plugins/block-extension-tooltip/src/index.ts
@@ -8,7 +8,7 @@
  * @fileoverview Custom tooltip block extension.
  * @author samelh@google.com (Sam El-Husseini)
  */
-import * as Blockly from 'blockly/core';
+import * as Blockly from 'blockly';
 import './tooltip_monkey_patch';
 
 type TooltipRender = (block: Blockly.BlockSvg) => HTMLElement;
--- a/plugins/block-extension-tooltip/src/index.ts
+++ b/plugins/block-extension-tooltip/src/index.ts
@@ -8,6 +8,10 @@
  * @fileoverview Custom tooltip block extension.
  * @author samelh@google.com (Sam El-Husseini)
  */
+
+// TODO(maribethb): This should be from 'blockly/core'; fix asap when this
+// plugin is made compatible with v7 of Blockly.
+// See https://github.com/google/blockly-samples/issues/805
 import * as Blockly from 'blockly';
 import './tooltip_monkey_patch';
 
--- a/plugins/block-extension-tooltip/package.json
+++ b/plugins/block-extension-tooltip/package.json
@@ -12,6 +12,7 @@
     "prepublishOnly": "npm run clean && npm run dist",
     "start": "blockly-scripts start"
   },
+  "private": true,
   "main": "./dist/index.js",
   "module": "./src/index.js",
   "unpkg": "./dist/index.js",
--- a/plugins/block-extension-tooltip/tsconfig.json
+++ b/plugins/block-extension-tooltip/tsconfig.json
@@ -4,7 +4,11 @@
     "target": "es5",
     "allowJs": true,
     "sourceMap": true,
-    "lib": ["es6", "dom"]
+    "lib": ["es6", "dom"],
+    // Tell tsc to use the version of blockly local to this plugin.
+    // See https://github.com/google/blockly-samples/issues/805
+    "baseUrl": ".",
+    "paths": {"blockly": ["node_modules/blockly"]}
   },
   "include": [
     "src", "test"


plugins/block-extension-tooltip/src/index.ts
```diff
@@ -8,7 +8,7 @@
  * @fileoverview Custom tooltip block extension.
  * @author samelh@google.com (Sam El-Husseini)
  */
-import * as Blockly from 'blockly/core';
```
samelhusseini, line 11: Importing the entire `blockly` module would mean including the JS generator and built-in blocks when a user may not need them.
This package doesn't depend on the JS generator bring there, hence why it was only importing `blockly/core`

samelhusseini (approved):
LGTM, but can we replace "blockly" in paths by "blockly/core" and leave importing as is.

plugins/block-extension-tooltip/src/index.ts
maribethb: Yeah, it's not ideal. But this is inexplicably causing one of the errors. If I change both the src and test files to use `blockly/core` then I get this error:

```
ERROR in /Users/maribethb/github/maribethb/blockly-samples/plugins/block-extension-tooltip/test/index.ts(79,53):
TS2345: Argument of type '(blocklyDiv: HTMLElement, options: import("/Users/maribethb/github/maribethb/blockly-samples/plugins/block-extension-tooltip/node_modules/blockly/blockly").BlocklyOptions) => import("/Users/maribethb/github/maribethb/blockly-samples/plugins/block-extension-tooltip/node_modules/blockly/blockly").WorkspaceSvg' is not assignable to parameter of type '(blocklyDiv: HTMLElement, options: import("blockly_options").BlocklyOptions) => import("workspace_svg").WorkspaceSvg'.
  Type 'WorkspaceSvg__Class' is missing the following properties from type 'WorkspaceSvg': metricsManager_, componentManager_, blockDragSurface_, workspaceDragSurface_, and 47 more.
```

which is very similar to the error that you get if you leave them how they are (`blockly/core` in src and `blockly` in test) but just on a different line. I'm not sure why this occurs. dev-tools is already using `blockly/core` so it's not just purely that they don't match. Also the .d.ts file in dev-tools says the type is `Workspace` rather than `WorkspaceSvg` so I tried updating that but it doesn't fix this error.

So, yeah... I don't know why this only works if you import from `blockly` instead of `blockly/core` but empirically that's what I'm seeing. It's not ideal that it bloats the module size but it's better than not building at all.

I'm going to work on fixing the plugin so that is is compatible with v7 soon, so this should be temporary (also no idea why this error goes away if the versions of blockly match...). I can add a todo comment with a link back to the issue to remind me to change this back once we don't need it.

maribethb: Oh sorry, I just re-read your comment below and understood what you meant. Unfortunately the "paths" change only fixes the "conflicting files" error and doesn't have any effect on this error. I did just test it to see if that would work but no such luck.

alschmiedt (approved):

PR #988, opened 2022-02-09
fix: override old blocks in dynamic connections

### LINKED ISSUE #844: `overrideOldBlockDefinitions` in blocks-dynamic-connection breaks existing projects
**Category**

 - Plugins

**Component**

blocks-dynamic-connection

**Describe the bug**

If you call `overrideOldBlockDefinitions` for the blocks-dynamic-connections plugin It will not properly load existing saves, even though the [documentation](https://github.com/google/blockly-samples/tree/master/plugins/block-dynamic-connection) says that it will.

The issue is that the mutators for the blocks-dynamic-connections plugin are different than those for the equivalent blocks in core:
  * [controls_if](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/blocks/logic.js#L309), [dynamic_if](https://github.com/google/blockly-samples/blob/832f3f51db2d291dd05df1bae113ba7ff3cc740d/plugins/block-dynamic-connection/src/dynamic_if.js#L52)
  * [lists_create_with](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/blocks/lists.js#L137), [dynamic_list_create](https://github.com/google/blockly-samples/blob/832f3f51db2d291dd05df1bae113ba7ff3cc740d/plugins/block-dynamic-connection/src/dynamic_list_create.js#L46)
  * [text_join](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/blocks/text.js#L684), [dynamic_text_join](https://github.com/google/blockly-samples/blob/832f3f51db2d291dd05df1bae113ba7ff3cc740d/plugins/block-dynamic-connection/src/dynamic_text_join.js#L46)

**To Reproduce**

1. Run `npm run start` from the blocks-dynamic-connection directory.
2. Paste the following XML into the XML tab, and import it.
```
<xml xmlns="https://developers.google.com/blockly/xml">
  <block type="controls_if" id="OE?Ny/Oa,2R%=M55EEBu" x="187" y="112">
    <mutation elseif="1" else="1"></mutation>
    <value name="IF0">
      <block type="logic_boolean" id="9+TYl*=-y%y4q|TPSU3y">
        <field name="BOOL">TRUE</field>
      </block>
    </value>
    <statement name="DO0">
      <block type="text_print" id="H*h8pu6:dN68a#[Q}bJ,">
        <value name="TEXT">
          <shadow type="text" id="pBKz}FjbfcNs?sLRf$DQ">
            <field name="TEXT">first</field>
          </shadow>
        </value>
      </block>
    </statement>
    <value name="IF1">
      <block type="logic_boolean" id="p5L3)9^oT:-TpOP3?=Ks">
        <field name="BOOL">TRUE</field>
      </block>
    </value>
    <statement name="DO1">
      <block type="text_print" id="%J@Ln}Fcao5tEedPHHFE">
        <value name="TEXT">
          <shadow type="text" id="qWm/P]n#!h%XxWt*:wwC">
            <field name="TEXT">secod</field>
          </shadow>
        </value>
      </block>
    </statement>
    <statement name="ELSE">
      <block type="text_print" id="uRW(~uiyHwyWBCF#]HQ!">
        <value name="TEXT">
          <shadow type="text" id="13oCd+(pjXg5=/ejA`uN">
            <field name="TEXT">third</field>
          </shadow>
        </value>
      </block>
    </statement>
  </block>
</xml>
```
3. Observe how the block only has a single statement input, when it should have three.

**Additional problem**

There is a secondary issue, which is that when these blocks get reserialized, they are saved as `controls_if`, `list_create_with`, or `text_join` blocks, but they have the mutation of their dynamic counter parts:
```
<xml xmlns="https://developers.google.com/blockly/xml">
  <block type="controls_if" id="RAPyQ#S1$@r$UUpIrX~Y" x="88" y="188">
    <mutation inputs="0" else="false" next="1"></mutation>
  </block>
</xml>
```

This makes them incompatible with the built-in blocks. It also means we cannot remove the `overrideOldBlockDefinitions` function without breaking those saves.

**Recommended solution**
All of the dynamic blocks should be updated to properly deserialize the mutations of their built-in counter parts. I think that ideally they would also serialize to the same format, but that may not be possible/worth it.

For now we should add documentation not to use the `overrideOldBlockDefinitions` function.

BeksOmega:
### Description

Closes #844

Adds support for deserializing core's standard XML format for the controls_if, lists_create_with, and text_join blocks.

### Tests

Adds tests for deserializing the standard XML format, and maintaining the minimum inputs on list_create_with blocks and text_join blocks.

Also tested these things manually.

### Additional Info

Updates the readme with now-accurate information.

### DIFF
--- a/plugins/block-dynamic-connection/src/dynamic_text_join.js
+++ b/plugins/block-dynamic-connection/src/dynamic_text_join.js
@@ -91,7 +91,7 @@ Blockly.Blocks['dynamic_text_join'] = {
     const itemCount = Math.max(
         parseInt(xmlElement.getAttribute('items'), 10), this.minInputs);
     // Two inputs are added automatically.
-    for (let i = 2; i < itemCount; i++) {
+    for (let i = this.minInputs; i < itemCount; i++) {
       this.appendValueInput('ADD' + i);
     }
     this.inputCounter = itemCount;
--- a/plugins/block-dynamic-connection/src/dynamic_text_join.js
+++ b/plugins/block-dynamic-connection/src/dynamic_text_join.js
@@ -57,6 +57,14 @@ Blockly.Blocks['dynamic_text_join'] = {
    * @this {Blockly.Block}
    */
   domToMutation: function(xmlElement) {
+    if (xmlElement.getAttribute('inputs')) {
+      this.deserializeInputs_(xmlElement);
+    } else {
+      this.deserializeCounts_(xmlElement);
+    }
+  },
+
+  deserializeInputs_: function(xmlElement) {
     const items = xmlElement.getAttribute('inputs');
     if (items) {
       const inputNames = items.split(',');
@@ -67,6 +75,16 @@ Blockly.Blocks['dynamic_text_join'] = {
     }
     const next = parseInt(xmlElement.getAttribute('next'));
     this.inputCounter = next;
+  },
+
+  deserializeCounts_: function(xmlElement) {
+    const itemCount = Math.max(
+        parseInt(xmlElement.getAttribute('items'), 10), this.minInputs);
+    // Two inputs are added automatically.
+    for (let i = 2; i < itemCount; i++) {
+      this.appendValueInput('ADD' + i);
+    }
+    this.inputCounter = itemCount + 1;
   },
 
   /**
--- a/plugins/block-dynamic-connection/src/dynamic_text_join.js
+++ b/plugins/block-dynamic-connection/src/dynamic_text_join.js
@@ -64,6 +64,11 @@ Blockly.Blocks['dynamic_text_join'] = {
     }
   },
 
+  /**
+   * Parses XML based on the 'inputs' attribute (non-standard).
+   * @param {!Element} xmlElement XML storage element.
+   * @this {Blockly.Block}
+   */
   deserializeInputs_: function(xmlElement) {
     const items = xmlElement.getAttribute('inputs');
     if (items) {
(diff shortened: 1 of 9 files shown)


plugins/block-dynamic-connection/src/dynamic_text_join.js
```diff
@@ -69,6 +82,21 @@ Blockly.Blocks['dynamic_text_join'] = {
...
+  /**
+   * Parses XML based on the 'items' attribute (standard).
+   * @param {!Element} xmlElement XML storage element.
+   * @this {Blockly.Block}
+   */
+  deserializeCounts_: function(xmlElement) {
+    const itemCount = Math.max(
+        parseInt(xmlElement.getAttribute('items'), 10), this.minInputs);
+    // Two inputs are added automatically.
+    for (let i = 2; i < itemCount; i++) {
```
alschmiedt: Can we use `this.minInputs` here?

alschmiedt (approved):

plugins/block-dynamic-connection/src/dynamic_text_join.js
BeksOmega: We can, but it's a bit weird because `minInputs` not being used [by what actually determines what this number needs to be](https://github.com/google/blockly-samples/blob/3832abc63faaf5fe5d40a2e4c2dcfea90267e391/plugins/block-dynamic-connection/src/dynamic_text_join.js#L34). So it kinda felt right to keep it consistently magical? But saying it out loud seems weird lol

I'll switch it.

alschmiedt: Ah, I see what you mean. 

Yeah that should probably be cleaned up a bit, but we can do that in a separate PR. This fix LGTM.

Thanks!

PR #1001, opened 2022-02-14
Fixed to avoid deprecated schedule warning

Taikono-Himazin:
fix:
The following warning occurred while using backpack.
`Blockly.utils.genUid was deprecated on September 2021 and will be deleted on September 2022. Use Blockly.utils.idGenerator.genUid instead.`

The version of blockly has been upgraded to Q4 2021.
Perhaps the following changes are involved:
https://github.com/google/blockly/pull/5441

PR #1002, opened 2022-02-14
fix: Blockly.utils.genUid was deprecated

Taikono-Himazin:
fix: The following warning occurred while using backpack.
Blockly.utils.genUid was deprecated on September 2021 and will be deleted on September 2022. Use Blockly.utils.idGenerator.genUid instead.

The version of blockly has been upgraded to Q4 2021.
Perhaps the following changes are involved:
google/blockly#5441

### DIFF
--- a/plugins/workspace-backpack/package.json
+++ b/plugins/workspace-backpack/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@blockly/workspace-backpack",
-  "version": "1.0.15",
+  "version": "2.0.0",
   "description": "A Blockly plugin that adds Backpack support.",
   "scripts": {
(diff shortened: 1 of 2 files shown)


Taikono-Himazin:
Same content as #1001. 
Changed the commit comment.

alschmiedt:
Since this uses a new API that is not in version 6 of Blockly can you update the peerDependency to be `^7.20211209.0` and the version number to be 2.0.0? We update the version number when we make a change that is not backwards compatible and we update the peerDependency when we make a change that is not compatible with one of the versions listed.

alschmiedt (approved):

alschmiedt:
Thanks @Taikono-Himazin!

PR #1039, opened 2022-03-07
feat(migration): add a script for automatically performing renaming in developers' files

### LINKED ISSUE #924: Create renamings script
**Category**
Plugins

**Component**
Migration package

We want a central package where we can hold multiple scripts that help developers update to newer versions of Blockly. The package should be able to hold multiple scripts to help with updating. Beka has already started on a [package](https://github.com/google/blockly-samples/pull/871) that we can use as a starting point.

_Renaming script_
- The renaming script in particular should rely on an external file with the format specified in [renamings.js](https://github.com/google/blockly/tree/develop). Where we get this file from should be configurable.
- It should have a set of unit tests that we can run before we release.

In addition, we can sanity check the script by running it on samples before a release. Our release notes documentation should be updated when this package is released.

- We could attach individual scripts to releases. This is not as easy to find, the developer would have to download it, and it would be difficult for it to have any dependencies. 
- We could have separate packages for each script. This is more work for us to create a new package for every script and runs into problems 

**Additional context**
This is related to google/blockly#5645.

BeksOmega:


plugins/migration/test/rename.mocha.js
```diff
@@ -0,0 +1,527 @@
...
+
+/**
+ * @fileoverview Tests for the rename script.
+ */
+
+import {assert} from 'chai';
+import {doRenamings} from '../bin/rename.js';
+
+
+suite('Rename', function() {
```
BeksOmega, line 16: I think this should be a pretty good start to a test suite. If you can think of any other cases that need to be covered I'm happy to add them.

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Upgrades pieces of the Blockly API that have been renamed by modifying
+ * developers' local files.
+ * @param {!Object} database The database of renames.
+ * @param {string} currVersion The version to migrate from.
+ * @param {string} newVersion The version to migrate to.
+ * @param {!Array<string>} strings The strings to apply the renamings in.
+ * @return {!Array<string>} The strings with renamings applied.
+ */
+export function doRenamings(database, currVersion, newVersion, strings) {
```
BeksOmega: I separated this part of the algorithm out into an exported function to make the system easier to test.

cpcallen: Yes, this looks good, at the very least as a place to start.

plugins/migration/test/rename.mocha.js
cpcallen: So far so good.  We've already thought of a few more and I'm sure will think of more yet, but a solid start!

plugins/migration/bin/rename.js
```diff
@@ -246,6 +256,18 @@ class VersionRenamer {
...
+          console.log(`NOTE: ${entry.old} has been removed.`);
+          if (entry.get) {
+            console.log(`    - Call ${entry.get}() instead of reading it.`);
+          }
+          if (entry.set) {
+            console.log(`    - Call ${entry.set}(/* new value */) instead of ` +
+                'setting it.');
+          }
+          console.log('You will need to manually verify this update.');
+          return (entry.get ?? entry.set) + '()' + str.slice(entry.old.length);
```
BeksOmega: When I was adding the `-i` option I learned that `console.log` outputs to stdout. I'm not sure if that is desirable in this case. Eg people could be doing `npx @blockly/migrate rename 6 7 my.js > new.js` and they will get these logs in their new.js file.

cpcallen: I agree that that is not desirable.  We should use `process.stderr.write` or the like instead.

plugins/migration/bin/command.js
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * @fileoverview The command building utilities for the migration script.
+ */
+
+import chalk from 'chalk';
+import semver from 'semver';
+import {Command} from 'commander';
+
+// TODO: Not sure the best way to access the JSON from node.
```
cpcallen: I think the code is fine, but maybe a comment like this would be useful:

```suggestion
// require() is not available in ES modules by default, so we must obtain it.
```

plugins/migration/bin/command.js
```diff
@@ -0,0 +1,111 @@
...
+const SCRIPT_NAME = scriptPackageJson.name;
+
+const ROOT_COMMAND = new Command(SCRIPT_NAME)
+    .version(
+        scriptPackageJson.version,
+        '-v, --version',
+        'output the version of this script')
+    .usage(`${chalk.blue('<migration>')} ` +
+        `${chalk.green('<from-version> <to-version>')} [options] ` +
+        `${chalk.green('<file...>')}`);
```
cpcallen: Out of interest, did you pick these colours based on some standard?

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+ * @const {string}
+ */
+const DEV_VERSION = 'develop';
+
+createAndAddSubCommand(
+    'rename', '>=5', 'Perform renamings for breaking changes')
+    .option(
+        '-i [suffix]',
+        'do renamings in-place, optionally create backup files with the ' +
+        'given suffix')
```
cpcallen: Perhaps mention that otherwise modified code will be written to stdout.

plugins/migration/bin/command.js
```diff
@@ -0,0 +1,111 @@
...
+    name: name,
+    description: description,
+    range: targetVersionRange,
+  };
+  migrations.push(migration);
+
+  const subCommand = new Command(name)
+      .description(description)
+      .argument('<from-version>', 'Blockly version to migrate from')
+      .argument('<to-version>', 'Blockly version to migrate to')
```
cpcallen: I'd suggest that this should be optional, and default to the most recently released version.  I'd be ok with requiring `--from` and `--to` or the like

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+    .action(async function() {
+      const fromVersion = this.processedArgs[0];
+      const toVersion = this.processedArgs[1];
+      const fileGlobs = this.processedArgs[2];
+      const fileNames = fileGlobs.flatMap((fileGlob) =>
+        glob.sync(fileGlob, {nodir: true, nonull: false}));
```
cpcallen: Why are we doing globbing?  Is this for Windows users whose shell doesn't do it for them?  I'd prefer not to glob on platforms where the shell does—it makes it difficult to migrate files that have asterisks in their name…  ;-)

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+
+      if (!fileNames.length) {
+        console.log(`No matching files found for ${fileGlobs}. ` +
+            `Aborting rename.`);
+        return;
+      }
+
+      const renamer = new Renamer(await getDatabase(), fromVersion, toVersion);
+      fileNames.forEach((name) => {
+        readFile(name, 'utf8', (err, contents) => {
```
cpcallen: Consider using `readFileSync`, and later `writeFileSync`, since we don't actually need async here.

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+/**
+ * The version associated with renamings that have not been released yet.
+ * @const {string}
+ */
+const DEV_VERSION = 'develop';
+
+createAndAddSubCommand(
+    'rename', '>=5', 'Perform renamings for breaking changes')
+    .option(
+        '-i [suffix]',
```
cpcallen: Can we also accept `--in-place`?

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+      });
+    });
+
+/**
+ * Throws the error if the error is received, otherwise noop.
+ * @param {!Error|null} err A possible error to throw.
+ */
+function throwError(err) {
+  if (err) throw err;
+}
```
cpcallen: Do the errors thrown by this function contain an informative stack trace?  Do they cause the program to terminate?

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+      const renamer = new Renamer(await getDatabase(), fromVersion, toVersion);
+      fileNames.forEach((name) => {
+        readFile(name, 'utf8', (err, contents) => {
+          if (err) throw err;
+          const newContents = renamer.rename(contents);
+          const i = this.opts().i;
+          if (i) {
+            if (typeof i == 'string') writeFile(name + i, contents, throwError);
+            writeFile(name, newContents, throwError);
+            console.log(`Migrated renamings in ${name}`);
```
cpcallen: This will get printed too soon, before the write completes successfully.

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+          } else {
+            process.stdout.write(newContents);
```
cpcallen: Is it guaranteed that this can't result in writes becoming interleaved, even though we are not waiting for the first `write` to complete?

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+function throwError(err) {
+  if (err) throw err;
+}
+
+/**
+ * Gets the database of renames.
+ * @return {!Promise<Object>} The database of renames as an object.
+ */
+export async function getDatabase() {
+  const response = await fetch(DATABASE_URL);
```
cpcallen: Oh yeah: this should probably be a command line option (but with the existing value as default, obviously).

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+
+const DATABASE_URL = `https://raw.githubusercontent.com/google/blockly/develop/scripts/migration/renamings.json5`;
+
+/**
+ * The version associated with renamings that have not been released yet.
+ * @const {string}
+ */
+const DEV_VERSION = 'develop';
+
+createAndAddSubCommand(
```
cpcallen: Is there a way that we could make this module side-effect free, and instead export the subcommand?  (That would also give `migrate.js` a bit more purpose.)

plugins/migration/bin/command.js
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env node
```
cpcallen: Unless I'm mistaken, migrate.js is the only file in this package that's actually an executable script in its own right (i.e., should be +x and have a `#!` line); the rest are just regular JS files.  (Here and most files below.)

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+    return renamers;
+  }
+
+  /**
+   * Coerces the given string into a valid version (semver compliant or
+   * develop).
+   * @param {string} version  The version to coerce.
+   * @return  {string} The coerced version.
+   */
+  static coerce(version) {
```
cpcallen: Perhaps `coerceVersion`?

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+        version : semver.coerce(version).toString();
+  }
+
+  /**
+   * Compares the given versions. Compatible with Array.sort.
+   * @param {string} v1 The first version to compare.
+   * @param {string} v2 The second version to compare.
+   * @return {number} A number indicating the relationship between the versions.
+   */
+  static compare(v1, v2) {
```
cpcallen: Perhaps `compareVersions`?

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+
+  /**
+   * Returns true if the first version is less than or equal to the second
+   * version.
+   * @param {string} v1 The version to compare.
+   * @param {string} v2 The version to compare against.
+   * @return {boolean} True if the first version is less than or equal to the
+   *     second one.
+   */
+  static lte(v1, v2) {
```
cpcallen: I'd prefer a more descriptive name (and for `gt`), but I can't actually think of anything I like better.  :-/

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+    return semver.gtr(v1, `^${v2}`);
+  }
+
+  /**
+   * Applies the given renamings directly to a JavaScript string
+   * (presumably the contents of a developer's file).
+   * @param {string} str The string to apply the renamings in.
+   * @return {string} The file with renamings applied.
+   */
+  rename(str) {
```
cpcallen, line 122: I'd suggest keeping this together with `calculateRenamings`, to make it easier to read both at the same time.

cpcallen (commented):
I've had a look through the code that you wrote.  Overall looks great, though I have some suggestions.

* `migrate.js` is so small that I'd probably try to fold `command.js` into it.
* It would be good to have an automated test that actually exercises the script via the command line, perhaps via [`child_process.execSync`](https://nodejs.org/api/child_process.html#child_processexecsynccommand-options).
* `GETSTARTED.md` and `README.md` need to be updated appropriately.
* Do we want to include `manual-test-data.js`?  If so, an explanatory comment at the top about how to use it would be helpful.

Can I leave you to finish this up, and then get one of our colleagues to review?

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,286 @@
...
+    /** 
+     * List of pre-compiled renamings.
+     * @private @const {
+     *     !Array<{old: string, new: ?string, get: ?string, set: ?string}>}
+     */ 
+    this.renamings_ = [];
+    // See the sample entry in renamings.json5 for explanation of the
+    // meaning of the different properties on database entries.
+    for (const module of entry) {
+      const oldModulePath = module.oldPath ?? module.oldName;
```
BeksOmega: Apparently null coalescing [doesn't exist in node twelve](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator#browser_compatibility) :/ So I'm going to have to shuffle this around a bit.

plugins/migration/bin/command.js
BeksOmega: Nope!

plugins/migration/bin/rename.js
BeksOmega: Why?

plugins/migration/bin/rename.js
BeksOmega: Should no longer be a problem if we write synchronously correct?

plugins/migration/bin/command.js
BeksOmega: `--from <from-version>` is now a required option rather than an argument. `--to <to-version>` is an optional argument with a required value. It defaults to `'latest'`. Working with `'latest'` and `'develop'` versions is now handled in a separate `versions.js` file so that it can be reused by future migrations.

plugins/migration/bin/rename.js
BeksOmega: Made these descriptive by moving them to a new versions file. The renamer now uses `versionUtils.lte` which I think is clear enough.

plugins/migration/bin/rename.js
BeksOmega: I was trying to make sure that `ROOT_COMMAND` could only have subcommands with the given form, by making it so that the only external access to it was via `createAndAddSubcommand` and `parseAndRunMigrations`. IE no `addSubCommand` function was available.

I could also create a subclass of `Command` and use type annotations to "ensure" those are the only things that get added to the `ROOT_COMMAND`... but that seems like overkill. I'll just add an `addSubCommand` function to remove side effects, and hope nobody decides to break the pattern in the future.

BeksOmega:
> `migrate.js` is so small that I'd probably try to fold `command.js` into it.

They're separated to avoid circular dependencies. The individual scripts rely on `command.js`. `migrate.js` relies on both the individual scripts, and `command.js`. We have to at least keep the command creation and running separate, and the way it is currently separated makes sense in my head.

>It would be good to have an automated test that actually exercises the script via the command line, perhaps via [child_process.execSync](https://nodejs.org/api/child_process.html#child_processexecsynccommand-options).

I'm going to punt on this for now (I don't think commander leaves a lot of room for messing things up), but if it becomes a problem in the future we can always add it.

plugins/migration/bin/command.js
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * @fileoverview The command building utilities for the migration script.
+ */
+
+import chalk from 'chalk';
+import semver from 'semver';
+import {Command} from 'commander';
+
+// require() is not available in ES modules by default, so we must obtain it.
```
maribethb, line 15: I'm a little confused by this part -- why do we need to mix `import` and `require`? The comment should explain that.

 `@blockly/create-package` is executable via `npx` and just uses `require` directly without doing this, so I guess that's not the reason. perhaps because it only uses `require` and not `import` so it is automatically understood as a CommonJs module instead of an ES module? which leads me back to why are we mixing them?

plugins/migration/bin/command.js
```diff
@@ -0,0 +1,135 @@
...
+        `${chalk.green('<file...>')}`);
+
+const HELP_COMMAND = new Command('help')
+    .argument('[version]', 'the version of Blockly to provide info about')
+    .description('display this help, or help for the specific version')
+    .action(function() {
+      if (this.args.length) {
+        showVersionHelp(this.args[0]);
+      } else {
+        ROOT_COMMAND.help();
```
maribethb, line 40: What does this call? it looks like it would be an infinite loop but I don't know that much about Commander

plugins/migration/bin/command.js
```diff
@@ -0,0 +1,135 @@
...
+ * @type {!Array<{range: string, name: string, description: string}>}
+ */
+const migrations = [];
+
+/**
+ * Logs information about migrations for the given version.
+ * @param {string} version The version to show migrations for.
+ */
+function showVersionHelp(version) {
+  const targetRange = `${version} - ${version}`;
```
maribethb: In terms of semver ranges, I don't see how this is different than just `version`?

plugins/migration/bin/command.js
```diff
@@ -0,0 +1,135 @@
...
+ * @param {Command} command The migration to run.
+ */
+export function addSubCommand(command) {
+  ROOT_COMMAND.addCommand(command);
+}
+
+/**
+ * Extracts the from-version, to-version, and file names from the options and
+ * arguments.
+ * @param {Command} command The command to extract the info from.
```
maribethb, line 113: `!Command`

plugins/migration/bin/rename.js
maribethb: +1, one reason is so that we can test upcoming changes to the renamings file before publishing them to this default location.

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,212 @@
...
+'use strict';
+
+import {createSubCommand, extractRequiredInfo} from './command.js';
+import fetch from 'node-fetch';
+import {readFileSync, writeFileSync} from 'fs';
+import * as versionUtils from './versions.js';
+import JSON5 from 'json5';
+
+
+const DATABASE_URL = `https://raw.githubusercontent.com/google/blockly/develop/scripts/migration/renamings.json5`;
```
maribethb: are we sure this should be develop and not master? on the one hand i guess if changes are gated by the release you're updating to, it's fine to have unreleased changes here, but on the other hand develop is a lot more active and what if we break the renamings file during the quarter? then we've broken this script "in production" despite not doing a release

i also haven't read far enough to know how the default value of "to" version is chosen to know if that might be a problem from using develop branch

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,212 @@
...
+
+      const renamer = new Renamer(await getDatabase(), fromVersion, toVersion);
+      fileNames.forEach((name) => {
+        const contents = readFileSync(name, 'utf8');
+        const newContents = renamer.rename(contents);
+        const i = this.opts().i;
+        if (i) {
+          if (typeof i == 'string') writeFileSync(name + i, contents);
+          writeFileSync(name, newContents);
+          process.stderr.write(`Migrated renamings in ${name}`);
```
maribethb: Why does this one go to stderr?

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,212 @@
...
+    .action(async function() {
+      const {fromVersion, toVersion, fileNames} = extractRequiredInfo(this);
+
+      const renamer = new Renamer(await getDatabase(), fromVersion, toVersion);
+      fileNames.forEach((name) => {
+        const contents = readFileSync(name, 'utf8');
+        const newContents = renamer.rename(contents);
+        const i = this.opts().i;
+        if (i) {
+          if (typeof i == 'string') writeFileSync(name + i, contents);
```
maribethb: I know we discussed allowing single-line if statements, but I thought the consensus was only for things like `if (a) return;` and not for actual statements that do things. I would strongly prefer this be on its own line, though we can revisit this as a team if opinions are split.

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,212 @@
...
+          process.stderr.write(`Migrated renamings in ${name}`);
+        } else {
+          process.stdout.write(newContents);
+        }
+      });
+    });
+
+/**
+ * Gets the database of renames.
+ * @return {!Promise<Object>} The database of renames as an object.
```
maribethb: `!Promise<!Object>`

unless parse can return null? in which case `?`

i feel like we would want some error messaging here in case the file can't be parsed, instead of just crashing

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,212 @@
...
+
+/**
+ * Upgrades pieces of the Blockly API that have been renamed by modifying
+ * developers' local files.
+ * @param {!Object} database The database of renames.
+ * @param {string} currVersion The version to migrate from.
+ * @param {string} newVersion The version to migrate to.
+ * @param {!Array<string>} strings The strings to apply the renamings in.
+ * @return {!Array<string>} The strings with renamings applied.
+ * @deprecated Use Renamer instance instead.
```
maribethb: Why add a function that is already marked deprecated? i'm a little confused here

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,212 @@
...
+      renamers.push(new VersionRenamer(database[version]));
+    }
+    return renamers;
+  }
+
+  /**
+   * Applies the given renamings directly to a JavaScript string
+   * (presumably the contents of a developer's file).
+   * @param {string} str The string to apply the renamings in.
+   * @return {string} The file with renamings applied.
```
maribethb: nit: to be consistent with the other language here, you're returning a string, not a file

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,212 @@
...
+      return match;
+    });
+  }
+}
+
+/**
+ * An object which can apply a given set of renamings for a single
+ * Blockly version to the individual dotted-identifier paths.
+ */
+class VersionRenamer {
```
maribethb, line 136: This marks the end of what I've reviewed so far, I'm taking a break for now but wanted to get the above feedback in so we can work simultaneously. if anything i've said would just make more sense if i waited to review the rest of it, my bad!

plugins/migration/bin/migrate.js
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env node
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview A script for helping to migrate between versions of Blockly.
```
maribethb, line 9: I don't know what the purpose of this file is... if we create the subcommand in `rename.js` why can't we add it there? why do we call `parseAndRunMigrations` with no args?

i'm sure there are good reasons :) but could you add them to the comments in this file?

plugins/migration/bin/command.js
BeksOmega: > perhaps because it only uses require and not import so it is automatically understood as a CommonJs module instead of an ES module?

We instruct npm to interpret it as an ES module by adding `"type": "module"` to the `package.json`. 

> which leads me back to why are we mixing them?

We don't need to mix them, we could just use CommonJS modules, but it seems like we're trying to move toward ES modules, so we may not want to. However, `import` doesn't allow you to import JSON, which we need to do to get access to the script name and version. So we need to import require to do that.

If you want to switch everything over to CommonJS we can do that. But if we want to keep it as ES Modules afaict this is the standard solution.

plugins/migration/bin/command.js
BeksOmega: It calls commanders automatically generated help. It has information about sub commands and their descriptions, but they're just organized alphabetically.

plugins/migration/bin/rename.js
BeksOmega: Whoops, this was how we were structuring the API before, and it should be removed how that we added the classes. Thanks for the catch!

plugins/migration/bin/migrate.js
BeksOmega: > why can't we add it there?

There seems to be disagreement between you and Christopher about whether modules should have side effects or not haha. That's what we did before, but Christopher wanted `rename.js` side-effect-less (which I think is reasonable) so I changed it.

> why do we call parseAndRunMigrations with no args?

It defaults to using `process.argv` if you pass no args. I'll add comments to indicate that!

plugins/migration/bin/rename.js
BeksOmega: > it's fine to have unreleased changes here, but on the other hand develop is a lot more active and what if we break the renamings file during the quarter?

Christopher wanted it to be develop so that if we forgot to add renamings to the database before a release we could add them to the database without having to make a commit on master. Wrt breaking the file, the schema that we added to core should help with that (theoretically?) But I agree, that is pretty worrisome. 

> i also haven't read far enough to know how the default value of "to" version is chosen to know if that might be a problem from using develop branch

The default value of "to" is a special value "latest" which performs all renames up to the latest released version (so up to the version before "develop" in the database

plugins/migration/bin/command.js
BeksOmega: Ahah! You are correct :D I didn't think the package would accept that, but it does. Thank you Maribeth :D

plugins/migration/bin/rename.js
BeksOmega: All the informative messages in `rename.js` go to stderr so that people get the messages even when they're redirecting the command output to new files. Eg `npx @blockly/migrate rename --from 7 ./file.js > ./newFile.js`

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+    return str;
+  }
+}
+
+/**
+ * RegExp matching a dotted identifier path like "foo.bar.baz".  Note
+ * that this only matches such paths containing at least one dot, as
+ * we expect to be looking for string like "Blockly.<something>" and
+ * don't want to try to rename every singe variable and every word
```
rachel-fenichel: nit: singe -> single

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,208 @@
...
+      // TODO: switch all of the || to ?? if/when we drop Node12.
+      const oldModulePath = module.oldPath || module.oldName;
+      const newExport = module.newExport ? '.' + module.newExport : '';
+      const newModulePath =
+          module.newPath || (module.newName || oldModulePath) + newExport;
+
+      if (module.exports) {
+        for (const [oldExportName, info] of Object.entries(module.exports)) {
+          const oldExportPath =
+              info.oldPath || `${oldModulePath}.${oldExportName}`;
```
rachel-fenichel, line 160: Why are you mixing template strings and string concatenation? (here vs next line and line 145).

plugins/migration/bin/rename.js
```diff
@@ -0,0 +1,208 @@
...
+            renaming.get = info.getMethod ? newBase + info.getMethod : null;
+            renaming.set = info.setMethod ? newBase + info.setMethod : null;
+          } else {
+            renaming.new = newBase + (info.newExport || oldExportName);
+          }
+          this.renamings_.push(renaming);
+        }
+      }
+
+      this.renamings_.push({old: oldModulePath, new: newModulePath});
```
rachel-fenichel, line 184: I expected this to be directly after the lines that construct `oldModulePath` and `newModulePath`. is it down here because the order of the renamings matters?

Also, why is it safe to use `new` as a key in your object here? It's normally a reserved keyword.

rachel-fenichel (approved):
A few nits, but overall LGTM.

plugins/migration/bin/rename.js
BeksOmega: I believe it is down here because we need to rename properties before renaming modules, otherwise we won't be able to properly detect properties that need to be renamed.

Eg:
moduleA -> moduleB
moduleA.exportA -> moduleC.exportC

If we rename moduleA first we won't be able to find moduleA.exportA because it will have become moduleB.exportA

@cpcallen  to check. If I'm right, I'll add a comment.

> Also, why is it safe to use new as a key in your object here? It's normally a reserved keyword.

I don't think there's any conflict because we're using it as a property, not a local identifier. Different grammar rules afaik.

rachel-fenichel (approved):

plugins/migration/bin/rename.js
maribethb: I think it should be master. Changing this should be considered the same as a "release" so if we forget to add renamings then we should accept the consequences and cherry-pick the additions into master. We don't even have to actually release a patch version if we do that, so there's not a lot of downside imo.

plugins/migration/bin/rename.js
maribethb: I also still think we need this.

maribethb (approved):

PR #1040, opened 2022-03-08
feat: Backwards execution sample

NeilFraser:
Also fixes and updates the existing two interpreter samples.

Not using NPM for JS-Interpreter or DMP libraries since they aren't published to NPM (well, they are, but by unknown 3rd parties).

Not using NPM for Acorn, since we are frozen at version version 0.4.1 (from late 2013) instead of the current version 8.7.0.  The ancient 0.4.1 is the last version to be exclusively ES5, is smaller, and faster.


rachel-fenichel:
@NeilFraser can you give a suggested review order for these files, and note which ones are wholesale pulled in from other projects?

NeilFraser:
The three commits are separate and in order.

Files pulled in from JS-Interpreter:
* acorn_interpreter.js (already existing, just being updated)
* acorn.js + interpreter.js (same code as above package, but needs to be uncompressed since serialization reaches deep beyond the public API and mucks with properties that the compiler renames).
* serialize.js (the serialization/deserialization functions)

Additionally, there's one optional file from DMP:
* diff-match-patch.js (compressed, reduces each step from 300 KB to a few bytes)

examples/interpreter-demo/backwards.html
```diff
@@ -0,0 +1,320 @@
...
+  <h1><a href="https://developers.google.com/blockly/">Blockly</a> &gt;
+    <a href="../index.html">Demos</a> &gt; Backwards Stepping with JS-Interpreter</h1>
+
+  <p>This is a demo of executing code step-by-step with a sandboxed JavaScript
+  interpreter -- both forwards and backwards.</p>
+
+  <p>Each step forwards saves the current state to a stack.  Each state
+  backwards restores to the previous state on the stack.  Compression is used
+  to keep only the deltas between stack frames.  Note that because serialization
+  reaches deep beyond the public API for the JS-Interpreter, the uncompiled
```
BeksOmega, line 42: I assume that it's fine to have something in the demo reaching beyond the public API because it is also provided? IE it's not like we're recommending people reach beyond the public API

examples/interpreter-demo/backwards.html
```diff
@@ -0,0 +1,320 @@
...
+
+  <div style="width: 100%">
+    <div id="blocklyDiv"
+        style="display: inline-block; height: 480px; width: 58%"></div>
+    <textarea id="output" disabled="disabled"
+        style="display: inline-block; height: 480px; width: 38%;">
+    </textarea>
+  </div>
+
+  <xml xmlns="https://developers.google.com/blockly/xml" id="toolbox" style="display: none">
```
BeksOmega: It would be nice if this used JSON instead of XML, but I don't think that's a priority :P

examples/interpreter-demo/backwards.html
```diff
@@ -0,0 +1,320 @@
...
+      demoWorkspace.highlightBlock(null);
+      highlightPause = false;
+
+      if (clearOutput) {
+        outputArea.value = 'Program output:\n=================';
+      }
+      myInterpreter = null;
+    }
+
+    function stepBackwards() {
```
BeksOmega, line 189: I don't understand why the outputArea doesn't need to be cleared when we step backwards :/

examples/interpreter-demo/backwards.html
```diff
@@ -0,0 +1,320 @@
...
+    //   .highlight: The ID of the block highlighted in this state.
+    var serializationStack = [];
+
+    // Global variable to keep track of the currently highlighted block.
+    var highlightedBlockId = '';
+
+    // The forward step button disables for 2s after program completion, then
+    // re-enables.  If the backwards step button is pressed during this time
+    // then cancel the scheduled re-enabling.
+    var disablePid = 0;
```
BeksOmega: I found this name kind of confusing, because it is actually tracking when the button gets re-enabled. Maybe `enablePid` instead?

examples/interpreter-demo/backwards.html
```diff
@@ -0,0 +1,320 @@
...
+    Blockly.Xml.domToWorkspace(document.getElementById('startBlocks'),
+                               demoWorkspace);
+
+    var outputArea = document.getElementById('output');
+    var myInterpreter = null;
+
+    // Stack of serializations.  Each object contains three properties:
+    //   .json: A full serialization of the interpreter's state.
+    //   .delta: Instead of a full serialization, a delta from the next state.
+    //   .highlight: The ID of the block highlighted in this state.
```
BeksOmega: Not exactly true. If dmp is enabled only the top of the stack has a .json property right?

examples/interpreter-demo/backwards.html
```diff
@@ -0,0 +1,320 @@
...
+      var wrapper = function(id) {
+        id = String(id || '');
+        highlightedBlockId = id;
+        return highlightBlock(id);
+      };
+      interpreter.setProperty(globalObject, 'highlightBlock',
+          interpreter.createNativeFunction(wrapper));
+    }
+
+    var highlightPause = false;
```
BeksOmega, line 172: Maybe `foundHighlightBlock`? or add some comments to clarify what this is doing.

examples/interpreter-demo/backwards.html
```diff
@@ -0,0 +1,320 @@
...
+      highlightedBlockId = serialization.highlight;
+      highlightBlock(highlightedBlockId);
+      var json = serialization.json;
+      if (dmp && serializationStack.length) {
+        // Uncompress the previous serialization.
+        var previousStack = serializationStack[serializationStack.length - 1];
+        var previousDiff = dmp.diff_fromDelta(json, previousStack.delta);
+        previousStack.delta = null;
+        previousStack.json = dmp.diff_text2(previousDiff);
+      }
```
BeksOmega: How would you feel about splitting the stack compression out into separate functions, since it is optional? I think that would make the demo faster to digest.

examples/interpreter-demo/backwards.html
```diff
@@ -0,0 +1,320 @@
...
+            resetStepUi(false);
+
+            // Cool down, to discourage accidentally restarting the program.
+            document.getElementById('stepForwardsButton').disabled = 'disabled';
+            disablePid = setTimeout(function() {
+              document.getElementById('stepForwardsButton').disabled = '';
+              document.getElementById('stepBackwardsButton').disabled = 'disabled';
+              disablePid = 0;
+            }, 2000);
+
```
BeksOmega: How do you feel about splitting this into a separate function?

BeksOmega (changes requested):
Love this demo :D Took me a bit to figure it out, but once I did it was super cool! Comments are mostly about clarifying/making it easier to read

Also [blocklyDemoConfig](https://github.com/google/blockly-samples/blob/1be2b748414ef147c8f7d657fa440442c0ef49a8/examples/interpreter-demo/package.json#L5) needs to be updated, but otherwise I think this should work fine with github pages.

examples/interpreter-demo/backwards.html
```diff
@@ -0,0 +1,320 @@
...
+      // Add an API function for the alert() block, generated for "text_print" blocks.
+      var wrapper = function alert(text) {
+        text = arguments.length ? text : '';
+        outputArea.value += '\n' + text;
+      };
+      interpreter.setProperty(globalObject, 'alert',
+          interpreter.createNativeFunction(wrapper));
+
+      // Add an API function for the prompt() block.
+      var wrapper = function prompt(text) {
```
rachel-fenichel: Use different names for the wrappers (or include them inline) instead of redeclaring a var with the same name each time.

examples/interpreter-demo/serialize.js
```diff
@@ -0,0 +1,411 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
rachel-fenichel, line 3: Update copyright year if this is new code.

examples/interpreter-demo/serialize.js
```diff
@@ -0,0 +1,411 @@
...
+ */
+'use strict';
+
+
+// Constructors for objects within Acorn.
+var NODE_CONSTRUCTOR;
+var NODE_LOC_CONSTRUCTOR;
+var LINE_LOC_CONSTRUCTOR;
+
+function recordAcornConstructons(interpreter) {
```
rachel-fenichel: Add JSDoc-style comments on top level functions.

examples/interpreter-demo/serialize.js
```diff
@@ -0,0 +1,411 @@
...
+
+/**
+ * @fileoverview Saving and restoring the state of a JS-Intepreter.
+ * @author fraser@google.com (Neil Fraser)
+ */
+'use strict';
+
+
+// Constructors for objects within Acorn.
+var NODE_CONSTRUCTOR;
```
rachel-fenichel, line 15: Use const and let instead of var.

examples/interpreter-demo/serialize.js
```diff
@@ -0,0 +1,411 @@
...
+  // Constructors for objects within Acorn.
+  NODE_CONSTRUCTOR = interpreter.ast.constructor;
+  NODE_LOC_CONSTRUCTOR = interpreter.ast.loc &&
+      interpreter.ast.loc.constructor;
+  LINE_LOC_CONSTRUCTOR = interpreter.ast.loc &&
+      interpreter.ast.loc.end.constructor;
+}
+
+function deserialize(json, interpreter) {
+  function decodeValue(value) {
```
rachel-fenichel: This function can be defined at the top level if you pass in `objectList`.

examples/interpreter-demo/serialize.js
```diff
@@ -0,0 +1,411 @@
...
+      for (var i = 0; i < names.length; i++) {
+        if (isAcornNode && names[i] === 'loc') {
+          continue;  // Skip over node locations, they are specially handled.
+        }
+        try {
+          objectHunt_(node[names[i]], objectList);
+        } catch (e) {
+          // Accessing some properties may trigger a placeholder getter.
+          // Squelch this error, but re-throw any others.
+          if (e.message !== 'Placeholder getter') {
```
rachel-fenichel, line 465: Which properties? You don't have a try/catch around the initial calls to `objectHunt_`. Why is that safe?

examples/interpreter-demo/serialize.js
```diff
@@ -0,0 +1,411 @@
...
+  objectHunt_(stack, objectList);
+  var functionHash = Object.create(null);
+  for (var i = 0; i < objectList.length; i++) {
+    if (typeof objectList[i] === 'function') {
+      functionHash[objectList[i].id] = objectList[i];
+    }
+  }
+  // First pass: Create object stubs for every object.
+  objectList = [];
+  for (var i = 0; i < json.length; i++) {
```
rachel-fenichel, line 73: Break these passes out into separate functions for readability.

examples/interpreter-demo/serialize.js
```diff
@@ -0,0 +1,411 @@
...
+      case 'State':
+        obj = new Interpreter.State(undefined, undefined);
+        break;
+      case 'Node':
+        obj = new NODE_CONSTRUCTOR();
+        delete obj.start;
+        delete obj.end;
+        var locText = jsonObj['loc'];
+        if (locText) {
+          var loc = new NODE_LOC_CONSTRUCTOR();
```
rachel-fenichel: Break this out into a helper function and add comments, writing for a reader who doesn't know the structure of Nodes in JS-Interpreter.

examples/interpreter-demo/serialize.js
```diff
@@ -0,0 +1,411 @@
...
+    }
+  }
+  // First object is the interpreter.
+  var root = objectList[0];
+  for (var prop in root) {
+    interpreter[prop] = root[prop];
+  }
+}
+
+function serialize(interpreter) {
```
rachel-fenichel: Also decompose this function.

examples/interpreter-demo/backwards.html
NeilFraser: Correct, only serialize.js (which comes from the JS-Interpreter project) access these properties.  The Blockly code stays well within the public API.

Serialization requires getting and setting every property in the interpreter -- public or not.  So it is not possible to run under advanced compilation.

examples/interpreter-demo/backwards.html
NeilFraser: Well that's an interesting question.  There are three things that the output area could do on a backwards step:
1. Clear completely.  This seems a bit brutal to lose all output just because of one backwards step.
2. Restore to the output at that step.  This involves storing the output on each step object (just like we do for highlighted block).
3. Leave the output alone, and let new print lines accumulate (current behaviour).  This is more like a console log.

I can see a strong argument for the second option, however I chose not to do it since it adds complexity to the demo in a way that doesn't help developers who want to implement backwards debugging on their apps.  Each app would need to store different data to restore to.  A maze game for example would need to store the x, y, and direction of the player on each step object.

BeksOmega: > doesn't help developers who want to implement backwards debugging on their apps.

I see your point about how each app is going to need to store different data, however I disagree that option 2 it doesn't help. I can't think of a situation where an app (especially an educational one) *wouldn't* want to store data about the output of the previous step. Because if the user doesn't see the output of the previous step, it doesn't really help them with debugging, and it doesn't "feel" like the program has been stepped backwards. I think that giving some indication of how this would be done is important for external developers who want to implement the backwards stepping.

I'd also like to note that if the demo doesn't return to the previous output state, people may be confused about whether it is actually working as intended or not.

I feel like it should be possible to separate all of the output storage into separate functions so that it exists, gives the demo the "expected" functionality, and doesn't add an excessive amount of complexity. However, I didn't write the demo, so I'm happy for you to disagree on this.

examples/interpreter-demo/serialize.js
NeilFraser: This is the focus-point of the design.  The offending `encodeValue` function needs access to `objectList`.  And one of the two calls is from `Array.map`.  Thus there's no way to pass `objectList` in as a second argument.  The options are:

1. Wrap the call as a new function: `map(function(element) {return encodeValue(element, objectList);})` This creates a new function for every array in the interpreter, which is a lot.
2. Reverse the order of the arguments and bind the function: `map(encodeValue.bind(null, objectList))` This creates a new function for every array in the interpreter, and also would be much harder to understand.
3. Push `objectList` into a global variable.  Works fine.
4. Push `encodeValue` into `serialize`.  Also works fine.

The fourth option is what we currently have.  I that's preferable to a global variable, let alone the other options.  What do you think?

examples/interpreter-demo/serialize.js
NeilFraser: While this function can be trivially broken out, I think it's best for it to exactly mirror its `encodeValue` counterpart which cannot be easily broken out (pending comment below).

examples/interpreter-demo/serialize.js
NeilFraser: JS-Interpreter is ES5 only.  It can execute itself.  Using ES6 would break this.

examples/interpreter-demo/serialize.js
NeilFraser: This is an unmodified file from the JS-Interpreter project.

examples/interpreter-demo/serialize.js
rachel-fenichel: I see. 3 is fine then, particularly since (as previously discussed) this code is not new code written for this particular demo.

At a higher level, this is what modules are for: letting you put things in file-global variables without having them be accessible everywhere.

examples/interpreter-demo/backwards.html
NeilFraser: Done.

examples/interpreter-demo/backwards.html
NeilFraser: Added 'or...'

examples/interpreter-demo/backwards.html
NeilFraser: One button gets enabled, one gets disabled, so `disablePid` and `enablePid` are equally incorrect.  Changed to `resetButtonsPid`.

examples/interpreter-demo/backwards.html
NeilFraser: Renamed in this and all existing example files.

examples/interpreter-demo/backwards.html
NeilFraser: Added a comment.  Also removed the property from async-execution since it is not used.

examples/interpreter-demo/backwards.html
NeilFraser: Done.

examples/interpreter-demo/backwards.html
NeilFraser: Done.

examples/interpreter-demo/backwards.html
NeilFraser: Done.

examples/interpreter-demo/serialize.js
NeilFraser: Done.

examples/interpreter-demo/serialize.js
NeilFraser: Done.

examples/interpreter-demo/serialize.js
NeilFraser: Done.

examples/interpreter-demo/serialize.js
NeilFraser: Done.

examples/interpreter-demo/serialize.js
NeilFraser: Rewritten to make it clear that it's not the objectHunt_ call that might throw, but just the `node[names[i]]` property lookup.

examples/interpreter-demo/serialize.js
NeilFraser: Done.

examples/interpreter-demo/backwards.html
BeksOmega: Wow awesome! tysm!

BeksOmega (approved):
All changes I requested LGTM! Might want to wait on Rachel for hers (not sure)

rachel-fenichel (approved):

PR #1064, opened 2022-03-29
fix: Remove host option to work on Windows.

### LINKED ISSUE #832: npm run start does not work on windows
**Category**

All

**Component**

All

**Describe the bug**

Windows does not redirect 0.0.0.0 to localhost (127.0.0.1) So when you run `npm run start` you have to redirect it yourself.

**To Reproduce**

1. Be on windows
2. Navigate into a plugin directory
3. Run `npm install`
4. Run `npm run start`
5. Observe how you do not get redirected to a plugin page

Note that I haven't reproduced this myself, because I'm currently on chromeOS.

**Expected behavior**

The `npm run start` script works on windows

**Additional context**

Originally reported on the forums: https://groups.google.com/g/blockly/c/9ed-Do6p_CY

yamadayutaka:
### Resolves

#832

### Proposed Changes

Remove host option from webpackDevServer.config.js.

This config works on both Windows and Linux.

### DIFF
--- a/plugins/dev-scripts/config/webpackDevServer.config.js
+++ b/plugins/dev-scripts/config/webpackDevServer.config.js
@@ -12,7 +12,6 @@
 module.exports = () => {
   return {
     port: 3000,
-    host: '0.0.0.0',
     hot: true,
     static: ['./test'],
     watchFiles: {


BeksOmega:
@yamadayutaka Even though we can't accept the changes you suggested in core (https://github.com/google/blockly/issues/5000), I think we can accept this fix =)

Based on the [webpack docs](https://webpack.js.org/configuration/dev-server/) the host property is just necessary to make the server available externally (which is semi-required for testing fixes on mobile devices). But I don't think we often have mobile-specific bugs in this repo. So if we ever need do test that, we can just temporarily modify the dev-server config. 

Thank you for the fix @yamadayutaka ! We really appreciate it!

yamadayutaka:
Thank you for your approval.
I have confirmed it only on Windows and Linux (Ubuntu), but it seems that it is possible to connect from the outside without specifying the host option.
You can probably connect and test from the outside without modifying the config.

PR #1074, opened 2022-03-31
chore: update dependencies on Blockly for new core release

BeksOmega:
### Description

Updates the dev dependencies in plugins to point to the latest release of Blockly.

Updates dependency ranges to have an upper limit of version 8.

Updates demos to have upper bounds on their dependency ranges.

### Additional info

Did not update the tooltip plugin because it is getting deleted.

Did not update the plane demo because I was told it is also getting deleted.


maribethb (approved):
I'm a little skeptical of the changes to the "examples" that aren't "demos", since they don't get run on the samples site they probably haven't been tested with the new version (unless you ran them?). But those also use the most basic APIs like `Blockly.inject` which we haven't changed, so they're probably fine... but I'd also be fine with not updating those and filing an issue to check on and update each example.

BeksOmega:
> I'd also be fine with not updating those and filing an issue to check on and update each example.

sgtm

BeksOmega:
Created #1076

PR #1078, opened 2022-04-01
chore: update name for cross tab copy

alschmiedt:
This PR: 
- Renames copy-by-storage --> cross-tab-copy-paste
- Calls workspace.paste instead of using the json serializer to add the block directly to the workspace.
- Adds the plugin to gh-pages.
- Adds info to the plugin page.


plugins/cross-tab-copy-paste/test/index.html
```diff
@@ -12,6 +12,10 @@
 </head>
 
 <body>
+  <p>This plugin can be tested by opening up multiple tabs of this page
+    and copy/pasting the block from one tab to another. Keep in mind you can
+    not copy a block from this workspace and paste into another plugin page.
+  </p>
```
alschmiedt: If someone has better wording/explanation for this I am definitely open to suggestions.

plugins/cross-tab-copy-paste/package.json
```diff
@@ -45,7 +44,7 @@
     "blockly": "^8.0.0"
   },
   "peerDependencies": {
-    "blockly": "^8.0.0"
```
maribethb, line 48: I thought this relied on some change to paste that went in this release? or am i misremembering?

plugins/cross-tab-copy-paste/src/index.js
```diff
@@ -100,8 +103,9 @@ export class CopyByStorage {
           /** @type {!Blockly.ContextMenuRegistry.Scope} */ scope) {
         const blockText = localStorage.getItem('blocklyStash');
         const saveInfo = JSON.parse(blockText);
-        Blockly.serialization.blocks.append(
-            saveInfo['saveInfo'], scope.workspace, {recordUndo: true});
+        scope.workspace.paste(saveInfo['saveInfo']);
+        // Blockly.serialization.blocks.append(
```
maribethb: Remove this instead of commenting out

plugins/cross-tab-copy-paste/test/index.html
maribethb: "You can only paste into a workspace that also has this plugin enabled." ?

maribethb (approved):

plugins/cross-tab-copy-paste/package.json
BeksOmega: I think the change your thinking of is the one that changed the return type of paste. If so, that's for fixing keyboard nav so that it doesn't access internal properties.

plugins/cross-tab-copy-paste/test/index.html
alschmiedt: Perfect, Thanks!

plugins/cross-tab-copy-paste/package.json
alschmiedt: Confirming that I did not add anything to core for this plugin and I tested it works with v7 before merging.

PR #1080, opened 2022-04-04
fix: misprint

Taikono-Himazin:


gonfunko (approved):
Thank you for catching and fixing this! We really appreciate it.

PR #1093, opened 2022-04-05
feat: add a demo of the custom tooltip api

maribethb:
Adds a demo of using the new custom tooltip API.

Has two blocks: one has an image which will be rendered into the tooltip and one is regular. Also, the tooltip div takes the color of the block, in order to be similar to the existing block-extension-tooltip which we will deprecate.

(Ignore the branch name, this version of the demo does not use markdown)


examples/custom-tooltips-demo/index.js
```diff
@@ -0,0 +1,89 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
```
alschmiedt: 2022

examples/custom-tooltips-demo/index.js
```diff
@@ -0,0 +1,89 @@
...
+      // You can access the block being moused over.
+      // Here we get the color of the block to set the background color.
+      div.style.backgroundColor = element.getColour();
+    }
+    const tip = Blockly.Tooltip.getTooltipOfObject(element);
+    const text = document.createElement('div');
+    text.textContent = tip;
+    const container = document.createElement('div');
+    container.style.display = 'flex';
+    // Check to see if the custom property we added is present.
```
alschmiedt: Maybe add information on where it was added? '...we added in the block definition is present.'

examples/custom-tooltips-demo/index.js
```diff
@@ -0,0 +1,89 @@
...
+          .appendField('Mouse over me.');
+      this.setColour(150);
+      this.setTooltip('Tip: This tooltip has an image.');
+      // We will check for this property in our custom rendering code.
+      this.tooltipImg = 'lightbulb.png';
+      this.setHelpUrl('');
+    },
+  };
+  const defaultOptions = {
+    toolbox: document.getElementById('toolbox'),
```
alschmiedt: Might be better to have the toolbox definition defined in line using json? Then everything is in the same file and the user doesn't have to jump around to the different files as much to put everything together.

alschmiedt (approved):
LGTM. A few suggestions but feel free to take them or leave them.

maribethb:
Thanks! Added all your suggestions.

PR #1095, opened 2022-04-05
chore: update license year in create-package templates

maribethb:
Update all the template files so someone using the `@blockly/create-package` template gets the correct year in the copyright notice.


BeksOmega (approved):

BeksOmega:
Do you think we could make the create-package script do this dynamically? Like when you run it it would find the year and insert it.

maribethb:
> Do you think we could make the create-package script do this dynamically? Like when you run it it would find the year and insert it.

Probably somehow? but right now it just copies these files directly I'm pretty sure so it would be a bigger change.

Issue #1031, opened 2022-02-24
Add skip-lint option to start script.


IBlackI:
**Category**

 - Plugins

**Component**

blockly-scripts

**Is your feature request related to a problem? Please describe.**
I am using a different set of linting rules than provided by the blockly-scripts. I run my linter seperately from the blockly-scripts commands.

**Describe the solution you'd like**
Allow the start script to run with the skip-lint option like the build or test scripts.

**Additional context**
Example: https://github.com/google/blockly-samples/blob/master/plugins/dev-scripts/scripts/test.js#L31

PR #1116, opened 2022-04-19
Add skip-lint option to start script (#1031)

derwehr:
Haven't found any documentation that needs to be updated for this, so I believe this is it
(implements #1031)


alschmiedt (approved):

alschmiedt:
Thank you @derwehr for the fix!

PR #1122, opened 2022-04-21
feat: add bitmap input block

GregoryCannon:
Adds a field for a user-inputted pixel grid, as requested in https://github.com/google/blockly-samples/issues/370.

Sample: https://screenshot.googleplex.com/3GiLPouwcvw4f3T.png

Supports any image size. Value is stored as a 2D array of 1s and 0s.


plugins/field-bitmap/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+This field accepts up to 3 parameters:
+- `"value"` to specify an initial value. Must be a 2D rectangular array of 1s and 0s.
+If not provided, the default is an empty grid of the specified size.
+- `"width"` to specify an initial width, if there is no initial value.
+If not provided, the default is a width of 5.
+- `"height"` to specify an initial height, if there is no initial value.
+If not provided, the default is a height of 5.
+
+### JavaScript
```
BeksOmega, line 29: Could we have an example with the width and height? Since they use the config paramter and might be a bit confusing.

plugins/field-bitmap/package.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "blockly-field-bitmap",
```
BeksOmega: `@blockly/field-bitmap`

(This is a problem with our create-package script, sorry about that.)

plugins/field-bitmap/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "version": "0.0.0",
```
BeksOmega: I think we start with version 1.0.0 usually (@maribethb to confirm)

plugins/field-bitmap/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "description": "A Blockly field.",
```
BeksOmega: Update description plox.

plugins/field-bitmap/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "module": "./src/index.js",
+  "unpkg": "./dist/index.js",
+  "author": "",
+  "keywords": [
+    "blockly",
+    "blockly-plugin",
+    "blockly-field",
+    "bitmap"
+  ],
+  "homepage": "https://github.com/GregoryCannon/blockly-samples/tree/master/field-bitmap#readme",
```
BeksOmega: github.com/google/blockly

Here and below

(again our bad not yours)

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+ */
+
+/**
+ * @fileoverview Bitmap input field.
+ * @author gregoryc@outlook.com (Greg Cannon)
+ */
+
+
+import Blockly from 'blockly/core';
+import {assert} from 'chai';
```
BeksOmega: nit: 2 blank lines between imports and logic. (only [one is required](https://google.github.io/styleguide/jsguide.html#file-implementation), but we use two)

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+const PIXEL_SIZE = 15;
+const FILLED_PIXEL_COLOR = '#363d80';
+const EMPTY_PIXEL_COLOR = 'white';
+
+/**
+ * Field for inputting a small bitmap image.
+ * Includes a grid of clickable pixels that's exported as a bitmap.
+ */
+export class FieldBitmap extends Blockly.Field {
+  constructor(value = undefined, validator = undefined, config = undefined) {
```
BeksOmega, line 28: Add a comment to make the lint happy please. The description can just be like "constructor for the bitmap field"

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+   * @param {!Object} options A JSON object with options.
+   * @return {!FieldBitmap} The new field instance.
+   * @package
+   * @nocollapse
+   */
+  static fromJson(options) {
+    return new FieldBitmap((options && options['value']), undefined, options);
+  }
+
+  getImageWidth() {
```
BeksOmega, line 79: JSDoc on all public functions. Here and below.

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+    this.fieldGroup_ = null;
+    this.blockDisplayPixels_ = null;
+
+    /** Stateful variables */
+    this.mouseIsDown_ = false;
+    this.valToPaintWith_ = undefined;
+  }
+
+  /**
+   * Constructs a FieldTemplate from a JSON arg object.
```
BeksOmega:
```suggestion
   * Constructs a FieldBitmap from a JSON arg object.
```

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+  getImageHeight() {
+    return this.imgHeight_;
+  }
+
+  /**
+   * Validates that a new value meets the requirements for a valid bitmap array.
+   * @param {*} newValue
+   * @return the new value if it's valid, or null
+   */
+  doClassValidation_(newValue = undefined) {
```
BeksOmega, line 96: Love this validation function!

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+  }
+
+  /**
+   * Called when a new value has been validated and is about to be set.
+   * @param {*} newValue
+   */
+  doValueUpdate_(newValue) {
+    super.doValueUpdate_(newValue);
+    if (newValue) {
+      assert(Array.isArray(newValue));
```
BeksOmega: I don't think you need an assertion here. Should be covered by the checks in `doClassValidation_`

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+      }
+
+      this.imgHeight_ = newValue.length;
+      this.imgWidth_ = newValue[0] ? newValue[0].length : 0;
+    }
+  }
+
+  /**
+   * Show the bitmap editor dialog
+   * @param {Event=} e Optional mouse event that triggered the field to
```
BeksOmega: nit: I think we want `!Event=`

(Sorry, the type system is terrible haha)

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+   *     open, or undefined if triggered programmatically.
+   * @param {boolean=} _quietInput Quiet input.
+   * @protected
+   */
+  showEditor_(e = undefined, _quietInput = undefined) {
+    const editor = this.dropdownCreate_();
```
BeksOmega, line 160: Thank you for separating this out into a separate function! Organized code makes me so happy :D

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+    newGrid[r][c] = newValue;
+    this.setValue(newGrid);
+  }
+
+  /**
+   * Calls a given function for all cells in the image, with the cell
+   * coordinates as the arguments.
+   * @param {*} func a function to be applied
+   */
+  forAllCells_(func) {
```
BeksOmega, line 423: Yesssss a custom iterator! You're my new best friend.

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+  /**
+   * Disposes of events belonging to the slider editor.
+   * @private
+   */
+  dropdownDispose_() {
+    for (const event of this.boundEvents_) {
+      Blockly.browserEvents.unbind(event);
+    }
+    this.sliderInput_ = null;
+  }
```
BeksOmega: Update this to properly dispose of the elements in the `editorPixels` array. They get [cleaned up automatically](https://github.com/google/blockly/blob/d2329f8cba661999cc7f03fcd6c0196770b05cdc/core/dropdowndiv.js#L232), but I believe they need to be dereferenced as well, otherwise there are performance impacts. Setting `this.editorPixels_ = []` should work.

Also update docs.

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+        this.createElementWithClassname_('div', 'dropdownEditor');
+    const pixelContainer =
+        this.createElementWithClassname_('div', 'pixelContainer');
+    dropdownEditor.appendChild(pixelContainer);
+
+    this.bindEvent_(dropdownEditor, 'mouseup', this.onMouseUp_);
+    this.bindEvent_(dropdownEditor, 'mouseleave', this.onMouseUp_);
+
+    this.editorPixels_ = [];
+    for (let r = 0; r < this.imgHeight_; r++) {
```
BeksOmega, line 218: Love the variable names, very helpful!

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+        });
+      }
+      pixelContainer.appendChild(rowDiv);
+    }
+
+    // Add control buttons below the pixel grid
+    this.addControlButton_(dropdownEditor, 'Randomize', this.randomizePixels_);
+    this.addControlButton_(dropdownEditor, 'Clear', this.clearPixels_);
+
+    this.render_();
```
BeksOmega: I'm confused by this call to render.

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+  /**
+   * Called when a mousedown event occurs within the bounds of a pixel.
+   * @private
+   */
+  onMouseDownInPixel_(r, c) {
+    // Toggle that pixel to the opposite of its value
+    const newPixelValue = 1 - this.getValue()[r][c];
+    this.setPixel_(r, c, newPixelValue);
+    this.mouseIsDown_ = true;
+    this.valToPaintWith_ = newPixelValue;
```
BeksOmega, line 357: I like using the new value to set the paint color. I'm obvs not UX, but seems very intuitive!

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -0,0 +1,464 @@
...
+  /**
+   * Resets mouse state (e.g. after either a mouseup event or if the mouse
+   * leaves the editor area).
+   * @private
+   */
+  onMouseUp_() {
+    this.mouseIsDown_ = false;
+    this.valToPaintWith_ = undefined;
+    this.forAllCells_((r, c) => {
+      this.editorPixels_[r][c].alreadyToggledThisDrag = false;
```
BeksOmega: I don't think this is used anymore.

plugins/field-bitmap/test/field_test.mocha.js
```diff
@@ -0,0 +1,168 @@
...
+   */
+  const validValueTestCases = processTestCases([
+    {title: '3x3 solid', value: [[1, 1, 1], [1, 1, 1], [1, 1, 1]]},
+    {
+      title: '4x4 solid',
+      value: [[1, 1, 1, 1], [1, 1, 1, 1], [1, 1, 1, 1], [1, 1, 1, 1]],
+    },
+    {title: '3x3 checkerboard', value: [[1, 0, 1], [0, 1, 0], [1, 0, 1]]},
+    {
+      title: '4x4 solid',
```
BeksOmega: 4x4 blank?

plugins/field-bitmap/test/field_test.mocha.js
```diff
@@ -0,0 +1,168 @@
...
+    [0, 0, 0, 0, 0], [0, 0, 0, 0, 0], [0, 0, 0, 0, 0], [0, 0, 0, 0, 0],
+    [0, 0, 0, 0, 0],
+  ];
+  /**
+   * Asserts that the field property values are set to default.
+   * @param {FieldBitmap} field The field to check.
+   */
+  const assertFieldDefault = function(field) {
+    assert(
+        JSON.stringify(field.getValue()) === JSON.stringify(defaultFieldValue));
```
BeksOmega: I think you can use [assert.deepEqual](http://aaronsofaly.github.io/chai-docs/api/assert/#method_deepequal) here.

plugins/field-bitmap/test/field_test.mocha.js
```diff
@@ -0,0 +1,168 @@
...
+  /**
+   * Asserts that the field properties are correct based on the test case.
+   * @param {FieldBitmap} field The field to check.
+   * @param {FieldValueTestCase} testCase The test case.
+   */
+  const validTestCaseAssertField = function(field, testCase) {
+    const expectedValue = testCase.value;
+    assert(JSON.stringify(field.getValue()) === JSON.stringify(expectedValue));
+    assert(field.getImageHeight() === expectedValue.length);
+    assert(field.getImageWidth() === expectedValue[0].length);
```
BeksOmega: You can use [assert.equal](http://aaronsofaly.github.io/chai-docs/api/assert/#method_equal) here

plugins/field-bitmap/test/field_test.mocha.js
```diff
@@ -0,0 +1,168 @@
...
+  };
+
+  /**
+   * Re-implementation of the method from TestHelpers that supports array
+   * equality checking.
+   * @return void
+   */
```
BeksOmega: I think this comment can be removed

plugins/field-bitmap/test/field_test.mocha.js
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * Reimplementation of the function from TestHelpers that supports array
+   * equality testing.
```
BeksOmega, line 105: Hmm. Do you think it would be worth modifying the underlying test helper? (not necessarily you doing it, but like, someone doing that haha)

BeksOmega (changes requested):
Overall looks great! Your code organization makes me so genuinely happy :D

The main top-level thing that needs fixes is JSDocs:
  - Every public function needs docs
  - Including a description
  - And descriptions & types for every parameter
  - And a return description & type if it returns something
  - If the function doesn't return anything, you can just leave off the `@return` line.

It's a lot :/ but it's good for people using the plugin. 

Also, some comments related to serialization:
  - Could you add a note to the README that this doesn't support serializing to XML?
  - Could you do a quick manual check that this serializes and deserializes to/from JSON? A 2D array of numbers is a JSON serializable type, so it should be fine. I'd just like to know we checked. I think by default there should be a JSON tab on your `npm run start` page, but if not I can give it a look.

Thanks for all of your hard work on this Greg! It looks so cool :D

plugins/field-bitmap/test/field_test.mocha.js
GregoryCannon: Absolutely, it's not obvious that the test helpers wouldn't support array values, and I think would be an easy enough change.

plugins/field-bitmap/src/field-bitmap.js
GregoryCannon: Was originally needed to set the initial values of the editor pixels. Instead, now it sets the value during their creation.

plugins/field-bitmap/src/field-bitmap.js
GregoryCannon: :D

plugins/field-bitmap/src/field-bitmap.js
GregoryCannon: Thank you!

plugins/field-bitmap/package.json
GregoryCannon: Updated here and everywhere else the string appeared

plugins/field-bitmap/test/field_test.mocha.js
BeksOmega: Filed https://github.com/google/blockly/issues/6180 to track =)

BeksOmega (approved):
Thanks for the fixes! LGTM!

BeksOmega:
Ah one last thing from the top level comment:

> Could you add a note to the README that this doesn't support serializing to XML?

GregoryCannon:
> Ah one last thing from the top level comment:
> 
> > Could you add a note to the README that this doesn't support serializing to XML?

Done!

PR #1143, opened 2022-05-25
feat: add cross-tab-copy-paste plugin type declaration

johnnyoshika:
## Problem

The `cross-tab-copy-paste` plugin doesn't include a type declaration file. As a result, importing it into a TypeScript project results in this error:

image

Searching on [DefinitelyTyped[(https://github.com/DefinitelyTyped/DefinitelyTyped) for a type declaration file yields no results.

## PR Summary:

* Adds a type declaration file in `plugins/cross-tab-copy-paste/src/index.d.ts`
* Adds `"types": "./src/index.d.ts"` to cross-tab-copy-paste's package.json. If all contents of `src` is copied to `dist` during build, than this change is not required. However I'm not familiar with the build pipeline so I added this to instruct code editors to look in the `src` folder for the type declaration file.

## Results

This change allows `cross-tab-copy-paste` plugin to be imported into a TypeScript project and improves intellisense:

image

image

## Note

I'm not familiar with the build system, so it should be confirmed that the `index.d.ts` file gets properly copied before this PR is merged.

### DIFF
--- a/plugins/cross-tab-copy-paste/package.json
+++ b/plugins/cross-tab-copy-paste/package.json
@@ -13,6 +13,7 @@
     "test": "blockly-scripts test"
   },
   "main": "./dist/index.js",
+  "types": "./src/index.d.ts",
   "module": "./src/index.js",
   "unpkg": "./dist/index.js",
   "author": "",
--- a/plugins/cross-tab-copy-paste/src/index.d.ts
+++ b/plugins/cross-tab-copy-paste/src/index.d.ts
@@ -1,0 +1,21 @@
+/**
+ * A Blockly plugin that adds context menu items and keyboard shortcuts
+ * to allow users to copy and paste a block between tabs.
+ */
+export class CrossTabCopyPaste {
+  /**
+   * Initializes the cross tab copy paste plugin. If no options are selected
+   * then both context menu items and keyboard shortcuts are added.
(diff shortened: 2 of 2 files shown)


maribethb:
Hey Johnny, thanks for the PR, and sorry this slipped through the cracks. I think this will need some work with the build system, but I'll have a look this week and follow up.

maribethb:
OK, sorry for the delay. I had to do a bit of reading into what gets published, what should get published, and checking on our webpack config. I think this is actually good to go with no changes. We publish both the contents of `dist` and `src` so the type declarations will be available even though they aren't copied to `dist` since you updated the field in `package.json`. This is similar to how it works in `dev-tools`. Tested this change by:

- Running `npm run build` and then `npm pack` to check the contents of the tarball that would be published on npm, and it does contain the `index.d.ts` file.
- Ran `npm link` in the `dist` directory of this plugin, and then `npm link @blockly/plugin-cross-tab-copy-paste` into the TS repo that Johnny previously provided for debugging (blockly-typescript-libcheck) and observed that the error about missing types is no longer present (and it is present if you don't npm link).

So I think this is good to go and it will get released during our next publish which is likely to be next week. Thanks again for contributing this and sorry it took us a while to look into it!

maribethb (approved):

johnnyoshika:
@maribethb Fantastic, thank you

PR #1148, opened 2022-05-26
feat: adds an automated github action to publish all of the packages.

BeksOmega:
### Description

Implementation of the github action as described in the internal document: https://docs.google.com/document/d/1zy5NSPxcLUKguhCCfNH7trFc3rhFY9g4DfdJmYrljAg/edit?usp=sharing

The npm release token secret has been created in the `release` environment. So this should be gtg assuming I implemented it correctly.

### Testing

Tested the `lerna publish` command with `--no-push` and the below options. This correctly bumped the version number, but I wasn't able to test that releases got correctly published. I think `--no-push` skips that because it doesn't have a commit to associate the release with, because it doesn't get pushed.


.github/workflows/publish.yml
```diff
@@ -0,0 +1,40 @@
...
+      run: npm run build -- -- -- --skip-lint
+
+    - name: Test
+      run: npm run test -- -- -- --skip-lint
+
+    - name: Publish
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.RELEASE_BACKED_NPM_TOKEN }}
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: npm lerna publish --conventional-commits --create-release github --yes
```
maribethb: is this line correct? it should either be `lerna publish` or `npx lerna publish` I would think? (in other words should the `npm` be there in the front?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Setup Node
+      uses: actions/setup-node@v3
+      with:
+        node-version: 16
+
+    - name: NPM install
+      run: npm install
```
maribethb: i think this should probably be `npm ci` instead so it doesn't update dependencies during this step.

maribethb (approved):

.github/workflows/publish.yml
BeksOmega: Thanks for the catch, fixed!

.github/workflows/publish.yml
BeksOmega: Makes sense :D

PR #1147, opened 2022-05-26
fix: update cross-tab-copy-paste plugin peer dependency to support minor and patch updates to Blockly version 8

johnnyoshika:
# Problem

When installing `@blockly/plugin-cross-tab-copy-paste` into a project with `blockly@8.0.1` or `blockly@8.0.2`, npm throws this error:

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: blockly-paste@0.1.0
npm ERR! Found: blockly@8.0.2
npm ERR! node_modules/blockly
npm ERR!   blockly@"^8.0.2" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer blockly@"^7 - 8" from @blockly/plugin-cross-tab-copy-paste@1.0.5
npm ERR! node_modules/@blockly/plugin-cross-tab-copy-paste
npm ERR!   @blockly/plugin-cross-tab-copy-paste@"*" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

A temporary workaround is to use the `--legacy-peer-deps` flag:

```
npm install @blockly/plugin-cross-tab-copy-paste --legacy-peer-deps
```

# Solution

This PR fixes this problem by replacing this peer dependency:

```
  "peerDependencies": {
    "blockly": "^7 - 8"
  },
```

...with this:

```
  "peerDependencies": {
    "blockly": "^7 || ^8"
  },
```

I'm not proficient with npm version dependency ranges, so if there's a better way to set this range, please feel free to change it. I tried `^7 - ^8` to be more consistent with the original code, but it didn't work.

PR #1150, opened 2022-05-26
fix: package versions to support patch releases

### LINKED ISSUE #1149: Update package versions to support patches to v8 of blockly core
**Category**

 - Plugins

**Component**

All of 'em.

**Describe the bug**

When you try to update a plugin while using blockly v8.0.2 npm is not able to resolve the dependency (as reported here: https://github.com/google/blockly-samples/pull/1147).

**Expected behavior**

NPM should be able to resolve the dependencies.

**Additional context**

Instead of a version range like `^7 - 8` we should use `>=7 <9`. And when we release v9 of Blockly it should change to `>=7 <10`.

BeksOmega:
### Description

Closes #1149 except for cross-tab-copy-paste, which is covered by #1147


plugins/cross-tab-copy-paste/package.json
```diff
@@ -44,7 +44,7 @@
     "blockly": "^8.0.0"
   },
   "peerDependencies": {
-    "blockly": "^7 - 8"
+    "blockly": ">=^7 <9"
```
maribethb: this one's weird, does this still work

maribethb (approved):

plugins/cross-tab-copy-paste/package.json
BeksOmega: Fixed!

PR #1159, opened 2022-06-06
fix: expand excludes for plugins directory

GregoryCannon:
This makes it so VSCode Javascript/Typescript features can handle the project size.

Not sure if I'm missing some context as to why this hasn't been updated, or if it just got lost by the wayside.


BeksOmega (approved):
LGTM! Thanks for removing the field-date exclude. I checked the git-blame and it seems that was added by accident.

PR #1160, opened 2022-06-10
fix: requiring the incorrect package file

BeksOmega:
### Description

The migration script worked using `node` but failed using `npx` with the error `Cannot find module ../package-lock.json`. I think requiring `package.json` instead will fix the issue. But this needs to be tested after releasing (because we can't test with `npx` before release.


maribethb (approved):

BeksOmega:
Tested and confirmed that this fix worked :D

Issue #1174, opened 2022-06-21
bitmap field is missing JsDoc


maribethb:
**Category**

 - Plugins

**Component**

field-bitmap

**Describe the bug**

Many methods in the plugin are missing jsdoc for the parameters.

**To Reproduce**

Run `npm run lint` and observe lint errors

**Expected behavior**

**Screenshots**

**Additional context**

Priyankcoder:
Hi @maribethb, Can you please assign this to me? I am working on it.

maribethb:
Fixed by #1176 , thanks @Priyankcoder !

PR #1176, opened 2022-06-24
fix(#1174): bitmap field JsDoc and formatting

Priyankcoder:
This PR fix issue #1174. Also, formatted the file by running the lint fix command. Now the field-bitmap file has no linting issue.
image

### DIFF
--- a/plugins/field-bitmap/src/field-bitmap.js
+++ b/plugins/field-bitmap/src/field-bitmap.js
@@ -24,7 +24,12 @@ const EMPTY_PIXEL_COLOR = 'white';
  * Includes a grid of clickable pixels that's exported as a bitmap.
  */
 export class FieldBitmap extends Blockly.Field {
-  /** Constructor for the bitmap field. */
+  /**
+   *Constructor for the bitmap field.
+   * @param {Array.<Array<number>>} value 2D rectangular array of 1s and 0s
+   * @param {Function} validator A function that is called to validate
+   * @param {Object} config config A map of options used to configure the field
+   */
   constructor(value = undefined, validator = undefined, config = undefined) {
     super(value, validator, config);
 
@@ -69,7 +74,7 @@ export class FieldBitmap extends Blockly.Field {
    * @nocollapse
    */
   static fromJson(options) {
-    return new FieldBitmap((options && options['value']), undefined, options);
+    return new FieldBitmap(options && options['value'], undefined, options);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -24,7 +24,12 @@ const EMPTY_PIXEL_COLOR = 'white';
  * Includes a grid of clickable pixels that's exported as a bitmap.
  */
 export class FieldBitmap extends Blockly.Field {
-  /** Constructor for the bitmap field. */
+  /**
+   *Constructor for the bitmap field.
+   * @param {Array.<Array<number>>} value 2D rectangular array of 1s and 0s
```
maribethb: Annotating types can be tricky! There's a few relevant rules here from the [Google JS style guide](https://google.github.io/styleguide/jsguide.html#jsdoc-nullability)
- Non-primitive types (like `Array` or `HTMLElement`) need to be marked with a `!` to denote that they aren't allowed to be null. Primitives (like string or number) don't need to be marked that way, so most of the rest of this file is fine, there's just the `Array`s and `Object`  that need to be marked.
- we don't use the `.` when specifying generics (you might see it in some of our old code but it shouldn't be in new code)
- Optional parameters get a `=` at the end. You can tell they are optional because they provide a default value in the constructor signature.

So overall this type would be `!Array<!Array<number>>=`

and the other parameters should be fixed up similarly

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -332,7 +352,7 @@ export class FieldBitmap extends Blockly.Field {
 
   /**
    * Constructs an array of zeros with the specified width and height
-   * @return the new value
+   * @return {!Array.<Array<?>>}the new value
```
maribethb: In this array, we know the values are all 0s, so instead of `?` you can use `number` like

`!Array<!Array<number>>`

plugins/field-bitmap/test/field_test.mocha.js
```diff
@@ -11,8 +11,8 @@ const {FieldBitmap, DEFAULT_HEIGHT, DEFAULT_WIDTH} = require('../src/index');
 /**
  * Helper method to reformat raw test cases to a format that the
  * various tests can use.
- * @param {*} testCases
- * @return the processed test cases
+ * @param {Array.<Object>} testCases unformatted testcase
```
maribethb: this looks like a stray comment added here

maribethb (commented):
Overall this looks great! It will be super helpful to not have lint warnings in our CI again. :)

I have just a few comments. Annotating types can be tricky so feel free to ask if you have questions about any of them. Thanks!

Priyankcoder:
Thanks a lot for the reviews @maribethb. I am improving the code as per your suggestions.

Priyankcoder:
Hey @maribethb @BeksOmega updated the code as per your suggestions. Please let me know if it needs some more changes, I'll be happy to do so.

maribethb:
Awesome, this looks great. Thank you for the contribution!

PR #1181, opened 2022-07-14
fix: undo and redo in plus-minus blocks

### LINKED ISSUE #1161: block-plus-minus plugin breaks undo/redo
**Category**

 - Plugins

**Component**

block-plus-minus

**Describe the bug**

The undo/redo doesn't work correctly after adding/removing user-defined fields (arguments) with block-plus-minus plugin.

**To Reproduce**

1. Open [@blockly/block-plus-minus Demo](https://google.github.io/blockly-samples/plugins/block-plus-minus/test/).
2. Add and move around some blocks; check that undo/redo is working correctly by pressing `ctrl+z`.
3. Add "procedure" block.
4. Add/remove some user-defined arguments by pressing `+` / `-` button.
5. Try to undo last actions with `ctrl+z`.
6. Notice that change of arguments list wasn't undone.
7. Notice error messages in console.

Problem can be reproduced with "condition" block with the same way.

**Expected behavior**

Undo/redo should work correctly; changes of arguments list should be undone on undo action.

**Screenshots**

Error messages

**Additional context**

Probably there are some issues with serialization. I've noticed some chunks of XML inside of undo stack of workspace after modifying blocks fields with plugin (see at `newValue` and `oldValue` fields on screenshot). There are no such XML chunks when using vanilla blocks, without block-plus-minus plugin.

Part of undo stack

BeksOmega:
### Description

Fixes #1161 

So normally in core the mutator fires change events. That doesn't work for the +/- blocks because... they don't use mutators. So the +/- fields have to fire the change events.

This means that the +/- fields have to duplicate the behavior of the mutator, which changed with the addition of the new serialization system. This pulls those changes into the plugin.

### Additional Information

This is not sustainable.


plugins/block-plus-minus/src/procedures.js
```diff
@@ -210,6 +210,7 @@ const procedureDefMutator = {
    * @this {Blockly.Block}
    */
   mutationToDom: function(isForCaller = false) {
+    console.trace('mutation to dom');
```
gonfunko: Remove this and the other added logging in this file?

gonfunko (approved):

Issue #882, opened 2021-09-14
Remove version-probing checks (and old-version support) from dev-tools


cpcallen:
**Category:**  Plugins

**Component:** dev-tools

**Cleanup to be done:**

In the process of doing the `goog.module` migration of Blockly, it was necessary to add code to dev-tools to probe for various changes that were made in core.  Once the next release (end of 2021Q3) of Blockly is out these can be removed.

**Describe the solution you'd like:**

* In `plugins/dev-tools/package.json`:
  * [ ] Bump required version of `blockly` as appropriate (e.g. `"blockly": "^5.202109__._"`)
* In `plugins/dev-tools/src/index.js`:
  * [ ] Remove probe for `Blockly.utils.global.globalThis` (and always use it) (see #881).
* In `plugins/dev-tools/src/block_test_helpers.mocha.js`:
  * [ ] In `runSerializationTestSuite`, remove probe for `Blockly.utils.idGenerator`; always stub `Blockly.utils.idGenerator.TEST_ONLY`  (see #876)

PR #1183, opened 2022-07-14
fix: bump the blockly peer dependency in dev-tools and remove version probing

BeksOmega:
### Description

Work on #882 Also fix https://github.com/google/blockly/pull/6204

Requires us to bump devtools to only work with v8+.

### DIFF
--- a/plugins/dev-tools/package.json
+++ b/plugins/dev-tools/package.json
@@ -57,7 +57,7 @@
     "blockly": "^8.0.0"
   },
   "peerDependencies": {
-    "blockly": ">=7 <9"
+    "blockly": ">=8 <9"
   },
   "publishConfig": {
     "access": "public",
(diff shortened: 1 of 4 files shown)


maribethb:
If this requires us to update the dev-tools dependency then do that in this PR and mark this as a breaking change, then lgtm

maribethb (approved):

BeksOmega:
The id generator fix was already reviewed by Aaron in #1184 No need to re-review. Just moving everything into one PR.

maribethb:
This should be a breaking change. If someone is using blockly v7 then we would not want npm to automatically update them to this version of the plugin since it is no longer compatible. It has to be a major update so that they manually update when needed. You can probably still update the title of this PR to affect lerna's reading of the commit, but I'm not sure.

And since this would be a major version update of dev-tools I think that means we need to manually update all of the other plugins to depend on the new version because lerna won't keep them up to date after a major change is what we learned last time. So we should take extra care after the next samples release (which I won't be here next week)

BeksOmega:
> This should be a breaking change.

Yeah :/ whoops.

> You can probably still update the title of this PR to affect lerna's reading of the commit, but I'm not sure.

Pretty sure it won't, because lerna will just be looking at the git history :/ Will revert and reapply with the correct conventional commits.

> So we should take extra care after the next samples release (which I won't be here next week)

Sounds good, made a note to myself.

PR #1184, opened 2022-07-14
fix: idGenerator stub

BeksOmega:
### Description

Work on https://github.com/google/blockly-samples/issues/882

Requires us to bump devtools to only work with v8+.


gonfunko (approved):

BeksOmega:
Didn't realize this would also need to bump the peer dependency when I posted it. Moving this into #1183

Issue #857, opened 2021-08-02
continuous-toolbox not created function 


YangJiawei:
**Describe the bug**

https://google.github.io/blockly-samples/plugins/continuous-toolbox/test/

No created ‘abcd’ block.
**Expected behavior**
created abcd block

**Screenshots**
![20210802181607](https://user-images.githubusercontent.com/24451400/127846383-a8e022b6-9523-49f9-b7ac-701743eccd62.png)

BeksOmega:
> I have use continuous-toolbox,Is there a simple solution?

 There is no great solution at the moment. I did come up with a work-around that uses the events system:
```javascript
myWorkspace.addChangeListener((e) => {
  if (e.type == Blockly.Events.BLOCK_CREATE
      || e.type == Blockly.Events.BLOCK_DELETE
      || e.type == Blockly.Events.BLOCK_MOVE) {
    myWorkspace.getToolbox().refreshSelection();
  }
});
```

But it is laggy and when you drag procedures from the toolbox it temporarily shows bad callers blocks (probably because we're not accounting for insertion markers). Sadly this needs a real fix with dev time dedicated to it :/

If you do go with this temporary solution, make sure to add this listener *before* deserializing any saved state! Eg before `Blockly.Xml.domToWorkspace`.

To test this in the demo you can open the console and add the following code:
```javascript
Blockly.mainWorkspace.addChangeListener((e) => {
  if (e.type == Blockly.Events.BLOCK_CREATE
      || e.type == Blockly.Events.BLOCK_DELETE
      || e.type == Blockly.Events.BLOCK_MOVE) {
    Blockly.mainWorkspace.getToolbox().refreshSelection();
  }
});
```

I hope that work-around works for you temporarily!

YangJiawei:
Is there a better way now?Has this bug been fixed?

BeksOmega:
I added the change listener fix as a temporary solution. If we ever get procedure-specific events, the change listener should be modified to use those instead.

BeksOmega:
Ok, the change listener I implemented kind of worked. Callers are being added to the toolbox when defs are created, but when the procedure defs are mutated the callers in the toolbox are not updated to match.

PR #1188, opened 2022-07-15
fix: procedure callers not appearing in continuous toolbox

BeksOmega:
### Description

Temporary fix for #857 so external devs don't have to register the change listener themselves.

If we ever get procedure-specific events, this plugin should be changed to use those instead, so I'm leaving that issue open.

### DIFF
--- a/plugins/continuous-toolbox/src/ContinuousToolbox.js
+++ b/plugins/continuous-toolbox/src/ContinuousToolbox.js
@@ -30,7 +30,6 @@ export class ContinuousToolbox extends Blockly.Toolbox {
     flyout.recordScrollPositions();
 
     this.workspace_.addChangeListener((e) => {
-      console.log('triggered');
       if (e.type === Blockly.Events.BLOCK_CREATE ||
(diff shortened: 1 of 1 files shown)


plugins/continuous-toolbox/src/ContinuousToolbox.js
```diff
@@ -28,6 +28,14 @@ export class ContinuousToolbox extends Blockly.Toolbox {
     const flyout = this.getFlyout();
     flyout.show(this.getInitialFlyoutContents_());
     flyout.recordScrollPositions();
+
+    this.workspace_.addChangeListener((e) => {
+      console.log('triggered');
```
maribethb: please remove console.log statement

maribethb (approved):

plugins/continuous-toolbox/src/ContinuousToolbox.js
BeksOmega: Sorry about that. In the future I will be more careful about checking the code before submitting the PR.

PR #1189, opened 2022-07-15
fix: removed unnecessary .tgz file from field-slider plugin

btw17:
### Description

Removed an unnecessary .tgz file from the field-slider plugin


BeksOmega:
Oh forgot to say, thank you for the find and the fix!

PR #1192, opened 2022-07-19
chore: convert field-slider plugin to TypeScript

btw17:
### Description

Updated `plugins/field-slider` to use TypeScript for its src code and demo test code. 

**Note:** do not submit until pending discussion notes are reviewed.

**Note:** the `test/*.mocha.js` files are left unchanged since the dev scripts do not support TS currently.

### Discussion Notes

1. Should `noFocus`, `quietInput_`, or nothing be provided for `showEditor_`? See the original [field_slider.js](https://github.com/google/blockly-samples/blob/master/plugins/field-slider/src/field_slider.js#L81). (It seems like `quietInput_` should be provided, though the exported type either needs to be fixed or I need to over set up the file to ignore the type there.)
  - The inheritance path ([blockly/core/field_number.js](https://github.com/google/blockly/blob/master/core/field_number.js) -> [blockly/core/field_textinput.js](https://github.com/google/blockly/blob/master/core/field_textinput.js#L298) -> [blockly/core/field.js](https://github.com/google/blockly/blob/master/core/field.js#L718)) would not properly use `noFocus` as a second param. It seems **field_textinput** uses the second param as a boolean, like with `quietInput_`.
  - Although `quietInput_` is passed in as a param, it seems like it's not being used
  - The type information only includes `showEditor_` from [blockly/core/field.js](https://github.com/google/blockly/blob/master/core/field.js#L718)'s types, excluding the same field in [blockly/core/field_textinput.js](https://github.com/google/blockly/blob/master/core/field_textinput.js#L298)'s types. Because of this, TypeScript does not see a valid use of `super.showEditor` with a second param provided.
2. **ESLint** is configured to want **camelCase** or **PascalCase**. This is broken by values ending in `_`, like `showEditor_`. Would it be fine to update it like `protected override showEditor_(` -> `protected showEditorInternal(` and just call the parent's internal function in place of super, or should the the name remain as it was? (If it should remain, we can just leave ESLint ignored for those lines as I currently have.)
3. Anybody know if there's a VSCode setting to indent overflowing parameters further than the line? 😂 For example, my formatter wants to align ([in this example](https://github.com/btw17/blockly-samples/blob/master/plugins/field-slider/src/field_slider.js#L36)) `value` and `precision` with `super`, which breaks the linting rules.
4. I've added some TODO notes for adding more precise type information in place of `Object` and `Function`. Any thoughts for or against the proposed TODOs?
5. For versioning, should this change to `3.1.0` instead of `3.0.19`?

Final note, I still need to update the package to properly use the built js files instead of using `./src/index.js`, and I'll need to test the published package to make sure the type information is properly included and the default script is properly used.


BeksOmega:
Thanks for the pull request! This looks awesome :D I haven't done and in-depth look at the code (I'll leave that to @cpcallen for now). But I do have opinions about the discussion topics =)

> Should noFocus, quietInput_, or nothing be provided for showEditor_?

I think getting rid of `noFocus` is correct because that is handled by the [FieldTextInput](https://github.com/google/blockly/blob/61322294da70c770b32abb28b9d695af75f61a16/core/field_textinput.js#L302).

Wrt `_quietInput` I think that should be passed to `super.showEditor_`. Pretty sure not passing it was a bug.

Wrt the types being unhappy, hopefully that should be fixed once we convert core. Is that currently blocking?

> ESLint is configured to want camelCase or PascalCase. This is broken by values ending in _, like showEditor_. Would it be fine to update it like protected override showEditor_( -> protected showEditorInternal(

Sadly that would be a breaking change :/ Modifying `@protected` functions could break people who are overriding classes. So unless we want this to be a major bump it has to stay the same, and eslint should be silenced.

Personally I think if we're going to fix this, we should probably fix it in core too so that it's consistent. But I also don't think we want to change this in core right now. So my guess is it should stay the same. @maribethb can give a final opinion.

> Anybody know if there's a VSCode setting to indent overflowing parameters further than the line?

I think I just turned auto indent off haha

> I've added some TODO notes for adding more precise type information in place of Object and Function. Any thoughts for or against the proposed TODOs?

I like all of them! I think I actually have a branch somewhere that fixes the field config types in core (I remember writing out interfaces for all of them). And we have a [tracking issue](https://github.com/google/blockly/issues/6075) for the options problem.

We like to make sure all `TODO`s have associated github issues though (eg `// TODO (123): Fix the types or whatever`. So as long as no one else disagrees with them, issues should be filed against core before this gets merged.

> For versioning, should this change to 3.1.0 instead of 3.0.19?

Versioning is actually [done automatically based on the commit message/PR title](https://developers.google.com/blockly/guides/modify/contribute/commits). So right now with the `feat` keyword, it will get bumped to `3.1.0` automatically. However, I don't think this actually adds a feature (it just changes the internal structure of the file) so it should probably be `chore` (even though that sounds much less exciting haha).

---

Thanks again for the PR! It's so exciting to see typescript actually happening in Samples :D

btw17:
> Wrt the types being unhappy, hopefully that should be fixed once we convert core. Is that currently blocking?

Nope! :) I created a type in the meantime to tell the compiler "I know you think this other thing is correct, but listen to me instead"

> Sadly that would be a breaking change :/ Modifying `@protected` functions could break people who are overriding classes. So unless we want this to be a major bump it has to stay the same, and eslint should be silenced.

That makes sense about the `protected` functions. Even though the `private` functions should be safe to change, I'll leave those as is for now as well until it's decided to update the naming convention and what that convention should be.

> We like to make sure all `TODO`s have associated github issues though (eg `// TODO (123): Fix the types or whatever`. So as long as no one else disagrees with them, issues should be filed against core before this gets merged.

Works for me! I'll leave them as TODOs for now and look into spinning up issues for them later if there's no issues with those issues 😉 

> Versioning is actually [done automatically based on the commit message/PR title](https://developers.google.com/blockly/guides/modify/contribute/commits). So right now with the `feat` keyword, it will get bumped to `3.1.0` automatically. However, I don't think this actually adds a feature (it just changes the internal structure of the file) so it should probably be `chore` (even though that sounds much less exciting haha).

Good to know regarding versioning! Neat that it's done automatically. Without changing function names or parameters, a patch version bump should be no big deal since it shouldn't break anything. 

Also, I'll update this when I get the chance to `chore` instead of `feat`.

btw17:
Closing this in favor of an updated pull request (#1217)

PR #1202, opened 2022-07-22
fix: scroll-options when drag surface is disabled

HollowMan6:
Patch for integration of https://github.com/mit-cml/workspace-multiselect/pull/1

Signed-off-by: Hollow Man <hollowman@opensuse.org>

### DIFF
--- a/plugins/scroll-options/src/index.js
+++ b/plugins/scroll-options/src/index.js
@@ -76,8 +76,11 @@ export class ScrollOptions {
       return;
     }
 
-    let element = this.workspace_.getBlockDragSurface().getSvgRoot();
-    if (!Blockly.utils.svgMath.is3dSupported()) {
+    let element;
+    const dragSurface = this.workspace_.getBlockDragSurface();
+    if (Blockly.utils.svgMath.is3dSupported() && dragSurface) {
+      element = dragSurface.getSvgRoot();
+    } else {
       element = this.workspace_.svgGroup_;
     }
 
--- a/plugins/scroll-options/src/index.js
+++ b/plugins/scroll-options/src/index.js
@@ -75,9 +75,14 @@ export class ScrollOptions {
       // Already enabled.
       return;
     }
-    const dragSurface = this.workspace_.getBlockDragSurface();
+
+    let element = this.workspace_.getBlockDragSurface().getSvgRoot();
+    if (!Blockly.utils.svgMath.is3dSupported()) {
+      element = this.workspace_.svgGroup_;
+    }
+
     this.wheelEvent_ = Blockly.browserEvents.conditionalBind(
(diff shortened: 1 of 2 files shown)


BeksOmega:
I've read through this PR, and the info from your [other PR](https://github.com/mit-cml/workspace-multiselect/pull/1), and I'm not following what this is fixing :/ Could you explain what this change is doing and why it's necessary?

HollowMan6:
> I've read through this PR, and the info from your [other PR](https://github.com/mit-cml/workspace-multiselect/pull/1), and I'm not following what this is fixing :/ Could you explain what this change is doing and why it's necessary?

@BeksOmega 

Say that when the drag surface is disabled (`Blockly.utils.svgMath.is3dSupported()` returns false), then the scroll-options plugin won't work without this patch.

As https://github.com/mit-cml/workspace-multiselect/pull/1 make drag surface disabled as the default, and you also mentioned that https://github.com/google/blockly/issues/6160 the drag surface has some performance issue, so I guess it would be the best to merge it into the upstream if someone choose to disable the drag surface while still want to use this plugin.

HollowMan6:
You can check the bugs when you use the monkey patch for disabling drag surface like here: https://github.com/HollowMan6/workspace-multiselect/blob/f2584f477903e2c3789f7f6f43577bc5dc4795ce/src/multiselect_patch.js#L13-L19

image
For wheel scroll, it doesn't work anymore although no error is in the console logs.
For edge scroll, errors will occur:
image

plugins/scroll-options/src/index.js
```diff
@@ -75,9 +75,14 @@ export class ScrollOptions {
       // Already enabled.
       return;
     }
-    const dragSurface = this.workspace_.getBlockDragSurface();
+
+    let element = this.workspace_.getBlockDragSurface().getSvgRoot();
```
maribethb: if the drag surface is not being used, I think `getBlockDragSurface` can return null. so before calling `getSvgRoot` we should check for null, and/or we should only do this check if `is3dSupported` is true.

So i think this is safest?

```js
let element;
const dragSurface = this.workspace_.getBlockDragSurface();
if (Blockly.utils.svgMath.is3dSupported() && dragSurface) {
  element = dragSurface.getSvgRoot();
} else {
  element = this.workspace_.svgGroup_;
}
```

maribethb (approved):
Thank you,  sorry for the delay in reviewing, I needed to recall how this plugin worked to begin with :) Congrats on the multi-select plugin!

One thing to fix then this is ready.

plugins/scroll-options/src/index.js
HollowMan6: Great, thanks for reviewing! Yeah, I agree. It's also a little bit strange as of now, since `getBlockDragSurface` doesn't return null even when it is not supported.

maribethb (approved):

PR #1204, opened 2022-07-22
chore: bump typescript from ^3.2.1 to ^4.7.4 in dev-scripts 

btw17:
### Description 

Bumps the `typescript` peer dependency from `^3.2.1` to `4.7.4` in dev-scripts

This should be updated since the previous version is quite outdated at this point. TypeScript [3.2.1](https://www.npmjs.com/package/typescript/v/3.2.1) was published 4 years ago, and `^3.2.1` prevents using `4+` versions. The first `4+` version, [4.0.2](https://www.npmjs.com/package/typescript/v/4.0.2), was published 2 years ago.

### Discussion

- We could do an earlier 4.X.X version that will still allow using the latest typescript dependencies. For converting other blockly plugins to TS, I plan to use the [override](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#override-and-the---noimplicitoverride-flag) feature from `4.3.X` to provide compiler safety if inherited functions drift from the parents, though all that matters is the version rule for this dev-scripts peer dependency supports this


btw17:
Discussing this with @maribethb, I'm going to instead update TS to `^4.3.2` to keep in sync with [google/blockly](https://github.com/google/blockly/blob/master/package.json#L100). To keep things easy I'm going to close this pull request and spin up a new one with that change.

btw17:
See #1210 for the pull request with the finalized change

PR #1205, opened 2022-07-22
fix: don't allow anything in the bitmap dropdown to be dragged

GregoryCannon:
Small fix to prevent UI elements in the bitmap dropdown from being dragged (and showing a globe icon under the mouse)


BeksOmega (approved):
LGTM! Thanks for the fix :D

PR #1208, opened 2022-07-29
chore: update root TS dependency to align with the google/blockly

btw17:
The TypeScript version in [google/blockly](https://github.com/google/blockly/blob/master/package.json#L100) supports `^4.3.2`, so this package should support the same.

This will be needed to:
1) safely update the `dev-scripts` plugin to support later TS
2) upgrade plugin packages to TS (which rely upon `dev-scripts`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
                 "json-to-pretty-yaml": "^1.2.2",
                 "lerna": "^4.0.0",
                 "rimraf": "^3.0.2",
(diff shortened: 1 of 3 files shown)


btw17:
After further review, it looks like the typescript package here isn't used at all. I'm going to close this request and open a new one with TypeScript removed entirely

btw17:
See #1209 for the follow-up pull request

PR #1209, opened 2022-07-29
chore: remove unused typescript dependency at root

btw17:
### Description

Two things are included in this package.
1. In the root `./package.json` file, I removed the unused `typescript` dependency
2. Elsewhere, `*/package-lock.json` files were updated with more recent blockly package versions


BeksOmega (approved):
So generally I think this is a good move. I don't like having dependencies in the root, unless we have documented reasons for having them there.

For context I think #238 is the original PR where this dependency was added.

Just want to make sure a few things got tested.
  1) Did you run a clean, fresh install, and then build of a plugin? (`npm ci`, `npm install`, `cd some plugin`, `npm run test`)
  2) Did you run `npm run lint` (since that's what the original PR was enabling)?

If all that works then I think it's fine to remove this =) If we run into problems later we can always add it back.

(approving on the condition that test cases pass)

btw17:
Good to know the original context!

1. Yes - tests in plugins still work as expected! (I just spot checked a few this morning)
2. Yes - I also ran `npm run build` and `npm run test` from the root without any issues

PR #1212, opened 2022-08-02
release: update dev-tools to major version and update all plugin dependencies on it

maribethb:
## Background

https://github.com/google/blockly-samples/pull/1191 was merged but lerna did not correctly pick this up as a major version bump. This is either because
a) Something went wrong because this PR was actually after reverting an identical early PR that wasn't marked as breaking change, so we confused lerna
b) Lerna is looking at the individual commits in the PR (none of which is marked breaking) rather than the squashed commit title (which is marked breaking)

Which of these two scenarios it is needs further testing but regardless, samples is currently borked until this PR sorting things out.

## This PR
1. Manually bumps the version of dev-tools to a new major version
2. Manually bumps the version of all plugins to a new minor version (in theory, because they all depend on dev-tools; in reality because lerna always bumps all the packages for a reason I can't figure out). This was done using `lerna version` but with enough command line arguments that it didn't push and tag them automatically, so that's why I say "manually"
3. Updates every plugin to depend on v4.0.1 of dev-tools. The previous version 3.0.9 does not actually exist (it's a relic of the failed update previously) so that's why samples is currently borked.

## Details
All the plugins already relied on blockly@8 in dev-dependencies, which is required now to use this version of dev-tools, except for `field-bitmap` so I updated that plugin's versions and manually tested, everything looks good.

## Next steps
1. Publish all the plugins using the `from-git` command so that lerna doesn't bump them *again*
2. Update the templates in `dev-create` to use the new version of dev-tools. I wanted to do that in a separate publish cycle in case this one doesn't work for some reason, then we don't have inaccurate templates.
3. Update the samples site. This isn't really related to this PR but it needs to be done after this stuff is all taken care of.

### DIFF
--- a/plugins/field-bitmap/package.json
+++ b/plugins/field-bitmap/package.json
@@ -42,13 +42,13 @@
   ],
   "devDependencies": {
     "@blockly/dev-scripts": "^1.2.25",
-    "@blockly/dev-tools": "^3.1.9",
-    "blockly": "^7.20211209.2",
+    "@blockly/dev-tools": "^4.0.1",
+    "blockly": "^8.0.0",
     "chai": "^4.3.6",
     "mocha": "^9.2.1"
   },
   "peerDependencies": {
-    "blockly": "^7.20211209.2"
+    "blockly": ">=7.20211209.2 <9"
   },
   "publishConfig": {
     "access": "public",
--- a/plugins/field-bitmap/package.json
+++ b/plugins/field-bitmap/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@blockly/field-bitmap",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "description": "A field that lets users input a pixel grid with their mouse.",
   "scripts": {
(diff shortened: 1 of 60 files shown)


BeksOmega (approved):
LGTM, just want to confirm that in the ~~"manually bump plugins to their next version" commit, the devscripts dependency was also supposed to be bumped?~~ Nevermind, makes sense to bump it at the same time haha.

plugins/field-bitmap/package.json
```diff
@@ -42,13 +42,13 @@
...
-    "@blockly/dev-tools": "^3.1.9",
-    "blockly": "^7.20211209.2",
+    "@blockly/dev-tools": "^4.0.1",
+    "blockly": "^8.0.0",
     "chai": "^4.3.6",
     "mocha": "^9.2.1"
   },
   "peerDependencies": {
-    "blockly": "^7.20211209.2"
+    "blockly": ">=7.20211209.2 <9"
```
BeksOmega, line 51: Nice catch!

PR #1215, opened 2022-08-04
fix: Remove the deprecated block-extension-tooltip plugin

maribethb:
Removes the `@blockly/block-extension-tooltip` plugin.

This plugin hasn't been compatible with Blockly v7 and above, and keeping the code in the repository creates difficulty for us when testing and when updating github-pages, so it's better to remove it. The plugin has been marked deprecated on npm: https://www.npmjs.com/package/@blockly/block-extension-tooltip

Note: This is obviously a breaking change, but I haven't marked it as such because 
a) The conventional commits don't matter here since this plugin won't ever be published again
b) I am worried that because of the change at the `plugins/` root level, lerna might interpret it as a breaking change for every plugin, which I do not want, and because of the shiny new CI pipeline we don't have a chance to provide input on the versions it chose, and while this deserves further investigation I don't want to tie it up in this PR since it does not matter.

## If you were using this plugin

You can continue to use it for now, but it won't receive any updates. It is not compatible with the latest versions of Blockly. We recommend that you update to Blockly v8, where you can use the new custom tooltip API. That API is simpler than the approach used in this plugin and you will not have to include an extension with every block definition. Find more info
- [at the demo](https://google.github.io/blockly-samples/examples/custom-tooltips-demo/)
- [in the developer docs](https://developers.google.com/blockly/guides/create-custom-blocks/define-blocks#customizing)


BeksOmega (approved):
LGTM assuming CI passes.

PR #1217, opened 2022-08-05
chore: convert field-slider plugin to TypeScript

btw17:
### Description

- Updated `plugins/field-slider` to use TypeScript for its src code and demo test code. 
- Updated `plugins/dev-tools` to require Blockly > 8.0.3 to be compatible.

### Note

- The `blockly.d.ts` export in blockly core is not compatible with this package prior to `8.0.3`.
- The `test/*.mocha.js` files are left unchanged since the dev scripts do not support TS currently.
- The `.d.ts` files for `plugins/field-slider` will not be generated due to an issue with `plugins/dev-scripts`. A fix was found for that issue, though it won't be accepted until the 'core Blockly to TS' is resolved since there is some crossover in the fix's implementation.

### Discussion

- There are some `TODO` notes related to type information for core Blockly packages, though there's a large TS conversion in that incoming package which will likely impact the type information. Should:
  - Individual issues be created for each TODO?
  - One group issue be created for all of those TODOs?
  - Those TODOs be attributed to the incoming TS conversion?
  - The TODOs be removed for now?


BeksOmega:
I'm assigning @maribethb  as well, because this needs her opinions on bumping the `dev-tools` dependency. Bumping the min version of Blockly requires a major bump for the `dev-tools` version, and I just want to double check we're cool with doing that.

btw17:
The build issue is an interesting one. When running a clean install from `plugins/field-slider`, the build works fine. When running a clean install from the root, though, there's an issue with the blockly package for building. (The resulting `plugins/field-slider/package-lock.json` also varies based on where the install is from.) My guess is it's related to how lerna is linking the packages, though I'll keep investigating.

BeksOmega (changes requested):
Can you insert a commit that changes the file extensions `.js` to `.ts` without actually modifying the contents of the files? That way we can ensure that git understands we're just renaming files, not deleting them and creating new ones, which is important for maintaining our `git blame`.

btw17:
> Can you insert a commit that changes the file extensions `.js` to `.ts` without actually modifying the contents of the files? That way we can ensure that git understands we're just renaming files, not deleting them and creating new ones, which is important for maintaining our `git blame`.

Done!

btw17:
After discussion this morning, the general plan is to delay submitting this change until after the major Blockly version bump where it's converted mostly to TypeScript. I'll close this PR in the meantime and open a new PR at that time.

rachel-fenichel:
This will need the change in #1299 as well.

btw17:
Looks like `blockly-scripts` doesn't work with the `override` keyword. I removed it from `render_` and `showEditor_`, though I'd originally added it to those since an override comment was present.

In TypeScript, `override` ensures that the method we're trying to override exists in the parent. This is pretty handy if, for example, the `render_()` method were updated to `renderInternal()` or something similar since the compiler would catch it. I'm not sure if this is a blocking issue, though. (See the [TypeScript 4.3 documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#override-and-the---noimplicitoverride-flag) for more about override.)

It's also probably worth investigating why `blockly-scripts` doesn't work with override.

btw17:
For the failing tests, it looks like the updated Blockly imports break outside of my changes.

For example, going to `blockly-samples/plugins/block-plus-minus/test/if.mocha.js`, `Blockly.Dart` does not exist.

plugins/field-slider/src/field_slider.ts
```diff
@@ -10,155 +10,165 @@
...
+/**
+ * Options used to define a field slider from JSON.
+ */
+export interface FieldSliderOptions extends FieldSliderConfig {
+  value?: string | number;
+}
+
+/* eslint-disable @typescript-eslint/ban-types */
+/**
+ * NOTE: `Function` is banned by eslint. Eventually, a more precise
```
BeksOmega, line 29: Can you file a bug for this in core for fixing this in `field.ts` and add a note to fix in plugins as well?

BeksOmega (approved):
LGTM! The comment I left below is nonblocking. So I'm going to go ahead and get this merged! :D

plugins/field-slider/src/field_slider.ts
btw17: Done! Here's the issue: https://github.com/google/blockly/issues/6505

PR #1219, opened 2022-08-11
fix: update workflow to fetch all tags for lerna publish

maribethb:
The checkout workflow by default only fetches the most recent commit. This behavior is why the lerna version command in CI says "assuming all versions changed" and just publishes a patch release for every plugin.

See https://github.com/lerna/lerna/issues/2542

(the second `git fetch` command is no longer required since the `checkout` workflow will now fetch tags if you set the fetch-depth to 0, according to its readme.)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,6 +18,9 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


BeksOmega (approved):
LGTM! Might be nice to put a comment on the lerna issue about not needing the separate `fetch` call (for future peoples).

PR #1221, opened 2022-08-17
feat: add codelab demonstrating validation and warnings

### LINKED ISSUE #1133: Create a Warnings codelab
**Category**

 - Codelabs

We had requests at the 2022 Blockly Summit for more info about [Warnings](https://developers.google.com/blockly/reference/js/Blockly.Warning) and [setWarningText](https://developers.google.com/blockly/reference/js/Blockly.BlockSvg#setWarningText). It seemed like many people hadn't discovered it, but thought it would be a cool feature to add to their application once they had seen it.

A codelab explaining what warnings are, and giving an example of how to use them. The example should probably either apply warnings based on [validators](https://developers.google.com/blockly/guides/create-custom-blocks/fields/validators) or [move events](https://developers.google.com/blockly/guides/configure/web/events#blocklyeventsblock_move) (since those are the situations we normally see warnings being applied in).

Some docs about warnings, but I'm not sure where they would go.

**Additional context**

@maribethb I think this relates to the work you're doing wrt docs.

johnnesky:
Fixes #1133. I wrote up a codelab demonstrating how to validate fields, or display warnings when validation can't automatically fix problems.

The codelab explains how to create a custom block representing a range between two numbers and outputs a list of integers. Then it explains how to automatically fix problems with individual fields using validators, then how to display a warning message and disable a block if errors can't be corrected automatically.


codelabs/validation_and_warnings/validation_and_warnings.md
```diff
@@ -0,0 +1,280 @@
...
+    ],
+    'output': 'Array',
+    'style': 'list_blocks',
+  },
+]);
+```
+
+This definition indicates that this block has two numerical fields called `FIRST` and `LAST`, and that it outputs a list. Later on, we'll refine this definition to add validation to it.
+
+However, if you reload `index.html` and open the toolbox, you won't find the new block in it yet. To be able to use the new block in a workspace, the toolbox definition needs to include it. Find the toolbox definition in `index.js` and insert this code at the beginning of the list of available blocks, right before the one named `controls_forEach`, making sure that the `'type'` here matches the custom block's `'type'`:
```
maribethb: They need to understand the basics of creating and using custom blocks before taking this codelab. So I would simplify this section to just instruct them to add it to the toolbox definition before checking the browser.

codelabs/validation_and_warnings/validation_and_warnings.md
```diff
@@ -0,0 +1,280 @@
...
+});
+```
+
+Reload `index.html` now, then drag the custom block to your workspace and try setting the `LAST` field to various values. Notice how it always turns into an odd number!
+
+## Displaying warnings
+
+Both Blockly's built-in validators and custom validators are nice because they immediately correct any errors so that there should never be any interruption in the validity of the blocks in the workspace or in the validity of the code that it generates. This results in a smooth, pleasant experience for the user, and you should take advantage of these validators whenever possible. 
+
+However, there may be invalid conditions that can't be corrected automatically because it's ambiguous what the desired result is. For example, it doesn't make much sense for our custom block to have a `FIRST` field with a greater value than the `LAST` field, but it's not obvious which of the two fields is "wrong". The best we can do is warn the user about the problem, and let them decide how to fix it.
```
maribethb: super nit: "wrong." (period inside quotes)

codelabs/validation_and_warnings/validation_and_warnings.md
```diff
@@ -0,0 +1,280 @@
...
+```js
+  // Validate the entire block whenever any part of it changes,
+  // and display a warning if the block cannot be made valid.
+  this.setOnChange(function(event) {
+    const first = this.getFieldValue('FIRST');
+    const last = this.getFieldValue('LAST');
+    const valid = (first <= last);
+    this.setWarningText(valid
+      ? null
+      : `The first number (${first}) cannot be greater than the last number (${last}).`);
```
maribethb: nit (especially since this is an example and not a real error): tell the user what to do, not what not to do. "The first number must be smaller than the last number."

codelabs/validation_and_warnings/validation_and_warnings.md
```diff
@@ -0,0 +1,280 @@
...
+
+![A Blockly workspace with disabled break and continue blocks.](disabled_break.png)
+
+You can disable a block using `this.setEnabled(false)`, although there are some caveats: disabled blocks can't be dragged out of the toolbox flyout, and the act of disabling a block usually adds an event to Blockly's undo history. That's probably not the behavior you want when validating a block, so you can avoid both of these effects with the following code, which you should put inside the change listener function after setting the warning text:
+
+```js
+    // Disable invalid blocks (unless it's in a toolbox flyout,
+    // since you can't drag disabled blocks to your workspace).
+    if (!this.isInFlyout) {
+      const group = Blockly.Events.getGroup();
```
maribethb: `initialGroup` to clarify that afterwards we are resetting the state?

codelabs/validation_and_warnings/validation_and_warnings.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+Reload `index.html` one last time, drag out the custom block, and edit the fields to make the `FIRST` field greater than the `LAST` field again. This time, the block should be disabled, and it won't generate code even if it is combined with other blocks:
+
+![A Blockly workspace with disabled custom blocks.](disabled_block.png)
+
+## Summary
+In this codelab, you learned:
+* How to use the field validators in Blockly's JSON API.
+* How to create extension to access Blockly's JavaScript API.
```
maribethb: "extensions" or "an extension"

I'm not too sure about "to access Blockly's JavaScript API." What about "How to create an extension to add functionality to a block."?

examples/validation-and-warnings-codelab/complete-code/index.html
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
+<html>
+  
+<head>
+  <meta charset="utf-8">
+  <title>Validation and Warnings Codelab</title>
+  <script src="https://unpkg.com/blockly/blockly.min.js"></script>
+  <script src="https://unpkg.com/@blockly/dev-tools@2.0.0/dist/index.js"></script>
+  <script src="index.js"></script>
```
maribethb, line 8: Are the dev-tools actually being used here? I don't see where. If not, remove them.

If they are, v2.0.0 is out of date; we are on v4 now. but since `blockly` is not versioned here, in the future this could get into a state where the latest version of blockly is not compatible with the exact version of dev-tools listed here. so they should either both be versioned or neither of them (since hopefully the latest version of dev-tools is always compatible with the latest version of blockly).

maribethb (commented):
Overall lgtm, thanks for this!

examples/validation-and-warnings-codelab/complete-code/index.html
johnnesky: Done, I don't think it's used here.

FYI, the custom toolbox codelab, which I copied this from, has the same outdated version and potential version compatibility problem, see: https://github.com/google/blockly-samples/blob/master/examples/custom-toolbox-codelab/starter-code/index.html

maribethb (approved):

PR #1227, opened 2022-08-20
fix: clean up links and documentation

johnnesky:
* In codelabs/README.md I fixed links to use proper relative paths.
* In codelabs/custom_generator/custom_generator.md I removed underscores from parameters... I see that the canonical names of these parameters in https://developers.google.com/blockly/reference/js/Blockly.Generator#scrub_ contain underscores, but the following example in the codelab does not and I don't see why a user would choose to include the underscores unless it helped avoid shadowing some other symbol in their code. Alternatively, if you think it makes more sense, we could add the underscores back to the codelab but then we should add them to the other example in the codelab as well.
* In codelabs/custom_toolbox/custom_toolbox.md I fixed a broken link to the complete code for the custom toolbox codelab.
* In examples/custom-toolbox-codelab/README.md I fixed a broken link to the codelab.
* In examples/README.md I added a link to my new validation and warnings codelab (which hasn't been deployed to https://blocklycodelabs.dev/ yet).
* In examples/turtle-field-demo/index.html I fixed a broken link to documentation on customizing fields.
* In several codelab starter and complete index.html files, I either deleted a link to https://unpkg.com/@blockly/dev-tools@2.0.0/dist/index.js if it was unused, or I removed the (outdated) version number, so that it would use the latest version which would be more likely to be compatible with the latest version of blockly that was loaded.

### DIFF
--- a/examples/context-menu-codelab/complete-code/index.html
+++ b/examples/context-menu-codelab/complete-code/index.html
@@ -5,7 +5,7 @@
   <meta charset="utf-8">
   <title>Context Menu Codelab</title>
   <script src="https://unpkg.com/blockly/blockly.min.js"></script>
-  <script src="https://unpkg.com/@blockly/dev-tools@2.0.0/dist/index.js"></script>
+  <script src="https://unpkg.com/@blockly/dev-tools"></script>
   <script src="./index.js"></script>
 
   <style>
--- a/codelabs/README.md
+++ b/codelabs/README.md
@@ -23,12 +23,12 @@ You can find additional details on the
 Put your image assets associated with your codelab in the same folder as your
 markdown file. Any images you add can be linked from the Codelab using:
 ```md
-![image_caption](./codelabexample.png
+![image_caption](codelabexample.png)
 ```
 
 
 ## Resources
 
-[Sample Codelab](/sample.md)
+[Sample Codelab](sample.md)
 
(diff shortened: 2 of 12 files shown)


examples/context-menu-codelab/complete-code/index.html
```diff
@@ -5,7 +5,7 @@
   <meta charset="utf-8">
   <title>Context Menu Codelab</title>
   <script src="https://unpkg.com/blockly/blockly.min.js"></script>
-  <script src="https://unpkg.com/@blockly/dev-tools@2.0.0/dist/index.js"></script>
```
BeksOmega, line 8: @maribethb was there a reason we wanted to lock these?

BeksOmega (commented):
Thanks for finding these! Generally LGTM :D Just going to wait for Maribeth to see if she has context about locking the dev-tools version.

examples/context-menu-codelab/complete-code/index.html
maribethb: No, I think the first codelab just happened to do it and all the other ones just copied it, which is why it's in some of the codelabs that don't even use it. v2 is already incompatible with certain changes in latest blockly so we've just gotten lucky that things continue to work right now.

BeksOmega (approved):

PR #1228, opened 2022-08-23
chore: some changes to improve release process

maribethb:
1. Update lerna to v5. v5 of lerna is supposed to contain the fix that causes the package-locks to be incorrectly updated. We'll see if this is true :)
2. Removed the workaround that tried to updated the package-locks (it didn't work anyway) (the `version` script)
3. Pass `--ignore-scripts` in the github workflow to publish packages. This will mean we don't re-build each plugin during the publish step (it is already built earlier in the workflow). should speed up automated plugin publishing.
4. Reconfigure the manual publishing scripts.
    - `publish:dryrun` previously didn't do a lot. worst, it would not even catch errors in plugins that would later break the publishing step. now it clones blockly-samples, runs all of the builds and tests, and logs into the npm publishing service.
    - `publish:manual` now uses the correct lerna command that will use conventional commits and create the github release just as if you had used the automated workflow
    - added a `publish:unpublishedOnly` script which can be used to recover from errors
    - added a `publish:checkVersions` script which will create but not push the versions. useful for debugging.
    - I left `publish:force` alone but I hope that we don't need it anymore.
5. Updated the script documentation. I've also updated our internal documentation to match the new steps and will send that out separately.

I tested the `publish:prepare` and `publish:checkVersions` scripts but did not test the other ones because I did not want to actually publish anything. I think we should try the manual publish this week instead of using the automated one. The only thing I am not sure about is the authentication for creating the github release. I think you will just be prompted for your github un/pw but I'm not certain.

If these changes are successful then this is the last major change I plan to make to the publishing process in samples. 🥳

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -91,6 +91,22 @@ function prepareForPublish(done) {
 }
 
 /**
+ * Exit early if the release directory does not exist.
+ * @param {string} releaseDir release directory to check
+ * @param {Function} done Gulp callback.
+ */
+function exitIfNoReleaseDir(releaseDir, done) {
+  // Check that release directory exists.
+  if (!fs.existsSync(releaseDir)) {
+    console.error(
+        `No release directory ${releaseDir} exists. ` +
+            `Did you run 'npm run publish:prepare'?`);
+    done();
+    process.exit(1);
+  }
+}
+
+/**
  * This script does not log into the npm publish service. If you haven't run
  * the prepare script recently, publishing will fail for that reason.
  * @param {boolean=} force True for forcing all plugins to publish, even ones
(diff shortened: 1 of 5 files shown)


gulpfile.js
```diff
@@ -49,75 +49,147 @@ function checkLicenses() {
...
-    if (fs.existsSync(releaseDir)) {
-      console.log('Removing previous `dist/` directory.');
-      rimraf.sync(releaseDir);
+    // Check that release directory exists.
+    if (!fs.existsSync(releaseDir)) {
+      console.err(
+          `No release directory ${releaseDir} exists. ` +
+          `Did you run 'npm run publish:prepare'?`);
+      process.exit(1);
     }
```
BeksOmega: Could you pull this out into its own function?

BeksOmega (approved):
LGTM! Love all of the docs :D Just one code health thing.

PR #1234, opened 2022-09-02
feat: created plugin for colour hsv sliders field.

### LINKED ISSUE #356: Can i use color slider for LED?
Hi guys I'm new to here and i'm not good at English. 

 I wanna choose LED color with color slider but i can't find the way to do that.

I found Scratch 3.0 has the function that i want but i don't know how to apply it to the blockly.
Untitled

I found that there's no field_colour_slider.js in blockly/core/ so i paste it from scratch but it doesn't work. I hope someone helping me. thank you

johnnesky:
Fixes #356. I created a Blockly plugin using the existing "field_slider" plugin as a reference. It defines a class called FieldColourHsvSliders that extends Blockly.FieldColour and is registered as `'field_colour_hsv_sliders'`. 

When the field is clicked, it displays a dropdown container with three HSV sliders, and each slider's track is a gradient that represents how the color would change if you dragged that slider. There's also an eyedropper button below the sliders that activates the browser's eyedropper feature if available, and falls back on an `input` element of `type="color"`. The field mostly ignores the color swatch options from the parent class, although I haven't overridden the behavior of using the first swatch as the default value. This field doesn't add any new configuration options.

This code uses trailing underscores on private fields and methods but I could change that if desired. It also uses Google Closure Compiler comment syntax. 

I added unit tests for setting the field's value via constructor, JSON, or setValue(). I haven't added any tests for click events, are there any examples of something like that?


plugins/field-colour-hsv-sliders/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@blockly/field-colour-hsv-sliders",
```
BeksOmega, line 2: Thanks for correcting the package.json!

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,639 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Colour input field with HSV sliders.
+ * @author nesky@google.com (John Nesky)
```
BeksOmega: Feel free to remove this if you don't want bots to scrape it.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,639 @@
...
+class RgbColour {
+  /**
+   * The RgbColour constructor.
+   * @param {number=} r The initial amount of red. Defaults to 0.0.
+   * @param {number=} g The initial amount of green. Defaults to 0.0.
+   * @param {number=} b The initial amount of blue. Defaults to 0.0.
+   * @constructor
+   */
+  constructor(r = undefined, g = undefined, b = undefined) {
+    this.r = r || 0.0;
```
BeksOmega: Nit: In JS you can just write `0`. All numbers are implicitly floats =)

Here + everywhere below.

BeksOmega: Public properties also need JSDoc.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,639 @@
...
+   * Updates the properties of this instance to copy the provided HsvColour.
+   * @param {!HsvColour} other An HSV representation of a colour to copy.
+   * @return {!HsvColour} This instance after updating it.
+   */
+  copy(other) {
+    this.h = other.h;
+    this.s = other.s;
+    this.v = other.v;
+    return this;
+  }
```
BeksOmega, line 171: Love filling out the API =)

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,639 @@
...
+
+  /**
+   * Sets the hue to the provided value.
+   * @param {number} h The hue of the colour.
+   * @return {!HsvColour} This instance after updating it.
+   */
+  setHue(h) {
+    this.h = h;
+    return this;
```
BeksOmega: Method chaining!

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,639 @@
...
+
+/**
+ * Class for a colour input field that displays HSV slider widgets when clicked.
+ * @extends {Blockly.FieldColour}
+ * @alias Blockly.FieldColourHsvSliders
+ */
+export class FieldColourHsvSliders extends Blockly.FieldColour {
+  /**
+   * Class for an number slider field.
+   * @param {(string|!Sentinel)=} value The initial value of the
```
BeksOmega:
```suggestion
   * @param {(string|!Blockly.Sentinel)=} value The initial value of the
```

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,639 @@
...
+  constructor(
+      value = undefined, validator = undefined, config = undefined) {
+    super(value, validator, config);
+
+    /**
+     * Array holding info needed to unbind events. Used for disposing.
+     * @type {!Array<?>}
+     * @private
+     */
+    this.boundEvents_ = [];
```
BeksOmega, line 251: I think it makes sense to leave the `_`s since this is in JS. There's no harm in renaming them if/once we convert this to TS, because they're private anyway. I.e. it's not a breaking change.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,639 @@
...
+   */
+  showEditor_() {
+    this.createDropdownSliders_();
+    Blockly.DropDownDiv.getContentDiv().appendChild(this.dropdownContainer_);
+
+    Blockly.DropDownDiv.showPositionedByField(
+        this, this.dropdownDisposeSliders_.bind(this));
+
+    // Focus so we can start receiving keyboard events.
+    this.hueSlider_.focus({preventScroll: true});
```
BeksOmega, line 316: Are you receiving keyboard events?

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,639 @@
...
+
+  /**
+   * Creates the colour picker slider editor and adds event listeners.
+   * @private
+   */
+  createDropdownSliders_() {
+    const container = document.createElement('div');
+    container.classList.add('fieldColourSliderContainer');
+
+    function addLabel(name) {
```
BeksOmega: Generally it's "better" to break inner functions out into private methods. Justification being: (1) not everyone knows you can define inner functions in JS, (2) methods are slightly more efficient than inner functions b/c they don't need to be redefined for every invocation.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,639 @@
...
+      slider.type = 'range';
+      slider.min = String(0);
+      slider.max = String(max);
+      slider.step = String(step);
+      container.appendChild(slider);
+      return slider;
+    }
+
+    this.hueReadout_ = addLabel('Hue');
+    this.hueSlider_ = addSlider(360, 5);
```
BeksOmega: Did you try making this value smaller? (eg 1, 2, or 3)? Having 72 distinct hues is *probably* fine, but if we can give people more I'd love to :D

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,639 @@
...
+    const tempHsv = new HsvColour();
+    const tempRgb = new RgbColour();
+    const trackRadius = '8px';
+
+    let hueGradient = 'linear-gradient(to right, ';
+    tempHsv.copy(hsv).setHue(0.0/6.0);
+    hueGradient += tempRgb.loadFromHsv(tempHsv).toHex() +
+        ' ' + trackRadius + ', ';
+    tempHsv.copy(hsv).setHue(1.0/6.0);
+    hueGradient += tempRgb.loadFromHsv(tempHsv).toHex() + ', ';
```
BeksOmega: I think the stateful string building is a bit difficult to parse. Maybe you could separate this into a helper function and modify the below code to use template strings? Eg `hsvToHex(hue, sat, vib): string`

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,639 @@
...
+  /**
+   * Updates the value of this field and editor sliders using an eyedropper.
+   * @param {?Event} event Unused.
+   * @private
+   */
+  onEyedropperEvent_(event) {
+    const eyeDropper = new window.EyeDropper();
+    eyeDropper.open().then((result) => {
+      this.setValue(result.sRGBHex);
+      this.updateSliderValues_();
```
BeksOmega, line 481: Instead of calling this explicitly, I think you can use `doValueUpdate_`.

plugins/field-colour-hsv-sliders/README.md
```diff
@@ -0,0 +1,72 @@
...
+### npm
+```
+npm install @blockly/field-colour-hsv-sliders --save
+```
+
+## Usage
+This plugin adds a field type `FieldColourHsvSliders` that is registered to the name `'field_colour_hsv_sliders'`. This field is an extension of the Blockly.FieldColour field. See the [Blockly.FieldColour documentation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/colour#creation) on what parameters and configurations this field supports. However, unlike Blockly.FieldColour, FieldColourHsvSliders does not use colour swatches and thus ignores options related to the swatches such as `'colourOptions'`.
+
+To use it, you'll need to add this field to a block type definition, and add that block to your toolbox. See below for an example of defining a block type that uses this field.
+
```
BeksOmega, line 23: Can you add a comment that even though the UI is HSV the value is `'#rrggbb'`?

BeksOmega (changes requested):
Overall looks lovely! Thank you for your hard work on this :D I absolutely love custom fields so I'm so excited to see this be available to developers.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: Done, I removed the author but left the file overview.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: I still like to add the decimal as an indication to myself when reading code that the number is intended to be used with non-integer math, but that doesn't seem to be common in Blockly so I'll remove the decimal.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: I was following Scratch's example here but there's no reason why the step size can't be smaller, I'll set it to 2.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: Done.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: Done.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: Done.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: Done, I added a private static helper method hsvToHex_.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
BeksOmega: Looks like the github UI hid some of the comments in here =)

plugins/field-colour-hsv-sliders/README.md
johnnesky: Done.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,707 @@
...
+ */
+class RgbColour {
+  /**
+   * The RgbColour constructor.
+   * @param {number=} r The initial amount of red. Defaults to 0.
+   * @param {number=} g The initial amount of green. Defaults to 0.
+   * @param {number=} b The initial amount of blue. Defaults to 0.
+   * @constructor
+   */
+  constructor(r = undefined, g = undefined, b = undefined) {
```
BeksOmega: Instead of setting these to `undefined` by default, and then using `|| 0` below, can you just set them to `0` by default? (same for HsvColor below)

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,707 @@
...
+    /**
+     * The red component of the color, ranging from 0 to 1.
```
BeksOmega: Most of the docs consistently use `colour` (thank you!) but it looks like a few instances of `color` snuck in here.

johnnesky:
Thanks for the feedback! I've pushed an update that I think addresses all your comments.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,707 @@
...
+        FieldColourHsvSliders.SATURATION_SLIDER_MAX;
+    const v = parseFloat(this.brightnessSlider_.value) /
+        FieldColourHsvSliders.BRIGHTNESS_SLIDER_MAX;
+
+    // The gradient control point positions should align with the center of the
+    // slider thumb when the corresponding color is selected. When the slider is
+    // at the minimum or maximum value, the distance of center of the thumb from
+    // the edge of the track will be the thumb's radius, so that's how far the
+    // minimum and maximum control points should be.
+    const thumbRadius = '8px';
```
BeksOmega: Is this controlled by the CSS  below, or is this the browser default? If it's controlled by the CSS could you make this a constant and use it in both places?

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
```diff
@@ -0,0 +1,707 @@
...
+    // colours of the rainbow.
+    let hueGradient = 'linear-gradient(to right, ';
+    hueGradient += FieldColourHsvSliders.hsvToHex_(0/6, s, v) +
+        ` ${thumbRadius}, `;
+    hueGradient += FieldColourHsvSliders.hsvToHex_(1/6, s, v) + ', ';
+    hueGradient += FieldColourHsvSliders.hsvToHex_(2/6, s, v) + ', ';
+    hueGradient += FieldColourHsvSliders.hsvToHex_(3/6, s, v) + ', ';
+    hueGradient += FieldColourHsvSliders.hsvToHex_(4/6, s, v) + ', ';
+    hueGradient += FieldColourHsvSliders.hsvToHex_(5/6, s, v) + ', ';
+    hueGradient += FieldColourHsvSliders.hsvToHex_(6/6, s, v) +
```
BeksOmega, line 523: Thank you! I find this much more readable :D

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: Is this a yea or a nay?

Also I got rid of this particular method because it wasn't needed anymore.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: The sliders are implemented as HTML input elements of type="range", which have built-in keyboard handlers for the arrow keys. When an arrow key is pressed, the value changes, which results in the same input events as clicking on the slider, which the code is already handling.

BeksOmega (changes requested):

BeksOmega:
> I added unit tests for setting the field's value via constructor, JSON, or setValue(). I haven't added any tests for click events, are there any examples of something like that?

Nope! We don't generally do UI tests for fields, we just test them manually.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
BeksOmega: It was a yay :D I love method chaining.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: Done.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: Done, I created the constant and used it in the CSS.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.js
johnnesky: Done,

BeksOmega (approved):
LGTM! Just confirm that you're happy with it and I'll get this merged =)

johnnesky:
Confirmed! Thanks for the help!

Issue #1267, opened 2022-09-10
Blockly v9 causes `TypeError: Class constructors cannot be invoked without 'new'` in many plugins + demos due to ES5-style subclasses


cpcallen:
**Categories**

 - Plugins
 - Examples
 - (Codelabs?)

**Components**

So far I have found (but not yet fixed) this error in:
* [x] [`plugins/field-date`](https://google.github.io/blockly-samples/plugins/field-date/test/)
* [x] [`examples/pitch-field-demo`](https://google.github.io/blockly-samples/examples/pitch-field-demo/)
* [x] [`examples/turtle-field-demo`](https://google.github.io/blockly-samples/examples/turtle-field-demo/)
* [ ] Maybe others?

**Describe the bug**

**To Reproduce**

1. Link the demo or plugin-test against Blockly v. `9.0.0-beta.1`
   * Or use the version so linked on GitHub Pages (as of 2022-09-08) via the URLs above
0. Open the browser console.
0. See error (e.g. from field-date):
   ```
   Uncaught (in promise) TypeError: Class constructors cannot be invoked without 'new'
       at new Blockly.FieldDate (field_date.js?fdfc:51:45)
       at Blockly.FieldDate.fromJson (field_date.js?fdfc:71:10)
       at Object.fromJsonInternal$$module$build$src$core$field_registry [as fromJsonInternal] (field_registry.ts?42f1:82:22)
       at fromJson$$module$build$src$core$field_registry (field_registry.ts?42f1:61:20)
       at BlockSvg$$module$build$src$core$block_svg.fieldFromJson_ (block.ts?6432:1813:33)
       at BlockSvg$$module$build$src$core$block_svg.interpolate_ (block.ts?6432:1713:28)
       at BlockSvg$$module$build$src$core$block_svg.jsonInit (block.ts?6432:1554:12)
       at BlockSvg$$module$build$src$core$block_svg.eval [as init] (common.ts?52f0:237:10)
       at BlockSvg$$module$build$src$core$block_svg.doInit_ (block.ts?6432:287:14)
       at new BlockSvg$$module$build$src$core$block_svg (block_svg.ts?8904:208:10)
   ```
0. Note also that Blockly has in consequence failed to load/inject successfully.

**Expected behaviour**

No errors occur.

**Additional context**

[...]

PR #1276, opened 2022-09-16
fix: update pitch field to es6 class

rachel-fenichel:
Part of https://github.com/google/blockly-samples/issues/1267

Converts the pitch field example to an ES6 class with const and let. Mostly leaves the rest of the code in place.

### DIFF
--- a/examples/pitch-field-demo/field_pitch.js
+++ b/examples/pitch-field-demo/field_pitch.js
@@ -27,6 +27,11 @@ var CustomFields = CustomFields || {};
  * @constructor
  */
 class FieldPitch extends Blockly.FieldTextInput {
+  /**
+   * All notes available for the picker.
+   */
+  static NOTES = 'C3 D3 E3 F3 G3 A3 B3 C4 D4 E4 F4 G4 A4'.split(/ /);
+
   constructor(text) {
     super(text);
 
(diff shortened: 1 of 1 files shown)


NeilFraser:
The only issue is with the Closure Compiler.  It doesn't support public class fields:
`field_pitch.js:33:2: ERROR - [JSC_LANGUAGE_FEATURE]`

I just moved the offending field to the bottom of the file, don't know if there's a better solution:
```
/**
 * All notes available for the picker.
 */
FieldPitch.NOTES = 'C3 D3 E3 F3 G3 A3 B3 C4 D4 E4 F4 G4 A4'.split(/ /);
```

NeilFraser:
Also, change @return to @returns?

rachel-fenichel:
I opened up the field in the demo page to test it and everything worked fine. Where is it being fed to Closure Compiler?

NeilFraser:
> Where is it being fed to Closure Compiler?

In Blockly Games:
https://github.com/google/blockly-games/blob/master/appengine/music/src/field_pitch.js

rachel-fenichel:
Since you've already applied fixes on the Blockly Games side, and since we're possibly getting rid of Closure Compiler there too, I think it's fine to leave the public fields in here. What do you think?

NeilFraser (approved):

PR #1290, opened 2022-09-16
fix: Updating graph-demo to use json style block defs, resolves 1258

### LINKED ISSUE #1258: Update graph-demo to use JSON toolbox definitions instead of xml
**Category**
 - Examples

**Component**

graph-demo

Many of the examples and plugins use XML toolboxes instead of the new JSON API. We should have our example code use the new APIs to best model what other developers can do with Blockly.

Update the code in blockly-samples to use the JSON toolbox API.

**Additional context**

To work on this issue, first read the [background information](https://developers.google.com/blockly/guides/contribute/samples) about how to contribute to samples and verify your changes.

1. Learn about [defining a toolbox](https://developers.google.com/blockly/guides/configure/web/toolbox) and read about the XML format vs the JSON format.
2. In blockly-samples, navigate to the graph-demo example's directory. Some examples can be started with `npm run start` and some can be opened directly by opening the `index.html` file in a browser. The README may give specific instructions, or you can look in the `package.json` to see what scripts are defined. Note you may need to run `npm install` in the example directory. Either way, open the example's demo page to observe the toolbox and its contents.
3. Write a new copy of the toolbox definition in the JSON format instead of the XML format. Where exactly it should be declared will depend on the example. It will most likely make sense to declare the toolbox definition close to the location where it is used, i.e. where `Blockly.inject` is called.
4. Refresh the demo page and make sure the toolbox looks the same as it did before. You can compare against the version in master on the [samples site](https://google.github.io/blockly-samples/) to make sure the behavior hasn't changed. Note that not all demos are published on the samples site so it may not be there.
5. Follow the rest of the instructions in the Contributing docs to verify your change and open a PR.

quoo:
- Converts both toolbox xml and startBlocks xml to JSON
- Moves deletable flag to custom block definition as it is not supported in json (https://github.com/google/blockly/issues/2285)
- updates blockly version

fixes #1258

### DIFF
--- a/examples/graph-demo/index.html
+++ b/examples/graph-demo/index.html
@@ -335,7 +335,6 @@ const startBlocks = {
     "blocks": [{
       "kind": "block",
       "type": "graph_set_y",
-      "deletable": false,
       "x": 100,
       "y": 100,
       "inputs": { 
--- a/examples/graph-demo/index.html
+++ b/examples/graph-demo/index.html
@@ -55,119 +55,6 @@
     ...
   </div>
 
-  <xml xmlns="https://developers.google.com/blockly/xml" id="toolbox" style="display: none">
-    <category name="Math" colour="%{BKY_MATH_HUE}">
-      <block type="math_number">
-        <field name="NUM">123</field>
-      </block>
-      <block type="math_arithmetic">
-        <value name="A">
-          <shadow type="math_number">
-            <field name="NUM">1</field>
-          </shadow>
-        </value>
-        <value name="B">
-          <shadow type="math_number">
-            <field name="NUM">1</field>
-          </shadow>
-        </value>
-      </block>
-      <block type="math_single">
-        <value name="NUM">
-          <shadow type="math_number">
-            <field name="NUM">9</field>
-          </shadow>
-        </value>
-      </block>
-      <block type="math_trig">
-        <value name="NUM">
-          <shadow type="math_number">
-            <field name="NUM">45</field>
-          </shadow>
-        </value>
-      </block>
-      <block type="math_constant"></block>
-      <block type="math_number_property">
-        <value name="NUMBER_TO_CHECK">
-          <shadow type="math_number">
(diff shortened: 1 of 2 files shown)


quoo:
This should resolve https://github.com/google/blockly-samples/issues/1258

examples/graph-demo/index.html
```diff
@@ -201,6 +88,297 @@
...
+    }
+  ]
+};
+
+const startBlocks = {
+  "blocks": {
+    "blocks": [{
+      "kind": "block",
+      "type": "graph_set_y",
+      "deletable": false,
```
maribethb: I think this can be removed since it doesn't apply in json

examples/graph-demo/index.html
```diff
@@ -343,9 +524,9 @@
       {collapse: false,
        disable: false,
        media: 'https://unpkg.com/blockly/media/',
-       toolbox: document.getElementById('toolbox')});
-  Blockly.Xml.domToWorkspace(document.getElementById('startBlocks'),
-                             Graph.workspace);
+       toolbox: toolbox});
+
```
maribethb: These extra spaces may be a lint error when CI finishes running:)

maribethb (commented):
Looks great, just a couple small things! Thanks so much for the PR!

examples/graph-demo/index.html
maribethb: well, maybe not?

maribethb: we don't actually lint examples 😅

examples/graph-demo/index.html
quoo: Ah, darn, I thought I removed that.

examples/graph-demo/index.html
quoo: I removed them:D

maribethb (approved):

rachel-fenichel:
@quoo Welcome to Blockly, and thanks for the fix!

PR #1291, opened 2022-09-16
feat: add error callback to cross-tab copy-paste

### LINKED ISSUE #1218: Add error callback to the cross tab copy paste plugin
**Category**

 - Plugins

**Component**

cross-tab-copy-paste

There is the possibility that two unrelated projects could both have the cross tab copy paste plugin installed. This would allow people to try to paste blocks into a workspace where the definitions for those blocks do not exist (causing an error). External developers may want to handle those errors (e.g. presenting a message to the user when they happen). But currently this is not possible because there is no where for external developers to wrap the input event listeners in a `try... catch`.

External developers should be able to pass a callback to the cross-tab-copy-paste plugin that is called when the plugin triggers an error when pasting.

N/A

**Additional context**

N/A

debaraj-barua:
fixes: #1218

### DIFF
--- a/plugins/cross-tab-copy-paste/README.md
+++ b/plugins/cross-tab-copy-paste/README.md
@@ -32,7 +32,9 @@ const options = {
 
 // Initialize plugin.
 const plugin = new CrossTabCopyPaste();
-plugin.init(options);
+plugin.init(options, () => {
+    console.log('Use this error callback to handle TypeError while pasting');
+  });
 
 // optional: Remove the duplication command from Blockly's context menu.
 Blockly.ContextMenuRegistry.registry.unregister('blockDuplicate');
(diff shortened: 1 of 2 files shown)


debaraj-barua:
Can someone please re-run the pipeline?

alschmiedt (commented):
This looks good to me! Just one thing before I merge, would you mind updating the [README](https://github.com/google/blockly-samples/blob/master/plugins/cross-tab-copy-paste/README.md) to include an example of how you can use this new callback? I think you can just update the usage section.

debaraj-barua:
@alschmiedt  Done, PTAL

alschmiedt (approved):
LGTM, thanks for picking up this issue! 🎊

PR #1295, opened 2022-09-20
fix!: update field-date to es6 class

rachel-fenichel:
Fixes part of https://github.com/google/blockly-samples/issues/1267

Notes:
- I had to move the default value after calling the super constructor, because I can't use `this` before the super. I think this is a behaviour change, and I'm not sure the best way around it: skip initialization? Call `setValue` again after the super constructor if `value` was null? Make `DEFAULT_VALUE` static? Note that it still works when the user clicks the `today` option in the calendar dropdown, but I think that's a different code path.
- This update originally failed because closure compiler was not outputting es2015 code, so I updated it in the gulpfile.
  - I tried to just update `google-closure-compiler` and `google-closure-library` to the latest version, but the compiler complained about having too low an input version. That might be because of code in the newest version of `google-closure-library`. I can test this by only updating the compiler, and/or by setting the input version to something newer.


plugins/field-date/gulpfile.js
```diff
@@ -59,6 +59,8 @@ function build() {
         entry_point: './src/field_date.js',
         js_output_file: 'date_compressed.js',
         language_in: 'ECMASCRIPT_2015',
+        // TODO: look into updating closure compiler--this should just be ES NEXT
```
cpcallen: I note that you subsequently deleted this, but: it wasn't clear which entry this comment relates to.  I think setting `language_in` to `ECMASCRIPT_NEXT ` is reasonable (default is `STABLE`) but we should probably be conservative in choosing `language_out`, and ES6 is indeed the reasonable choice here.

plugins/field-date/src/field_date.js
```diff
@@ -26,380 +28,378 @@ goog.require('goog.i18n.DateTimeSymbols');
...
+
+.blocklyDatePicker button:hover,
+.blocklyDatePicker .goog-date-picker-date:hover {
+  background-color: rgb(218, 220, 224, .5);
+}
+`);
+
+Blockly.fieldRegistry.register('field_date', FieldDate);
+
+Blockly.FieldDate = FieldDate;
```
cpcallen: This works only because we transpile Blockly to a non-ESM module, and is unlikely to work in future.  Is it not sufficient to add it to the registry?

plugins/field-date/src/field_date.js
```diff
@@ -26,380 +28,378 @@ goog.require('goog.i18n.DateTimeSymbols');
...
-      new goog.date.Date().toIsoString(true);
+  constructor(value = undefined, validator = undefined, textEdit = false) {
+    super(value, validator);
+
+    /**
+     * The default value for this field (current date).
+     * @type {*}
+     * @protected
+     */
+    this.DEFAULT_VALUE = (new goog.date.Date()).toIsoString(true);
```
cpcallen: As far as I can tell, setting `DEFAULT_VALUE` here is useless:

- It appears only to be read by the `Field` constructor, which has already been invoked.
- The `Field` constructor never uses `this.DEFAULT_VALUE`, but instead uses `new.target` to read `FieldDate.prototype.DEFAULT_VALUE` (which is not set, so falls back to `FieldTextInput.prototype.DEFAULT_VALUE`, which is `''`).

I don't think there's any reason to try to set this on the instance at all.

The minimally-invasive fix is probably to change the `super` call:

```
    super(value ?? (new goog.date.Date()).toIsoString(true), validator);
````

This (like the original design) has the issue that multiple `FieldDate`s on the same page could get different initial values if the page is being loaded at midnight.  To avoid that the default value could be a module-local variable or (since this is still a `goog.provides` module) a static property on the `FieldDate` constructor.

Since we're already making breaking changes here, I can't help but think it might be better to fix the whole design by having `Field`s get their initial value from a static property on the constructor or by calling a static or instance method.  But that would be a much larger and more invasive change.

**Review by cpcallen — CHANGES_REQUESTED** [review]

File: plugins/field-date/src/field_date.js (hunk 3 above)
Comment (rachel-fenichel) [reply plugins/field-date/src/field_date.js]: Yeah, that placement of default value is useless, as the tests told me.

The problem here is that default value is used by the base class if validation fails. So using that change to the super call doesn't work, because it doesn't handle cases where the input date was invalid.

The tests all pass if I just move it to a static property on the class. 

But as you pointed out, that fails if the page is loaded at midnight (or, more commonly, if the page is left open for a long time) and I'm very leery of making quiet changes in how default dates work. I'll look at the alternatives you suggested.

File: plugins/field-date/src/field_date.js (hunk 2 above)
Comment (rachel-fenichel) [reply plugins/field-date/src/field_date.js]: If someone is using it through `Blockly.FieldDate` they'll be broken if I delete this--but they'll be able to use it through the registry instead. That is, it won't disappear, but they will have to change code.

**Comment by rachel-fenichel** [comment]
Okay, I think I've got this fully sorted. Changes here:
- Make test work when run outside of the US
- Update the field to an ES6 class
- Change how we're calling the output wrapper to make it return the date field class, not Blockly

There are still failing tests. That's because I rebased onto #1297, which already had failing tests. The counts are the same:
```
162 passing (980ms)
  2 pending
  35 failing
  ```

cpcallen (approved):

plugins/field-date/src/field_date.js
maribethb: Rachel, do you remember why you ended up deleting this? I think this did end up breaking people using `Blockly.FieldDate`.

rachel-fenichel: Adding properties directly to the Blockly module from outside of Blockly will eventually fail if/when we seal the module, and I thought the workaround (using it from the registry) was sufficient.

PR #1297, opened 2022-09-21
chore!: Update plugins to `blockly` version `9.0.0`

cpcallen:
This branch is intended to collect all the changes to the `blockly-samples` repository needed to update samples to Blockly v9.

- Use `clang-format` to fix (whitespace) formatting of `package.json` and `package-lock.json` files so they are consistent.
- (breaking change) Bump `blockly` `peerDependencies` for `plugins/*/` from `<9` to `^9`.
    - The reason we are doing this unilaterally for all plugins is that we don't test plugins with older releases of Blockly. Some of the plugins had to be changed to be compatible with v9, and some of them may already be broken on older releases that they claim to be compatible with, because when we change the plugins we don't check which version of Blockly an API was added in, etc.
- Bump `blockly` `devDependencies` for `plugins/*/` to `^9.0.0`.
- Fixes the generator imports to match v9 style for all plugins using generators (most in tests)
- Fixes field-slider tests that used to use private properties
- Fixes field defaults not being declared correctly
- Fixes keyboard navigation tests that relied on monkeypatching Blockly
- Updates usage of deprecated functions from Blockly

🚧 If you're seeing this PR from release notes about breaking changes, know that if you've upgraded to Blockly v9 and are using any plugins, you should also update the plugin to the next major version. There are no other breaking changes in this release of any plugins. If you haven't yet upgraded to Blockly v9 but are using plugins, don't upgrade to the next major version of the plugin until you've updated Blockly.

~~Note that because of the way lerna handles package dependencies within the monorepo, this PR must not be released using the standard release process. Do not click the auto-release button until the release including this change has already been handled manually.~~ This concern seems not to apply anymore after careful testing with `lerna version`.


rachel-fenichel (commented):
Changes so far make sense. It's a lot smaller if I ignore the whitespace changes!

rachel-fenichel:
Current breakages:
- field-date (fixed by #1295)
- field-bitmap (probably fixed by https://github.com/google/blockly/pull/6453 but needs confirmation)
- blocks-plus-minus (fixed by #1305)
- generator imports in dev tools (fixed by #1305)
- keyboard-navigation

rachel-fenichel:
Remaining issues:
- Keyboard nav tests inject a workspaceSvg, but it's actually using jsdom. The tests fail on these lines in `core/renderers/marker_svg.ts`:
```
    // Ensures the marker will be visible immediately after the move.
    const animate = this.currentMarkerSvg!.childNodes[0];
    if (animate !== undefined) {
      animate instanceof SVGAnimationElement && animate.beginElement();
    }
```

My proposed solution is to swap it back from an `instanceof` check to a cast or `AnyDuringMigration`, at least for release. That requires a change in core, a beta, and installing the beta over here.

- Keyboard nav monkeypatches functions on fields and then [tries to call functions on the superclass](https://github.com/google/blockly-samples/blob/blocky-9-beta/plugins/keyboard-navigation/src/navigation_controller.js#L116) in a way that's no longer allowed. 

I don't have a workaround yet.

rachel-fenichel:
The next issue is also related to running rendered workspaces in jsdom for testing. Canvas doesn't exist, and Blockly uses it for text width computation.

`Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package)`

The old workaround was to monkeypatch `Blockly.dom.utils.getFastTextWidthWithSizeString`, but that's no longer accessible for monkeypatching.

Based on the error message, the new workaround is to install a package to add this.

rachel-fenichel:
@maribethb fyi I just updated this to a full PR instead of a draft, and I think it's ready to go:
- Every commit has been reviewed individually
- Initial changes are primarily whitespace
- We've tested the results of these changes on my gh-pages.

package.json
```diff
@@ -1,49 +1,52 @@
...
+    "rimraf": "^3.0.2"
+  },
+  "publishConfig": {
+    "access": "public",
+    "registry": "https://wombat-dressing-room.appspot.com"
+  },
+  "eslintConfig": {
+    "extends": "@blockly/eslint-config"
+  },
+  "dependencies": {
```
maribethb: The blockly dependency in root of samples should be removed.

maribethb:
Normally after the release, we'd update the branch to the full non-beta 9.0.0 release before merging the branch. That should be done first as we don't want the main version of plugins to have a dependency on a beta version.

maribethb (approved):

PR #1299, opened 2022-09-21
fix: slider field dispose

### LINKED ISSUE #1272: Slider Field
**Category**
 Examples

**Component**
field-slider

**Describe the bug**
Cannot edit the slider

**To Reproduce**

Try to add a parent component with a slider as connected entry. One time you did it, update the value with the slider remembering the original value and affect the same value that originaly. Then, you will see it's no possible to edit the slider field anymore and all others too.

Test in the sandbox here : https://google.github.io/blockly-samples/plugins/field-slider/test/

**Expected behavior**
After editing a value, if it's wrong we could change it always as we wish.

rachel-fenichel:
Fixes #1272 

The issue was that on dropdown dispose we unbound events but didn't clear the array containing the bind info.

I didn't fix any other issues, because https://github.com/google/blockly-samples/pull/1217 makes extensive changes to the code as it updates to TypeScript.

### DIFF
--- a/plugins/field-slider/src/field_slider.js
+++ b/plugins/field-slider/src/field_slider.js
@@ -136,6 +136,7 @@ export class FieldSlider extends Blockly.FieldNumber {
     for (const event of this.boundEvents_) {
       Blockly.unbindEvent_(event);
     }
+    this.boundEvents_.length = 0;
     this.sliderInput_ = null;
   }
 


gonfunko (approved):

plugins/field-slider/src/field_slider.js
```diff
@@ -136,6 +136,7 @@ export class FieldSlider extends Blockly.FieldNumber {
     for (const event of this.boundEvents_) {
       Blockly.unbindEvent_(event);
     }
+    this.boundEvents_.length = 0;
```
btw17, line 139: Interesting tech! I hadn't realized you could clear an array in this way.

PR #1302, opened 2022-09-23
feat: created plugin to reify shadow blocks.

### LINKED ISSUE #878: Add a plugin to convert shadow blocks to real blocks when they are modified
**Category**
 - Plugins

**Component**

Not sure what a good name for this is.

It seems some projects prefer having real blocks rather than shadow blocks. Having the shadows turn into real blocks is nice because it makes it seem like your values are "filling in".

A plugin that turns shadow blocks into real blocks when the blocks are edited.

N/A

**Additional context**

Some complications that would need to be handled:
* Undo  and redo
* Shadow blocks with shadow children

Also, we may not want this as a first-party plugin. But I'm filing here to track it.

The original tracking issue for this is:  https://github.com/google/blockly/issues/311
A PR that was previously created for this is: https://github.com/google/blockly/pull/319

johnnesky:
Fixes #878. This plugin exports a function `shadowBlockConversionChangeListener` that, when added to a workspace as a change listener, detects BlockChange events on shadow blocks and responds by converting the block (and any parent shadow blocks via output or previous connections) into real blocks. 

It also exports a new undoable BlockShadowChange event class. If this plugin is ever merged into the main repository, the new event type can probably be merged with BlocklyChange events, and the new change listener function logic can probably be mostly moved into Blockly.Field.prototype.setValue. 

The change listener attempts to group the new shadow change events with the initiating block change event. ~~As a hack, it retroactively overwrites the group property of the initiating event if it doesn't already have a group name, in order to be able to group events with it. We can delete this hack, but then changes to~~ boolean fields of blocks will not be grouped with the shadow change events for undo purposes, because those events do not have group names by default. Other field types such as text and number will still be grouped with shadow change events. 

As a subtlety, the change listener ignores changes that occur while undoing or redoing. Only new user events should result in shadow blocks getting converted to real blocks.

See also the closed PR https://github.com/google/blockly/pull/319 to add this feature to the main Blockly repository. It came with a feature flag to enable the feature, which may be useful if this plugin is ever merged into the main repository.


johnnesky:
Obviously I'm wrestling with test failures in my PR, although I don't know how to reproduce the failure locally. It seems to have something to do with me using async functions in the test.mocha.js file, resulting in the error message: "ReferenceError: regeneratorRuntime is not defined". I see some articles online suggesting additions to the webpack configuration. Should I try to mess with that? 

I noticed the existing test file https://github.com/google/blockly-samples/blob/master/plugins/migration/test/rename.mocha.js uses async successfully, although I think it is configured somewhat differently.

Uh I guess worst case scenario I could just return explicit promises?

The reason why I'm using async functions is because I'm testing how a workspace change listener responds to events, and events are broadcast asynchronously. Is there a way to broadcast them synchronously? I saw that there was a "fireNow" function in event utils but I couldn't figure out a way to call it from the test file.

johnnesky:
At the suggestion of @rachel-fenichel and @BeksOmega I used sinon fake time instead of async functions, and that seems to resolve test failures, thanks!

gonfunko (approved):
This LGTM, but I'd suggest also getting a review from someone who's been on the team longer/is more familiar with the details.

plugins/shadow-block-converter/src/shadow_block_converter.js
```diff
@@ -0,0 +1,194 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Colour input field with HSV sliders.
```
rachel-fenichel: Update fileoverview.

plugins/shadow-block-converter/src/shadow_block_converter.js
```diff
@@ -0,0 +1,194 @@
...
+  run(forward) {
+    const workspace = this.getEventWorkspace_();
+    const block = workspace.getBlockById(this.blockId);
+    if (!block) {
+      console.warn('Can\'t change non-existent block: ' + this.blockId);
+      return;
+    }
+
+    const value = forward ? this.newValue : this.oldValue;
+    block.setShadow(!!value);
```
rachel-fenichel, line 95: What happens if `block` has children and you're making `block` into a shadow? E.g. the addition block here:
image

plugins/shadow-block-converter/src/shadow_block_converter.js
```diff
@@ -0,0 +1,194 @@
...
+    // The initiating event wasn't part of any named group, so the shadow events
+    // can't be grouped with it, but at least they can be grouped with each
+    // other. This happens when editing e.g. boolean fields.
+    Blockly.Events.setGroup(true);
+
+    // HACK: Retroactively overwrite the initiating event's group property in
+    // order to include it in the same group as the shadow events anyway. It
+    // wasn't previously grouped with any other events, but now it can be
+    // grouped with the new events.
+    event.group = Blockly.Events.getGroup();
```
rachel-fenichel: Oh. Yeah, this is pretty gross. It will work if no one else is messing around with the event stream, but it's definitely a hack.

plugins/shadow-block-converter/src/shadow_block_converter.js
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  // If the changed shadow block is, itself, a child of another shadow block,
```
rachel-fenichel: What if the block is a shadow and has *children* that are shadows? For instance, if I change the dropdown on the first addition block here:
image

plugins/shadow-block-converter/src/shadow_block_converter.js
johnnesky: Nice catch, done.

plugins/shadow-block-converter/src/shadow_block_converter.js
johnnesky: Currently, nothing. There's nothing that's listening for shadow change events, and shadow change events themselves only affect the one block they're associated with. Shadow change events are only created in response to new block changed events, in which case one is created for that block and for any parents if necessary, and ignore children of the changed block.

plugins/shadow-block-converter/src/shadow_block_converter.js
johnnesky: Reverted. I'll file an issue on blockly about why editing boolean fields results in ungroupable change events.

plugins/shadow-block-converter/src/shadow_block_converter.js
johnnesky: I chose to only ensure that all parents of the changed block are real, and ignore the children. As you pointed out, "Doing it this way still maintains the invariants that no shadow block has a real child, while changing the minmum number of blocks."

Actually, that invariant is temporarily violated while in the middle of creating shadow change events, because the algorithm starts converting shadow blocks at the one that's edited before walking up the hierarchy. I guess technically it should walk up the hierarchy first, then start reifying blocks starting at the top and working back down to the one that was edited. 

In my testing it seems to work just fine so far. If the events are always executed as a group, then I assume nobody will notice. But, hypothetically, if the blockly source code is changed to depend on maintaining this invariant at all times then this plugin could get broken.

johnnesky: I could try to implement a version of the algorithm that maintains the invariant during the execution of the algorithm, but it might need to be a little complex. Is it safe to assume that every block will only have one parent (an outputConnection or a previousConnection but not both)? If not, then the set of parents might not be a linear hierarchy. I think the correct algorithm would be sort of an upside down, post-ordered depth-first-search (treating the changed block as the root node and treating the outputConnection and previousConnection as children). I think it would be fun, like a puzzle, to implement this, although it might feel unnecessarily complex to future readers of the plugin source code.

rachel-fenichel: > Is it safe to assume that every block will only have one parent (an outputConnection or a previousConnection but not both)? 

Yes, you can assume this. Every now and then someone asks to change it, but we've held the line :)

johnnesky: Great, thanks! I've committed a version that reifies the parent blocks first, along with a unit test that enforces the order of shadow block events.

BeksOmega: > Yes, you can assume this. Every now and then someone asks to change it, but we've held the line :)

Just to clarify, it is possible to have both an output and a previous connection! But only one of them can be connected to a parent at any given time.

rachel-fenichel (approved):
:tada:

PR #1304, opened 2022-09-26
fix: avoid copy from inside context menus

### LINKED ISSUE #1169: cross-tab-copy-paste can copy a block from inside the bubble to outside the bubble.
cross-tab-copy-paste can copy a block from inside the bubble to outside the bubble.
image

image

This will cause an error when generating the code.
Right-clicking is prohibited in the bubble, but you can copy using the shortcut key.

debaraj-barua:
fixes #1169


alschmiedt (approved):
lgtm, thanks for fixing this!

Issue #1211, opened 2022-08-02
Add support for adding imports via the migration script


BeksOmega:
**Category**

 - Plugins

**Component**

migration

**Is your feature request related to a problem? Please describe.**

We moved the blockly library blocks and generators to no longer cause side effects. Instead you need to use namespace imports to access them. This causes problems for people who were previously accessing them as properties of `Blockly`, which were added via side effects.

I want the migration script to add the imports when necessary. I want it to support named imports, default imports, and namespace imports (so that it is future proof).

**Option 1: Modify renamings script**

We can achieve this by modifying the renamings script.  In this case I would add an `addImport` property to the renamings.json5 schema at the `exports` level, which takes in a value of `named`, `default`, or `namespace` and inserts the correct import. The name of the import will be the value of `newPath` and the name of the module will be `newModule`.

For example, if I wanted to transform:
```
import ‘blockly/javascript’;
Blockly.JavaScript.whatever
```

Into:
```
import * as javascriptGenerator from ‘blockly/javascript’;
javascriptGenerator.whatever
```

I would specify:
```
‘0.0.0’: {
  {
    oldName: ‘Blockly’,
    exports: {
      ‘JavaScript’: {
        oldPath: ‘Blockly.JavaScript’,
        newModule: ‘blockly/javascript’,
        newPath: ‘blockly/javascript’,
        addImport: 'namespace',
      }
    }
  }
}
```

The issue with this is the actual implementation. Currently we [perform renamings using regex replaces](https://github.com/google/blockly-samples/blob/d03a87fc0130d6874d0242445f1746e23e72597c/plugins/migration/bin/rename.js#L123), which doesn't allow us to insert imports when we detect things that need new imports.

**Option 2: Create new script**

So alternatively, we could create a new script entirely separate from the renaming script for performing these specific migrations. In this case I'm not sure what the database should look like. Options include:
  * Just using the same `renamings.json5` file, and only looking at the relevant entries (as defined by the `addImport` property defined above)
  * Create a new database that has the same schema as the `renamings.json5` file
  * Create a new database that has a different schema from the `renamings.json5 file.

Then instead of doing renamings using regex replaces, we would:
```
pseudofunction rename() {
  var matches = findMatches();
  if (!matches) return;
  replaceMatches();
  insertImport();
}
```

**Additional context**

[...]

PR #1306, opened 2022-09-27
feat: add a migration script to fix imports

BeksOmega:
### Description

Work on https://github.com/google/blockly-samples/issues/1211

Adds a new migration script `npx @blockly/migrate fix-imports` that will automatically add new imports, and rename references to the old location of the import on the blockly namespace tree to now reference the import.

### Testing

Created some manual test data and ran the migration script against it. Everything seems to look good!

### Additional Info

This isn't how I would ideally like the system to function. In particular:
  1) I would like the database to be separate from the script, and live in core.
  2) I don't want to have to specify the import string explicitly. I would rather be able to specify if it is `named`, `default`, or `namespace`.

But since we want to get this out before the release, I don't think there's any reason to add that extra complexity.


plugins/migration/bin/fix-imports.js
```diff
@@ -0,0 +1,173 @@
...
+export const fixImports = createSubCommand(
+  'fix-imports',
+  '>=9',
+  'Add imports for modules that have been moved out of the Blockly namespace')
+  .option(
+      '-i, --in-place [suffix]',
+      'fix imports in-place, optionally create backup files with the ' +
+      'given suffix. Otherwise output to stdout')
+  .action(function() {
+    // TODO: In the future we should use the fromVersion and toVersion so that
```
cpcallen: I believe it is team policy that there should be an open issue  and the issue number should be listed here (e.g. `TODO(#1234): ...`).

plugins/migration/bin/fix-imports.js
```diff
@@ -0,0 +1,173 @@
...
+  });
+
+/**
+ * @typedef {{
+ *   oldIdentifier: string,
+ *   newIdentifier: string,
+ *   import: string,
+ * }}
+ */
+const MigrationData = {};
```
cpcallen: Examples in the styleguide (and Closure Compiler docs) do not provide a value for `@typedef` identifiers:
```suggestion
let MigrationData;
```

plugins/migration/bin/fix-imports.js
```diff
@@ -0,0 +1,173 @@
...
+
+// TODO: Make this database format more robust.
+/**
+ * @type {MigrationData[]}
+ */
+const database = [
```
cpcallen: I think this can go all on one line:
```suggestion
const /** @type {MigrationData[]} */database = [
```

Additionally, there should be an entry for `libraryBlocks`.

cpcallen (changes requested):
Looks good so far, but:

* It didn't occur to me when we discussed this yesterday, but any file that needs this migration should already contain an `import 'blockly/javascript';` statement (or the like).
  * Files that rely on `Blockly` and e.g. `Blockly.JavaScript` having already been loaded via a `<script>` tag won't have those `import`s _and don't need this migration_.
  * Files that do have the import need this migration _but the developer has already identified where the new import should go_, since it should replace the existing one.
* It would be very helpful if the script printed some diagnostics to indicate whether the file needs migration or not—e.g. "References to Blockly.JavaScript not migrated as it looks like this module is loaded via a <script> tag" vs. "Migrated references to …" vs. "File contains no references to generators or libraryBlocks".
* We should also support CJS and AMD `require()` statements as well as `import`.  This will require a bit of auto-detecting; probably best see which is used to import the generator and stick with that.

BeksOmega:
> It didn't occur to me when we discussed this yesterday, but any file that needs this migration should already contain an `import 'blockly/javascript';` statement (or the like).
> * Files that rely on `Blockly` and e.g. `Blockly.JavaScript` having already been loaded via a `<script>` tag won't have those `import`s _and don't need this migration_.
> * Files that do have the import need this migration _but the developer has already identified where the new import should go_, since it should replace the existing one.

I don't think this is true for the javascript generator or the library blocks (although it is true for the other generators). You can access `Blockly.JavaScript` when you do `import Blockly from 'blockly';` or `import Blockly from 'blockly/core';`. So in those cases references would need to be renamed, and the import would need to be added.

Or at least that's my understanding anyway.

cpcallen:
> I don't think this is true for the javascript generator or the library blocks (although it is true for the other generators). You can access `Blockly.JavaScript` when you do `import Blockly from 'blockly';` or `import Blockly from 'blockly/core';`. So in those cases references would need to be renamed, and the import would need to be added.
> 
> Or at least that's my understanding anyway.

Oh yes, I forgot about our "some batteries included" default.  Hmm…  🤔

I guess the best we can do is to try to add the new imports after the Blockly import/require).  (We can definitely at least check check for the existence of such to figure out if we need to apply the migration, anyway.)

BeksOmega:
I think this is ready for another look @cpcallen 

I really hate the organization of this (i.e. using `if` statements everywhere to change the behavior based on the type of import) but I don't know that reorganizing it rn is worth it.

plugins/migration/bin/fix-imports.js
```diff
@@ -0,0 +1,237 @@
...
+    newImport: `import {pythonGenerator} from 'blockly/python';`,
+    newRequire: `const {pythonGenerator} = require('blockly/python');`,
+  },
+  {
+    import: 'blockly/blocks',
+    oldIdentifier: 'Blockly.libraryBlocks',
+    newIdentifier: 'libraryBlocks',
+    newImport:  `import {libraryBlocks} from 'blockly/blocks';`,
+    newRequire: `const {libraryBlocks} = require('blockly/blocks');`,
+  }
```
cpcallen: I'd put this above the other entries, especially if doing so results in the corresponding `import` appearing first, because it doesn't really make sense to use the generators if you're not using the library blocks.

Also: there is no reason not to have a trailing `,` on the last item of this array.  (Is this not being linted, or is eslint getting soft on us?)

plugins/migration/bin/fix-imports.js
```diff
@@ -0,0 +1,237 @@
...
+]
+
+/**
+ * Migrates the contents of a particular file, renaming references and adding
+ * imports.
+ * 
+ * @param {string} contents The string contents of the file to migrate.
+ * @return {string} The migrated contents of the file.
+ */
+function createNewContents(contents) {
```
cpcallen: Maybe `migrateFile`, or `migrateContents`?  The amount of actual creation going on here is hopefully minimal…

plugins/migration/bin/fix-imports.js
```diff
@@ -0,0 +1,237 @@
...
+}
+
+function getIdentifier(contents, migrationData) {
+  if (usesImportStatements(contents)) {
+    const identifierMatch = contents.match(
+        new RegExp(`\\s(\\S*) from '${migrationData.import}'`));
+    if (!identifierMatch) return migrationData.oldIdentifier ?? null;
+    return identifierMatch[1];
+  }
+  if (usesRequireStatements(contents)) {
```
cpcallen: I was going to say that even though it is not necessary and does spoil the aesthetics, I think it is helpful for readability to use an `else` here:

```suggestion
    return identifierMatch[1];
  } else if (usesRequireStatements(contents)) {
```
because otherwise one is left to puzzle out the consequences for files that use both`import` and `require`.

But then upon further consideration I think that there may actually be a bug here, because AFAIK it is quite legitimate for a node module to use both `import` and `require`, and it looks like this code might do the wrong thing in that case.

(Here and below at `addImport`.)

cpcallen (commented):
I was so close to approving this with just a couple of small suggestions but then I realised there was a situation I would have completely overlooked except for an oddity of your coding style…

plugins/migration/bin/fix-imports.js
BeksOmega: With the way I rearranged the code, being at the end actually inserts it directly after the `Blockly` import.

BeksOmega:
@cpcallen Should be good for another look =)

plugins/migration/bin/fix-imports.js
```diff
@@ -33,6 +33,7 @@ export const fixImports = createSubCommand(
       const newContents = createNewContents(contents);
       const inPlace = this.opts().inPlace;
 
+
```
cpcallen: [The styleguide says](https://google.github.io/styleguide/jsguide.html#formatting-vertical-whitespace) that multiple blank lines are "not encouraged".

plugins/migration/bin/fix-imports.js
```diff
@@ -0,0 +1,246 @@
...
+ * Returns the identifier a given import is assigned to.
+ * 
+ * @param {string} contents The string contents of the file to migrate.
+ * @param {MigrationData} migrationData Data defining what to migrate and how.
+ * @return The identifier associated with the import associated with the
+ *     migration data.
+ */
+function getIdentifier(contents, migrationData) {
+  const importMatch = contents.match(
+      new RegExp(`\\s(\\S*) from ('|")${migrationData.import}('|")`));
```
cpcallen: Why not use character classes—and while we're at it, be more flexible about whitespace?:
```suggestion
      new RegExp(`\\s(\\S*)\\s+from\\s+['"]${migrationData.import}['"]`));
```
Or are you extracting information about what quote character they used, too, and I'm just not seeing how?

(Here and several places below.)

plugins/migration/bin/fix-imports.js
```diff
@@ -0,0 +1,246 @@
...
+  if (importMatch) return importMatch[1];
+  const requireMatch = contents.match(
+      new RegExp(`(\\S*) = require\\(('|")${migrationData.import}('|")\\)`));
```
cpcallen:
```suggestion
      new RegExp(`(\\S*)\\s+=\\s+require\\(['"]${migrationData.import}['"]\\)`));
```
(This could yet have a few more `\\s*`s added to it too, if we wanted to be very thorough.)

plugins/migration/bin/fix-imports.js
```diff
@@ -0,0 +1,246 @@
...
+    newIdentifier: 'pythonGenerator',
+    newImport: `import {pythonGenerator} from 'blockly/python';`,
+    newRequire: `const {pythonGenerator} = require('blockly/python');`
+  },
+  {
+    import: 'blockly/blocks',
+    oldIdentifier: 'Blockly.libraryBlocks',
+    newIdentifier: 'libraryBlocks',
+    newImport:  `import {libraryBlocks} from 'blockly/blocks';`,
+    newRequire: `const {libraryBlocks} = require('blockly/blocks');`
```
cpcallen: I think you may have gotten the wrong end of the stick about how the library blocks are exported.  This will be mostly irrelevant as until fairly recently we only exported the actual blocks dictionary and AFAIK most developers only imported `blockly/blocks` for side effects, but I believe this section should be as follows:

```suggestion
    import: 'blockly/blocks',
    oldIdentifier: 'Blockly.libraryBlocks',
    newIdentifier: 'libraryBlocks',
    newImport:  `import * as libraryBlocks from 'blockly/blocks';`,
    newRequire: `const libraryBlocks = require('blockly/blocks');`
```

cpcallen (approved):
A few small suggestions and a correction to the `libraryBlocks` section.  Otherwise LGTM.

plugins/migration/bin/fix-imports.js
BeksOmega: Whoops! thanks for the catch.

plugins/migration/bin/fix-imports.js
BeksOmega: I did think about matching the quotes, but decided not to for simplicity. Thank you for the catch!

PR #1307, opened 2022-09-28
fix: field slider test that used private properties

rachel-fenichel:
Delete tests that relied on accessing and changing the field's `htmlInput_` property, which is private.

The validators are still tested by the "When not editing" suite of tests.


rachel-fenichel:
I tested this PR locally by running `npx lerna run test --no-bail ` at the root level and confirming that nothing in this plugin failed. Note that the keyboard navigation plugin still fails tests.

BeksOmega (approved):

PR #1311, opened 2022-10-01
fix: keyboard navigation tests

rachel-fenichel:
## The changes

Actual keyboard nav implementation: Adjust how it's calling a superclass function in a strange bit of monkeypatching code. This change is necessary because of the switch to ES6 classes.

Test code:
- Stop overriding `Blockly.utils.dom.getFastTextWidthWithSizeString`. 
  - This is no longer accessible for replacement. Unfortunately, without this monkeypatch, the tests fail any time they try to render a field with text that has to be measured.
- Replace text fields with colour fields in test blocks.
  - Colour fields can render just fine, because they don't need text width measurement. The fields are mostly used to test that the correct block has been selected after a series of key presses. That happens by looking at the value of the field, which is set in the hardcoded toolbox definition. I updated the block definitions, toolbox definitions, and test assertions to all be for my colour fields.
- Don't access a private field in `MarkerManager` for a test assert. Particularly since that field's name has changed. Use the getter instead.
- Call `this.navigation.removeWorkspace(this.workspace)` in the test teardown for each suite. Otherwise the workspace is disposed before being removed from `this.navigation`, which means that the `navigation` object is manipulating an already-disposed `WorkspaceSvg` object. 
  - **I'm not sure why this worked before.**

## Tested
In the keyboard nav playground and with automated tests.

### DIFF
--- a/plugins/keyboard-navigation/src/navigation_controller.js
+++ b/plugins/keyboard-navigation/src/navigation_controller.js
@@ -113,7 +113,8 @@ export class NavigationController {
           return false;
       }
     }
-    return Blockly.FieldColour.superClass_.onShortcut.call(this, shortcut);
+    // Field is the superclass of FieldColour.
+    return Blockly.Field.prototype.onShortcut.call(this, shortcut);
   }
 
   /**
@@ -139,7 +140,8 @@ export class NavigationController {
           return false;
       }
     }
-    return Blockly.FieldDropdown.superClass_.onShortcut.call(this, shortcut);
+    // Field is the superclass of FieldDropdown.
+    return Blockly.Field.prototype.onShortcut.call(this, shortcut);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


plugins/keyboard-navigation/src/navigation_controller.js
```diff
@@ -113,7 +113,8 @@ export class NavigationController {
           return false;
       }
     }
-    return Blockly.FieldColour.superClass_.onShortcut.call(this, shortcut);
+    // Field is the superclass of FieldColour.
```
maribethb: nit, but outside the context of this code review where i can see the line you're removing, this comment doesn't help me understand what's going on.

"If we haven't already handled the shortcut, let the default Field handler try" or something might be more clear. same below.

maribethb (approved):

plugins/keyboard-navigation/src/navigation_controller.js
rachel-fenichel: Done.

BeksOmega (approved):

PR #1322, opened 2022-10-07
fix: scripts plugin now properly builds TypeScript declarations

btw17:
### Description

- Updated `plugins/dev-scripts` to use `ts-loader` for building TypeScript
- Updated `plugins/field-slider` to exclude `test` from built types

### Note

- The current babel plugin had issues with building the declarations. I originally investigated this in August and don't recall the exact reason why it didn't work, though using `ts-loader` seemed to be the preferred, more standard option anyway. I could re-investigate looking into babel, though, if that's the preferred long term solution.

### DIFF
--- a/plugins/dev-scripts/config/webpack.config.js
+++ b/plugins/dev-scripts/config/webpack.config.js
@@ -11,9 +11,7 @@
 
 const fs = require('fs');
 const path = require('path');
-const resolve = require('resolve');
 const webpack = require('webpack');
-const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
 
 const appDirectory = fs.realpathSync(process.cwd());
 const resolveApp = (relativePath) => path.resolve(appDirectory, relativePath);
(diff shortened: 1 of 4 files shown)


plugins/dev-scripts/config/webpack.config.js
```diff
@@ -120,7 +118,7 @@ module.exports = (env) => {
           enforce: 'pre',
         },
         // Run babel to compile both JS and TS.
```
BeksOmega: Please remove this comment (since it should be clear from the actual code). Otherwise update it to be correct.

I would also be in favor of moving the typescript case above the JS case.

btw17:
I updated the target to `ES2015` to address [this build failure](https://github.com/google/blockly-samples/actions/runs/3207813005/jobs/5243082274).

Of note, [Blockly core's tsconfig](https://github.com/google/blockly/blob/master/tsconfig.json) has target set to `ES2020` which may be an issue depending on if it plans to support older browsers.

BeksOmega (changes requested):
Just one comment fix, then LGTM!

plugins/dev-scripts/config/webpack.config.js
btw17: Done! **Edit:** with the comment removal.

btw17: Moved the TS case above the JS case!

BeksOmega (approved):

PR #1329, opened 2022-10-12
fix: Reverting invalid changes to hsv slider css.

johnnesky:
For some reason, https://github.com/google/blockly-samples/pull/1313 deleted some arbitrary characters from the CSS in the hsv colour field plugin code. Some other CSS was altered in the same PR but I don't think the rest was broken in the same way. This PR brings back the deleted characters (while preserving the whitespace formatting changes).


BeksOmega:
Whoops! Looks like my macro broke some things and I didn't catch it :/ Thank you for finding this John!

PR #1339, opened 2022-10-15
chore: Convert the shadow-block-converter plugin to typescript.

johnnesky:
I used Blake's example of field-slider as a reference in: https://github.com/google/blockly-samples/pull/1217 and https://github.com/google/blockly-samples/pull/1322

I noticed that I had to add `'block': undefined` to several inputs in a toolbox, because the field is mandatory even for shadow blocks in the interface `ConnectionState` defined in `core/serialization/blocks.ts`. That seems undesirable, perhaps `ConnectionState` should be defined as a union between two types, one that only requires a `'block'` field and one that only requires a `'shadow'` field.

I also had to temporarily disable `@typescript-eslint/naming-convention` because it didn't like that I had used SCREAMING_SNAKE_CASE to name a constant.


plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -164,15 +184,15 @@ export function shadowBlockConversionChangeListener(event) {
...
     // If connected blocks need to be converted too, add them to the list.
-    if (shadowBlock.outputConnection != null &&
-        shadowBlock.outputConnection.isConnected() &&
-        shadowBlock.outputConnection.targetBlock().isShadow()) {
-      shadowBlocks.push(shadowBlock.outputConnection.targetBlock());
+    const outputBlock: Block | null =
+        shadowBlock.outputConnection?.targetBlock();
+    const previousBlock: Block | null =
+        shadowBlock.previousConnection?.targetBlock();
+    if (outputBlock != null && outputBlock.isShadow()) {
```
rachel-fenichel: Can this be `if (outputBlock?.isShadow())`?

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -164,15 +184,15 @@ export function shadowBlockConversionChangeListener(event) {
...
+      shadowBlocks.push(outputBlock);
     }
-    if (shadowBlock.previousConnection != null &&
-        shadowBlock.previousConnection.isConnected() &&
-        shadowBlock.previousConnection.targetBlock().isShadow()) {
-      shadowBlocks.push(shadowBlock.previousConnection.targetBlock());
+    if (previousBlock != null && previousBlock.isShadow()) {
```
rachel-fenichel: Same here.

plugins/shadow-block-converter/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+      "baseUrl": "./",
+      "outDir": "dist",
+      "declaration": true,
+      "declarationMap": true,
+      "downlevelIteration": true,
```
rachel-fenichel: Remove `downlevelIteration`: It's only useful if you expect this code to be used in an ES5 environment where a `Symbol.iterator` implementation *might( exist. Since we're compiling to ES2015, it's not necessary.

Reference: https://www.typescriptlang.org/tsconfig#downlevelIteration

plugins/shadow-block-converter/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+      "experimentalDecorators": true,
```
rachel-fenichel: Remove `experimentalDecorators`: I don't think you're using them.

rachel-fenichel (commented):
I agree that the `ConnectionState` type isn't working quite right. I filed https://github.com/google/blockly/issues/6576 to track figuring out what it should be and fixing it in core.

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: Done.

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: Done.

plugins/shadow-block-converter/tsconfig.json
johnnesky: Done. This file was copied verbatim from the field-slider example. I've taken the liberty of applying the same change to the example I took this from.

plugins/shadow-block-converter/tsconfig.json
johnnesky: Done. Again I did the same thing to the other file.

PR #1343, opened 2022-10-18
Draft Update json xml

mikebarr24:
Not working. Workspace is returning "undefined"

### DIFF
--- a/examples/headless-demo/index.html
+++ b/examples/headless-demo/index.html
@@ -26,61 +26,117 @@
   </style>
 </head>
 <body>
-  <p>This is a simple demo of generating Python code from XML with no graphics.
+  <p>This is a simple demo of generating Python code from JSON with no graphics.
   This might be useful for server-side code generation.</p>
 
   <table style="width: 100%">
     <tr>
       <td style="width:50%">
-        <textarea id="xml_input">
-<xml xmlns="https://developers.google.com/blockly/xml">
-  <block type="controls_if" inline="false" x="20" y="20">
-    <mutation else="1"></mutation>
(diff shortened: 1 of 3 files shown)


examples/headless-demo/index.html
```diff
@@ -97,21 +153,18 @@
...
-      var xmlText = document.getElementById('xml_input').value;
+      var json = document.getElementById('json_input').value;
+      // Create a headless workspace.
       try {
-        var xml = Blockly.Xml.textToDom(xmlText)
-      } catch (e) {
-        alert(e);
-        return;
+        var workspace = new Blockly.Workspace();
+        Blockly.serialization.workspaces.load(JSON.parse(json), workspace);
```
mikebarr24, line 161: I'm struggling to get this working. Workspace seems to have a value if you log it out, but `Blockly.serialization.workspace.load()` doesn't seem to recognise it.

BeksOmega: Hmm that's weird! Could you post the exact text of the error you're getting?

mikebarr24:
```
TypeError: Cannot read properties of undefined (reading 'workspaces')
    at generate (index.html:160:31)
    at HTMLButtonElement.onclick (index.html:151:34)
```

BeksOmega: Ah ok! You need to update the [dependency on Blockly](https://github.com/google/blockly-samples/blob/master/examples/headless-demo/package.json) so that it's using v9, which includes the `serialization` namespace =)

You probably want it to look like: `blockly: '^9.0.0'`

PR #1361, opened 2022-11-02
fix: field-bitmap now serializing correctly

### LINKED ISSUE #1312: Bitmap field cleanup
**Category**
 - Plugins

**Component**

field-bitmap

**Describe the bug**

Console warnings in the bitmap test page: 
- Not setting SERIALIZABLE correctly
- Using `css.register` with an array of strings instead of a multiline string.

**To Reproduce**
Check out `blocky-9-beta` (after this week, probably just check out `master`

**Expected behavior**

No console warnings

**Additional context**

mikebarr24:
fix #1312 

- Serialisation issue should be sorted but unsure about changing CSS to an array of singles lines of CSS. 

- I'm not too familiar with Typescript but I the function definition for CSS.register specifies only type string. I can't supply this function with an array without it complaining. Is there a way round this?

### DIFF
--- a/plugins/field-bitmap/src/field-bitmap.js
+++ b/plugins/field-bitmap/src/field-bitmap.js
@@ -34,6 +34,9 @@ export class FieldBitmap extends Blockly.Field {
   constructor(value = undefined, validator = undefined, config = undefined) {
     super(value, validator, config);
 
+    this.SERIALIZABLE = true;
+
+
     // Configure value, height, and width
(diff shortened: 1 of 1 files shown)


BeksOmega:
Heylo! The css issue was already fixed in https://github.com/google/blockly-samples/pull/1313 so I think this is actually gtg =)

BeksOmega (approved):

PR #1363, opened 2022-11-04
fix: scroll-options stops using is3dSupported

HollowMan6:
This is a follow up on:
- #1202
- https://github.com/google/blockly/pull/6400
- https://github.com/mit-cml/workspace-multiselect/pull/6

As `is3dSupported` has been removed in Blockly 9, the [workspace-multiselect](https://github.com/mit-cml/workspace-multiselect) plugin v0.1.3 will choose to use "`Blockly.WorkspaceSvg.getBlockDragSurface` always return `null`" to disable dragsurface, which means now it's OK to remove `is3dSupported` here in scroll-options.

Also fix the plugin to get the workspace translation when there's no getBlockDragSurface.

Signed-off-by: Hollow Man <hollowman@opensuse.org>

### DIFF
--- a/plugins/scroll-options/src/utils.js
+++ b/plugins/scroll-options/src/utils.js
@@ -1,0 +1,24 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Gets the current location of the workspace considering
+ * when there's no drag surface.
+ * @param {!Blockly.WorkspaceSvg} ws The workspace to calculate.
+ * @return {!Blockly.utils.Coordinate} The current workspace coordinate.
+ */
+export const getTranslation = (ws) => {
+  const dragSurface = ws.getBlockDragSurface();
+  if (dragSurface) {
+    return dragSurface.getWsTranslation();
+  } else {
+    const translation = ws.svgBlockCanvas_.getAttribute('transform');
+    const splitted = translation.split(',');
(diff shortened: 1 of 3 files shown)


plugins/scroll-options/src/utils.js
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Gets the current location of the workspace when there's no drag surface.
+ * @param {!Blockly.WorkspaceSvg} ws The workspace to calculate.
+ * @return {!Blockly.utils.Coordinate} The current workspace coordinate.
+ */
+export const getTranslation = (ws) => {
+  const translation = ws.svgBlockCanvas_.getAttribute('transform');
```
rachel-fenichel: How about moving the check for `dragSurface` into this function and having it automatically return either `dragSurface.getWsTranslation` or the rest of this function? That lets you avoid duplicating code in the three places where this is called.

plugins/scroll-options/src/utils.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
```
rachel-fenichel: 2022

rachel-fenichel (changes requested):

HollowMan6:
Fixed, thanks for reviewing!

rachel-fenichel (approved):

PR #1371, opened 2022-11-09
fix: disable copy for immovable and undeletable blocks

mikeharv:
The context menu option for "Copy" was set to always be enabled. To match core Blockly behavior, and to align to this plugin's behavior for the copy shortcut (below), the precondition function should be modified. 
Compare to copy shortcut:
https://github.com/google/blockly-samples/blob/4a5f0368e21ea90784041e4daafe5a43ebba3b39/plugins/cross-tab-copy-paste/src/index.js#L136-L143

New behavior:

https://user-images.githubusercontent.com/43474485/200906473-09a7264c-81c4-49bb-87f5-3f1c62562ceb.mp4

### DIFF
--- a/plugins/cross-tab-copy-paste/src/index.js
+++ b/plugins/cross-tab-copy-paste/src/index.js
@@ -66,7 +66,12 @@ export class CrossTabCopyPaste {
       },
       preconditionFn: function(
           /** @type {!Blockly.ContextMenuRegistry.Scope} */ scope) {
-        return 'enabled';
+        if (
+          Blockly.selected.isDeletable() &&
+          Blockly.selected.isMovable()) {
+          return 'enabled';
+        }
+        return 'disabled';
       },
       callback: function(
           /** @type {!Blockly.ContextMenuRegistry.Scope} */ scope) {


maribethb:
For additional context, in core I believe this behavior is to prevent users from accidentally copying & pasting a block that is not deletable. Also, this behavior is already present in the behavior for copying via the keyboard shortcut, so I would definitely consider this a bug and not sure that it's a breaking change.

Interestingly, because movable and deletable are not serialized by default to json, using this plugin by itself would result in pasting a block that is movable and deletable. But if you also use the serialize-disabled-interactions plugin, then the pasted block would not be deletable. This is highly undesirable, so it makes sense that the default behavior here would match the behavior of copy in core.

cpcallen (approved):
It seems to me that being able to copy a non-movable or non-deleteable block, but having the copy be movable or deletable, could be a feature not a bug—but it might be a rather surprising feature given that we don't let that happen in any other situation.  So I'll approve this for consistency if nothing else.

maribethb:
> It seems to me that being able to copy a non-movable or non-deleteable block, but having the copy be movable or deletable, could be a feature not a bug

Agreed, if it were intentional, but if you use the serialize-disabled-interactions plugin then this won't be the case anymore, so we'd have to do additional work to enable this (and probably have it be configurable)

maribethb:
Hey @mikeharv I don't think this is a breaking change, just a bug fix. Could you update the commit so that it doesn't create a new major version? https://developers.google.com/blockly/guides/contribute/get-started/commits#fixing_non-conventional_commits

mikeharv:
I've updated the pr title to show that it's not a breaking change. @maribethb and @cpcallen are either of you able to merge? I don't have write access to this repository.

Issue #522, opened 2020-12-16
Time Picker Field


alschmiedt:
**Category**
Plugins

**Component**
A field that will allow the user to pick the time.

**Is your feature request related to a problem? Please describe.**
There is not an easy way to configure the date field to also allow time.

**Describe the solution you'd like**
A field that will allow the user to choose the time. They could use this as well as the date field on a block to get the time on a specific day.

**Describe alternatives you've considered**
A field that will allow a user to pick the date as well as the time. I think this would be a bit more complicated and less flexible than the other solution, however I think either way would be acceptable.

**Additional context**
https://groups.google.com/g/blockly/c/Q5NdVWX1Ke4

rachel-fenichel:
This could be done with an [HTML time input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time) with appropriate validation.

MatoPlus:
@alschmiedt My group (for a university project) would like to work on this!

BeksOmega:
Hi @MatoPlus I've assigned you, but feel free to split the work up amongst yourselves! Best of luck, and we are here to help if you need it =)

PR #1385, opened 2022-11-21
feat: time picker field

JankyCS:
image

https://github.com/google/blockly-samples/issues/522

This field is implemented differently from some of the other field plugins because it does not use the `DropDownDiv` for it's time picker popup. This is because it uses [HTML time input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time) as per https://github.com/google/blockly-samples/issues/522#issuecomment-1259830986, and the time picker popup can not be placed into the dropdown div(as far as I can tell). Instead, `widgetCreate_` is overridden to produce an invisible HTML time input.

It currently displays the time in 24-hour format, although the time picker may use AM/PM, depending on the browser.


BeksOmega:
Woo that looks sweet! I think you should be able to put in the dropdowndiv though, because the dropdowndiv is just a fancy div. Could you explain a bit more about what issues you ran into?

JankyCS:
image
 The actual time picker is what appears when you click the clock on the right side of a `<input type='time'/>` and I programmatically force it to appear using a call to `showPicker()`. I don't think this popup is an HTML element that can be placed inside of a div.

plugins/field-time/GETSTARTED.md
```diff
@@ -0,0 +1,24 @@
+This package was bootstrapped with [@blockly/create-package](https://www.npmjs.com/package/@blockly/create-package).
```
cpcallen: N.B.: I've queried with my colleagues about whether we actually want this file committed or not.  Leave it for now, but we might delete it (and its many sibling `plugins/field-*/GETSTARTED.md` files).

plugins/field-time/README.md
```diff
@@ -0,0 +1,46 @@
+# blockly-plugin-field-time [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
+
+<!--
+  - TODO: Edit plugin description.
+  -->
+A [Blockly](https://www.npmjs.com/package/blockly) plugin that ...
```
cpcallen: Obviously this (and the Usage and API sections below) will need to be written.

plugins/field-time/src/index.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './field_time';
```
cpcallen: This file is a little useless.  I'd suggest getting rid of it and just adjusting the package's entry point, but evidently the `build` script depends on it existing—so why not replace it with the contents of `field_time.js` (which can then be deleted)?

plugins/field-time/src/field_time.js
```diff
@@ -0,0 +1,115 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
cpcallen:
```suggestion
 * Copyright 2022 Google LLC
```

plugins/field-time/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "blockly-plugin-field-time",
+  "version": "0.0.0",
+  "description": "A Blockly plugin.",
```
cpcallen: This will need to be written.

plugins/field-time/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "blockly-plugin-field-time",
```
cpcallen:
```suggestion
  "name": "@blockly/field-time",
```

plugins/field-time/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "version": "0.0.0",
```
cpcallen: This will need to be set to `1.0.0` before this PR is merged.

plugins/field-time/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "author": "",
+  "keywords": [
+    "blockly",
+    "blockly-plugin",
+    "field-time"
+  ],
+  "homepage": "https://github.com/team-vibe-labs/blockly-samples/tree/master/plugins\\field-time#readme",
+  "bugs": {
+    "url": "https://github.com/team-vibe-labs/blockly-samples/issues"
+  },
```
cpcallen: These will need to be updated as well:
```suggestion
  "homepage": "https://github.com/google/blockly-samples/tree/master/field-time#readme",
  "bugs": {
    "url": "https://github.com/google/blockly-samples/issues"
  },
```

plugins/field-time/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/team-vibe-labs/blockly-samples.git",
+    "directory": "plugins\\field-time"
```
cpcallen:
```suggestion
    "url": "https://github.com/google/blockly-samples.git",
    "directory": "plugins/field-time"
```

plugins/field-time/package.json
```diff
@@ -0,0 +1,62 @@
...
+  ],
+  "devDependencies": {
+    "@blockly/dev-scripts": "^1.2.28",
+    "@blockly/dev-tools": "^5.0.2",
+    "blockly": "^9.1.0"
+  },
+  "peerDependencies": {
+    "blockly": "^9.1.0"
+  },
+  "publishConfig": {},
```
cpcallen:
```suggestion
  "publishConfig": {
    "access": "public",
    "registry": "https://wombat-dressing-room.appspot.com"
  },
```
(Don't ask.)

plugins/field-time/GETSTARTED.md
cpcallen: Update: this file should be deleted (once you are familiar with its contents).

plugins/field-time/src/field_time.js
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Class for a time input field.
+ * @extends {Blockly.FieldTextInput}
+ */
+export class FieldTime extends Blockly.FieldTextInput {
+  /**
+   * Class for a time input field. Built using <input type='time'/> time picker
+   * @param {string=} value The initial value of the field. Should be in
+   *    'HH:MM' format. Defaults to 12:00.
```
cpcallen: Probably worth clarifying that this will be in 24h format.

plugins/field-time/src/field_time.js
```diff
@@ -0,0 +1,115 @@
...
+   * @param {*=} newValue The input value.
+   * @return {?string} A valid time, or null if invalid.
+   * @protected
+   */
+  doClassValidation_(newValue = undefined) {
+    if (!newValue) {
+      return null;
+    }
+    const isValid =
+      /^([0-1]?[0-9]|2[0-4]):([0-5][0-9])(:[0-5][0-9])?$/.test(newValue);
```
cpcallen: * Do you want to accept `24:00`?  Do you want to accept `24:59`?
* Do you want to accept seconds?  The description suggests it's strictly HH:MM.
* The middle set of parens appear to be superfluous:
  ```suggestion
        /^([0-1]?[0-9]|2[0-4]):[0-5][0-9](:[0-5][0-9])?$/.test(newValue);
  ```
* An alternative approach might be to match against `/^(\d\d?):(\d\d)(?::(\d\d))?$/` and then check the submatches numerically.  (Probably not simpler, but worth keeping in mind if the regexp starts getting out of hand.)

plugins/field-time/src/field_time.js
```diff
@@ -0,0 +1,115 @@
...
+    if (!isValid) {
+      return null;
+    }
```
cpcallen: I might try:
```JS
  if (!/…detail omitted…/.test(newValue) {
    return null;
  }
```
(If it fits in 80 columns!)

plugins/field-time/src/field_time.js
```diff
@@ -0,0 +1,115 @@
...
+    return newValue;
+  }
+
+  /**
+   * Create the time input editor widget.
+   * Override the default text input to be a time input
+   *
+   * @return {!HTMLElement }The newly created time input editor.
+   */
+  widgetCreate_() {
```
cpcallen: Does this need to be marked `protected`?  I'm slightly unclear: I know `public` is the default, and that `protected` method can be overridden to make it `public`, but not sure if that happens in the absence of a visibility modifier.

(Same for `showEditor_` below.)

plugins/field-time/src/field_time.js
```diff
@@ -0,0 +1,115 @@
...
+   * @param {Event=} e Optional mouse event that triggered the field to
+   *     open, or undefined if triggered programmatically.
+   * @protected
+   * @override
+   */
+  showEditor_(e = undefined) {
+    super.showEditor_(e);
+    if (this.picker) {
+      this.picker.showPicker();
+    }
```
cpcallen: Do you intend to show both an editor pop-up and the picker?  Testing on macOS I see only the picker.

cpcallen (changes requested):
There's a bunch of (mostly trivial) comments about the code below.  Overall, my thoughts are:

* This is great and looks great!
* Unfortunately, it does not actually work.  🙁

I suspect it's a locale issue (I'm in the UK and have 24h time turned on).  The picker looks like this:
Screenshot 2022-12-02 at 19 49 36
and I can pick times no problem, but any time I pick a (and thereby change the time) I get the following error in the console:
```
field_textinput.ts?dc62:512 The specified value "undefined" does not conform to the required format.  The format is "HH:mm", "HH:mm:ss" or "HH:mm:ss.SSS" where HH is 00-23, mm is 00-59, ss is 00-59, and SSS is 000-999.
setEditorValue_	@	field_textinput.ts?dc62:512
eval	@	field_time.js?6b7f:79
onClick_	@	unknown
```
and the time field is left blank.

I'm on macOS Monterey 12.6.1, using Chrome 107.0.5304.121, with my system local settings as follows:
- Region: United Kingdom (Custom)
- Calendar: Gregorian
- Time Format: 24h time


Additionally: although popping up the picker should clearly be the default method of entry when one clicks on the field, it would be nice if there were also a way to edit the time using the keyboard.  (Maybe this is possible if one does not encounter the above error?)


Overall very promising.

plugins/field-time/src/index.js
JankyCS: Will do. I actually copied this from the `field_slider` plugin so it should probably be refactored there as well?

plugins/field-time/src/field_time.js
JankyCS: The editor has the picker within it (its opacity is 0 so you can't see it), and we call `showPicker` in `showEditor_` just to trigger the picker to appear.

JankyCS:
> Additionally: although popping up the picker should clearly be the default method of entry when one clicks on the field, it would be nice if there were also a way to edit the time using the keyboard. (Maybe this is possible if one does not encounter the above error?)

I tried to allow the user to edit the time using the keyboard but ultimately decided against it. The reason is that when the `<input type=time/>` popup appears, it prevents any keyboard strokes from being read. I always want this popup to appear on click, so then the user would have to click again in the text area to make the popup disappear in order to actually start typing input. This behavior felt unintuitive to me.

It also comes with the side-effect that the user would be unable to reopen the popup once they start typing input (right now additional clicks just close and re-open the popup, but I would have to disable this behavior). The user would need to deselect the block and then click the area again to get the popup, which also isn't desirable.

I don't think a text input is very important for a time picker. For example, the date picker needs text input because its popup only allows traversing by one month at a time, so it can take a very long time to get to a desired date. The time-picker does not have this issue, since it is limited by the 24-hour clock and it is quick to get any desired time.

Happy to discuss more about this.

plugins/field-time/src/index.js
```diff
@@ -0,0 +1,120 @@
...
+   * @param {string=} value The initial value of the field. Should be in
+   *    '24-hour HH:MM' format. Defaults to 12:00.
+   * @param {Function=} validator A function that is called to validate
+   *    changes to the field's value. Takes in a time string & returns a
+   *    validated time string ('HH:MM' format), or null to abort the
+   *    change.
+   */
+  constructor(value = '12:00', validator = undefined) {
+    const initial = timeValidator(value) || '12:00';
+    super(initial, validator);
```
JankyCS, line 45: Originally I just called `super('12:00', validator)` followed by `this.setValue(value)`, which seemed to work fine. However, the mocha test wouldn't let me do this (can't call setValue in the constructor?) so I took this approach of factoring out the validator, and now it seems fine.

Mocha error:
image

JankyCS:
@cpcallen

BeksOmega:
I'm going to go ahead and close this becausae we're not planning to merge as is, but we can always reopen if necessary =)

BeksOmega:
Reopening per christopher's request.

BeksOmega:
I'm closing this again since it's been 3 months.

PR #1398, opened 2022-12-03
feat: create plugin with renderer that treats dummy inputs as row separators.

johnnesky:
Related to: https://github.com/google/blockly/issues/2062

Creates a plugin that registers a renderer that treats dummy inputs as row separators. This is useful for putting adjacent inline value input connectors on separate rows, which was previously only possible for external value input connectors in the default renderer. 

This plugin also defines some alternatives to built-in blocks that demonstrate the feature, and optionally allows overriding the built-in blocks themselves.

The renderer and the new block definitions override methods with names that end in an underscore, so I had to disable eslint for those names.

This plugin is a proof-of-concept. I suggest that Blockly should eventually support putting \n characters in block JSON message0 to accomplish the same effect. Note that I originally intended to create a renderer that supports indented external value input connectors, but decided that inline value input connectors are easier to implement and more idiomatic for some use-cases. (Although I still think that indented external value input connectors would be useful for growable lists of values.)


johnnesky:
Beka pointed out that there's a request at https://github.com/google/blockly/issues/216 for adding line-breaks between inline inputs (or in this case fields?) that I think could be addressed by this plugin.

plugins/inline-row-separators/src/inline_row_separators_renderer.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+import * as Blockly from 'blockly/core';
+
+/**
+ * A subclass of the geras RenderInfo that renders all dummy inputs like line
+ * breaks, forcing any following input to be rendered on a separate row, even
+ * when the block type definition has `"inputsInline": true`.
+ */
+export class RenderInfo extends Blockly.geras.RenderInfo {
```
BeksOmega: Since the logic in here is so minimal, it would be really nice if we could make this a decorator instead of a subclass. But I think the interface of `RenderInfo` is too large to make a decorator really practical :/

plugins/inline-row-separators/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@blockly/inline-row-separators",
```
BeksOmega: Based on https://developers.google.com/blockly/guides/contribute/samples/naming I think we probably want this to be `renderer-inline-row-separators`. But happy to receive push back on this.

plugins/inline-row-separators/test/inline_row_separators_test.mocha.js
```diff
@@ -0,0 +1,47 @@
...
+    });
+    this.renderer = this.workspace.getRenderer();
+  });
+
+  teardown(function() {
+    this.workspace.dispose();
+    this.jsdomCleanup();
+  });
+
+  test('a text join block should start new rows for value inputs', function() {
```
BeksOmega: This test is testing a lot of things, `overrideOldBlockDefinitions`, the new `text_join` definition, and `should StartNewRow_`. This will make it hard to figure out what exactly is wrong if any of those things break in the future.

Can you separate those into different tests? In particular I think `shouldStartNewRow_` should be tested separately from the other bits.

plugins/inline-row-separators/test/index.ts
```diff
@@ -0,0 +1,112 @@
...
+import {createPlayground} from '@blockly/dev-tools';
+import {overrideOldBlockDefinitions} from '../src/index';
+
+overrideOldBlockDefinitions();
+
+const toolbox: Blockly.utils.toolbox.ToolboxDefinition = {
+  'kind': 'categoryToolbox',
+  'contents': [
+    {
+      'kind': 'category',
```
BeksOmega: I like the category organization here!

plugins/inline-row-separators/package.json
```diff
@@ -0,0 +1,67 @@
...
+  "version": "1.0.0-beta.0",
```
BeksOmega: Does this need a beta?

BeksOmega (changes requested):
At a high level, I'm wondering if this would be better as a demo than a plugin? Since really, external devs might want to apply this to whatever their custom renderer is.

Inline I left a comment about using a decorator, which I think would also solve the problem of wanting to apply this to multiple renderers. But I don't think a decorator pattern makes sense when the `RenderInfo` interface is so large.

Besides that high level decision, looks really good! Also, thanks for going through and cleaning up the package.json, looks great =)

plugins/inline-row-separators/src/inline_row_separators_renderer.ts
johnnesky: Done! Using the mixin technique that Rachel suggested: https://www.typescriptlang.org/docs/handbook/mixins.html

plugins/inline-row-separators/package.json
johnnesky: Done!

plugins/inline-row-separators/test/inline_row_separators_test.mocha.js
johnnesky: I've split it into a bunch of tests and added a new one for the ability to add row separators to custom renderers.

plugins/inline-row-separators/test/index.ts
johnnesky: Thanks!

plugins/inline-row-separators/package.json
johnnesky: Rachel had originally requested this plugin be marked as beta, serving as a proof-of-concept for eventually adding a similar feature to core Blockly. I think that's still the plan? But other stuff about the design has changed, not sure if that makes a difference to how it should be tagged here.

johnnesky: Update: she says to call it: 0.1.0

johnnesky:
I believe I was able to use TypeScript's mixin feature to allow this plugin to add the inline-row-separator feature to any Blockly Renderer!

plugins/inline-row-separators/src/inline_row_separators_renderer.ts
```diff
@@ -10,45 +10,72 @@
...
+   * A subclass of the provided RenderInfo that renders all dummy inputs like
+   * line breaks, forcing any following input to be rendered on a separate row,
+   * even when the block type definition has `"inputsInline": true`.
    */
-  protected shouldStartNewRow_(
-      input: Blockly.Input, lastInput?: Blockly.Input): boolean {
-    /* eslint-enable @typescript-eslint/naming-convention */
-    if (lastInput?.type === Blockly.inputTypes.DUMMY) {
-      return true;
+  class RenderInfo extends renderInfoBase {
```
BeksOmega: whoa these mixins are beautiful! I love this language feature :D

BeksOmega: Could you rename this to `InlineRenderInfo` for clarity?

plugins/inline-row-separators/src/inline_row_separators_renderer.ts
```diff
@@ -10,45 +10,72 @@
...
-   *
-   * @param block The block to measure.
-   * @return The render info object.
+   * A subclass of the provided Renderer that uses the new RenderInfo object.
    */
-  protected makeRenderInfo_(block: Blockly.BlockSvg): RenderInfo {
-    /* eslint-enable @typescript-eslint/naming-convention */
-    return new RenderInfo(this, block);
-  }
+  return class Renderer extends rendererBase {
```
BeksOmega: Similarly, could this be `InlineRenderer`?

plugins/inline-row-separators/src/inline_row_separators_renderer.ts
```diff
@@ -10,45 +10,72 @@
...
+  class RenderInfo extends renderInfoBase {
+    /* eslint-disable @typescript-eslint/naming-convention */
+    /**
+     * Decide whether to start a new row between the two Blockly.Inputs.
+     *
+     * @param input The first input to consider
+     * @param lastInput The input that follows.
+     * @return True if the next input should be rendered on a new row.
+     */
+    protected shouldStartNewRow_(
```
BeksOmega: I think these protected functions should have `override` applied to them. Here and below.

Not sure if the tsc actually checks that for dynamically created classes, but just in case 🤷

plugins/inline-row-separators/README.md
```diff
@@ -48,19 +48,26 @@ Blockly.common.defineBlocksWithJsonArray([
 
 ![A Blockly block demonstrating inline value input connectors on separate rows.](readme-media/inline_row_example_screenshot.png)
 
-This plugin also defines some alternative versions of built-in Blockly blocks that demonstrate the feature: `"inline_text_join"`, `"inline_lists_create_with"`, and `"inline_procedures_defreturn"`. You can replace the built-in blocks `"text_join"`, `"lists_create_with"`, and `"procedures_defreturn"` with these alternatives (which allows you to use Blockly's built-in code generators with these new block types) by importing and calling `overrideOldBlockDefinitions()`:
+This plugin also exports a function called `addInlineRowSeparators` that you can use to add the same inline row separator feature to any class extending Blockly's Renderer, including your own.
```
BeksOmega: Can you split this into two sub headings?

Eg:
```
## Usage

### Pre-defined thrasos renderer

// etc...

### Extending a different renderer

// etc...
```

plugins/inline-row-separators/test/inline_row_separators_test.mocha.js
```diff
@@ -19,29 +20,86 @@ suite('inlineRowSeparates', function() {
     this.jsdomCleanup =
         jsdomGlobal('<!DOCTYPE html><div id="blocklyDiv"></div>');
     this.workspace = Blockly.inject('blocklyDiv', {
-      renderer: 'inline-row-separators',
+      renderer: 'thrasos-inline-row-separators',
     });
```
BeksOmega: Oh huh, I don't think we currently use `jsdom` in our tests. This might be a good pattern to use in the future (e.g. to help us stop using webdriver for mocha tests in core -.- )

plugins/inline-row-separators/test/inline_row_separators_test.mocha.js
```diff
@@ -19,29 +20,86 @@ suite('inlineRowSeparates', function() {
...
 
   teardown(function() {
+    delete Blockly.Blocks['test_block'];
     this.workspace.dispose();
     this.jsdomCleanup();
   });
 
-  test('a text join block should start new rows for value inputs', function() {
-    const block = this.workspace.newBlock('text_join');
+  test('shouldStartNewRow_ returns true if last input is a dummy', function() {
```
BeksOmega: Wonderful! Thank you for splitting this out into multiple tests :D I think this makes your behavioral contract much more clear (or at least personally easier for me to understand :P )

plugins/renderer-inline-row-separators/test/inline_row_separators_test.mocha.js
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  test('inline_text_join is inline and inserts dummies', function() {
+    const block = this.workspace.newBlock('inline_text_join');
+    assert.isTrue(block.inputList[0].type === Blockly.inputTypes.DUMMY);
+    assert.isTrue(block.inputList[1].type === Blockly.inputTypes.VALUE);
+    assert.isTrue(block.inputList[2].type === Blockly.inputTypes.DUMMY);
+    assert.isTrue(block.inputList[3].type === Blockly.inputTypes.VALUE);
+    const renderInfo = this.renderer.makeRenderInfo_(block);
+    assert.isTrue(renderInfo.isInline);
```
BeksOmega: What is this assertion guarunteeing? Is it to ensure that the base `shouldStartNewRow_` maintains its current behavior? Could you leave a comment on one of these assertions explaining why you're asserting it? (no need to duplicate the comment to all of them).

BeksOmega (approved):
LGTM! Approximately 50% nits and 50% compliments :P Thank you for your work on this John!

Once the nits are fixed up I'll merge this, and it'll either go out next Thursday, or before then if we can figure out #1452

plugins/inline-row-separators/test/inline_row_separators_test.mocha.js
johnnesky: Actually I searched the tests in blockly-samples for something that'll let me test a renderer and found that some plugin tests use it, like keyboard-navigation and workspace-search. I have no idea about what core blockly uses though.

plugins/inline-row-separators/README.md
johnnesky: Done. I also swapped the order of the sections for extending a different renderer and using the demonstration blocks.

plugins/inline-row-separators/src/inline_row_separators_renderer.ts
johnnesky: Done, although last time I tried this it upset the automatic lint check on this pull request (and I wasn't able to replicate the lint error locally), let's see what happens this time...

plugins/inline-row-separators/src/inline_row_separators_renderer.ts
johnnesky: Done.

plugins/inline-row-separators/src/inline_row_separators_renderer.ts
johnnesky: Done.

plugins/renderer-inline-row-separators/test/inline_row_separators_test.mocha.js
johnnesky: This is just testing that the structure of the new demonstration block definitions meet the requirements to demonstrate the new feature, in other words that they insert dummies between values and enable inline input mode, unlike the old blocks. The behavior of the renderer is tested elsewhere, and I've stopped referring to it here now in favor of checking the block's inputsInline property directly. I've added some comments.

plugins/inline-row-separators/src/inline_row_separators_renderer.ts
johnnesky: Yep, you can see that the check failed again:

```
lerna info run Ran npm script 'lint' in '@blockly/renderer-inline-row-separators' in 3.2s:

> @blockly/renderer-inline-row-separators@0.1.0 lint
> blockly-scripts lint

Running lint for @blockly/renderer-inline-row-separators

/home/runner/work/blockly-samples/blockly-samples/plugins/renderer-inline-row-separators/src/inline_row_separators_renderer.ts
  50:23  error  Parsing error: ';' expected

✖ 1 problem (1 error, 0 warnings)
```

Any suggestions on how to proceed?

BeksOmega: haha yeah, no need to keep it. Feel free to revert!

johnnesky: Done!

PR #1440, opened 2022-12-10
fix: Add DevSite demo to Blockly Samples

NeilFraser:
Resolves some 12 different issues with the current demo on DevSite.
https://github.com/google/blockly/issues/6609


examples/devsite-demo/script.js
```diff
@@ -0,0 +1,359 @@
...
+/**
+ * @fileoverview JavaScript for Blockly's DevSite demo.
+ */
+'use strict';
+
+
+/**
+ * Lookup for names of languages.  Keys should be in ISO 639 format.
+ */
+const LANGUAGE_NAME = {
```
rachel-fenichel: Suggestion: put language names and RTL lookup in a separate file from the rest of this.

rachel-fenichel:
Overall looks good to me. I had one suggestion on file structure, and we discussed whether the play button should be visible if the generated language is not JavaScript (Neil says yes but it should show that it runs the JS; I say no, but not strongly).

examples/devsite-demo/script.js
NeilFraser: I've moved all the language constants to msgs.js.  Don't really want to create yet another file -- there's no compiler so each additional file is a significant performance hit.

rachel-fenichel (approved):

PR #1454, opened 2022-12-14
fix: generator imports in examples

BeksOmega:
### Description

Fixes the generator imports in examples.

### Additional info

Sadly this could not be fixed with the migration script, because we didn't think anyone was using side effect imports to access generators off of `Blockly`, because we thought we never told them to. Apparently we did in these examples :/

### DIFF
--- a/examples/blockly-node/index.js
+++ b/examples/blockly-node/index.js
@@ -22,6 +22,7 @@
  */
 
 var Blockly = require('blockly');
+var {pythonGenerator} = require('blockly/python');
 
 var json = {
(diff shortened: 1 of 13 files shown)


rachel-fenichel:
Are these changes backwards compatible, or do we (technically) also need to update the minimum version of blockly for these examples?

rachel-fenichel (approved):

BeksOmega:
Ah yep, we need to update the min version, will do o7

rachel-fenichel (approved):

PR #1458, opened 2022-12-14
fix: Don't use private fields to access the workspace flyout buttons are on.

### LINKED ISSUE #1457: Some flyout buttons in test blocks are broken
**Category**
Plugins

**Component**
block-test

**Describe the bug**
The `add blocks to workspace`, `set random style`, `toggle enabled`, `toggle shadow`, and`toggle collapsed` flyout buttons provided by the block-test package are broken because they attempt to access the private `workspace_` property of the flyout button which has been renamed to remove the underscore and made truly private.

**To Reproduce**
1. Open the [playground](https://blockly-demo.appspot.com/static/tests/playground.html)
2. Choose the Test Blocks option
3. Click any of the above listed buttons
4. Observe the console errors

**Expected behavior**
The buttons do what they say they do

gonfunko:
Fixes #1457.

### DIFF
--- a/plugins/block-test/src/fields/defaults.js
+++ b/plugins/block-test/src/fields/defaults.js
@@ -280,7 +280,7 @@ export const category = {
  */
 export function onInit(workspace) {
   const randomizeLabelText = function(button) {
-    const blocks = button.targetWorkspace_.getBlocksByType(
+    const blocks = button.getTargetWorkspace().getBlocksByType(
         'test_fields_label_serializable');
     const possible = 'AB';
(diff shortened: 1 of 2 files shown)


BeksOmega (approved):
Lint :D Otherwise LGTM!

I would ask if you've tested this - but if I remember correctly that requires linking both block-test and dev-tools to the latest blockly, which can be tricky to get right. So I'm just going to assume this is working (cuz it should work) =)

gonfunko:
Fixed the lint issue, and correct, this hasn't been tested in practice but should work.

PR #1464, opened 2022-12-17
Chore: Converted the field-colour-hsv-sliders plugin to TypeScript.

johnnesky:
I had to temporarily disable @typescript-eslint/naming-convention for some SCREAMING_SNAKE_CASE constants and overriding a method that ends in an underscore, but I removed the underscores from new methods.

There are some properties that default to null values. I used "!" to access them in places that should be safe to do so, but there are a bunch of lint warnings:
```
Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion
```
Any preference on whether I leave the "!" in or explicitly test for null values before accessing them?


gonfunko:
> I had to temporarily disable @typescript-eslint/naming-convention for some SCREAMING_SNAKE_CASE constants and overriding a method that ends in an underscore, but I removed the underscores from new methods.
> 
> There are some properties that default to null values. I used "!" to access them in places that should be safe to do so, but there are a bunch of lint warnings:
> 
> ```
> Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion
> ```
> 
> Any preference on whether I leave the "!" in or explicitly test for null values before accessing them?

I believe we do generally want to keep lint/the styleguide happy, so checking for null before access or optional chaining if it makes sense would be great. The lint disable for the constants is fine though.

johnnesky:
Okay, I've removed the ! and added explicit null tests!

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.ts
```diff
@@ -0,0 +1,639 @@
...
+    return this;
+  }
+}
+
+/**
+ * Class for a colour input field that displays HSV slider widgets when clicked.
+ */
+export class FieldColourHsvSliders extends Blockly.FieldColour {
+  /* eslint-disable @typescript-eslint/naming-convention */
+  /** The maximum value of the hue slider range. */
```
gonfunko, line 181: IIRC styleguide guidance is to omit types for simple direct assignments

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.ts
```diff
@@ -0,0 +1,639 @@
...
+  /** HTML span element to display the current brightness. */
+  private brightnessReadout: HTMLSpanElement | null = null;
+
+  /** HTML range input element for editing brightness. */
+  private brightnessSlider: HTMLInputElement | null = null;
+
+  /** HTML div element containing all the labels and sliders. */
+  private dropdownContainer: HTMLDivElement | null = null;
+
+  /**
```
gonfunko: Do we need this constructor at all if it's just a passthrough for the superclass?

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.ts
```diff
@@ -0,0 +1,639 @@
...
+   * Create and show the colour field's editor.
+   * @override
+   */
+  protected showEditor_(): void {
+    /* eslint-enable @typescript-eslint/naming-convention */
+    this.createDropdownSliders();
+    if (!this.dropdownContainer || !this.hueSlider) {
+      throw new Error('Failed to initialize the HSV sliders.');
+    }
+    Blockly.DropDownDiv.getContentDiv().appendChild(this.dropdownContainer);
```
gonfunko, line 241: Maybe import DropDownDiv directly vs accessing through the Blockly namespace?

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.ts
```diff
@@ -0,0 +1,639 @@
...
+        'Saturation', container);
+    this.saturationSlider = FieldColourHsvSliders.createSliderInContainer(
+        FieldColourHsvSliders.SATURATION_SLIDER_MAX, 1, container);
+    this.brightnessReadout = FieldColourHsvSliders.createLabelInContainer(
+        'Brightness', container);
+    this.brightnessSlider = FieldColourHsvSliders.createSliderInContainer(
+        FieldColourHsvSliders.BRIGHTNESS_SLIDER_MAX, 1, container);
+
+    this.boundEvents.push(
+        Blockly.browserEvents.conditionalBind(
```
gonfunko, line 308: As with DropDownDiv, import browserEvents?

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.ts
```diff
@@ -0,0 +1,639 @@
...
+
+    const h: number = parseFloat(this.hueSlider.value) /
+        FieldColourHsvSliders.HUE_SLIDER_MAX;
+    const s: number = parseFloat(this.saturationSlider.value) /
+        FieldColourHsvSliders.SATURATION_SLIDER_MAX;
+    const v: number = parseFloat(this.brightnessSlider.value) /
+        FieldColourHsvSliders.BRIGHTNESS_SLIDER_MAX;
+
+    // The hue slider needs intermediate gradient control points to include all
+    // colours of the rainbow.
```
gonfunko, line 421: Up to you, but this might be a bit cleaner if the whole thing was made a template literal? I could see that potentially being worse as well though...

gonfunko (approved):

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.ts
johnnesky: There's a bunch of expression parsing in here, and this method isn't creating any new elements anyway, so I'll leave it the way it is.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.ts
johnnesky: Hm, actually I'm not sure how to do that in this context. If I add the import:
import {DropDownDiv} from 'blockly/core/dropdowndiv';
Upong running the test I get the error:
Error: Cannot find module 'blockly/core/dropdowndiv'

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.ts
johnnesky: Removed.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.ts
johnnesky: Done.

plugins/field-colour-hsv-sliders/src/field_colour_hsv_sliders.ts
gonfunko: Oh wait nevermind, I'm not sure if it's possible outside of core.

Issue #1326, opened 2022-10-12
Automatically update github pages after a samples release


maribethb:
**Category**

github actions

**Component**

**Is your feature request related to a problem? Please describe.**

We should keep the GitHub pages site up to date with the currently released plugins.

**Describe the solution you'd like**

A GitHub action that automatically updates GitHub pages after the samples release is done. It can probably use the existing deploy scripts.

There's also this information: https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site

**Describe alternatives you've considered**

**Additional context**

maribethb:
fixed by #1514

PR #1507, opened 2023-01-13
feat: Add workflow to deploy to GitHub Pages

maribethb:
Adds a GH Actions workflow that updates GitHub Pages. Currently, it's only triggered by manually running the workflow. If that works, I'll next set it up to run automatically after a publish workflow is successful.

I tested this on my fork and it worked, so I'm reasonably confident that this works.

The reason that I'm using `npm ci --force` when installing examples is some of the examples don't work without it and we don't have tests for examples that would catch that. Christopher also mentioned this issue recently. If we added a test for examples, then we could remove the --force flag. For now, personally I'm ok with it. The ones that currently don't have compatible dependencies aren't even shown on the website, so it's a bit unfortunate to block publishing the site just because the angular example is out of date.

work on #1326


BeksOmega:
@maribethb Do we have a tracking issue for adding that `npm ci` test to ci?

.github/workflows/update_gh_pages.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        ref: master
+
+    - name: Setup Node
+      uses: actions/setup-node@v3
+      with:
+        node-version: 16
```
cpcallen, line 24: Any reason not to use v18?  It will be supported longest.

.github/workflows/update_gh_pages.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: NPM install
+      # Use CI so that we don't update dependencies in this step.
+      # TODO: remove --force flag from installing examples when examples dependencies are routinely tested
```
cpcallen: I see @BeksOmega suggested a bug for this.  If you open one, please mention this TODO (and make it `TODO(#xxxx):…`).

.github/workflows/update_gh_pages.yml
```diff
@@ -0,0 +1,43 @@
...
+      # Until then, don't block updating just because some dependencies are out of date.
+      run: |
+        npm ci
+        cd examples
+        npm ci --force
+        cd ..
```
cpcallen: Presumably this works because `npm ci` runs the `postinstall` script, which runs `lerna boot`—which, as far as I can tell, runs `npm install` rather than `ci`.  Is `--force` getting passed through to the per-directory `install` invocations?  (I wouldn't expect it to even be passed to the `lerna boot` call, but maybe `lerna` is doing something clever?)

cpcallen: I think a subshell is usually the preferred way to do this, if it works (this is not a bash script, so I'm not sure):
```suggestion
        (cd examples && npm ci --force)
```

.github/workflows/update_gh_pages.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: Deploy to GH Pages
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      # deploys to the current repo, so you can use this workflow from a fork to update your personal gh-pages
```
cpcallen: Nice!

(Line length?)

.github/workflows/update_gh_pages.yml
```diff
@@ -0,0 +1,43 @@
...
+      run: |
+        npm run deploy:prepare
+        git remote set-url origin https://git:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
```
cpcallen, line 41: Will this work if one doesn't already have a remote named `origin`?  (I don't.)

cpcallen (commented):
Geerally looks good but a few questions.

(Feel free to reassign for re-review if you like or I will look again on Monday.)

.github/workflows/update_gh_pages.yml
maribethb: Nah, I just copied the publish workflow. Will update.

.github/workflows/update_gh_pages.yml
maribethb: Tbh I don't know why this works, when you point it out. The first time I tried it, without the force flag, it threw errors and failed the run due to incompatible dependencies in the angular example, so I just tried it with adding the flag without stopping to consider how lerna works, and it succeeded. 🤔

cpcallen: Huh.  TIL.  SGTM.

.github/workflows/update_gh_pages.yml
cpcallen: Oh, I realise that this is may be less of a concern than I thought, since I'm guessing _the repo on the CI machine_ has an `origin`, and that's what's relevant here.

(It's my local repo on my laptop that does not—because I kept getting confused between `origin` and `upstream`, and so have named them `google` and `cpcallen`.)

.github/workflows/update_gh_pages.yml
maribethb: Not all of the plugins are compatible with node 18. I think it might be the eslint config. Keeping it as node 16 for now but will file an issue for node 18 compatibility.

maribethb:
Ready for re-review. Tested again on my fork and everything works as expected.

.github/workflows/update_gh_pages.yml
cpcallen: SGTM.

cpcallen (approved):

PR #1514, opened 2023-01-19
feat: call the update gh-pages workflow after publishing succeeds

maribethb:
Calls the update gh-pages workflow from the publish workflow

Due to the way calling other workflows works, it has to be a separate job within the publish workflow, not just a step. Should only happen when publishing succeeds. Since skipping a job causes it to report "success", I have to repeat the if conditional to also skip the second job so that forks aren't updated every thursday

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -59,3 +59,9 @@ jobs:
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: npx lerna publish --conventional-commits --create-release github --yes --ignore-scripts
+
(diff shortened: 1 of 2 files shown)


maribethb:
GitHub Actions is experiencing failures: https://www.githubstatus.com/ The failing checks are due to that and not to the contents of this PR (the errors are about failing to download the source code for the external actions we use like checkout and setup-node)

@BeksOmega if you happen to review & approve this in the morning before 9am please go ahead and submit so we can see if it works during the regularly scheduled samples publish :) if not that's okay though!

BeksOmega (approved):

PR #1519, opened 2023-01-20
chore: Remove Jekyll templating.

NeilFraser:


maribethb:
Can you add more details about this change? Have you deployed it to your own fork's pages site to test it?

In particular I don't see where you're removing the Jekyll templating - so it seems like when the site is deployed then some of the page infrastructure would be added twice. 

Is the plan to remove Jekyll entirely? By default gh-pages uses Jekyll to build the site. So if we remove it we'd have to supply our own build system and reconfigure the GitHub Actions to not use the default build action.

NeilFraser:
My intention (though there seems to be some miscommunication over this) was to create the client-side replacement to Jekyll, and then Rachel would do the server-side removal of Jekyll.  I'm completely unable to publish or run Jekyll (which is why this issue arose in the first place), so I can't touch that side directly.

The principal benefits here are:
* An uncountable number of dependencies disappear (Jekyll obviously, but also Gem, Ruby, Sass, and who knows what else).  The current situation is unbuildable on my machine, for unknown reasons.
* Local testing becomes accurate.  Previously `npm install` would get each sample running, but without any of the framing, so we're not actually seeing what will be published.
* The current pages are wildly illegal, with a full html/head/body structure nested inside another body.  This sort of thing creates undefined behaviour and inconsistent rendering across browsers.

And during the course of making this change a whole bunch of unrelated issues were discovered and fixed.

You've got a good point, this PR should not be submitted until the corresponding server-side changes are made.  Accordingly I'll change this to a draft PR.

rachel-fenichel:
Superseded by https://github.com/google/blockly-samples/pull/1686

PR #1523, opened 2023-01-24
chore: add new plugin package for procedure blocks

BeksOmega:
### Description

Adds the boilerplate for the procedures plugin, but not any of the actual code.

Work on https://github.com/google/blockly/issues/6834

PR #1526, opened 2023-01-25
feat: add procedure data models

BeksOmega:
### Description

Moves the procedure data models & associated tests. from core into the procedures plugin.

Dependent on #1523


plugins/block-sharable-procedures/src/observable_parameter_model.ts
```diff
@@ -0,0 +1,129 @@
...
+   *     method is unimplemented.
+   */
+  setTypes(types: string[]): this {
+    throw new Error(
+        'The built-in ParameterModel does not support typing. You need to ' +
+        'implement your own custom ParameterModel.');
+  }
+
+  /**
+   * Returns the name of this parameter.
```
maribethb: This bothered me so I updated the eslint config to allow not having redundant descriptions: https://github.com/google/blockly-samples/pull/1537

if you want to rebase this PR on top of mine, you can remove these. but you'll also have to update your `@return` to `@returns`
otherwise I can update your return statements for you if you submit this before my PR is submitted.

plugins/block-sharable-procedures/src/observable_procedure_model.ts
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  /**
+   * Sets the human-readable name of the procedure.
+   * @param name The human-readable name of the procedure.
+   * @return This procedure model.
+   */
+  setName(name: string): this {
+    if (name === this.name) return this;
+    const prevName = this.name;
```
maribethb: nit: `oldName` for consistency

plugins/block-sharable-procedures/src/observable_parameter_model.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.shouldFireEvents = false;
+  }
+
+  /**
+   * Sets the procedure model this parameter is a part of.
+   * @param model The procedure model this parameter is a part of.
+   * @return This parameter model.
+   */
+  setProcedureModel(model: Blockly.procedures.IProcedureModel): this {
+    // TODO: Not sure if we want to do this, or accept it via the constructor.
```
maribethb: Making it be non-null seems better / developer can't forget this step and get weird bugs.

But I don't have a good sense of how the fluent API would be used, do you have an example?

maribethb (approved):

plugins/block-sharable-procedures/src/observable_parameter_model.ts
BeksOmega: It looks like #1537 is good to go in? So I'll just rebase on top of master once it's merged and update!

plugins/block-sharable-procedures/src/observable_parameter_model.ts
BeksOmega: So right now the API works like this (which is fluent):

```
const procedure = new ObservableProcedureModel('my procedure name')
    .setReturnTypes(['Number'])
    .insertParameter(new ObservableParameterModel('my param name'), 0);
```

While if we required the `ObservableProcedureModel` to be passed directly, it would have to look like this (which is not fluent):

```
const procedure = new ObservableProcedureModel('my procedure name')
    .setReturnTypes(['Number']);
 procedure.insertParameter(new ObservableParameterModel('my param name', procedure), 0);
```

I think it is better to preserve the fluent API in this case. After building it out, I don't think there's really a risk to making the `procedureModel` nullable, because the `ObservableProcedureModel` handles calling `setProcedureModel` inside `insertParameter`. External developers don't have to remember to call it.

PR #1536, opened 2023-01-31
feat: add procedure events

BeksOmega:
### Description

Pulls the procedure events from core into a plugin.

Dependent on #1526 

### Additional Info

For some reason, I could not get this plugin to properly link to `dev-tools` and blockly at the same time (even when linking devtools to blockly). So I was getting dependency problems that meant the assertion helpers in `dev-tools` would not work while the same code inside the plugin would work.

So for now the assertions just live in the plugin, but I might try to get it working again after #1526 is merged.

[Edit: Tried again now that we've got the beta, so I didn't have to link to Blockly. Still the same issue. The exact same code works in one directory, but not in another.]


plugins/block-sharable-procedures/src/events_procedure_create.ts
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * Constructs the procedure create event.
+   * @param workspace The workspace this event is associated with.
+   * @param procedure The model this event is associated with.
+   */
+  constructor(
+      workspace: Blockly.Workspace,
+      procedure: Blockly.procedures.IProcedureModel) {
+    super(workspace, procedure);
+  }
```
gonfunko: If we're just calling super, may as well omit this (and likewise for the other events that just pass through)?

plugins/block-sharable-procedures/src/events_procedure_enable.ts
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Notifies listeners that the procedure data model has been enabled or
+ * disabled.
+ */
+export class ProcedureEnable extends ProcedureBase {
+  /** A string used to check the type of the event. */
+  type = TYPE;
+
+  private oldState: boolean;
+  private newState: boolean;
```
gonfunko, line 23: Is there ever any situation where these wouldn't just be the inverse of one another, and therefore we could only store one? I can pretty easily be convinced that's being too clever for its own good, but the presence of both vaguely suggests that maybe it isn't that simple.

plugins/block-sharable-procedures/src/events_procedure_parameter_create.ts
```diff
@@ -0,0 +1,113 @@
...
+      procedureModel.deleteParameter(this.index);
+    }
+  }
+
+  /**
+   * Returns true if the given parameter is identical to this parameter.
+   * @param param The parameter to check for equivalence.
+   * @returns True if the parameter matches, false if it does not.
+   */
+  parameterMatches(param: Blockly.procedures.IParameterModel) {
```
gonfunko: Maybe add a return type annotation?

plugins/block-sharable-procedures/src/events_procedure_parameter_create.ts
```diff
@@ -0,0 +1,113 @@
...
+    return param && param.getId() === this.parameter.getId();
```
gonfunko: Is the param check here for the sake of sloppy JS? I feel like either the arg should be made optional, or the check should be omitted and allow an error to be thrown if called with null/undefined.

plugins/block-sharable-procedures/src/events_procedure_parameter_delete.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Returns true if the given parameter is identical to this parameter.
+   * @param param The parameter to check for equivalence.
+   * @returns True if the parameter matches, false if it does not.
+   */
+  parameterMatches(param: Blockly.procedures.IParameterModel) {
+    return param && param.getId() === this.parameter.getId();
+  }
```
gonfunko: Same comment as in the other file; perhaps it would make sense to move this method into events_procedure_parameter_base as well to reduce duplication?

plugins/block-sharable-procedures/src/events_procedure_parameter_rename.ts
```diff
@@ -0,0 +1,131 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+import {ProcedureParameterBase, ProcedureParameterBaseJson} from './events_procedure_parameter_base';
+
+
+const TYPE = 'procedure_parameter_rename';
+
+// TODO: Maybe delete this event since I don't think we're firing it.
```
gonfunko: It looks like this is now being fired, delete this TODO?

plugins/block-sharable-procedures/src/events_procedure_parameter_rename.ts
```diff
@@ -0,0 +1,131 @@
...
+    workspace: Blockly.Workspace,
+    modelId: string,
+    paramId: string
+): Blockly.procedures.IParameterModel|undefined {
+  const procedureModel = workspace.getProcedureMap().get(modelId);
+  if (!procedureModel) {
+    throw new Error(
+        'Cannot rename the parameter of a procedure that does not exist ' +
+        'in the procedure map');
+  }
```
gonfunko, line 104: Should this throw or just return undefined? The callers appear to be handling a failure to find the parameter in the model by throwing as well, so perhaps this should just also throw in that case as well and return a non-nullable IParameterModel?

plugins/block-sharable-procedures/src/events_procedure_rename.ts
```diff
@@ -0,0 +1,89 @@
...
+
+
+const TYPE = 'procedure_rename';
+
+/** Notifies listeners that a procedure model has been renamed. */
+export class ProcedureRename extends ProcedureBase {
+  /** A string used to check the type of the event. */
+  type = TYPE;
+
+  private newName: string;
```
gonfunko, line 18: Should this be readonly?

plugins/block-sharable-procedures/src/events_procedure_rename.ts
```diff
@@ -0,0 +1,89 @@
...
+   * Replays the event in the workspace.
+   * @param forward if true, play the event forward (redo), otherwise play it
+   *     backward (undo).
+   */
+  run(forward: boolean) {
+    const procedureModel =
+        this.getEventWorkspace_().getProcedureMap().get(this.procedure.getId());
+    if (!procedureModel) {
+      throw new Error(
+          'Cannot change the type of a procedure that does not exist ' +
```
gonfunko, line 45: Cannot *rename* a procedure that does not exist

plugins/block-sharable-procedures/test/event_procedure_change_return.mocha.js
```diff
@@ -0,0 +1,210 @@
...
+      });
+
+      test('one changing the return fires a change return event', function() {
+        const initial = this.createProcedureModel('test id');
+        const final = this.createProcedureModel('test id');
+        final.setReturnTypes(NON_DEFAULT_TYPES);
+        const event = this.createEventToState(final);
+        this.procedureMap.add(initial);
+
+        this.eventSpy.resetHistory();
```
gonfunko, line 77: Can/should this just be run in setup?

plugins/block-sharable-procedures/test/event_procedure_delete.mocha.js
```diff
@@ -0,0 +1,177 @@
...
+                'as the original model');
+          });
+
+      test('not creating a model does not fire a create event', function() {
+        const model = this.createProcedureModel('test name', 'test id');
+        const event = this.createEventToState(model);
+        this.procedureMap.add(model);
+
+        this.eventSpy.resetHistory();
+        event.run(/* forward= */ true);
```
gonfunko: Should this be false? Right now this is just doing a delete, and therefore not firing a create event, right?

plugins/block-sharable-procedures/test/event_procedure_parameter_create.mocha.js
```diff
@@ -0,0 +1,235 @@
...
+              index: 0,
+            },
+            this.workspace.id);
+      });
+
+      test(
+          'a parameter is not deleted if a parameter with a ' +
+          'matching ID and index does not exist',
+          function() {
+            // TODO: Figure out what we want to do in this case.
```
gonfunko, line 194: Implement or remove?

plugins/block-sharable-procedures/test/event_procedure_parameter_delete.mocha.js
```diff
@@ -0,0 +1,236 @@
...
+              index: 0,
+            },
+            this.workspace.id);
+      });
+
+      test(
+          'a parameter is not deleted if a parameter with a ' +
+          'matching ID and index does not exist',
+          function() {
+            // TODO: Figure out what we want to do in this case.
```
gonfunko, line 101: Handle this?

plugins/block-sharable-procedures/src/events_procedure_enable.ts
BeksOmega: It's just for readability. See [here](https://github.com/google/blockly/pull/6650/files#r1036377970).

plugins/block-sharable-procedures/src/events_procedure_parameter_delete.ts
BeksOmega: Moved!

plugins/block-sharable-procedures/src/events_procedure_parameter_rename.ts
BeksOmega: Good point! Changed `findMatchingParameter` to throw in the case of no parameter for consistency.

plugins/block-sharable-procedures/test/event_procedure_change_return.mocha.js
BeksOmega: It needs to be run after all of the code that actually creates the procedure models, to get rid of the events that those calls triggered.

plugins/block-sharable-procedures/test/event_procedure_delete.mocha.js
BeksOmega: Yes! Thank you for the catch :D

plugins/block-sharable-procedures/test/event_procedure_parameter_delete.mocha.js
BeksOmega: Implemented!

plugins/block-sharable-procedures/test/event_procedure_parameter_create.mocha.js
BeksOmega: Implemented!

BeksOmega:
Should be ready for another look @gonfunko

plugins/block-sharable-procedures/src/events_procedure_parameter_create.ts
```diff
@@ -49,23 +49,14 @@ export class ProcedureParameterCreate extends ProcedureParameterBase {
     }
     const parameterModel = procedureModel.getParameter(this.index);
     if (forward) {
-      if (this.parameterMatches(parameterModel)) return;
+      if (parameterModel && this.parameterMatches(parameterModel)) return;
```
gonfunko: Optional: I think it might be a bit nicer to make the arg to parameterMatches nullable, and do the non-null check in there, returning false if it is null?

BeksOmega: Ahhh ok I misunderstood what you were saying before. That makes much more sense haha. Sounds good!

gonfunko (approved):

PR #1537, opened 2023-02-01
chore: update eslint to allow returns for ts and omit redundant descriptions

maribethb:
Updates eslint config:

- Prefer `@returns` instead of `@return` in TS code. This is consistent with proper TsDoc and with core. Should be easier to move code between the two.
    - Also updates TS code to comply with this rule 
- Allow not having a description in a JsDoc block if there is a returns, param, or type block present. This is consistent with JS Style Guide section 7.8 and allows us to skip writing redundant function descriptions for simple functions.

### DIFF
--- a/plugins/eslint-config/index.js
+++ b/plugins/eslint-config/index.js
@@ -71,7 +71,13 @@ module.exports = {
         'forceRequireReturn': false,
       },
     ],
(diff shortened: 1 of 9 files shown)


rachel-fenichel (approved):
@NeilFraser FYI since you were looking at the `returns` part.

PR #1546, opened 2023-02-03
fix: Add the --in-place flag to the migration script README.

gonfunko:
By default, the migration script just dumps the changes to stdout. The docs made no reference to the `--in-place` flag, which causes the script to actually update the files, which is presumably the behavior most people would want and expect.


BeksOmega (approved):

maribethb:
related to #1359

PR #1550, opened 2023-02-08
feat: add workspace-backpack plugin type declarations

johnnyoshika:
## Problem

The `workspace-backpack` plugin is missing type declarations, which makes it difficult to add to TypeScript projects. Searching on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) yields no results.

## PR Summary:

* Adds the following type declaration files:
  * index.d.ts
  * backpack.d.ts
  * options.d.ts
* Adds `"types": "./src/index.d.ts"` to `workspace-backpack`'s package.json, as these type declarations aren't copied to `dist` during build, but the `src` folder is published along with `dist`

This PR is similar to https://github.com/google/blockly-samples/pull/1143

### DIFF
--- a/plugins/workspace-backpack/src/backpack.d.ts
+++ b/plugins/workspace-backpack/src/backpack.d.ts
@@ -62,7 +62,7 @@ export declare class Backpack extends Blockly.DragTarget {
      * @param {!Array<!Blockly.utils.Rect>} savedPositions List of rectangles that
      *     are already on the workspace.
      */
-    position(metrics: Blockly.MetricsManager.UiMetrics, savedPositions: Array<Blockly.utils.Rect>): void;
+    position(metrics: Blockly.MetricsManager.UiMetrics, savedPositions: Blockly.utils.Rect[]): void;
     /**
      * Returns the count of items in the backpack.
      * @return {number} The count of items.
@@ -72,7 +72,7 @@ export declare class Backpack extends Blockly.DragTarget {
      * Returns backpack contents.
      * @return {!Array<string>} The backpack contents.
      */
-    getContents(): Array<string>;
+    getContents(): string[];
     /**
      * Handles when a block or bubble is dropped on this component.
      * Should not handle delete here.
@@ -97,7 +97,7 @@ export declare class Backpack extends Blockly.DragTarget {
      * @param {!Array<!Blockly.Block>} blocks The blocks to be added to the
      *     backpack.
      */
-    addBlocks(blocks: Array<Blockly.Block>): void;
+    addBlocks(blocks: Blockly.Block[]): void;
     /**
      * Removes the specified block from the backpack.
      * @param {!Blockly.Block} block The block to be removed from the backpack.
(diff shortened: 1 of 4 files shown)


plugins/workspace-backpack/src/backpack.d.ts
```diff
@@ -0,0 +1,174 @@
...
+    getBoundingRectangle(): Blockly.utils.Rect;
+    /**
+     * Positions the backpack.
+     * It is positioned in the opposite corner to the corner the
+     * categories/toolbox starts at.
+     * @param {!Blockly.MetricsManager.UiMetrics} metrics The workspace metrics.
+     * @param {!Array<!Blockly.utils.Rect>} savedPositions List of rectangles that
+     *     are already on the workspace.
+     */
+    position(metrics: Blockly.MetricsManager.UiMetrics, savedPositions: Array<Blockly.utils.Rect>): void;
```
BeksOmega:
```suggestion
    position(metrics: Blockly.MetricsManager.UiMetrics, savedPositions: Blockly.utils.Rect[]): void;
```

Here and everywhere there's an array type.

plugins/workspace-backpack/src/options.d.ts
```diff
@@ -0,0 +1,32 @@
...
+        removeFromBackpack: boolean;
+        copyToBackpack: boolean;
+        copyAllToBackpack: boolean;
+        pasteAllToBackpack: boolean;
+        disablePreconditionChecks: boolean;
+    };
+};
+export type BackpackContextMenuOptions = {
+    allowEmptyBackpackOpen?: boolean;
+    contextMenu?: BackpackContextMenuOptions;
```
BeksOmega: I don't think this type should be recursively referencing itself.

plugins/workspace-backpack/src/options.d.ts
```diff
@@ -0,0 +1,32 @@
...
+    contextMenu: {
+        emptyBackpack: boolean;
+        removeFromBackpack: boolean;
+        copyToBackpack: boolean;
+        copyAllToBackpack: boolean;
+        pasteAllToBackpack: boolean;
+        disablePreconditionChecks: boolean;
+    };
+};
+export type BackpackContextMenuOptions = {
```
BeksOmega: I'm confused why this one is called `ContextMenuOptions` when the `BackpackOptions` include the context menu options.

This type is also only referenced within itself, and not by `parseOptions`.

BeksOmega (changes requested):
Thank you so much for putting up this contribution :D

Generally looks really good! But I think there needs to be a bit of cleanup on the options types =) If you could also fix the linting issues that would be wonderful!

BeksOmega:
Hiya @johnnyoshika are you still interested in working on this? If not I'm going to try to hand it off to another person =)

johnnyoshika:
@BeksOmega Yes, sorry. Will address this in the next couple of days.

plugins/workspace-backpack/src/backpack.d.ts
johnnyoshika: I fixed the TypeScript types from `Array<>` to `[]`, but left the JSDoc declarations as is. Let me know if I should be fixing JSDocs as well.

plugins/workspace-backpack/src/options.d.ts
johnnyoshika: Good catch. Not sure what happened to `BackpackContextMenuOptions `. Fixed and fixed.

johnnyoshika:
> If you could also fix the linting issues that would be wonderful!

I correct all of the linting issues. This should be ready for your review again.

BeksOmega (approved):
Thanks for the fixes! LGTM

PR #1544, opened 2023-02-03
feat: add procedure blocks

BeksOmega:
### Description

Moves over the procedure blocks that have an explicit backing data model.

Work on https://github.com/google/blockly/issues/6834

### Additional Info

Dependent on #1536

PR #1551, opened 2023-02-09
fix: running procedure events in other workspaces

BeksOmega:
### Description

When I hooked up the sharing procedures test page, I found some bugs with running the events in other workspaces.

Namely:
  * Some of the events didn't work because we weren't round-tripping the new value (it was basing the new value on inspection, which doesn't work when instantiating into a different worksapce)
  * The parameter create events weren't round-tripping the variable ID, so then if you renamed the variable in another workspace after instantiating it, the rename wouldn't properly propogate.

And I also made it consistent that events will throw when they run into erroneous situtations when deserializing /and/ when running.

### Additional Info

Dependent on #1544

### DIFF
--- a/plugins/block-sharable-procedures/src/events_procedure_parameter_base.ts
+++ b/plugins/block-sharable-procedures/src/events_procedure_parameter_base.ts
@@ -30,10 +30,41 @@ export abstract class ProcedureParameterBase extends ProcedureBase {
    * Returns true if the given parameter is identical to this parameter.
    * @param param The parameter to check for equivalence.
    * @returns True if the parameter matches, false if it does not.
+   * @internal
    */
   protected parameterMatches(
       param?: Blockly.procedures.IParameterModel): boolean {
     return param && param.getId() === this.parameter.getId();
+  }
+
+  /**
+   * Finds the parameter with the given ID in the procedure model with the given
+   * ID, if both things exist.
+   * @param workspace The workspace to search for the parameter.
+   * @param modelId The ID of the model to search for the parameter.
+   * @param paramId The ID of the parameter to search for.
+   * @returns The parameter model that was found.
+   * @internal
+   */
+  static findMatchingParameter(
+      workspace: Blockly.Workspace,
+      modelId: string,
+      paramId: string
+  ): ProcedureParameterPair {
+    const procedure = workspace.getProcedureMap().get(modelId);
+    if (!procedure) {
+      throw new Error(
+          'Cannot find the parameter of a procedure that does not exist ' +
+          'in the procedure map');
+    }
+    const parameter = procedure.getParameters()
+        .find((p) => p.getId() === paramId);
+    if (!parameter) {
+      throw new Error(
+          'Cannot find a parameter that does not exist in ' +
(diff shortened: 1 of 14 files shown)


plugins/block-sharable-procedures/src/events_procedure_parameter_base.ts
```diff
@@ -30,12 +30,43 @@ export abstract class ProcedureParameterBase extends ProcedureBase {
...
   protected parameterMatches(
       param?: Blockly.procedures.IParameterModel): boolean {
     return param && param.getId() === this.parameter.getId();
   }
 
+  /**
+   * Finds the parameter with the given ID in the procedure model with the given
+   * ID, if both things exist.
+   * @param workspace The workspace to search for the parameter.
+   * @param modelId The ID of the model to search for the parameter.
```
cpcallen: Naming: what is this "model" whose ID you speak of?  Is this a procedure model?  Does the procedure model potentially have a different ID than the procedure it models?  If not, would `procedureId` be a better name?

plugins/block-sharable-procedures/src/events_procedure_parameter_create.ts
```diff
@@ -28,7 +30,7 @@ export class ProcedureParameterCreate extends ProcedureParameterBase {
   constructor(
       workspace: Blockly.Workspace,
       procedure: Blockly.procedures.IProcedureModel,
-      parameter: Blockly.procedures.IParameterModel,
+      parameter: ObservableParameterModel,
```
cpcallen, line 33: This makes me sad.  😢

As discussed: why not generate a `VarCreate` event before each `ProcedureParameterCreate` event that that has the side-effect of creating a new variable, so that you can just look up the variable by name, rather than having to store its ID in the PPC event?

plugins/block-sharable-procedures/src/events_procedure_parameter_base.ts
```diff
@@ -79,8 +60,8 @@ export abstract class ProcedureParameterBase extends ProcedureBase {
 }
 
 export interface ProcedureParameterPair {
-  procedure: Blockly.procedures.IProcedureModel;
-  parameter: Blockly.procedures.IParameterModel;
+  procedure?: Blockly.procedures.IProcedureModel;
+  parameter?: Blockly.procedures.IParameterModel;
```
cpcallen, line 65: Kind of funny for this to be called `…Pair`, when it could be just one or indeed none!

cpcallen (approved):
As discussed I think the 'fix: unify events to throw when deserializing and noop when running' commit is a tiny bit dubious from a design perspective—but this is not an objection, more just a "have you thought this through carefully?"—I trust you to have a better perspective on this than I do.

Otherwise, see below.

plugins/block-sharable-procedures/src/events_procedure_parameter_base.ts
BeksOmega: Haha I don't have a better idea so I think we're sticking with this.

Although when I was ideating "ProcedureClump" did pop into my head, and I find this very amusing :P

PR #1553, opened 2023-02-10
feat: Add ability to create sample app

### LINKED ISSUE #1338: Improve create-package script
**Category**

scripts

**Component**

create-package

- create-package is only useful for creating plugins, but it would be great if it were used to bootstrap new applications using blockly
- it has too many templates that are too similar to each other, which is extra maintenance for us
- it doesn't handle first-party vs third-party packages that well

design doc incoming

**Additional context**

maribethb:
Adds the ability to create a clone of the blockly sample-app to the create-package script. Now, you can run `npx @blockly/create-package app hello-world` to create a new application, and the old `npx @blockly/create-package my-plugin-name` still works the same way as it did before (except you can't create plugins named `app` or `application`).

Fixes #1338 

- Reorganized the `create.js` file.
    - Moved the action of creating a plugin to its own file, largely unchanged except for extracting the common code.
    - Removed the validation of the plugin name, because commander handles that for us
    - Added a sub-command for creating applications
    - Pulled out some common code between the two modes into `common.js`
- Updated `commander` version to get to use new features, and updated minimum node engine version accordingly
- Updated README of create-package script with new instructions
- Update README of sample-app to tell you to use the script instead of cloning samples. 

Tested via

```
node ./plugins/dev-create/bin/create field-foo --skip-install --first-party --type field
node ./plugins/dev-create/bin/create app hello-world
node ./plugins/dev-create/bin/create --help
```

etc.


plugins/dev-create/README.md
```diff
@@ -1,17 +1,22 @@
 # @blockly/create-package [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
 
-A tool for creating a Blockly plugin based on a pre-existing template.
+A tool for creating a Blockly plugin or application based on a pre-existing template.
```
cpcallen, line 3: Line length (here and below).

plugins/dev-create/README.md
```diff
@@ -20,14 +25,31 @@ npm start
...
 
-## Adding TypeScript
+### Adding TypeScript
 ```
 npx @blockly/create-package my-plugin --type plugin --typescript
 ```
 
+## Creating a new application
+
+An application that uses Blockly is a standalone program, unlike a plugin. If you're just getting started with Blockly, you can create a new application based off of our sample using this script. The app that will be generated contains basic infrastructure for running, building, testing, etc. that you can use even if you don't understand how to configure the related tool yet. When your needs outgrow the functionality provided by it, you can replace the provided configuration or tool with your own.
```
cpcallen: "based off of" -> "based on"

plugins/dev-create/bin/app.js
```diff
@@ -0,0 +1,64 @@
...
+  const sampleDir = '../../../examples/sample-app';
+  fs.copySync(path.resolve(__dirname, sampleDir), appPath, {
+    filter(src) {
+      const excludes = [
+        'node_modules', 'dist', 'package-lock.json', 'package.json'];
+      for (const file of excludes) {
+        if (src.indexOf(file) > -1) return false;
+      }
+      return true;
+    },
```
cpcallen: My reading of `fs-extra`'s `copySync` is that [the filter returning false prunes the entire subtree](https://github.com/jprichardson/node-fs-extra/blob/f3a7f0beeb5858c628b10010ad819c813c7f3565/lib/copy/copy-sync.js#L29) so I think this whole section can be simplified:
```suggestion
    filter(src) {
      return !(['node_modules', 'dist', 'package-lock.json', 'package.json'].includes(file));
    },
```

If for some reason that doesn't work then I think:
* The formatting of `excludes` could be checked (did it get `clang-format`ted?)
* You should use `.includes` instead of `.indexOf`.

plugins/dev-create/bin/app.js
```diff
@@ -0,0 +1,64 @@
...
+
+  // Write the package.json to the new package.
+  fs.writeFileSync(
+      path.join(appPath, 'package.json'),
+      JSON.stringify(packageJson, null, 2));
+
+  // Run `npm install` to get things ready for user.
+  if (!options.skipInstall) {
+    console.log('Installing packages. This might take a couple of minutes.');
+    execSync(`cd ${dir} && npm install`, {stdio: [0, 1, 2]});
```
cpcallen: I think `'inherit'` is clearer:
```suggestion
    execSync(`cd ${dir} && npm install`, {stdio: 'inherit'});
```

plugins/dev-create/bin/app.js
```diff
@@ -0,0 +1,64 @@
...
+  // Print helpful instructions.
+  console.log(chalk.green('Success!'));
+
+  console.log(`You can start the development server now by typing:`);
+  console.log(chalk.blue(`  cd ${dir}`));
+  if (options.skipInstall) {
+    console.log(chalk.blue(`  npm install`));
+  }
+  console.log(chalk.blue(`  npm start`));
+  console.log(`See README.md for more commands and information.`);
```
cpcallen: Opinion: I'd avoid using template literals when plain strings would do.

plugins/dev-create/bin/common.js
```diff
@@ -0,0 +1,17 @@
+const fs = require('fs-extra');
+
+exports.createDir = function(path) {
```
cpcallen: A name (and JSDoc) distinguishing how this is different from `fs.mkdir` would be useful.  Not sure exactly what name would be best; `createNewDir` maybe?

plugins/dev-create/bin/plugin.js
```diff
@@ -0,0 +1,204 @@
...
+  let gitRoot = '';
+  let gitURL = '';
+  try {
+    gitRoot = execSync(`git rev-parse --show-toplevel`).toString().trim();
+    gitURL = execSync(`git config --get remote.origin.url`)
+        .toString()
+        .trim()
+        .replace(/\.git$/, '');
+  } catch (err) {
+  // NOP
```
cpcallen:
```suggestion
    // NOP
```

plugins/dev-create/bin/plugin.js
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  const isGit = !!gitURL;
+  const isFirstParty = options.firstParty || gitURL == 'https://github.com/google/blockly-samples';
```
cpcallen: Line length.

plugins/dev-create/bin/plugin.js
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * Gets the name of the plugin prefixed with the type.
+   * Default to the plugin name if the type=plugin, otherwise use [type]-[name].
+   * @param {string} name Plugin name.
+   * @param {string} type Plugin type.
+   * @return {string} Plugin name prefixed with type.
+   */
+  const getPrefixedName = function(name, type) {
+  // Don't add 'plugin' prefix for default type
```
cpcallen:
```suggestion
    // Don't add 'plugin' prefix for default type
```

plugins/dev-create/bin/plugin.js
```diff
@@ -0,0 +1,204 @@
...
+    console.log(`Available types: ${chalk.green(pluginTypes.join(', '))}`);
+    process.exit(1);
+  }
+
+  // Create the new directory if needed.
+  createDir(pluginPath);
+
+  console.log(`Creating a new Blockly\
+ ${chalk.green(pluginType)} with name ${chalk.green(pluginName)}\
+ in ${chalk.green(pluginPath)}.\n`);
```
cpcallen: [No line continuations](go/tsstyle#syntax-no-line-continuations):

```suggestion
  console.log(
      `Creating a new Blockly ${chalk.green(pluginType)} with name ` +
      `${chalk.green(pluginName)} in ${chalk.green(pluginPath)}.\n`);
```

plugins/dev-create/bin/app.js
```diff
@@ -0,0 +1,64 @@
...
+        'node_modules', 'dist', 'package-lock.json', 'package.json'];
+      for (const file of excludes) {
+        if (src.indexOf(file) > -1) return false;
+      }
+      return true;
+    },
+  });
+
+  // Create package.json file for app.
+  const templateJson = require(path.join(sampleDir, 'package.json'));
```
cpcallen, line 36: Any particular reason you opt to read this from the filesystem whereas for plugins it's hardcoded into the script?

cpcallen (changes requested):
Opinion: I think if we're introducing an `app` (sub)command, then the existing code should be accessed via a `plugin` (sub)command.  For backwards compatibility `plugin` could be the default command if none is given, but a warning could be printed.

maribethb:
> Opinion: I think if we're introducing an `app` (sub)command, then the existing code should be accessed via a `plugin` (sub)command. For backwards compatibility `plugin` could be the default command if none is given, but a warning could be printed.

I didn't know you could set a default command, and I didn't want to break existing users / muscle memory / documentation etc. You can set a default though so I did. I don't see any reason to print a warning (nor a way to do that easily)

plugins/dev-create/bin/app.js
maribethb: The sample app directory is a standalone directory we are copying and I hope to put that on the examples site soon, so it needs a real package.json. It wouldn't make sense to also have this information hardcoded into the script, because you'd have to update it in both places.

The plugin template used to have both a template file and some hardcoded sections, but the template files were not useful because there was one for each type of plugin, and all of them were entirely redundant with the hardcoded version. Those templates are not standalone packages so they didn't need the template file, and the plugin script does a lot more modification of the values present anyway, so it makes more sense for that to be generated entirely within the script.

cpcallen: Sounds sensible to me!

cpcallen (approved):
> I didn't know you could set a default command,

I didn't know that was a commander feature either, but I do know that where there's a will there a way.

PR #1555, opened 2023-02-10
feat: Create plugin for dependent dropdown fields.

### LINKED ISSUE #964: Pure validators and idiomatic support for handling dependent fields such as dependent dropdowns
**Category**

- Plugins

**Component**

- Field

It is currently possible for valid states to incorrectly be labeled as invalid or invalid states to be loaded from XML/JSON.  This can occur when correct validation of fields depends on the values of other fields, block connections, or other workspace state.  Related to this, developers sometimes need to resort to creating inelegant and potentially confusing error-prone workarounds to address validators not being explicitly given all the inputs they need to determine whether values are valid as illustrated in this forum post here: https://groups.google.com/g/blockly/c/uilMLN_VojA/m/sKd6LjNNAwAJ.

The ability to use pure validators in a standardized way to ensure that fields dependent on other workspace state always have valid values and that valid states can be deserialized from XML/JSON.  The following is a quick draft of some of the tasks that could facilitate such a plugin (Suggestions for improvements or refinements are very much welcome!):

- [ ] Add (protected) instance variables `virtualState` to the `Block` and `Workspace` classes representing "virtual" state that is used for determining/validating actual workspace state such as the values of fields, the descendants of blocks, etc.
- [ ] Add a `setState` method to the `Block` and `Workspace` classes to update the virtual state, if the change is valid and then update the actual state, accordingly:
  1. Freeze the actual workspace state to prevent inadvertent changes.
  2. Call the relevant pure `getNewState` listeners, to get additional updates to the state or reject the proposed changes.
  3. Call the relevant pure `validateState` validators to determine if the resulting virtual state from the final `getNewState` listener requesting changes is valid.
  4. Unfreeze the workspace.
  5. If all validators agree that the final virtual state is valid, set the workspace or block `virtualState` to be this new state and call the relevant virtual state => actual state function that makes the necessary changes to the block/workspace to reflect the new virtual state.
- [ ] Add a `Block`/`Workspace` method `addStateListener` to add a pure function `getNewState` listener.  These listeners would request additional changes through their outputs or defer additional changes until more properties of the desired new state are determined, accordingly.  They would do this by receiving the current state and currently constructed new state as inputs and return the new desired state as an output.  The resulting output would then be passed as the new state input to the next listener.  Alternatively, it can return `undefined` to defer its request to make changes to the new state, which would put it at the end of the listener queue, or return `null` to assert that this change is invalid and stop the process.  To prevent functions from perpetually deferring, `setState` would stop this process if a listener was returned to without any requested changes being made by the remaining listeners from its last deferment.
- [ ] Add a `Block`/`Workspace` method `addStateValidator` to add pure function validators.  After `setState` was called and all listeners were dispatched, the final new state output from a `getNewState` listener (or the input to `setState` if there were no listeners requesting changes) would be passed as an an input to each of these validators.  If every such validator returns `true`, the state change is accepted.  Otherwise, it is rejected.
- [ ] Add a `Block`/`Workspace` method `setVirtualToActualStateConverter` to set the function responsible for converting virtual state to actual state.  I.e., given a virtual state as an input, the function would call the appropriate methods such as `setFieldValue` to make the workspace reflect the new final accepted virtual state from `setState`.
- [ ] Add a workspace/block method to freeze/unfreeze all actual workspace/block state to ensure the virtual state listeners/validators don't inadvertently change actual state.  This should be a developer freeze in that methods such as `setFieldValue` should throw an error when trying to use it to change a field value, as opposed to a user freeze such as calling `setEditable(false)` on a block which only prevents a field value from being changed by clicking on the field.  This could be done by adding protected `Boolean` properties `isFrozen` that are checked by any methods that change state.
- [ ] Allow fields to have a `disableSaving` method so that their field values would not be serialized or deserialized (even if they were present).  The point is to have these values set separately after they have gone through the validation process as described above.
- [ ] Extend serialization methods to include this virtual workspace/block state.  Extend deserialization methods to call all relevant validators on the saved virtual state before calling the virtual => actual state converters, if validation is successful.

Providing extra documentation to standardize best practices around handling common use cases such as dependent dropdowns or providing plugins for each of these use cases.

**Additional context**

For current examples in which you can get unavailable or invalid field values after serializing and surrounding discussion, see https://github.com/google/blockly/issues/5787 and https://github.com/google/blockly/issues/5793.  If this seems like a viable/useful idea, I could try to have a few of my students work on this next semester.

johnnesky:
Fixed #964.

Defines a new field called FieldDependentDropdown that automatically updates its own available options depending on the current value of a parent field on the same block. It supports undoing and deserializing dependent fields. The parent field must be attached to the block before the child field.


plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
```diff
@@ -0,0 +1,354 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview A tool for making a dropdown field's options depend on the
+ * value of another field.
+ */
+
+import * as Blockly from 'blockly/core';
+import {Block} from 'blockly/core/block';
```
maribethb: Importing specific modules from outside of core is not currently supported. Instead, in a plugin, each of these should be referenced as `Blockly.Block` etc.

It's interesting to me that this seems to work for you, but this isn't something we currently recommend that people do, and it may change in the future depending on Christopher's work. @cpcallen any thoughts on this?

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
```diff
@@ -0,0 +1,354 @@
...
+  oldOptions: MenuOption[];
+  newOptions: MenuOption[];
+}
+
+/** The type of the mapping from parent value to child options. */
+export interface ChildOptionMapping{
+  [key: string]: MenuOption[];
+}
+
+declare module 'blockly/core/field_dropdown' {
```
maribethb: This doesn't feel right to me... I don't think we should be reaching into Blockly core's namespace and adjusting it. We would normally have a subclass of the Field to do this. Or you could define your own interface.

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
```diff
@@ -0,0 +1,354 @@
...
+ * A change event representing a simultaneous change to a dropdown field's
+ * options and value. The old value must be one of the old options, and the new
+ * value must be one of the new options. However, unlike a normal value change
+ * event, it's possible for this event to change the value to something that
+ * wasn't previously one of the valid options, in either direction, by also
+ * changing the options at the same time.
+ */
+class DependentDropdownOptionsChange extends Blockly.Events.BlockBase {
+  /** The name to register with Blockly for the type of event. */
+  /* eslint-disable @typescript-eslint/naming-convention */
```
maribethb: nit/fyi: you can use // eslint-disable-next-line so that you don't have to re-enable the rule after

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
```diff
@@ -0,0 +1,354 @@
...
+
+/**
+ * A change event representing a simultaneous change to a dropdown field's
+ * options and value. The old value must be one of the old options, and the new
+ * value must be one of the new options. However, unlike a normal value change
+ * event, it's possible for this event to change the value to something that
+ * wasn't previously one of the valid options, in either direction, by also
+ * changing the options at the same time.
+ */
+class DependentDropdownOptionsChange extends Blockly.Events.BlockBase {
```
maribethb: Can you move the event stuff to its own file?

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
johnnesky: Done.

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
johnnesky: Okay, I've redesigned the plugin as a subclass of FieldDropdown, and I think it turned out to be a much better API! It no longer requires using the JavaScript interface!

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
johnnesky: Done.

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
johnnesky: Done.

plugins/field-dependent-dropdown/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@blockly/field-dependent-dropdown",
+  "version": "1.0.0",
+  "description": "A tool for making a dropdown field's options depend on the value of another field.",
```
maribethb: nit: A Blockly dropdown field that depends on the value of another field

(or something else that says this is now a field and not a tool)

plugins/field-dependent-dropdown/package.json
```diff
@@ -0,0 +1,64 @@
...
+  "scripts": {
+    "audit:fix": "blockly-scripts auditFix",
+    "build": "blockly-scripts build",
+    "clean": "blockly-scripts clean",
+    "lint": "blockly-scripts lint",
+    "predeploy": "blockly-scripts predeploy",
+    "prepublishOnly": "npm run clean && npm run build",
```
maribethb: Can you remove the "prepublishOnly" script? We recently made some changes and this has been removed from all plugins now

plugins/field-dependent-dropdown/src/dependent_dropdown_options_change.ts
```diff
@@ -0,0 +1,201 @@
...
+    this.newOptions = json['newOptions'];
+  }
+
+  /**
+   * Does this event leave all state as it was before?
+   * @returns False if something changed.
+   */
+  isNull(): boolean {
+    return !this.oldValue ||
+        !this.newValue ||
```
maribethb: Is this right? This would mean if oldValue doesn't exist and newValue does exist, we'd say it's a null event. Or if one of the values was the empty string, which you set as the default option if you can't find a configured parent mapping. I think if you were in that situation where the old value was a default / "not found" option due to not being in the mapping, then the subsequent change event would be marked as null. Or the change marking getting into that state.

It seems like the fields in core just check for equality. That would allow it to work for field values of '' as well.

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
```diff
@@ -0,0 +1,266 @@
...
+
+import * as Blockly from 'blockly/core';
+import {DependentDropdownOptionsChange} from './dependent_dropdown_options_change';
+
+/** The type of the mapping from parent value to child options. */
+export interface ChildOptionMapping {
+  [key: string]: Blockly.MenuOption[];
+}
+
+// This type isn't exported from Blockly so we have to derive it from the API.
```
maribethb, line 20: Would you mind filing a bug for this? We should probably be exporting it.

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
```diff
@@ -0,0 +1,266 @@
...
+    const parentField: Blockly.Field<string> | null =
+        block.getField(this.parentName);
+
+    if (!parentField) {
+      throw new Error('Could not find a parent field with the name ' +
+          this.parentName + ' for the dependent dropdown.');
+    }
+
+    if (parentField.getValidator()) {
+      throw new Error('The parent field with the name ' +
```
maribethb: Instead of throwing an error, can you let the parent field have its own validator, and then call the old validator from the new one you set? Because especially if the parent field's value should be restricted to a specific set of values (that are supported in the dependent dropdown), it would be intuitive for that work to occur in the validator originally passed to the parent field.

As it is, you are allowing the value to be set to something not in the config options list (by returning undefined from the validator in that case, which allows the newValue to be set as is), so there is no way to actually restrict the value of the field, which you could imagine someone wanting to do.

plugins/field-dependent-dropdown/test/field_dependent_dropdown.mocha.js
```diff
@@ -0,0 +1,229 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const chai = require('chai');
+const sinon = require('sinon');
+const Blockly = require('blockly');
```
maribethb: this might need to be `blockly/node` since we run mocha tests in node environment, and the default blockly package is browser (not 100% confident on that, maybe our default package is umd, but `blockly/node` is more explicit)

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
johnnesky: Done, although for timing reasons it only works if you use the full JavaScript API. If a user's extension adds a validator to the parent field, that will happen after the dependency has been initialized along with its built-in validator, in which case it doesn't know yet about the one the extension wants to add and the extension would just end up replacing the one set up by the dependency. 

I guess the user's extension could compose the validator functions together similar to how this plugin does it, if they know what they're doing, but it wouldn't be obvious. 

I'll add a description to the README about how to use the JavaScript API to add a validator to the parent.

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
johnnesky: Whoops, looks like you already did this while I was on vacation! https://github.com/google/blockly/issues/6870

johnnesky:
Thanks for the feedback, I believe I've addressed it so far!

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
maribethb: If you could add a TODO here with the type it should be so that it's easier to follow up after the next Blockly release, that would be great. e.g. `// TODO: Use Blockly.DropdownFieldConfig instead`

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
```diff
@@ -0,0 +1,274 @@
...
+    // place to do this because it is called immediately while deserializing
+    // workspaces before the following fields are deserialized, so when the
+    // child value is deserialized the appropriate options will already be
+    // available. If the parent already had a validator function, it will be
+    // composed with this one and the new value returned from it will be the
+    // basis for determining the new available options.
+    parentField.setValidator((newValue) => {
+      if (oldValidator) {
+        const validatedValue = oldValidator(newValue);
+        // If a validator returns undefined, that means no change. Otherwise,
```
maribethb, line 208: This is not quite right. The validator can return one of three types of values:
- `null` means the value was invalid. If the original parent validator returns null, we also want to return null from this validator, and not update the options, because the value is not actually going to change. The parent field's `doValueInvalid_` method will be called, but that's outside the scope of the child field. From the child field's perspective, the value isn't changing and we shouldn't update anything.
- `undefined` means set the `newValue` exactly it was proposed, e.g. `newValue = newValue`
- any other value means set the `newValue` to the value returned from the validator, e.g. `newValue = validatedValue`.

So you need to handle all three cases here, by adding

```
if (validatedValue === null) return null;
```

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
```diff
@@ -0,0 +1,274 @@
...
+    });
+    this.updateOptionsBasedOnNewValue(parentField.getValue());
+  }
+
+  /**
+   * Updates the options of this child dropdown field based on the new value of
+   * the parent field.
+   * @param newValue The newly assigned value.
+   */
+  private updateOptionsBasedOnNewValue(newValue: string | undefined): void {
```
maribethb, line 225: Is `newValue` actually allowed to be undefined? I don't think it's a valid value that would ever get here, because the class validator for field rejects `null` and `undefined` values, in which case the local validator defined above is never called and the field's value is not updated.

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
```diff
@@ -0,0 +1,274 @@
...
+ * A dropdown field that automatically updates its own options based on a
+ * mapping from a parent field's value. The parent field must be attached to the
+ * block before this child field.
+ *
+ * When this field is attached to a block, it will find the parent field and
+ * attach a validator to the parent field that intercepts changes to its value
+ * to update the options on this field. If the new validator is later removed\
+ * or replaced, then this dependent field will no longer function.
+ */
+export class FieldDependentDropdown extends Blockly.FieldDropdown {
```
maribethb, line 60: As a follow up feature request (you don't have to do as part of this PR) it would be great if you could configure a default set of options to use if the parent field's value is not in the optionMapping. That would enable this to more easily be used with parent fields that aren't necessarily dropdown fields.

plugins/field-dependent-dropdown/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+```js
+import * as Blockly from 'blockly';
+import {FieldDependentDropdown} from '@blockly/field-dependent-dropdown';
+Blockly.Blocks['dependent_dropdown_example'] = {
+  init: function() {
+    const parentFieldName = 'ANIMAL_CATEGORY';
+    const childFieldName = 'ANIMAL';
+    const parentOptions = [['Mammal', 'mammal'], ['Bird', 'bird']];
+    const childOptions = {
```
maribethb: nit: could you call this `optionMapping` to be more consistent with the json api?

maribethb (commented):
Overall, looks good. There is just one remaining problem with the validator.

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
johnnesky: The type of FieldValidator is:
export type FieldValidator<T = any> = (value?: T) => T | null | undefined;
...Wait no it's not anymore, it was updated in https://github.com/google/blockly/commit/0fb64a6772bfa3a57c0736f551d96a99fced98ec
export type FieldValidator<T = any> = (newValue: T) => T | null | undefined;
But the argument was still optional in the latest release 9.2.1 of Blockly on npm, and this plugin is configured to depend on ^9.0.0 anyway.

plugins/field-dependent-dropdown/src/field_dependent_dropdown.ts
johnnesky: Good idea, I went ahead and implemented it here, thanks!

johnnesky:
Done! I even added the default options.

maribethb (approved):
Thanks, John!

PR #1556, opened 2023-02-10
fix: create events not fired

BeksOmega:
### Description

Fixed create events not being fired when dragging blocks from the flyout or deserializing.

This doesn't properly order events. Ideally the procedure create event would happen before the block create event so that it would work properly with real time collaboration. But this is not possible due to https://github.com/google/blockly/issues/6840

### Testing

Added unit tests that check that events are properly fired when they are supposed to be. Surprised I didn't have this before but 🤷

### DIFF
--- a/plugins/block-sharable-procedures/src/blocks.ts
+++ b/plugins/block-sharable-procedures/src/blocks.ts
@@ -218,7 +218,7 @@ const procedureDefGetDefMixin = function() {
       Blockly.Procedures.findLegalName(this.getFieldValue('NAME'), this));
 
   // Events cannot be fired from instantiation when deserializing or dragging
-  // from the flyout. So make this consistant and never fire from instantiation.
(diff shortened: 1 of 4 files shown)


plugins/block-sharable-procedures/src/blocks.ts
```diff
@@ -215,7 +216,12 @@ const procedureDefGetDefMixin = function() {
   mixin.model_ = new ObservableProcedureModel(
       this.workspace,
       Blockly.Procedures.findLegalName(this.getFieldValue('NAME'), this));
+
+  // Events cannot be fired from instantiation when deserializing or dragging
+  // from the flyout. So make this consistant and never fire from instantiation.
```
rachel-fenichel: Nit: consistant -> consistent

rachel-fenichel (approved):

PR #1559, opened 2023-02-11
fix: Remove double entry of 'kind' in toolbox-demo

instance01:
### Description

Remove double entry of 'kind' key in toolbox-demo for the math_random_int block.


BeksOmega (approved):
Thank you for the fix @instance01 ! This looks perfect =) And thank you for dealing with the CLA and conventional commits hehe.

PR #1561, opened 2023-02-13
chore: properly export events and event types

BeksOmega:
### Description

Firstly events were not being exported from the index file, which meant they weren't accessible to external developers. This is fixed.

Secondly, we weren't exporting the static `TYPE` values for the events, so external developers wouldn't be able to use the `type` properties for event filtering. Now they can :D

### Additional information

Dependent on #1551 because I didn't want to run into merge conflicts.

### DIFF
--- a/plugins/block-shareable-procedures/src/events_procedure_base.ts
+++ b/plugins/block-shareable-procedures/src/events_procedure_base.ts
@@ -10,10 +10,12 @@ import * as Blockly from 'blockly/core';
  * The base event for an event associated with a procedure.
  */
 export abstract class ProcedureBase extends Blockly.Events.Abstract {
-  isBlank = false;
-
   static readonly TYPE: string = 'procedure_base';
+
+  /** A string used to check the type of the event. */
   type = ProcedureBase.TYPE;
+
+  isBlank = false;
 
   /**
    * Constructs the base procedure event.
--- a/plugins/block-shareable-procedures/src/events_procedure_base.ts
+++ b/plugins/block-shareable-procedures/src/events_procedure_base.ts
@@ -12,6 +12,9 @@ import * as Blockly from 'blockly/core';
 export abstract class ProcedureBase extends Blockly.Events.Abstract {
   isBlank = false;
 
+  static readonly TYPE: string = 'procedure_base';
+  type = ProcedureBase.TYPE;
+
   /**
    * Constructs the base procedure event.
    * @param workspace The workspace the procedure model exists in.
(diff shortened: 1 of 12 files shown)


plugins/block-sharable-procedures/src/index.ts
```diff
@@ -9,6 +9,15 @@ import {blocks} from './blocks';
...
 import {ObservableProcedureModel} from './observable_procedure_model';
+import {ProcedureBase, ProcedureBaseJson} from './events_procedure_base';
+import {ProcedureChangeReturn, ProcedureChangeReturnJson} from './events_procedure_change_return';
+import {ProcedureCreate, ProcedureCreateJson} from './events_procedure_create';
+import {ProcedureDelete, ProcedureDeleteJson} from './events_procedure_delete';
+import {ProcedureParameterBase, ProcedureParameterBaseJson} from './events_procedure_parameter_base';
+import {ProcedureParameterCreate, ProcedureParameterCreateJson} from './events_procedure_parameter_create';
+import {ProcedureParameterDelete, ProcedureParameterDeleteJson} from './events_procedure_parameter_delete';
+import {ProcedureParameterRename, ProcedureParameterRenameJson} from './events_procedure_parameter_rename';
+import {ProcedureRename, ProcedureRenameJson} from './events_procedure_rename';
```
cpcallen: Any particular reason you have opted to do the reexports this way, with a bunch of `import`s followed by a single `export`?  I'm aware that we have some usage of this style in core, but I think it is mostly a hold-over from when those files were closure modules.  The styleguide doesn't seem give any definitive rules AFACT but I note that [it has one example of `export {…} from '…'`](https://google.github.io/styleguide/tsguide.html#exports) and none of `export {…}`.

```suggestion
export {ProcedureBase, ProcedureBaseJson} from './events_procedure_base';
export {ProcedureChangeReturn, ProcedureChangeReturnJson} from './events_procedure_change_return';
export {ProcedureCreate, ProcedureCreateJson} from './events_procedure_create';
export {ProcedureDelete, ProcedureDeleteJson} from './events_procedure_delete';
export {ProcedureParameterBase, ProcedureParameterBaseJson} from './events_procedure_parameter_base';
export {ProcedureParameterCreate, ProcedureParameterCreateJson} from './events_procedure_parameter_create';
export {ProcedureParameterDelete, ProcedureParameterDeleteJson} from './events_procedure_parameter_delete';
export {ProcedureParameterRename, ProcedureParameterRenameJson} from './events_procedure_parameter_rename';
export {ProcedureRename, ProcedureRenameJson} from './events_procedure_rename';
```

cpcallen (changes requested):
LGTM overall, but please make the `events_*.ts` files consistent about line breaks, presence or absence of JSDoc on the `type` instance property, etc.

plugins/block-sharable-procedures/src/index.ts
BeksOmega: Fixed!

cpcallen (approved):

PR #1562, opened 2023-02-14
chore: update readme to point to dev docs

BeksOmega:
### Description

Updated the readme to point to the upcoming dev docs related to procedures. That way we only have to maintain information in one place (the prettier place!).

I kept everything that should be static (i.e. the project description and how to install).

### DIFF
--- a/plugins/block-sharable-procedures/README.md
+++ b/plugins/block-sharable-procedures/README.md
@@ -17,24 +17,12 @@ yarn add @blockly/block-sharable-procedures
 
 ### npm
 ```
-npm install @blockly/block-sharable-procedures --save
+npm install @blockly/block-sharable-procedures
 ```
 
 ## Usage
 
-### Import
-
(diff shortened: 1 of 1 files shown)


plugins/block-sharable-procedures/README.md
```diff
@@ -17,24 +17,12 @@ yarn add @blockly/block-sharable-procedures
...
-the blocks itself) which might be unexpected.
-
-```js
-import * as Blockly from 'blockly';
-import {blocks} from '@blockly/block-sharable-procedures';
-
-// Blocks must be explicitly defined after importing.
-Blockly.common.defineBlocks(blocks);
-```
+See developers.google.com/blockly/guides/create-custom-blocks/procedures/using-procedures for more information about using this plugin.
```
rachel-fenichel: Make this an actual link instead of plaintext url.

rachel-fenichel (approved):

PR #1564, opened 2023-02-14
fix!: remove serialization from disabled interaction serializer

BeksOmega:
### Description

Removes serialization of disabled properties from the disabled interactions serializer as per https://github.com/google/blockly/issues/6807#issuecomment-1430252324

Work on https://github.com/google/blockly/issues/6807

### Additional information

This should not be merged until after the next release of core which adds serialization of the disabled properties to core.


cpcallen (approved):

cpcallen:
@BeksOmega: I presume you're not planning to merge this before the next Blockly release?  Might be worth converting it back to draft as a precaution in the mean time.

maribethb (commented):
Can you add some instructions to the README so that I can link to that from the release notes for core, and so people know what to do with this plugin going forward?

maribethb (commented):
And also make sure the package.json is updated with the new minimum peer and dev dependency of 9.3.0. Thanks!

BeksOmega:
> And also make sure the package.json is updated with the new minimum peer and dev dependency of 9.3.0. Thanks!

Yesh! In the release doc to do before merging :D

plugins/serialize-disabled-interactions/README.md
```diff
@@ -4,6 +4,36 @@ A [Blockly](https://www.npmjs.com/package/blockly) plugin for the JSON
...
+### For Blockly versions < v9.3.0
+
+This plugin was created because the old XML serialization system supported
+serializing these disabled-interaction attributes, but the new JSON system did
+not (for versions of Blockly < v9.3.0). For more information about why this
+was, see [this forum post](https://groups.google.com/g/blockly/c/eP9PXVfCaHs/m/S4rPmwnTAQAJ).
+
+In v9.3.0 this was changed to fix https://github.com/google/blockly/issues/6793.
+
+It is **strongly** recommended that instead of using this plugin, you upgrade
```
maribethb, line 29: Add a note that they must use v2 of this plugin, as v3 does not actually serialize the attributes. This is implied by the peer dependency version, but better to be explicit.

Also maybe add a note that this plugin will be removed in the future, and by newly installing it now, they will be introducing a dependency they will be forced to remove in the future if they wish to remain up-to-date with new features. Or some other nicely worded reason for why it is we strongly recommend against this.

plugins/serialize-disabled-interactions/README.md
```diff
@@ -43,39 +73,19 @@ All of the keys are optional. For example, if all of the blocks on the workspace
...
-Blockly.Blocks['my_start_block'] = {
-  init: function() {
-    // Your normal init code...
-    this.setDeletable(false);
-    this.setMovable(false);
-    this.setEditable(false);
-  }
-}
-```
+## Usage for Blockly < v9.3.0
```
maribethb: I think you should just delete this section (and the installation section) for two reasons:
- For Blockly < 9.3 they'll need to use an old version of this plugin. If they insist on doing that, they can look at the old version of the code for instructions too.
- I want to strongly discourage people from using it, because it is being deprecated. I think giving people easy instructions on how to install and use a deprecated thing sends mixed messages.

maribethb (approved):

PR #1568, opened 2023-02-16
feat: Add third-party plugins section to site

maribethb:
Add a third-party plugins section to the samples site.

Didn't add any fancy css. If we are ripping out Jekyll soon then there is not a good way to add css classes using showdown, other than using straight html inside the markdown. Open to follow up PRs enhancing the style.

@mark-friedman are you ok with this? Using your plugin as the first example. We'd then accept PRs from folks wanting to add their own plugins.

Preview: https://maribethb.github.io/blockly-samples/ (note this is currently updating, if you view this within 10 min of me opening this PR)

### DIFF
--- a/gh-pages/index.md
+++ b/gh-pages/index.md
@@ -42,7 +42,7 @@ layout: index
 * [Theme Tritanopia](plugins/theme-tritanopia/test/index.html)
 * [Theme Deuteranopia](plugins/theme-deuteranopia/test/index.html)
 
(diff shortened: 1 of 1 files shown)


mark-friedman:
> @mark-friedman are you ok with this? Using your plugin as the first example.

Yes, I am honored to be chosen!  Thx.

rachel-fenichel (approved):

PR #1583, opened 2023-02-24
chore: add a test/demo playground for shareable procedures

BeksOmega:
### Description

Adds a test/demo page for shareable procedures. Procedure and variable events are shared between two workspaces to allow you to instantiate procedure definition blocks in one workspaces, and their caller blocks in another.

The JSON serialized state of both worksapces is made visible to the user, and a reload button is provided to showcase serialization working.

image

### Additional Info

Dependent on #1552

### DIFF
--- a/plugins/block-shareable-procedures/test/index.js
+++ b/plugins/block-shareable-procedures/test/index.js
@@ -3,34 +3,105 @@
  * Copyright 2022 Google LLC
  * SPDX-License-Identifier: Apache-2.0
  */
-
 /**
  * @fileoverview Block test.
  */
 
 import * as Blockly from 'blockly';
-import {createPlayground, toolboxCategories} from '@blockly/dev-tools';
-import {blocks, unregisterProcedureBlocks} from '../src/index';
+import {toolboxCategories} from '@blockly/dev-tools';
+import {blocks, unregisterProcedureBlocks, ObservableProcedureModel, ObservableParameterModel} from '../src/index';
+import {ProcedureBase} from '../src/events_procedure_base';
 
 
 unregisterProcedureBlocks();
 Blockly.common.defineBlocks(blocks);
 
-/**
- * Create a workspace.
- * @param {HTMLElement} blocklyDiv The blockly container div.
- * @param {!Blockly.BlocklyOptions} options The Blockly options.
- * @returns {!Blockly.WorkspaceSvg} The created workspace.
- */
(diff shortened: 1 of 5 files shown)


plugins/block-shareable-procedures/test/index.js
```diff
@@ -3,34 +3,105 @@
...
- * Create a workspace.
- * @param {HTMLElement} blocklyDiv The blockly container div.
- * @param {!Blockly.BlocklyOptions} options The Blockly options.
- * @returns {!Blockly.WorkspaceSvg} The created workspace.
- */
-function createWorkspace(blocklyDiv, options) {
-  const workspace = Blockly.inject(blocklyDiv, options);
-  return workspace;
-}
+Blockly.serialization.registry.unregister('procedures');
```
maribethb: This might be helpful as a provided helper function, similar to `unregisterProcedureBlocks`? 

unregistering and registering the new one together in one function, I mean

plugins/block-shareable-procedures/test/index.js
```diff
@@ -3,34 +3,105 @@
...
+Blockly.serialization.registry.register(
+    'procedures',
+    new Blockly.serialization.procedures.ProcedureSerializer(
+        ObservableProcedureModel, ObservableParameterModel));
+
+export let workspace1;
```
maribethb: Do these need to be exported?

plugins/block-shareable-procedures/test/index.js
```diff
@@ -3,34 +3,105 @@
...
+        !(e instanceof Blockly.Events.VarBase)) {
+      return;
+    }
+    let event;
+    try {
+      event = Blockly.Events.fromJson(e.toJson(), otherWorkspace);
+    } catch (e) {
+      console.log(
+          'Could not deserialize event. This is expected to happen. E.g. ' +
+          'when round-tripping parameter deletes, the delete in the ' +
```
maribethb: If this is expected to happen, we probably shouldn't log it in the console. Just a comment here would be better, I think.

plugins/block-shareable-procedures/test/index.js
```diff
@@ -3,34 +3,105 @@
...
+          'secondary workspace cannot be deserialized into the original ' +
+          'workspace.');
+      return;
+    }
+    console.log('running', event);
```
maribethb: Remove extra logging?

maribethb (approved):

PR #1586, opened 2023-02-25
fix: renaming params in other workspaces creating different vars

BeksOmega:
### Description

When we would rename parameters via the mutator in one workspace, the rename param even would create a new variable with the same name, but a different ID in the other workspace. This means that if the variable was then renamed in one worspace, the procedure blocks in the other workspace would not update to reflect the new variable name. This fixes that by round-tripping the variable ID through serialization as well.

### DIFF
--- a/plugins/block-shareable-procedures/src/observable_parameter_model.ts
+++ b/plugins/block-shareable-procedures/src/observable_parameter_model.ts
@@ -35,13 +35,15 @@ export class ObservableParameterModel implements
   /**
    * Sets the name of this parameter to the given name.
    * @param name The string to set the name to.
+   * @param id The optional ID the backing variable should have.
    * @returns This parameter model.
    */
-  setName(name: string): this {
+  setName(name: string, id?: string): this {
     if (name === this.variable.name) return this;
     const oldName = this.variable.name;
     this.variable =
-        this.workspace.getVariable(name) ?? this.workspace.createVariable(name);
+        this.workspace.getVariable(name) ??
+        this.workspace.createVariable(name, '', id);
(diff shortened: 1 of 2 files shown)


plugins/block-shareable-procedures/src/observable_parameter_model.ts
```diff
@@ -35,13 +35,15 @@ export class ObservableParameterModel implements
   /**
    * Sets the name of this parameter to the given name.
    * @param name The string to set the name to.
+   * @param id The optional ID the backing variable should have.
    * @returns This parameter model.
    */
-  setName(name: string): this {
+  setName(name: string, id?: string): this {
```
maribethb, line 41: Is this problem specific to the `ObservableParameterModel`, or put another way why not add this to the interface? Is it because the interface has already been released in core? If it is, I'm willing to use the "new APIs are beta" rule because there is 0 chance anyone has implemented these interfaces.

maribethb (commented):
nit: isn't this a "fix" not a chore?

plugins/block-shareable-procedures/src/observable_parameter_model.ts
BeksOmega: This is specific to the `ObservableParameterModel`. `ObservableParameterModel`s are backed by global variables, while `IParameterModel`s do not assume this (because using global variables to implement parameters is kiiinda sketchy). Adding an optional ID parameter to `setName` on the `IParameterModel` in core would add an assumption of using backing variable models to the interface in core.

maribethb (approved):

PR #1591, opened 2023-02-27
chore: Misc minor refactorings, doc updates, etc.

NeilFraser:
Incidental fixes while pursuing other work:

* Use `Array.prototype.includes` instead of `.indexOf` when checking for list membership.
* Fix formatting of some `README.md` files.
* Don't regenerate Blockly when in the middle of a drag in `devsite-demo`.
* Use `.key` instead of `.keyCode` in `workspace-search`.

### DIFF
--- a/examples/devsite-demo/script.js
+++ b/examples/devsite-demo/script.js
@@ -129,6 +129,9 @@ function languageChange() {
  * Called when the blocks change, or when the target language changes.
  */
 function regenerate(_e) {
+  if (Blockly.getMainWorkspace().isDragging()) {
+    return;  // Don't update code mid-drag.
+  }
   const generateLang = document.getElementById('generateDropdown').value;
   const generator = Blockly[generateLang];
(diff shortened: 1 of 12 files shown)


examples/devsite-demo/script.js
```diff
@@ -129,6 +129,9 @@ function languageChange() {
  * Called when the blocks change, or when the target language changes.
  */
 function regenerate(_e) {
+  if (Blockly.getMainWorkspace().isDragging()) {
+    return;  // Don't update code mid-drag.
+  }
```
cpcallen, line 134: Will this event handler get called again once the dragging is finished?

If not, does this change not create a bug?

cpcallen (approved):

examples/devsite-demo/script.js
NeilFraser: There should be a 'move' event which is fired after the 'drag' has completed.  Tested on playground, and the devsite demo.

Issue #1549, opened 2023-02-07
Stop using Jekyll


rachel-fenichel:
**Category**
gh-pages

**Proposal**

Stop using Jekyll for layout and templating for google.github.io/blockly-samples. Update the current predeploy step to create the entire web page, rather than creating the inner section of the web page and front matter.

Continue to deploy to gh-pages, just with a single build step (predeploy) rather than two (predeploy and build).

**Reasons to make this change:**
- Jekyll relies on Ruby and thus pulls in a chain of dependencies that is separate from the ones pulled in through npm. Removing it removes dependencies.
- Google removed the version of Ruby that we needed, and may make future breaking changes automatically.
- We’re primarily using Jekyll to add headers and footers around web pages that we already create and run when testing samples locally. That is, we’re doing something fairly straightforward with it. (However, note that the pages we have for running samples locally are actually different from and simpler than the versions on gh-pages.)
- We already have a pre-build step (predeploy) that prepends stuff onto the existing test pages.
- The scope of the project is small enough for Rachel to do it between meetings.

**Reasons not to make this change:**
Requires engineering time.
We currently have a build/deploy setup that works through GitHub Actions.

Maribeth’s primary goal: keep automation working for publishing.

- [x] Generate plugin pages
- [x]  Generate examples pages, which includes respecting the list of files that needs to be copied over (defined in package.json).
- [x]  Generate the index page, because its contents are currently generated from a markdown file that lists plugins/examples and links.
- [x]  Convert from sass to css.
- [x]  Fix remaining CSS issues. (Fixed in #1683)
- [x]  Remove the rest of the jekyll config. (Fixed in #1683)
- [x]  Update action for deploying to gh-pages. (Fixed in #1683)
- [x]  Remove jekyll from package.json. (Unnecessary--it wasn't there)
- [x]  Update documentation.  (Fixed in #1683)

PR #1595, opened 2023-02-28
chore: remove front matter for plugin pages on gh-pages

rachel-fenichel:
Part of https://github.com/google/blockly-samples/issues/1549.

Note that this is a PR against the `remove_jekyll` branch, not against `master`.

This change:
- Created a new gulpfile, `scripts/gh-predeploy.js`, with scripts for preparing examples and plugins for deployment to gh-pages.
- Moved the existing scripts for plugins and examples to the new gulpfile.
- Removed the Jekyll config for html pages in subdirectories of `plugins`.
- Changed the code for generating `index.html` pages for plugins:
  - New code is injected into the a copy of the plugin's `test/index.html` page, rather than wrapping it. This means we no longer have a full additional HTML document inside the body of the page, which was (I think) illegal anyway.
  - Stopped adding Jekyll front matter to the page, so Jekyll just treats it as a pass-through rather than processing it.
  - Example page: https://rachel-fenichel.github.io/blockly-samples/plugins/block-plus-minus/test/index
- Changed the code for generating `README.html` pages for plugins:
  - Used showdown to convert from markdown to html, and generated the rest of the page using the helpers I wrote for `index.html`.
  - Stopped adding Jekyll front matter, so Jekyll just treats it as a pass-through rather than processing it.
  - Example page: https://rachel-fenichel.github.io/blockly-samples/plugins/block-plus-minus/README
  - Note that the CSS is still not quite right for README pages.
- Added showdown as a dependency.

Remaining work:

- [ ] Make the same/similar changes to examples, which includes respecting the list of files that needs to be copied over (defined in `package.json`).
- [ ] Make an equivalent change to the index page, because its contents are currently generated from a markdown file that lists plugins/examples and links.
- [ ] Convert from sass to css.
- [ ] Fix remaining CSS issues.
- [ ] Remove the rest of the jekyll config.
- [ ] Update action for deploying to gh-pages.
- [ ] Remove jekyll from package.json.
- [ ] Update documentation.


gh-pages/_config.yml
```diff
@@ -8,12 +8,7 @@ defaults:
       type: "pages"
     values:
       article: true
-  -
-    scope:
```
BeksOmega, line 12: I can't find docs that describe what this is doing :/ I was looking at https://jekyllrb.com/docs/configuration/options/ but I don't think it has the information I need. Could you send me the relevant docs so I can make sur eI understand what this is doing before approving?

scripts/gh-predeploy.js
```diff
@@ -0,0 +1,355 @@
...
+    return fs.statSync(path.join(dir, file)).isDirectory() &&
+      fs.existsSync(path.join(dir, file, 'package.json')) &&
+      // Only prepare plugins with test pages.
+      fs.existsSync(path.join(dir, file, '/test/index.html'));
+  });
+  return gulp.parallel(folders.map(function (folder) {
+    return function preDeployPlugin() {
+      return preparePlugin(folder);
+    };
+  }))(done);
```
BeksOmega, line 269: Can you use an annonymous function expression here? I think that would make the fact that you're mapping each folder to a function that prepares the folder clearer.

```suggestion
  return gulp.parallel(
      folders.map((folder) => (() => return preparePlugin(folder)))
  )(done);
```

Similarly for `prepareToDeployExamples`

scripts/gh-predeploy.js
```diff
@@ -0,0 +1,355 @@
...
+    return function preDeployExample(done) {
+      return prepareExample(dir, folder, done);
+    };
+  }))(done);
+}
+
+module.exports = {
+  predeployPlugins: prepareToDeployPlugins,
+  predeployExamples: prepareToDeployExamples,
+};
```
BeksOmega: Newline at EOF

BeksOmega (commented):
Generally LGTM! Just want to make sure I understand the config changes before approving =)

gh-pages/_config.yml
rachel-fenichel: https://jekyllrb.com/docs/configuration/front-matter-defaults/

They're called "front matter defaults" and let you configure specific values as part of the front matter for all pages that match the glob. In this case, they were setting the value of the layout to default, for all pages in plugins that end in html.

More info on front matter here: https://jekyllrb.com/docs/front-matter/

BeksOmega: Sweet! Thank you for the link =) That makes sense to me!

scripts/gh-predeploy.js
rachel-fenichel: I tried this and it doesn't signal completion correctly. I think that means there's an issue with the layers of commands here. Rather than sort it out now, I'm going to look at it after I've settled the details of the rest of the prepare scripts. Ideally I can simplify everything.

maribethb (approved):

PR #1606, opened 2023-03-10
feat: remove front matter for example pages on gh-pages

rachel-fenichel:
Part of https://github.com/google/blockly-samples/issues/1549.

Note that this is a PR against the remove_jekyll branch, not against master.

This change continues the work I started in #1595.
- Removes Jekyll config for html pages in subdirectories of `examples`
- Changes the code for generating all html pages for examples:
  - New code is injected into a copy of the page, rather than wrapping it.
  - Stopped adding Jekyll front matter to the page, so Jekyll treats it as a pass-through rather than wrapping it.
- Fixed CSS issues by removing the devsite CSS--that was only supposed to be in the root index page, and I accidentally put it everywhere.

Edit: I tagged on a few more changes:
- Remove some parts of the Jekyll config in `_config.yml` that no longer apply to anything.
- Remove front matter tags from `custom.scss` and rename to `custom.css`. Jekyll now copies it over directly.

I combined tasks where I could, but in a lot of cases I have one function for injecting HTML into a plugin page and a different one for the example pages. I think I can eventually combine some parts, but want to wait until I've finished making all of the tweaks to parts that are different between the two.


scripts/gh-predeploy.js
```diff
@@ -128,13 +127,16 @@ function injectNavBar(inputString, packageJson, pluginDir) {
...
   <!-- END NAV BAR -->
-  ${createPageTabs(pluginDir)}`
+  ${createPluginTabs(pluginDir)}`
 
   // Find the start of the body and inject the nav bar just after the opening <body> tag,
   // preserving anything else in the tag (such as onload).
   // Also wrap all page content in a <main></main> tag.
   let modifiedContent = inputString.replace(
-      /(<body.*>)/, `<main id="main" class="has-tabs">$1${navBar}`);
+    /<body(.*)>/,
```
NeilFraser: Can you change this regexp to this: `/<body([^>]*)>/`
The existing one is greedy and will perform poorly if there's a second tag on the same line, e.g. `'<body class="foo"><div>' -> ' class="foo"><div'`

scripts/gh-predeploy.js
```diff
@@ -144,21 +146,57 @@ function injectNavBar(inputString, packageJson, pluginDir) {
...
+        <a href="${baseurl}/${pageRoot}/${page.link}">
+          ${page.label}
+        </a>
+      </li>
+      `;
+  }
+
+  let tabsString = ``;
+
+  for (let i = 0; i < pages.length; i++) {
```
NeilFraser: This isn't a high-performance loop.  It's ok to use:

```
for (const page of pages) {
  tabString += creatTab(page);
}
```

scripts/gh-predeploy.js
```diff
@@ -271,15 +309,79 @@ function prepareToDeployPlugins(done) {
...
+  let baseurl = '/blockly-samples';
+  // Assemble that information into a nav bar and tabs for getting to linked
+  // example pages.
+  let navBar = `
+  <!-- NAV BAR -->
+  <nav id="toolbar">
+    <a href="${baseurl}" id="arrow-back">
+      <i class="material-icons">close</i>
+      <img src="https://blocklycodelabs.dev/images/logo_knockout.png" class="logo-devs"
+        alt="Blockly logo" />
```
NeilFraser: The primary purpose of the ALT attribute is to provide text content for screen readers.  At the moment that 'a' link says "close Blockly logo", which isn't very clear as to what it does.  We can't change tho word "close" since it's a font.  But we can change the following ALT text.  Maybe change to "Blockly sample"?

scripts/gh-predeploy.js
```diff
@@ -271,15 +309,79 @@ function prepareToDeployPlugins(done) {
...
+    <a href="${codeLink}" class="button" target="_blank">View code</a>
+  </nav>
+  <!-- END NAV BAR -->
+  ${tabString}`
+
+  // Find the start of the body and inject the nav bar just after the opening <body> tag,
+  // preserving anything else in the tag (such as onload).
+  // Also wrap all page content in a <main></main> tag.
+  let modifiedContent = inputString.replace(
+      /<body(.*)>/,
```
NeilFraser: Likewise, `/<body([^>]*)>/`

scripts/gh-predeploy.js
```diff
@@ -350,7 +459,93 @@ function prepareToDeployExamples(done) {
...
+  <meta name="viewport" content="width=device-width,maximum-scale=2">
+  <link rel="stylesheet" href="https://blocklycodelabs.dev/styles/main.css" />
+</head>
+
+<body class="root">
+  <!-- HEADER -->
+  <nav id="toolbar">
+    <div class="site-width layout horizontal">
+      <a href="https://google.github.io/blockly-samples/"><img src="https://blocklycodelabs.dev/images/logo_knockout.png" class="logo-devs"
+          alt="Blockly logo" /></a>
```
NeilFraser: Suggest just `alt="Blockly"`, since that's literally what the logo image says.

scripts/gh-predeploy.js
```diff
@@ -350,7 +459,93 @@ function prepareToDeployExamples(done) {
...
+function predeployForGitHub(done) {
+  createIndexPage();
+  return gulp.parallel(prepareToDeployPlugins, prepareToDeployExamples)(done);
+}
+
 module.exports = {
   predeployPlugins: prepareToDeployPlugins,
   predeployExamples: prepareToDeployExamples,
+  prepareIndex: createIndexPage,
+  predeployAll: predeployForGitHub
```
NeilFraser: I think we like to use trailing commas.

scripts/gh-predeploy.js
```diff
@@ -350,7 +459,93 @@ function prepareToDeployExamples(done) {
...
+  </main>
+
+  <script src="https://cdn.jsdelivr.net/npm/fuse.js/dist/fuse.js"></script>
+  <script src="js/index.js"></script>
+</body>
+
+</html>
+`;
+
+  let contents = injectHeader(indexBase, "Plugins | blockly-samples");
```
NeilFraser: Double-quoted string?

scripts/gh-predeploy.js
```diff
@@ -350,7 +459,93 @@ function prepareToDeployExamples(done) {
...
+ * specified in in `index.md`.
+ */
+function createIndexPage() {
+  const initialContents = fs.readFileSync(`./gh-pages/index.md`).toString();
+
+  const converter = new showdown.Converter();
+  converter.setFlavor('github');
+  const text = initialContents;
+  const htmlContents = converter.makeHtml(text);
+  console.log(htmlContents);
```
NeilFraser: Just checking if this console.log is intentional.  It's developer-side, so no problem, but want to verify it wasn't some forgotten debugging log.

scripts/gh-predeploy.js
```diff
@@ -350,7 +459,93 @@ function prepareToDeployExamples(done) {
...
+ * Create the index page for the blockly-samples GitHub Pages site.
+ * This page has some nice wrappers, a search bar, and a link to every plugin or example
+ * specified in in `index.md`.
+ */
+function createIndexPage() {
+  const initialContents = fs.readFileSync(`./gh-pages/index.md`).toString();
+
+  const converter = new showdown.Converter();
+  converter.setFlavor('github');
+  const text = initialContents;
```
NeilFraser: What's this assignment for?

NeilFraser (approved):

PR #1626, opened 2023-03-15
Update README.md

### LINKED ISSUE #1625: Provide better documentation for suggested-blocks
🚧⚠️🚧 This issue is being saved for **Grace Hopper Open Source Day**. 🚧⚠️🚧 An event we're running to help girls and non-binary technologists make their first contributions to open source! Please see our list of [help wanted issues](https://github.com/google/blockly-samples/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22+-label%3A%22ghc-osd%22) if you'd like to contribute! 

---

**Category**

 - Plugins

**Component**

suggested-blocks

The README example is slicing into a toolbox XML string. That is not a good idea. There's a comment saying it's not a good idea, but still, we shouldn't model this for people. Also, we should use JSON.

Replace the existing [usage documentation](https://github.com/google/blockly-samples/tree/master/plugins/suggested-blocks#usage) with the following:
```js
import * as SuggestedBlocks from '@blockly/suggested-blocks';

const toolbox = {
  "kind": "categoryToolbox",
  "contents": [
    {
      "kind": "category",
      "name": "My Category",
      "contents": [ /* your category contents */ ]
    },
    {
      'kind': 'category',
      'name': 'Frequently Used',
      'custom': 'MOST_USED',
      'categorystyle': 'frequently_used_category',
    },
    {
      'kind': 'category',
      'name': 'Recently Used',
      'custom': 'RECENTLY_USED',
      'categorystyle': 'recently_used_category',
    },
  ]
};

// Inject Blockly.
const workspace = Blockly.inject('blocklyDiv', {
  toolbox: toolbox,
});

// Initialize the plugin
SuggestedBlocks.init(workspace);
```

N/A

**Additional context**

https://google.github.io/blockly-samples/plugins/suggested-blocks/README

martinyis:
Description
This pull request updates the example in the README for the @blockly/suggested-blocks package to use JSON instead of manipulating an XML string to define the new categories. The updated example defines the new categories as JSON objects and then generates the XML elements for each category and block using a loop. This approach is more flexible and maintainable and should make it easier to add or modify categories in the future.

Changes
Update the README example to use JSON instead of manipulating an XML string to define the new categories.
Define the new categories as JSON objects and generate the XML elements for each category and block using a loop.
Checklist
 I have tested the updated example and it works as expected.
 I have updated the pull request title to be descriptive of the changes.
 I have updated the pull request description to explain the changes and why they were made.

fixes #1625

### DIFF
--- a/plugins/suggested-blocks/README.md
+++ b/plugins/suggested-blocks/README.md
@@ -20,25 +20,53 @@ import * as Blockly from 'blockly';
 import {toolboxCategories} from '@blockly/dev-tools';
 import * as SuggestedBlocks from '@blockly/suggested-blocks';
 
-function getNewCategories() {
-  const mostUsedCategory = '<category name="Frequently Used" custom="MOST_USED"></category>';
-  const recentlyUsedCategory = '<category name="Recently Used" custom="RECENTLY_USED"></category>';
+// Define the new categories in JSON format
+const mostUsedCategory = {
+  custom: 'MOST_USED',
+  name: 'Frequently Used',
+  blocks: []
+};
 
-  // This type of insertion works in a pinch, but you would more so want to add
-  // the new categories wherever the rest of the categories are defined in the
-  // project (e.g. HTML, or a dedicated file for all the categories)
-  const indexToInsert = toolboxCategories.indexOf('</xml>');
-  // Insert the new categories into the existing toolbox XML string
-  return toolboxCategories.slice(0, indexToInsert) + mostUsedCategory + recentlyUsedCategory + toolboxCategories.slice(indexToInsert);
-}
+const recentlyUsedCategory = {
+  custom: 'RECENTLY_USED',
+  name: 'Recently Used',
+  blocks: []
+};
+
+// Add the new categories to the existing categories list
+const categories = [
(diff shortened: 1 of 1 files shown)


plugins/suggested-blocks/README.md
```diff
@@ -20,25 +20,53 @@ import * as Blockly from 'blockly';
...
+};
+
+// Add the new categories to the existing categories list
+const categories = [
+  ...toolboxCategories,
+  mostUsedCategory,
+  recentlyUsedCategory
+];
+
+// Convert the categories to XML
```
BeksOmega, line 43: Hello! Thanks for the pull request :D This looks great! But helpfully Blockly actually accepts toolboxes defined in JSON, so there's no need to convert =)

You can see the docs about this here: https://developers.google.com/blockly/guides/configure/web/toolbox#formats

If you could change this to just use a JSON definition, that would be awesome. Make sure the `kind` and `contents` properties!

I'll give the rest of a more in-depth look once this is fixed up! But generally LGTM =)

BeksOmega:
Closing in favor of #1632

PR #1635, opened 2023-03-17
chore: Migrate custom-generator-codelab to using `npx @blockly/create-package app`

jubishop:
- Update the initial instructions in `custom_generator.md` to use `npx @blockly/create-package app` instead of simply cloning the repo and copying the directory `examples/sample-app`.
- In the created code, the line `Object.assign(javascriptGenerator, generator);` no longer exists directly below the imports, so change the instructions to make clear that line for deletion is further down in the file.
- Add instructions to delete the line about the `outputDiv`.  This seemed like a good idea to me.  Since it's otherwise not used, VSCode starts calling it out as a problem as an unused variable, and as the reader of the codelab I start to wonder if I should still be using it somewhere and I missed it.
- Add the `README.md` file simply because `npx @blockly/create-package app` will create it, so the example matches our new directory more precisely.
- `package.json` and `webpack.config.js` now look a little different.  I don't know exactly why, but this is just what they look like now after running `npx @blockly/create-package app`.


examples/custom-generator-codelab/src/toolbox.js
maribethb, line 1: I don't believe this file should be changed. This file represent the end state of the toolbox after doing the codelab. In the codelab, they replace the default toolbox with one that uses the custom blocks and only the built-in blocks needed for this codelab.

maribethb (commented):
Looks good, other than the change to toolbox.js. Thanks!

examples/custom-generator-codelab/src/toolbox.js
jubishop: oh of course!  my bad.  the dang "large diff" thing meant i overlooked reviewing this change.  thanks.

jubishop:
hmm wait, i don't think this was ever merged.

codelabs/custom_generator/custom_generator.md
```diff
@@ -23,8 +24,9 @@ You will build a JSON generator that implements the [JSON language spec](https:/
 
 ### What you'll need
 
+- Comfort with the Blockly playground.
```
maribethb: I don't think we should add this line - this codelab no longer uses the playground. The sample app is distinct from the playground.

examples/custom-renderer-codelab/src/index.js
```diff
@@ -10,7 +10,7 @@ import {generator} from './generators/javascript';
 import {javascriptGenerator} from 'blockly/javascript';
 import {save, load} from './serialization';
 import {toolbox} from './toolbox';
-import './renderers/javascript';
+import './renderers/custom';
```
maribethb, line 13: Is this line supposed to be in this PR? Was the file already renamed in another PR?

jubishop: yeah this was a bug fix i slipped in.  i can break it out into it's own PR if you want?

codelabs/custom_generator/custom_generator.md
jubishop: gotcha.  the term playground is used a couple places in this codelab AND the renderer one.  i was thinking the main interface generated by the sample app is still the "playground".   i'll remove all references to playground from both.

examples/custom-renderer-codelab/src/index.js
maribethb: No that's fine, just double checking.

codelabs/custom_generator/custom_generator.md
maribethb: OK sounds good, I think we should remove them.

jubishop: Done

maribethb (approved):

PR #1640, opened 2023-03-24
chore: small changes for consistency between renderer/generator codelabs

jubishop:
Just breaking these out while working on the keyboard-navigation codelab.

- Add a space after the ### What you'll learn section so that it's consistent that all `###` entries have a newline spacer.
- Add the `Comfort with Blockly playground` need to both.
- Change `comfort with` to `basic understanding of`.  Comfort is really what you should have AFTER the codelab, but not necessary beforehand, IMO.  
- Also fixed a syntax error that crept in when i changed the imported file from javascript.js to custom.js but didn't change it in the example code.


jubishop:
do i need to do something about this check? seems pointless?

jubishop:
what'd i do?! 😂

PR #1643, opened 2023-03-29
chore: shareable release prep

BeksOmega:
Preps the `block-shareable-procedures` plugin for being released.


BeksOmega:
@maribethb Looks like the new calls to `requestAnimationFrame` in core break head-ful tests in node. I'm not sure why though when it should be polyfilled by JSDOM.

I've fixed the tests for this plugin by shimming it myself - so I think this is good to go out (please confirm). But I'm not sure what the priority should be for fixing this in general (i.e. is it a today thing or a tomorrow thing).

maribethb (approved):
I think let's publish this now since core is already released, but spend time after / during bug bash week figuring out why jsdom isn't polyfilling it

PR #1651, opened 2023-03-31
chore: updated `plugins/field-slider` to use Blockly 9.3.1's imports

btw17:
# Notes

- `FieldSlider` now imports type information from `blockly/core` instead of their definition files, where they previously hadn't been exported from prior to `9.3.1`
- This is a follow-up to https://github.com/google/blockly-samples/pull/1650 and should wait for it to be submitted.
- I cleared the `package-lock.json` in `@blockly/dev-tools` and re-ran `npm run boot` to pull the latest `blockly` changes
  - The current dev-scripts are still valid with `blockly@9.0.0`, though there was a Type update in `9.2.1` that results in `dev-scripts` and `field-slider` needing the same Blockly version.
    - This is a problem that is unique to the `blockly-samples` codebase. Installing both plugins would normally lead to installing the same Blockly version, avoiding the type problem. In `blockly-samples`, installing via `lerna` allows each to have their own modules that can deviate from each other.
    - The specific type problem between `9.0.0` and `9.2.1` has to do with the `Function` type not being compatible with the updated function type information. An actual function fulfilling this type information works across both, though `Function` is as a defined type is not compatible with `FieldSliderValidator`, for example.

### DIFF
--- a/plugins/field-slider/package.json
+++ b/plugins/field-slider/package.json
@@ -41,12 +41,12 @@
   "devDependencies": {
     "@blockly/dev-scripts": "^1.2.32",
     "@blockly/dev-tools": "^5.2.4",
-    "blockly": "^9.2.0",
+    "blockly": "^9.3.1",
     "chai": "^4.2.0",
     "sinon": "^9.0.1"
   },
   "peerDependencies": {
-    "blockly": "^9.2.0"
+    "blockly": "^9.3.1"
   },
   "publishConfig": {
     "access": "public",
--- a/plugins/field-slider/tsconfig.json
+++ b/plugins/field-slider/tsconfig.json
@@ -11,5 +11,5 @@
     },
     // NOTE: `test/**/*` is automatically included in `blockly-scripts start`.
     // Only src matters for production builds.
-    "exclude": ["test/**/*"]
+    "exclude": ["test/**/*", "dist/**/*", "build/**/*"]
 }
--- a/plugins/dev-tools/package-lock.json
+++ b/plugins/dev-tools/package-lock.json
@@ -28,9 +28,9 @@
       }
     },
     "node_modules/@sinonjs/commons": {
-      "version": "1.8.3",
-      "resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.8.3.tgz",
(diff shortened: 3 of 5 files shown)


plugins/field-slider/tsconfig.json
```diff
@@ -11,5 +11,5 @@
     },
     // NOTE: `test/**/*` is automatically included in `blockly-scripts start`.
     // Only src matters for production builds.
-    "exclude": ["test/**/*"]
```
BeksOmega, line 14: I'm a bit confused about why this is necessary. Are we not `clean`ing before `build`ing? If not maybe we should be.

Also a bit curious why the [shadow block converter](https://github.com/google/blockly-samples/blob/master/plugins/shadow-block-converter/tsconfig.json) plugin is working without this change.

plugins/field-slider/package.json
```diff
@@ -41,12 +41,12 @@
   "devDependencies": {
     "@blockly/dev-scripts": "^1.2.32",
     "@blockly/dev-tools": "^5.2.4",
-    "blockly": "^9.2.0",
+    "blockly": "^9.3.1",
     "chai": "^4.2.0",
     "sinon": "^9.0.1"
   },
   "peerDependencies": {
-    "blockly": "^9.2.0"
```
BeksOmega, line 49: I think technically this could be `^9.3.0` - but no real reason not to require the patch imo.

BeksOmega (approved):
Generally LGTM! Just one question about the tsconfig changes.

plugins/field-slider/tsconfig.json
btw17: Excluding `dist` and `build` is just so VS Code doesn't complain when looking at the TS Config with those files built. This doesn't impact build pipelines since `dist` and `build` are excluded from git.

Regarding `clean` and `build`, I'll usually `clean` and `build` before committing, though for dev workflow, I usually don't worry about cleaning if my changes aren't introducing or removing build artifacts. (In the past, I've setup fresh `build` calls to remove the built folder initially so I never have to think about it, though, which might be worth considering here.)

plugins/field-slider/package.json
btw17: That was my thought. Since this is being updated anyway, I might as well update it to the latest at that time.

BeksOmega:
Just confirming this is gtg @btw17 ?

btw17:
This is g2g! @BeksOmega

PR #1588, opened 2023-02-25
chore: update field grid dropdown to blockly 9.2.1

btw17:
**Note:** This is dependent on https://github.com/google/blockly-samples/pull/1587.

- Added validator function types to `FieldGridDropdown`
- Updated `FieldGridDropdown`'s test index to TS
- Updated `FieldGridDropdown`'s blockly dependencies to `9.2.1`

PR #1587, opened 2023-02-25
chore: field grid dropdown ts conversion

btw17:
- Updated `plugins/field-grid-dropdown` to TS
- Added note about using blockly imports properly for `field-grid-dropdown` and `field-slider`

- **Note:** The tests are not updated to TS due to `@blockly/dev-tools`'s `package-lock.json` syncing with `9.2.1` which is not compatible with a few niche types defined in `9.0.0`. The following `9.2.1` changes impact `field-grid-dropdown`:
  - Although a valid function supplied for `9.0.0` is valid for `9.2.1`, the `Function` type is not compatible with the updated function types
  - `Event` is replaced with `MouseEvent` in `9.2.1`, though usage remains the same
- Se follow-up PR which addresses updating the version: https://github.com/google/blockly-samples/pull/1588

PR #1652, opened 2023-03-31
chore: converted `plugins/field-grid-dropdown` to TypeScript

btw17:
## Notes

- This is a follow-up to https://github.com/google/blockly-samples/pull/1651
- This PR replaces https://github.com/google/blockly-samples/pull/1587 and https://github.com/google/blockly-samples/pull/1588


plugins/field-grid-dropdown/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+        "declaration": true,
+        "declarationMap": true,
+        "module": "ES2015",
+        "moduleResolution": "node",
+        "target": "ES2015",
+        "strict": true,
+    },
+    // NOTE: `test/**/*` is automatically included in `blockly-scripts start`.
+    // Only src matters for production builds.
+    "exclude": ["test/**/*", "dist/**/*", "build/**/*"]
```
BeksOmega, line 14: Similar question about excludes as in previous PR: https://github.com/google/blockly-samples/pull/1651#discussion_r1154982424

plugins/field-grid-dropdown/src/index.ts
```diff
@@ -9,169 +9,215 @@
  * @author kozbial@google.com (Monica Kozbial)
  */
 
-import Blockly from 'blockly/core';
+import * as Blockly from 'blockly/core';
+
+/**
+ * A config object for defining a field grid dropdown.
+ */
+export interface FieldGridDropdownConfig extends Blockly.FieldConfig {
```
BeksOmega: I /think/ this should extend the `FieldDropdownConfig` instead, because we are passing the config to `super`. So if we ever define more config options for the `FieldDropdown` those will be supported by the `FieldGridDropdown`

plugins/field-grid-dropdown/src/index.ts
```diff
@@ -9,169 +9,215 @@
...
-   * Sets the number of columns on the grid. Updates the styling to reflect.
-   * @param {number} columns The number of columns. Is rounded to
-   *    an integer value and must be greater than 0. Invalid
-   *    values are ignored.
-   * @private
-   */
-  setColumns(columns) {
-    this.setColumnsInternal_(columns);
-    this.updateColumnsStyling_();
+    return new FieldGridDropdown(config.options, undefined, config);
```
BeksOmega: As per comment, it looks like this should be doing `new this`. Is that possible to maintain?

plugins/field-grid-dropdown/src/index.ts
```diff
@@ -9,169 +9,215 @@
...
-  }
-
-  /**
-   * Sets the number of columns on the grid. Updates the styling to reflect.
-   * @param {number} columns The number of columns. Is rounded to
-   *    an integer value and must be greater than 0. Invalid
-   *    values are ignored.
-   * @private
-   */
-  setColumns(columns) {
```
BeksOmega: Why was this removed?

plugins/field-grid-dropdown/src/index.ts
```diff
@@ -9,169 +9,215 @@
...
+   *  must be greater than 0. Invalid values are ignored.
    */
-  setColumnsInternal_(columns) {
-    columns = parseInt(String(columns));
-    if (!isNaN(columns) && columns >= 1) {
-      this.columns_ = columns;
+  private setColumnsInternal_(columns: string | number) {
+    const cols = typeof columns === 'string' ? parseInt(columns) : columns;
+    if (!isNaN(cols) && cols >= 1) {
+      this.columns = cols;
```
BeksOmega, line 122: Yay goodbye underscore!

plugins/field-grid-dropdown/src/index.ts
```diff
@@ -9,169 +9,215 @@
...
-   * @private
+   *
+   * @param columns The number of columns. Is rounded to an integer value and
+   *  must be greater than 0. Invalid values are ignored.
    */
-  setColumnsInternal_(columns) {
-    columns = parseInt(String(columns));
-    if (!isNaN(columns) && columns >= 1) {
-      this.columns_ = columns;
+  private setColumnsInternal_(columns: string | number) {
```
BeksOmega: The underscore can actually be removed here as well (since this method is private).

Here and for all other private methods (or it can be left for a follow-up)

plugins/field-grid-dropdown/src/index.ts
```diff
@@ -9,169 +9,215 @@
...
-        this.menu_.getElement(), 'fieldGridDropDownContainer');
+    const menuElement = this.menu_?.getElement() ?? null;
+    if (menuElement) {
+      Blockly.utils.dom.addClass(menuElement,
+          'fieldGridDropDownContainer');
+    }
     this.updateColumnsStyling_();
 
+    // NOTE: dropdownDispose is 'private' in FieldDropdown. It is corrected
+    // in a later version to be 'protected', though until that is
```
BeksOmega: I think this is fixed in v9.3.0!

plugins/field-grid-dropdown/src/index.ts
```diff
@@ -9,169 +9,215 @@
...
+    return {
+      primary: this.primaryColour ?? colourSource.getColour(),
+      border: this.borderColour ?? colourSource.getColourTertiary(),
+    };
+  }
 }
 
+// IRegisterableField should accept `T extends Field`.
+// This `fromJson` doesn't match, though it should be compatible. The options
+// rightfully should be required here.
```
BeksOmega: Is this comment still relevant?

BeksOmega (commented):
Generally LGTM, just a few questions =)

plugins/field-grid-dropdown/tsconfig.json
btw17: Responded in the previous PR!

Tldr; this is a mostly innocuous change that just makes VS Code complain less when the TS Config file is open with artifacts built. It doesn't actually impact the pipeline build since those artifacts aren't included in git anyway.

plugins/field-grid-dropdown/src/index.ts
btw17: Good catch!

plugins/field-grid-dropdown/src/index.ts
btw17: Looks like it isn't - updated!

plugins/field-grid-dropdown/src/index.ts
btw17: Updated!

plugins/field-grid-dropdown/src/index.ts
btw17: Done!

plugins/field-grid-dropdown/src/index.ts
btw17: Looks like `this` works fine. Updated!

plugins/field-grid-dropdown/src/index.ts
btw17: I'll need to check back on this later this week when I've got time. Some of the changes were pulled from previous work I'd done before the `9.3.1` updates, so if this plugin changed in the meantime it's possible I missed something.

btw17: I see now - it was marked `private`, though it's not used anywhere internally. There's no purpose in it being here if it's private as it's currently not used and can't be used by subclasses.

I could add it back without marking it `private`, or I could leave it as is. WDYT?

BeksOmega: If it works (i.e. properly rerenders the dropdown div) I think it makes sense to readd it as public! Seems like a reasonable part of the API this field should have =)

But if it doesn't actually work properly (and can't trivially be fixed). Then I'd say leave as is (i.e. removed). That SGTY?

btw17: SGTM! It works just fine as public, so I'll add it back.

btw17:
All set @BeksOmega!

PR #1653, opened 2023-03-31
chore: modernize keyboard navigation codelab

jubishop:
This brings keyboard navigation inline with the standards established by the custom renderer and generator codelabs.

- Uses the sample app instead of the test/ folder in the plugin.
- Installs the keyboard navigation plugin using `npm install @blockly/keyboard-navigation --save`

Because there's less already set up for us compared to the file in `test/index.js`, the steps grew a bit.  In particular, the user needs to import and instantiate the `NavigationController` on their own.

I manually went through this codelab following the instructions as if I was a new user and am fairly confident it all works.


codelabs/keyboard_navigation/keyboard_navigation.md
```diff
@@ -67,22 +75,50 @@ export class CustomCursor extends Blockly.Cursor {
...
+```js
+import {CustomCursor} from './cursors/custom';
+```
+
+Now tell the workspace to use your new cursor by calling `setCursor()` on the `MarkerManager`:
 
 ```js
-import {CustomCursor} from './custom_cursor';
+// This line already exists, call `setCursor()` after this..
+const ws = Blockly.inject( // ...etc
```
maribethb: Instead of including this line (too easy to accidentally call inject twice if you indiscriminately copy/paste?) what about

After injecting the workspace, we will use its `MarkerManager` to set the new cursor:

```js
ws.getMarkerManager().setCursor(new CustomCursor());
```



---

rationale: if they paste this snippet in the wrong place (befpre inject), debugging that is a lot easier than debugging multiple inject calls. similar logic applies for the navigation controller step below. wdyt?

examples/keyboard-navigation-codelab/src/markers/custom_marker_svg.js
```diff
@@ -0,0 +1,88 @@
...
+   */
+  hide() {
+    super.hide();
+    // Hide the marker we created.
+    this.blockPath_.style.display = 'none';
+  }
+}
+
+/**
+ *
```
maribethb: Can you either add documentation or remove the comment block?

examples/keyboard-navigation-codelab/src/markers/custom_marker_svg.js
```diff
@@ -0,0 +1,88 @@
...
+ */
+class CustomRenderer extends Blockly.geras.Renderer {
```
maribethb, line 75: Optional: Would be good to use thrasos instead of geras if possible. If you're up for it, can you see if extending from thrasos works without any further changes?

maribethb (commented):
Generally looks good as a straightforward conversion, pending some minor comments.

---

I think this codelab could be improved in a few ways, but these are optional suggestions and they should be done in a separate PR or we could file an issue for them and do them later, but I wanted to write them down while we've both been looking at it:

- The organization could be improved a bit. Namely, the first thing we do is create a CustomCursor, but then we don't actually customize it in any way until after some other steps in the codelab. I would not add this file until we are ready to customize it.
- The introduction could be improved. I would probably explain in more detail about cursors and markers and how exactly we will customize them (e.g. the difference between customizing the appearance of a marker and the behavior of a cursor). Consider explaining the different types of cursors included in the plugin, or at least link to the explanation in the plugin's readme.
- After setting up the navigation controller, we could provide some instructions on exploring the basic keyboard navigation provided by the plugin, like telling the user to open the sample app, press ctl+shift+k, then navigate around by pressing buttons x, y, z, etc as a way of orienting what we are about to change.

codelabs/keyboard_navigation/keyboard_navigation.md
jubishop: sgtm.  done.

jubishop: small side note:  i noticed some inconsistency where sometimes the phrasing says things like "you will do" or just "do" and other times "we will do", and similarly will sometimes refer to the code as "your" vs "our".  i've standardized on the "you/your" language.   So like in this example I changed your lanugage to ```After injecting the workspace, use its `MarkerManager` to set your new custom cursor:```

examples/keyboard-navigation-codelab/src/markers/custom_marker_svg.js
jubishop: d'oh my bad.  my vscode autoformatter keeps adding these on me.  done.

examples/keyboard-navigation-codelab/src/markers/custom_marker_svg.js
jubishop: i think the problem is the two gifs in the "Test it out" sections below will no longer look right compared to what the user will be seeing, right?

since i have yet to investigate how to capture and replace the gifs in these articles i won't have time to do this today.  i've added this to #1657 so i can follow up and do it in 2 fridays (I'm OOO next friday).

maribethb (approved):
Thanks!

PR #1654, opened 2023-04-01
fix: local testing for github pages site

rachel-fenichel:
Part of https://github.com/google/blockly-samples/issues/1549

This change stops using `bundle exec jekyll serve` for local testing by just building the site locally and using `http-server` to serve it. To make that work I had to pipe an `isLocal` flag around and use it to set the base URL for all of the links.

Also removes an unused dependency (`json-to-pretty-yaml`) and cleans up some code, including inlining some portions of the template strings where I thought it made the resulting HTML easier to read.

I suggest reviewing one commit at a time.

You can test it by pulling the code down and running `npm ci && npm test:ghpages` to see a local version of all of the pages. The remote version is live at https://rachel-fenichel.github.io/blockly-samples/ and you can also test it by running `npm deploy`.

### DIFF
--- a/scripts/gh-predeploy.js
+++ b/scripts/gh-predeploy.js
@@ -19,37 +19,37 @@ const resolveApp = (relativePath) => path.resolve(appDirectory, relativePath);
 
 /**
  * Inject head HTML for a plugin or example page on gh-pages.
- * This finds the existing head tag and inserts a few additional lines of CSS,
- * as well as updating the title to match the plugin or example's name.
  * @param {string} initialContents The initial page HTML, as a string.
- * @param {string} title The title to use for the page, which may be generated from the package
- *    name or specified explicitly.
- * @param {boolean} isLocal True if building for a local test. False if building for gh-pages.
+ * @param {string} title The title to use for the page, which may be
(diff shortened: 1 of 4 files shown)


scripts/gh-predeploy.js
```diff
@@ -303,11 +306,32 @@ function prepareToDeployPlugins(done) {
...
 }
 
+/**
+ * Prepare plugins for deployment to gh-pages.
+ *
+ * For each plugin, copy relevant files to the gh-pages directory.
+ * @param {Function} done Completed callback.
+ * @return {Function} Gulp task.
+ */
+function prepareLocalPlugins(done) {
```
BeksOmega, line 320: This looks like it duplicates a lot of code from `prepareToDeplyPlugins`. Could these be unified using a subprocedure or another level of closure?

Same for examples.

BeksOmega (approved):
One comment about code duplication. Otherwise LGTM =)

scripts/gh-predeploy.js
rachel-fenichel: Beka and I took at look at this but couldn't find a good way to unify the functions that was actually more readable.

PR #1658, opened 2023-04-07
feat: created TypeScript sample app

btw17:
Created a mirror of `./examples/sample-app` converted to TypeScript at `./examples/sample-app-ts`

## Notes

- In it's current state, this app is translated to TS and is fully operational in the same ways the original sample-app is, though it's done nothing to update the code conventions or add additional type information beyond what is available in in the `blockly` core package.
- Additional considerations:
  - Updating the `css` implementation to `scss`

### DIFF
--- a/examples/sample-app-ts/package.json
+++ b/examples/sample-app-ts/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "sample-app",
+  "name": "sample-app-typescript",
   "version": "1.0.0",
   "description": "A sample app using Blockly",
   "main": "index.js",
--- a/examples/sample-app-ts/package.json
+++ b/examples/sample-app-ts/package.json
@@ -1,0 +1,31 @@
+{
+  "name": "sample-app",
+  "version": "1.0.0",
+  "description": "A sample app using Blockly",
+  "main": "index.js",
+  "private": true,
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "webpack --mode production",
+    "start": "webpack serve --open --mode development"
+  },
+  "keywords": [
+    "blockly"
+  ],
+  "author": "Blockly team",
+  "license": "Apache-2.0",
+  "devDependencies": {
+    "css-loader": "^6.7.1",
+    "html-webpack-plugin": "^5.5.0",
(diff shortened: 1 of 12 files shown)


examples/sample-app-ts/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "sample-app",
```
maribethb: can you make this `sample-app-typescript`, we use the package names for some things and I think there is a potential for breakage if we have duplicates

examples/sample-app-ts/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+        "declarationMap": true,
+        "module": "ES2015",
+        "moduleResolution": "node",
+        "target": "ES2015",
+        "strict": true,
+    },
+    // NOTE: `test/**/*` is automatically included in `blockly-scripts start`.
+    // Only src matters for production builds.
+    "exclude": ["test/**/*"]
+}
```
maribethb: needs newline at eof here

examples/sample-app-ts/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+        "baseUrl": "./",
+        "outDir": "dist",
+        "declaration": true,
+        "declarationMap": true,
+        "module": "ES2015",
+        "moduleResolution": "node",
+        "target": "ES2015",
+        "strict": true,
+    },
+    // NOTE: `test/**/*` is automatically included in `blockly-scripts start`.
```
maribethb: The comment is definitely wrong (we aren't using `blockly-scripts start`), but I think this entire property needs to go as currently there is no `test` directory. Eventually it should have tests, but if the tests are in ts too then i'm not sure if we would want this or not.

btw17: Good catch! Fixed.

examples/sample-app-ts/tsconfig.json
btw17: Done!

examples/sample-app-ts/package.json
btw17: Done!

maribethb (approved):

PR #1660, opened 2023-04-07
fix: examples vue3 javascript fix

### LINKED ISSUE #1594: Cannot set properties of undefined (setting 'stock_buy_simple')
I try to run the blockly-vue3 example, npm i & npm run start, but the web page is blank.
The console show the error message " Cannot set properties of undefined (setting 'stock_buy_simple')"
So I check it, found the "Blockly.JavaScript" is undefined, So what can I do now, thanks for help!

mikebarr24:
fixes #1594


BeksOmega (approved):
LGTM! Thanks for the fix Mike =)

PR #1666, opened 2023-04-12
fix!: make scroll options work without drag surfaces

### LINKED ISSUE #1664: Remove dependency on drag surfaces from the scroll options plugin
**Category**

 - Plugins

**Component**

scroll-options

**Describe the bug**

Currently the scroll options plugin depends on the drag surfaces to perate correctly. However, we are planning on removing those from core this quarter because they have bad performance. So we need to update the plugin so that it works with v10.

BeksOmega:
### Description

Fixes #1664 

Tested against https://github.com/google/blockly/pull/6758

### Additional information

Not mergable until after v10 release.

TODO: Need to bump minimum version of Blockly & master version of plugin.

### DIFF
--- a/plugins/scroll-options/src/index.js
+++ b/plugins/scroll-options/src/index.js
@@ -77,7 +77,7 @@ export class ScrollOptions {
       return;
     }
 
-    // TODO: We should maybe add an accessor for the svgGroup_?
+    // TODO(blockly/#7157): We should maybe add an accessor for the svgGroup_?
     this.wheelEvent_ = Blockly.browserEvents.conditionalBind(
         this.workspace_.svgGroup_, 'wheel', this, this.onMouseWheel_);
   }
--- a/plugins/scroll-options/src/index.js
+++ b/plugins/scroll-options/src/index.js
@@ -77,16 +77,9 @@ export class ScrollOptions {
       return;
     }
 
-    let element;
-    const dragSurface = this.workspace_.getBlockDragSurface();
-    if (dragSurface) {
-      element = dragSurface.getSvgRoot();
-    } else {
-      element = this.workspace_.svgGroup_;
-    }
-
+    // TODO: We should maybe add an accessor for the svgGroup_?
     this.wheelEvent_ = Blockly.browserEvents.conditionalBind(
-        element, 'wheel', this, this.onMouseWheel_);
+        this.workspace_.svgGroup_, 'wheel', this, this.onMouseWheel_);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


plugins/scroll-options/src/index.js
```diff
@@ -77,16 +77,9 @@ export class ScrollOptions {
...
 
-    let element;
-    const dragSurface = this.workspace_.getBlockDragSurface();
-    if (dragSurface) {
-      element = dragSurface.getSvgRoot();
-    } else {
-      element = this.workspace_.svgGroup_;
-    }
-
+    // TODO: We should maybe add an accessor for the svgGroup_?
```
cpcallen: Please open a bug about this (maybe with an explanation more comprehensible to the casual observer) and reference it here.  (Make the bug title along the lines of "consider whether to…" if you're not sure if we want to do this.)

plugins/scroll-options/src/utils.js
```diff
@@ -11,14 +11,10 @@
...
-  if (dragSurface) {
-    return dragSurface.getWsTranslation();
-  } else {
-    const translation = ws.svgBlockCanvas_.getAttribute('transform');
-    const splitted = translation.split(',');
-    const x = Number(splitted[0].split('(')[1]);
-    const y = Number(splitted[1].split(')')[0]);
-    return new Blockly.utils.Coordinate(x, y);
-  }
+  // TODO: We should maybe make getBlockCanvas public?
```
cpcallen: Ditto here.  (Maybe one bug for both TODOs?)

plugins/scroll-options/src/utils.js
```diff
@@ -11,14 +11,10 @@
...
+  const translation = ws.svgBlockCanvas_.getAttribute('transform');
+  const splitted = translation.split(',');
+  const x = Number(splitted[0].split('(')[1]);
+  const y = Number(splitted[1].split(')')[0]);
```
cpcallen, line 20: I think this deserves a comment with an example of what format `translation` is in.

(Also, consider if this is a place where a regexp with capture groups might actually be easier to read.)

cpcallen (approved):
Approved with two nits.

plugins/scroll-options/src/index.js
BeksOmega: Done! https://github.com/google/blockly/issues/7157

plugins/scroll-options/src/utils.js
BeksOmega: Done!

PR #1674, opened 2023-04-25
feat: Add angle, colour, and multiline fields.

NeilFraser:
In preparation for removing said fields from Blockly core.


plugins/field-angle/src/field_angle.ts
```diff
@@ -0,0 +1,553 @@
...
+      case Mode.PROTRACTOR:
+        // This is the default mode, so we could do nothing.  But just to
+        // future-proof, we'll set it anyway.
+        this.clockwise = false;
+        this.offset = 0;
+        break;
+    }
+
+    // Allow individual settings to override the mode setting.
+    if (config.clockwise) this.clockwise = config.clockwise;
```
rachel-fenichel, line 139: I think this fails if mode was `compass` but `config.clockwise` was false.

rachel-fenichel: (Which would be a weird case.)

plugins/field-angle/src/field_angle.ts
```diff
@@ -0,0 +1,553 @@
...
+
+  /**
+   * Create the block UI for this field.
+   *
+   * @internal
+   */
+  override initView() {
+    super.initView();
+    // Add the degree symbol to the left of the number,
+    // even in RTL (issue #2380).
```
rachel-fenichel: Change to a full link since it's going into a new repository: https://github.com/google/blockly/issues/2380

plugins/field-angle/src/field_angle.ts
```diff
@@ -0,0 +1,553 @@
...
+        break;
+      case 'ArrowUp':
+        // increment
+        multiplier = 1;
+        break;
+    }
+    if (multiplier) {
+      const value = this.getValue() as number;
+      this.displayMouseOrKeyboardValue(value + multiplier * this.round);
+      e.preventDefault();
```
rachel-fenichel, line 409: Why only call `preventDefault()` and `stopPropagation()` when there's a multiplier?

Edit: ah, I see, because otherwise it's another key and you ignore it. How about putting a default case in the `switch` that returns early, and then doing everything in the `if (multiplier)` section without the conditional?

plugins/field-angle/test/field_angle_test.mocha.js
```diff
@@ -0,0 +1,381 @@
...
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const {testHelpers} = require('@blockly/dev-tools');
+const {FieldAngle} = require('../src/index');
+const {assert} = require('chai');
+
+const {
```
rachel-fenichel, line 11: FYI I filed https://github.com/google/blockly-samples/issues/1676 about the type issues with the test helpers.

rachel-fenichel (commented):
First pass: I reviewed the angle files and left comments.

plugins/field-multilineinput/src/field_multilineinput.ts
```diff
@@ -0,0 +1,521 @@
...
+  }
+
+  /**
+   * Serializes this field's value to XML.
+   * Should only be called by Blockly.Xml.
+   *
+   * @param fieldElement The element to populate with info about the field's
+   *     state.
+   * @returns The element containing info about the field's state.
+   * @internal
```
rachel-fenichel: Remove `@internal` tag, here and below, since these will be called by code outside of this plugin.

plugins/field-angle/src/field_angle.ts
NeilFraser: Fixed.

plugins/field-angle/src/field_angle.ts
NeilFraser: Fixed.

plugins/field-angle/src/field_angle.ts
NeilFraser: There's a school of thought that each function should only have a single return statement (to the point where zealots disapprove of guard clauses).  I don't subscribe to this philosophy religiously, but it does have a point in often making program flow easier to understand.  With a single return one can add code anywhere in a function and know it will execute.  Whereas with multiple returns one has to read the whole function to know where one can add code.

While I think there are many cases where multiple returns makes code easier to read and maintain, the general case benefits from a single return.
http://wiki.c2.com/?SingleFunctionExitPointPlease

plugins/field-multilineinput/src/field_multilineinput.ts
NeilFraser: Done.

plugins/field-angle/src/field_angle.ts
NeilFraser: Lol.  Reverted.  The tests fail if this behaviour is altered.
Will add it to the proposed updates later.

plugins/dev-scripts/package.json
```diff
@@ -43,7 +43,7 @@
     "webpack-dev-server": "^4.0.0"
   },
   "peerDependencies": {
-    "typescript": "^4.3.2"
+    "typescript": "^5.0.2"
```
cpcallen: This should be replaced by PR #1684.

plugins/field-angle/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "dist",
+    "src"
+  ],
+  "devDependencies": {
+    "@blockly/dev-scripts": "^1.2.32",
+    "@blockly/dev-tools": "^5.2.4",
+    "blockly": "^9.3.1",
+    "chai": "^4.2.0",
+    "sinon": "^9.0.1",
+    "typescript": "^4.3.2"
```
cpcallen: You probably want to `npm install typescript@latest` here.

cpcallen (changes requested):
Once the `typescript` dependencies are updated per comments below, run `rm -rf plugins/*/node_modules` then `npm ci` (both from repository root, not anywhere in `plugins/`), and then check in any changes to any `package-lock.json` file (which should _mostly_ be deletions of references to other packages in `plugins/` and their dependencies).

BeksOmega:
Superceded by #1693

PR #1678, opened 2023-04-27
fix: replace scrollToVisible_ with standard centerOnBlock

### LINKED ISSUE #1677: plugin-workspace-search use centerOnBlock for scrollToVisible_
Can we use this to move the block to the center of the workspace:
https://developers.google.com/blockly/reference/js/blockly.workspacesvg_class.centeronblock_1_method.md
For the scrollToVisible() call:
https://github.com/google/blockly-samples/blob/master/plugins/workspace-search/src/WorkspaceSearch.js#L634

mrkprdo:
This is an attempt to fix #1677.

### DIFF
--- a/plugins/workspace-search/src/WorkspaceSearch.js
+++ b/plugins/workspace-search/src/WorkspaceSearch.js
@@ -637,7 +637,7 @@ export class WorkspaceSearch {
       return;
     }
 
-    if (this.workspace_.hasOwnProperty('centerOnBlock')) {
+    if (typeof this.workspace_.centerOnBlock === 'function') {
       // if workspace supports centerOnBlock, let's use it
       this.workspace_.centerOnBlock(block.id, false);
       return;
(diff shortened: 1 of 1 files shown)


plugins/workspace-search/src/WorkspaceSearch.js
```diff
@@ -636,6 +636,13 @@ export class WorkspaceSearch {
       // Cannot scroll to block in a non-movable workspace.
       return;
     }
+
+    if (typeof this.workspace_.centerOnBlock === 'function') {
```
BeksOmega: Hello! Thanks for the fix :D Instead of changing `scrollToVisible` could you just delete this method and instead call `centerOnblock` in its place? `centerOnBlock` should  always be defined so no need to worry about that!

BeksOmega (changes requested):

plugins/workspace-search/src/WorkspaceSearch.js
mrkprdo: Ok, I will do that.

BeksOmega (approved):
LGTM! Thank you for the fix :D

plugins/workspace-search/src/WorkspaceSearch.js
mrkprdo: Done

PR #1680, opened 2023-05-03
chore(deps): Update `typescript`, delete unneeded dependency on it

cpcallen:
## Changes

* Remove direct typescript dependency of `field-grid-dropdown` on `typescript`: all the other plugins depend on typescript only via dev-scripts, and that appears to work fine here too.  Done by running `npm rm typescript` in `plugins/field-grid-dropdown`.

* Update typescript to v5: this will prevent eslint from generating errors when we use the `overrides`keyword.  Done by running `npm install --save-peer typescript@latest` in `plugins/dev-scripts`.

## Notes
Both the `npm` commands had the unexpected side-effects:

* Removing `typescript` from `field-grid-dropdown` modified that package's `package-lock.json` as follows:
  ```diff
   {
     "name": "@blockly/field-grid-dropdown",
     "version": "2.0.14",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
       "": {
         "name": "@blockly/field-grid-dropdown",
         "version": "2.0.14",
         "license": "Apache 2.0",
         "devDependencies": {
  -        "blockly": "^9.3.1",
  -        "typescript": "^4.3.2"
  +        "@blockly/dev-scripts": "^1.2.33",
  +        "@blockly/dev-tools": "^5.2.5",
  +        "blockly": "^9.3.1"
         },
  ```
  followed by adding version lock information for `@blockly/dev-scripts` and `-tool` and all of their recursive dependencies.

* Similarly, upgrading `typescript` in `dev-scripts` modified that package's `package-lock.json` as follows:
  ```diff
   {
     "name": "@blockly/dev-scripts",
     "version": "1.2.33",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
       "": {
         "name": "@blockly/dev-scripts",
         "version": "1.2.33",
         "license": "Apache-2.0",
         "dependencies": {
           "@babel/code-frame": "^7.8.3",
           "@babel/core": "^7.8.6",
           "@babel/preset-env": "^7.8.6",
  +        "@blockly/eslint-config": "^2.1.21",
           "babel-loader": "^8.0.6",
           "chalk": "^4.0.0",
           "eslint": "^7.15.0",
           "eslint-webpack-plugin": "^3.0.0",
           "fork-ts-checker-webpack-plugin": "^4.1.3",
           "mocha": "^9.2.0",
           "rimraf": "^3.0.2",
           "source-map-loader": "^1.0.0",
           "ts-loader": "^9.4.1",
           "webpack": "^5.0.0",
           "webpack-cli": "^4.9.1",
           "webpack-dev-server": "^4.0.0"
         },
         "bin": {
           "blockly-scripts": "bin/blockly-scripts.js"
         },
         "engines": {
           "node": ">=8.0.0"
         },
         "peerDependencies": {
  -        "typescript": "^4.3.2"
  +        "typescript": "^5.0.4"
         },
         "peerDependenciesMeta": {
           "typescript": {
  -          "optional": true
  +          "optional": false
           }
         }
       },
  ```
  and again added copious amounts of additional lock information for `@blockly/eslint-config` and its dependencies.

## Questions

* What is the correct way to add, remove, or upgrade a dependency for a singe package?  Presumably we should not be pinning our dependencies on our own packages as appears to be happening here.  (`lerna add` seems to able to add deps to all packages, but not AFAICT to remove, or upgrade only where already installed.)
* Why is typescript a peer dependency of `dev-scripts`?  Is it done this way to let us to specify which version we want webpack and eslint to use?

### DIFF
--- a/plugins/dev-scripts/package-lock.json
+++ b/plugins/dev-scripts/package-lock.json
@@ -12,6 +12,7 @@
         "@babel/code-frame": "^7.8.3",
         "@babel/core": "^7.8.6",
         "@babel/preset-env": "^7.8.6",
+        "@blockly/eslint-config": "^2.1.21",
         "babel-loader": "^8.0.6",
         "chalk": "^4.0.0",
         "eslint": "^7.15.0",
@@ -32,11 +33,11 @@
         "node": ">=8.0.0"
       },
       "peerDependencies": {
-        "typescript": "^4.3.2"
+        "typescript": "^5.0.4"
       },
       "peerDependenciesMeta": {
         "typescript": {
-          "optional": true
+          "optional": false
         }
       }
     },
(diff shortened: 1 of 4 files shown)


maribethb:
After removing the dependency, you need to run `lerna bootstrap` again from the root of blockly-samples in order to fix the erroneous dependencies added by npm during the uninstall.

But I don't think this change is correct. Plugins should declare their dependencies accurately. In this case, the plugin has a dev dependency on typescript and it should declare that. That's why there is a peer dependency on `typescript` from `dev-scripts`, because the package using the script should also have `typescript` installed. The dev dependency on typescript is also supposed to be added by the `dev-create` template if you pass the `--typescript` flag when creating a new plugin.

So I think the correct thing is to add `typescript` to the plugins that don't correctly declare their dependencies instead. Did you run into problems doing that?


---

misc note: If we were using yarn, it is more strict about packages declaring their dependencies correctly, so plugins would definitely have to declare `typescript` as a dev dependency. Yarn also has built-in support for the inter-plugin dependencies so it wouldn't have added the erroneous dependency here that `lerna bootstrap` removes for us.

maribethb:
Additionally, I think you could modify the peer dependency on typescript to include both versions `^4 || ^5` and let plugins individually manage their minimum version of typescript, which is nicer to anyone using this script for a third-party plugin. The feature that requires ts 5 is only in one particular plugin, not in `dev-scripts`, and that's why that plugin should declare its own dependency on the particular version of ts it requires.

cpcallen:
> After removing the dependency, you need to run `lerna bootstrap` again from the root of blockly-samples in order to fix the erroneous dependencies added by npm during the uninstall.

I tried that but it had no effect, either after the `npm rm` or the `npm install` (except to create a bunch of spurious `.eslintcache` files, for some reason).

> But I don't think this change is correct. Plugins should declare their dependencies accurately. In this case, the plugin has a dev dependency on typescript and it should declare that. That's why there is a peer dependency on `typescript` from `dev-scripts`, because the package using the script should also have `typescript` installed. The dev dependency on typescript is also supposed to be added by the `dev-create` template if you pass the `--typescript` flag when creating a new plugin.

Hmm.  I agree with you in principle, but AFAICT `field-grid-dropdown` does not invoke `tsc` direct, instead only using `blockly-scripts` (from `@blockly/dev-scripts`, presumably).  Is `typescript` supposed to be a truly optional peer dependency of `dev-scripts`, separately installed where needed?  If so then it is all the other TS-defined packages that are wrong, because e.g. `block-shareable-procedures` is written in TypeScript but does not have `typescript` as a devDependency.  (I was removing it from `field-grid-dropdown` for consistency.)

> So I think the correct thing is to add `typescript` to the plugins that don't correctly declare their dependencies instead. Did you run into problems doing that?

Neil was getting version conflicts trying to install `typescript` 5.x in one of the field plugins he was porting from core, though your `^4 || ^5` might have fixed that.

> misc note: If we were using yarn, it is more strict about packages declaring their dependencies correctly, so plugins would definitely have to declare `typescript` as a dev dependency.

A useful feature, but I'm not sure how it could figure out that any of our plugins written in TypeScript actually depend on the `typescript` package, because none of them actually import it directly or use the `tsc` binary.

> Yarn also has built-in support for the inter-plugin dependencies so it wouldn't have added the erroneous dependency here that `lerna bootstrap` removes for us.

That will be great.

I should have noted that this PR is intended more as an emergency fix to facilitate Neil's work than an actual effort to improve our dependency situation…

> Additionally, I think you could modify the peer dependency on typescript to include both versions `^4 || ^5` and let plugins individually manage their minimum version of typescript, which is nicer to anyone using this script for a third-party plugin. The feature that requires ts 5 is only in one particular plugin, not in `dev-scripts`, and that's why that plugin should declare its own dependency on the particular version of ts it requires.

That makes a lot of sense but is as noted quite different from how things are currently set up.

maribethb:
Yes, all of the other TS plugins are wrong and should declare their dependency on TS IMO. I see what you're saying about how they could all just rely on the typescript version of dev-scripts, but I don't think that really makes sense, especially with typescript being a peer dependency of dev-scripts. (btw that's why yarn would complain - it would know that dev-scripts has an unmet peer dependency)

I think to unblock Neil all you need to do is add `|| ^5` to the `dev-scripts` peer dependency on typescript, and then Neil should be able to install v5 into his plugin.

In a separate PR once Neil is unblocked we can make the other plugins consistent by adding TS to their dev dependencies as well. 

As for lerna / package-lock issues, I'm not sure why bootstrap isn't working. I would run `npm run clean:node` from root next and then try bootstrap again, if this is still an issue.

As for the `.eslintcache` files, I ran into that when I was trying to get yarn working but I thought it was a yarn thing. They must have changed something about the default value of either whether to cache or where to store the cache file. I think I have a fix for this. For now, just delete them. Edit: I can't reproduce this now and I don't know why it would be happening. The default is to not use a cache and we don't change that when we invoke eslint. I was going to manually [set the cache location](https://eslint.org/docs/latest/integrate/nodejs-api#parameters) in the script but I don't think it's set up to cache in the first place.

cpcallen:
> Yes, all of the other TS plugins are wrong and should declare their dependency on TS IMO.

Am I out of touch?  No, it is the children who are wrong.

> I see what you're saying about how they could all just rely on the typescript version of dev-scripts, but I don't think that really makes sense, especially with typescript being a peer dependency of dev-scripts. (btw that's why yarn would complain - it would know that dev-scripts has an unmet peer dependency)

I agree that it probably makes sense for packages to provide their own version of `typescript`.  I note, however, that `npm` seems to ensure that _some_ version is installed even when the only dependency on it is an optional peer dependency.

> I think to unblock Neil all you need to do is add `|| ^5` to the `dev-scripts` peer dependency on typescript, and then Neil should be able to install v5 into his plugin.
> 
> In a separate PR once Neil is unblocked we can make the other plugins consistent by adding TS to their dev dependencies as well.

Sensible.

> As for lerna / package-lock issues, I'm not sure why bootstrap isn't working. I would run `npm run clean:node` from root next and then try bootstrap again, if this is still an issue.

Nope, that also had no effect.

cpcallen:
> > As for lerna / package-lock issues, I'm not sure why bootstrap isn't working. I would run `npm run clean:node` from root next and then try bootstrap again, if this is still an issue.
> 
> Nope, that also had no effect.

Actually, it may be that I spoke too soon: it seems that it did have an effect.

> > (btw that's why yarn would complain - it would know that dev-scripts has an unmet peer dependency)

I thought a bit more about this, and I don't think that makes sense, since plugins that don't use `typescript` will also  depend on `dev-scripts` and should be OK with having only the latter but not the former installed.

PR #1697, opened 2023-05-12
chore: Mostly remove Babel from dev-scripts.

gonfunko:
Since we're no longer supporting old browsers, we don't need Babel to cross-compile to old JS.

### DIFF
--- a/plugins/dev-scripts/config/webpack.config.js
+++ b/plugins/dev-scripts/config/webpack.config.js
@@ -97,21 +97,9 @@ module.exports = (env) => {
           use: [require.resolve('source-map-loader')],
           enforce: 'pre',
         },
-        isTypescript ? {
+        isTypescript && {
           test: /\.tsx?$/,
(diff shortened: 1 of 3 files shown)


plugins/dev-scripts/config/webpack.config.js
```diff
@@ -97,21 +97,9 @@ module.exports = (env) => {
           use: [require.resolve('source-map-loader')],
           enforce: 'pre',
         },
-        isTypescript ? {
+        isTypescript && {
```
rachel-fenichel, line 100: Side question for my own curiosity: why is the test here `.tsx` instead of `.ts`?

rachel-fenichel (approved):

plugins/dev-scripts/config/webpack.config.js
gonfunko: It's tsx*?*, so should match both I belive.

PR #1700, opened 2023-05-18
feat: Revise fields

NeilFraser:
* Remove static default fields in angle and colour (superseded by JSON configuration).
* Rebase angle onto number field (angle->value, round->precision).
* Add displayMin, displayMax, majorTick, minorTick, symbol, min, max to angle field.

### DIFF
--- a/plugins/field-angle/src/field_angle.ts
+++ b/plugins/field-angle/src/field_angle.ts
@@ -24,6 +24,8 @@ export class FieldAngle extends Blockly.FieldNumber {
    */
   static readonly RADIUS: number = FieldAngle.HALF - 1;
 
+  static readonly DEFAULT_PRECISION = 15;
+
   /**
    * Whether the angle should increase as the angle picker is moved clockwise
    * (true) or counterclockwise (false).
@@ -42,12 +44,6 @@ export class FieldAngle extends Blockly.FieldNumber {
    * Usually either 360 (for 0 to 359.9) or 180 (for -179.9 to 180).
    */
   private wrap = 360;
-
-  /**
-   * The amount to round angles to when using a mouse or keyboard nav input.
-   * Must be a positive integer to support keyboard navigation.
-   */
-  private round = 15;
 
   /**
    * Array holding info needed to unbind events.
@@ -86,6 +82,11 @@ export class FieldAngle extends Blockly.FieldNumber {
     if (value === Blockly.Field.SKIP_SETUP) return;
     if (config) {
       this.configure_(config);
+      if (config.precision === undefined || config.precision === null) {
+        this.setPrecision(FieldAngle.DEFAULT_PRECISION);
+      }
+    } else {
+      this.setPrecision(FieldAngle.DEFAULT_PRECISION);
     }
     this.setValue(value);
     if (validator) {
@@ -118,7 +119,6 @@ export class FieldAngle extends Blockly.FieldNumber {
     if (config.clockwise) this.clockwise = config.clockwise;
     if (config.offset) this.offset = config.offset;
     if (config.wrap) this.wrap = config.wrap;
-    if (config.round) this.round = config.round;
   }
 
   /**
@@ -304,8 +304,8 @@ export class FieldAngle extends Blockly.FieldNumber {
    * @param angle New angle.
    */
   private displayMouseOrKeyboardValue(angle: number) {
-    if (this.round) {
-      angle = Math.round(angle / this.round) * this.round;
+    if (this.precision_) {
+      angle = Math.round(angle / this.precision_) * this.precision_;
     }
     angle = this.wrapValue(angle);
     if (angle !== this.value_) {
@@ -384,7 +384,7 @@ export class FieldAngle extends Blockly.FieldNumber {
     }
     if (multiplier) {
       const value = this.getValue() as number;
-      this.displayMouseOrKeyboardValue(value + multiplier * this.round);
+      this.displayMouseOrKeyboardValue(value + multiplier * this.precision_);
       e.preventDefault();
       e.stopPropagation();
     }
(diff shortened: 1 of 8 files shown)


BeksOmega (approved):

NeilFraser:
Ok, this is ready now.
Honestly, I can't read the math in this PR.  It's write-only code.

plugins/field-angle/src/field_angle.ts
```diff
@@ -116,9 +149,21 @@ export class FieldAngle extends Blockly.FieldNumber {
...
-    if (config.clockwise) this.clockwise = config.clockwise;
-    if (config.offset) this.offset = config.offset;
-    if (config.wrap) this.wrap = config.wrap;
+    if (config.clockwise !== undefined) this.clockwise = config.clockwise;
+    if (config.offset !== undefined) this.offset = config.offset;
+    if (config.displayMin !== undefined) this.displayMin = config.displayMin;
+    if (config.displayMax !== undefined) this.displayMax = config.displayMax;
+    if (config.minorTick !== undefined) this.minorTick = config.minorTick;
+    if (config.majorTick !== undefined) this.majorTick = config.majorTick;
+    if (config.symbol !== undefined) this.symbol = config.symbol;
```
BeksOmega, line 158: This is configurable now to support radians?

plugins/field-angle/src/field_angle.ts
```diff
@@ -209,20 +256,45 @@ export class FieldAngle extends Blockly.FieldNumber {
...
-                (angle % 45 === 0 ? 10 : 5),
+            'x2': FieldAngle.HALF + FieldAngle.RADIUS - length,
             'y2': FieldAngle.HALF,
             'class': 'blocklyAngleMarks',
             'transform': 'rotate(' + angle + ',' + FieldAngle.HALF + ',' +
                 FieldAngle.HALF + ')',
           }, svg);
     }
 
+    if (minorTickAngle) {
```
BeksOmega, line 307: Is this to support disabling the ticks if someone sets the `minorTick` to 0?

plugins/field-angle/src/field_angle.ts
```diff
@@ -304,12 +387,9 @@ export class FieldAngle extends Blockly.FieldNumber {
    * @param angle New angle.
    */
   private displayMouseOrKeyboardValue(angle: number) {
-    if (this.precision_) {
-      angle = Math.round(angle / this.precision_) * this.precision_;
-    }
-    angle = this.wrapValue(angle);
-    if (angle !== this.value_) {
-      this.setEditorValue_(angle);
+    const validAngle = this.doClassValidation_(angle);
```
BeksOmega, line 435: I don't think we want to pass the new value through class validation before setting the editor value. This would mean that if a user entered a value outside the valid min/max, we would update the text input to show something they didn't type (which we generally do not do).

`setEditorValue_` will pass the angle to `setValue` which will pass it through `doClassValidation_`. So the value of the angle field will always be valid anyway.

plugins/field-angle/src/field_angle.ts
```diff
@@ -503,7 +632,11 @@ export interface FieldAngleConfig extends Blockly.FieldNumberConfig {
   mode?: Mode;
   clockwise?: boolean;
   offset?: number;
-  wrap?: number;
+  displayMin?: number;
+  displayMax?: number;
+  minorTick?: number;
+  majorTick?: number;
+  symbol: string;
```
BeksOmega: Symbol should be `?:` as well.

plugins/field-angle/test/index.ts
```diff
@@ -30,10 +30,39 @@ const toolbox = generateFieldTestBlocks('field_angle', [
...
+      'value': 0,
+      'min': 0,
+      'max': 2,
+      'displayMin': 0,
+      'displayMax': 8,
+      'minorTick': 0,
+      'majorTick': 1,
+      'symbol': '',
+    },
+  },
```
BeksOmega: Yay thank you for adding test blocks :D

plugins/field-angle/src/field_angle.ts
```diff
@@ -379,6 +391,34 @@ export class FieldAngle extends Blockly.FieldNumber {
     return angle;
   }
 
+  /**
+   * Convert a value for this field into an on-screen angle.
+   * @param angle Angle value for this field, scaled and offset as specified.
+   * @returns Radians where 0: East, π/2: North, π or -π: West, -π/2: South.
+   */
+  private fieldAngleToRadians(angle: number): number {
```
BeksOmega, line 404: Love separating this out into a helper method - much cleaner =)

BeksOmega (changes requested):
Only one actual change required! (The one related to `doClassValidation_`). Otherwise just questions =)

plugins/field-angle/src/field_angle.ts
NeilFraser: Correct.  Now that 360 isn't hardcoded into the field, we now have to consider radians, turns, gons, zams, etc.

plugins/field-angle/src/field_angle.ts
NeilFraser: Correct.  As it says in the comment: Zero to disable.

plugins/field-angle/src/field_angle.ts
NeilFraser: The overall math remains incomprehensible IMHO.

plugins/field-angle/src/field_angle.ts
NeilFraser: Right.

BeksOmega (approved):

plugins/field-angle/src/field_angle.ts
BeksOmega: We tested it, and for some reason this is necessary, but we're not sure why. Not worth digging into at the moment.

PR #1681, opened 2023-05-04
cleanup(examples): delete parcel example

rachel-fenichel:
Delete the example that uses [Parcel](https://www.npmjs.com/package/parcel).

Parcel is a build tool/bundler. The Blockly parts of this example and the same as any other use of Blockly:
```
import * as Blockly from 'blockly';
import {javascriptGenerator} from 'blockly/javascript';
```

Followed by a standard toolbox definition and workspace injection.

Developers who want to use Parcel should look at the [Parcel documentation](https://parceljs.org/docs/) for up to date instructions and examples of how to use the tool. Developers who want to get started with Blockly should look at the [Getting Started](https://developers.google.com/blockly/guides/get-started/web) page on the Blockly developer site.

Part of https://github.com/google/blockly-samples/issues/1679

PR #1702, opened 2023-05-19
chore: migrate custom generator codelab to using imperative language.

jubishop:


cpcallen (approved):
Just to be clear: my comment on #1681 re: use of imperative language was specifically about making sure that _instructions_ were in the imperative ("now twiddle the knob"); I had no objection to descriptive parts being written in a more informal/personal style ("we don't need to…" /  "later we will…").  Unlike a science lab report there is no requirement that codelabs be written in passive voice, and using a different style of writing for description vs. instructions might actually help with clarity.

However I have no objection to the changes in this PR and presume they adhere to the guidlines I understand you created (though which I am at present unable to locate).

PR #1705, opened 2023-05-22
feat!: updated `plugins/field-date` to TypeScript

btw17:
## Notes

- Recreated the Date plugin from scratch using the plugin creation dev script
- Updated to **use the browser's date input** rather than a text-input with a custom date picker
  - The goal was to deprecate the previous `goog.require` imports, which included the custom date picker
- Updated `package-lock.json` files after `npm run clean:node` and `npm run boot` at the root folder.
  - It updates the lock to include `blockly: ^10.0.0`, matching their corresponding blockly declarations that are already set to that version.

## Changes

1. Uses `TypeScript` now

## Breaking Changes

1. Can no longer customize date picker colors
    1. There are some styles that are available, such as `::-webkit-calendar-picker-indicator`, though they're inconsistent across browsers.
3. Format was updated from ISO (`yyyy-mm-dd`) to the browser default for date input.
    1. The ends up being the locale region in most cases, though this can differ depending on system settings and browser.
    2. As a result of this, there can be a desync between the format of the value displayed in the block vs the value in the input when clicking on the block
 4. Updated the config options:
    1. The `tooltip` config option is now passed to parent components, so it works instead of being ignored!
    2. Set `spellcheck?: never` since it isn't used for `FieldDate`
    3. Removed `textEdit` option. Previously, this allowed using the Date Picker while disabling the text input. This cannot be done with the default browser input, as disabling the input disables the picker associated with it.


plugins/field-date/src/field_date.ts
```diff
@@ -0,0 +1,263 @@
...
+    if (!this.htmlInput_) return;
+    Blockly.utils.dom.addClass(this.htmlInput_!, 'blocklyDateInput');
+
+    // NOTE: HTMLInputElement.showPicker() is not available in earlier
+    // TypeScript versions (like 4.7.4), so casting to `any` to be compatible
+    // with dev scripts. Additionally, it's not available for date inputs for 
+    // Safari. For browser compatibility of showPicker, see:
+    // https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/showPicker
+    /* eslint-disable @typescript-eslint/no-explicit-any */
+    (this.htmlInput_ as any).showPicker();
```
BeksOmega, line 149: You might be able to remove this `any` cast b/c we updated dev scripts.

plugins/field-date/src/field_date.ts
```diff
@@ -0,0 +1,263 @@
...
+/**
+ * Options used to define a field date from JSON.
+ */
+export interface FieldDateFromJsonConfig extends FieldDateConfig {
+  date?: string;
+  // NOTE: spellcheck is defined for FieldInput though irrelevant for FieldDate.
+  // TODO: Confirm 'never' options. Should these be allowed with FieldDate? If
+  // so, how would they be used?
+  spellcheck?: never;
+  tooltip?: never;
```
BeksOmega: Having spellcheck as `never` makes sense, but I don't understand `never`ing tooltip.

plugins/field-date/src/field_date.ts
```diff
@@ -55,18 +55,20 @@ export class FieldDate extends Blockly.FieldTextInput {
   /**
    * Ensures that the input value is a valid date.
    * @param newValue The input value. Ex: '2023-04-28'
```
BeksOmega, line 63: Can you add some additional info here stating that date strings including time zones will be converted ti ISO strings in the user's local time zone? I think this is just slightly surprising (even though it makes sense) so I would like to have it in English as well as code!

plugins/field-date/test/field_date_test.mocha.js
```diff
@@ -97,7 +119,7 @@ suite('FieldDate', function() {
       });
       test('New Value', function() {
```
BeksOmega: I know all of the validater tests already existed, but could you update them to have some more informative names? Not a high priority though if you just want to get this merged!

plugins/field-date/test/field_date_test.mocha.js
```diff
@@ -117,7 +139,112 @@ suite('FieldDate', function() {
...
         this.field.setValue('3030-03-30');
-        assertFieldValue(this.field, '3030-03-30');
+        assertFieldValue(this.field, '3030-03-30', '3/30/3030');
+      });
+    });
+  });
+
+  suite('Time Zones', function() {
```
BeksOmega, line 147: Thank you for adding time zone tests :D

plugins/field-date/README.md
```diff
@@ -49,8 +51,7 @@ Blockly.defineBlocksWithJsonArray([
     }]);
 ```
 
-[View the developer documentation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/date) for further usage examples.
+- [View the developer documentation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/date) for further usage examples.
```
BeksOmega: Does this still conform to the documented API? Even after your changes related to date parsing? My reading is that it does, but wanted to check!

plugins/field-date/package.json
```diff
@@ -41,10 +45,8 @@
     "@blockly/dev-scripts": "^2.0.0",
     "@blockly/dev-tools": "^6.0.0",
     "blockly": "^10.0.0",
-    "google-closure-compiler": "^20200920.0.0",
-    "google-closure-library": "^20200830.0.0",
```
BeksOmega, line 45: Yayay :D

plugins/field-date/src/field_date.ts
```diff
@@ -0,0 +1,262 @@
...
+   * editor.
+   * @param e Optional mouse event that triggered the field to
+   *     open, or undefined if triggered programmatically.
+   * @param _quietInput Quiet input.
+   * @override
+   */
+  protected showEditor_(e?: Event, _quietInput?: boolean) {
+    if (this.sourceBlock_) {
+      // NOTE: Disable modal inputs to use default browser inputs on mobile.
+      this.sourceBlock_.workspace.options.modalInputs = false;
```
BeksOmega: I'm not sure that we want to disable modalinputs for the whole workspace to get the date picker to work. I would rather:

1) Pass `quietInput` as `true` to super, but [focus on it anyway](https://github.com/google/blockly/blob/a677355f7319140f68c13c3468470b0ecd9d25b8/core/field_input.ts#L340C5-L345C6).
2) Tell external devs to disable modal inputs in their workspace config if they want to use this plugin. We could also log a warning from here if we run into enabled `modalInputs`.

plugins/field-date/src/field_date.ts
```diff
@@ -0,0 +1,262 @@
...
+    /* eslint-enable @typescript-eslint/no-explicit-any */
+  }
+
+  /**
+   * Create the html input and set it to type date.
+   *
+   * @returns The newly created date input editor.
+   */
+  protected widgetCreate_(): HTMLInputElement {
+    // NOTE: field_input should return HTMLInputElement for this.
```
BeksOmega, line 159: Can you file a bug to fix this in core? If you're busy I can instead :P

BeksOmega (commented):
So excited to remove the last last vestige of closure library :D

plugins/field-date/src/field_date.ts
btw17: Done! https://github.com/google/blockly/issues/7253

I created it as a feature request since it's not *technically* a bug, it could just have more type info.

plugins/field-date/src/field_date.ts
btw17: Since the TypeScript rule is "^4.3.2 || ^5" and the package-lock for dev-scripts set to `4.7.4`, unfortunately I can't without another update to dev-scripts.

plugins/field-date/src/field_date.ts
btw17: Looks like the previous implementation never passed the config, with tooltip, along. I updated it to do so now, though, and confirmed in manual tests the tooltip appears over the click target.

plugins/field-date/src/field_date.ts
btw17: Done!

plugins/field-date/test/field_date_test.mocha.js
btw17: Done!

plugins/field-date/test/field_date_test.mocha.js
btw17: Sure thing! They seemed important for conveying what I designed the timezone behavior around.

plugins/field-date/README.md
btw17: Looks like the link is no longer available, actually. I'm assuming this was originally in the core repo, though since it's in plugins now it's no longer documented there.

I went ahead and removed this link.

plugins/field-date/src/field_date.ts
btw17: I went ahead and implemented `option 1` of your suggestions. It seemed the least invasive. TBH, I'm not in tune with the use case of needing to set quietInput + avoid focusing on the input, though, so I'm not sure how important that is.

plugins/field-date/test/index.ts
```diff
@@ -10,14 +10,38 @@
...
+  [key: string]: unknown;
+}
+
+const basicConfig: FieldDateFromJsonConfig = {
+  date: '2020-02-20',
+};
+
+const tooltipConfig: FieldDateFromJsonConfig = {
+  date: '2021-03-13',
+  tooltip: 'This date block has a tooltip!',
```
BeksOmega, line 27: Thanks for adding a new test block for this =)

BeksOmega (approved):

PR #1709, opened 2023-06-05
fix(migration): Fix bugs in renaming script

cpcallen:
- Match only whole identifiers; i.e., if renaming `foo` to `baz`, don't rename `foobar` to `bazbar`.
- Embed FIXME comments where use of get/set pairs is needed.
- Add tests for these changes.


BeksOmega (approved):
LGTM!

PR #1716, opened 2023-06-15
feat!: Added a custom serializer to the backpack plugin so that its state is saved (when using JSON serialization):

### LINKED ISSUE #1662: Add serialization of content to backpack
**Category**

 - Plugins

**Component**

workspace-backpack

The backpack is a lot more useful if the contents of the backpack are actually saved between sessions! Then it can really act as a "custom toolbox" of the user's most useful blocks =)

Add a custom serializer to the backpack plugin so that its state is saved (when using JSON serialization): https://developers.google.com/blockly/guides/configure/web/serialization#serializer_hooks

N/A

**Additional context**

Might be necessary for App Inventor upgrade work.

cesarades:
**Issue**
Fixes #1662 

**Solution**
Converted the state of the backpack from XML to JSON and added a custom serializer to save a backpack between sessions

**Breaking changes**
The functions that previously had XML strings as parameters now only support JSON strings. The functions are addItem and addItems.

**Tests**
I tested the serializer by checking to see if the workspace state saved any time I made changes to the backpack. The advanced playground had the issue of loading a session based off of the XML (which is a separate issue to be addressed), so when reloading the page the backpack would not load the backpack contents from the previous session. To test around this, I copied the state of a full backpack and pasted it onto an empty one to see if it would load, and it did.

### DIFF
--- a/plugins/workspace-backpack/src/backpack.js
+++ b/plugins/workspace-backpack/src/backpack.js
@@ -445,13 +445,15 @@ export class Backpack extends Blockly.DragTarget {
    * Converts the provided block into a JSON string.
    * @param {!Blockly.Block} block The block to convert.
    * @returns {string} The JSON object as a string
+   * @private
    */
-  blockToJSONString_(block) {
+  blockToJsonString(block) {
     const json = Blockly.serialization.blocks.save(block);
     // Add a BLOCK parameter so that the flyout can categorize it
     json.kind = 'BLOCK';
-    const jsonString = JSON.stringify(json);
-    return jsonString;
+    // Does not allow the same block to be added to the backpack
+    delete json['id'];
+    return JSON.stringify(json);
   }
 
   /**
@@ -461,8 +463,7 @@ export class Backpack extends Blockly.DragTarget {
    *     provided block.
    */
   containsBlock(block) {
-    const blockJSONString = this.blockToJSONString_(block);
-    return this.contents_.indexOf(blockJSONString) !== -1;
+    return this.contents_.indexOf(this.blockToJsonString(block)) !== -1;
   }
 
   /**
@@ -470,7 +471,7 @@ export class Backpack extends Blockly.DragTarget {
    * @param {!Blockly.Block} block The block to be added to the backpack.
    */
   addBlock(block) {
-    this.addItem(this.blockToJSONString_(block));
+    this.addItem(this.blockToJsonString(block));
   }
 
 
@@ -480,8 +481,7 @@ export class Backpack extends Blockly.DragTarget {
    *     backpack.
    */
   addBlocks(blocks) {
-    const jsonStrings = blocks.map(this.blockToJSONString_);
-    this.addItems(jsonStrings);
+    this.addItems(blocks.map(this.blockToJsonString));
   }
 
 
@@ -490,7 +490,7 @@ export class Backpack extends Blockly.DragTarget {
    * @param {!Blockly.Block} block The block to be removed from the backpack.
    */
   removeBlock(block) {
-    this.removeItem(this.blockToJSONString_(block));
+    this.removeItem(this.blockToJsonString(block));
   }
 
   /**
@@ -835,7 +835,7 @@ class BackpackSerializer {
      * Should be after blocks, procedures, and variables.
      * @type {number}
      */
-    this.priority = Blockly.serialization.priorities.VARIABLES - 10;
+    this.priority = Blockly.serialization.priorities.BLOCKS - 10;
   }
 
   /**
@@ -844,7 +844,6 @@ class BackpackSerializer {
    * @returns {object|undefined} the serialized JSON if present
    */
   save(workspace) {
-    console.trace();
     const componentManager = workspace.getComponentManager();
     const backpack = componentManager.getComponent('backpack');
     return backpack.getContents().map((text) => JSON.parse(text));
@@ -856,7 +855,6 @@ class BackpackSerializer {
    * @param {Blockly.Workspace} workspace the workspace to load into
    */
   load(state, workspace) {
-    console.trace();
     const jsonStrings = state.map((j) => JSON.stringify(j));
     const componentManager = workspace.getComponentManager();
     const backpack = componentManager.getComponent('backpack');
--- a/plugins/workspace-backpack/src/backpack.js
+++ b/plugins/workspace-backpack/src/backpack.js
@@ -442,17 +442,36 @@ export class Backpack extends Blockly.DragTarget {
   }
 
   /**
-   * Converts the provided block into a JSON string.
+   * Converts the provided block into a JSON string and
+   * cleans the JSON of any unnecessary attributes
    * @param {!Blockly.Block} block The block to convert.
    * @returns {string} The JSON object as a string
    * @private
    */
   blockToJsonString(block) {
     const json = Blockly.serialization.blocks.save(block);
-    // Add a BLOCK parameter so that the flyout can categorize it
-    json.kind = 'BLOCK';
-    // Does not allow the same block to be added to the backpack
-    delete json['id'];
+
+    // The keys to remove
+    const keys = ['id', 'height', 'width', 'pinned', 'enabled'];
+
+    // Traverse the JSON recursively
+    const traverseJson = function(json, keys) {
+      for (const key in json) {
+        if (key) {
+          const val = json[key];
+          if (keys.includes(key)) delete json[key];
+
+          // Add a 'kind' key so the flyout can recognize it as a block
+          else if (key == 'type') json.kind = 'BLOCK';
+
+          if (val != null && typeof val == 'object') {
+            traverseJson(json[key], keys);
+          }
+        }
+      }
+    };
+
+    traverseJson(json, keys);
     return JSON.stringify(json);
   }
 
--- a/plugins/workspace-backpack/src/backpack.js
+++ b/plugins/workspace-backpack/src/backpack.js
@@ -445,26 +445,26 @@ export class Backpack extends Blockly.DragTarget {
    * Converts the provided block into a JSON string and
    * cleans the JSON of any unnecessary attributes
    * @param {!Blockly.Block} block The block to convert.
-   * @returns {string} The JSON object as a string
+   * @returns {string} The JSON object as a string.
    * @private
    */
   blockToJsonString(block) {
     const json = Blockly.serialization.blocks.save(block);
 
-    // The keys to remove
+    // Add a 'kind' key so the flyout can recognize it as a block.
+    json.kind = 'BLOCK';
+
+    // The keys to remove.
     const keys = ['id', 'height', 'width', 'pinned', 'enabled'];
 
-    // Traverse the JSON recursively
+    // Traverse the JSON recursively.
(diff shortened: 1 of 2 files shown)


plugins/workspace-backpack/src/backpack.js
```diff
@@ -817,3 +821,59 @@ Blockly.Css.register(`
...
+ */
+class BackpackSerializer {
+  /** Constructs the backpack serializer */
+  constructor() {
+    /**
+     * The priority for deserializing block suggestion data.
+     * Should be after blocks, procedures, and variables.
+     * @type {number}
+     */
+    this.priority = Blockly.serialization.priorities.VARIABLES - 10;
```
BeksOmega: If you want this to be after everything it should be `Blockly.serialization.priorities.BLOCKS - 10`, since blocks are last =)

You can see all of the built-in priority values here: https://github.com/google/blockly/blob/develop/core/serialization/priorities.ts

plugins/workspace-backpack/src/backpack.js
```diff
@@ -817,3 +821,59 @@ Blockly.Css.register(`
...
+  }
+
+  /**
+   * Saves a target workspace's state to serialized JSON.
+   * @param {Blockly.Workspace} workspace the workspace to save
+   * @returns {object|undefined} the serialized JSON if present
+   */
+  save(workspace) {
+    console.trace();
```
BeksOmega: Nit: remove stack traces here and below =)

plugins/workspace-backpack/src/backpack.js
```diff
@@ -440,14 +440,18 @@ export class Backpack extends Blockly.DragTarget {
...
-   * Converts the provided block into a cleaned XML string.
+   * Converts the provided block into a JSON string.
    * @param {!Blockly.Block} block The block to convert.
-   * @returns {string} The cleaned XML string.
-   * @private
+   * @returns {string} The JSON object as a string
    */
-  blockToCleanXmlString_(block) {
-    return cleanBlockXML(Blockly.Xml.blockToDom(block));
+  blockToJSONString_(block) {
```
BeksOmega: nit: The google javascript styleguide suggests `blockToJsonString`

https://google.github.io/styleguide/jsguide.html#naming

You'll also want to mark this `@private` unless you're calling it from a different class!

plugins/workspace-backpack/src/backpack_helpers.js
```diff
@@ -231,51 +231,3 @@ export function registerContextMenus(contextMenuOptions, workspace) {
...
-
-/**
- * Converts XML representing a block into text that can be stored in the
- * content array.
- * @param {!Element} xml An XML tree defining the block and any
- *    connected child blocks.
- * @returns {string} Text representing the XML tree, cleaned of all unnecessary
- * attributes.
- */
-export function cleanBlockXML(xml) {
```
BeksOmega, line 243: This method was actually important for making sure we can compare blocks to see if they're equivalent! It removed a lot of extranenous information that we don't care about in the context of the backpack (e.g. the block's old coordinates, its ID, etc).

You'll want to add this functionality to your `blockToJsonString` method.

plugins/workspace-backpack/src/backpack.js
```diff
@@ -440,14 +440,18 @@ export class Backpack extends Blockly.DragTarget {
...
+    const json = Blockly.serialization.blocks.save(block);
+    // Add a BLOCK parameter so that the flyout can categorize it
+    json.kind = 'BLOCK';
+    const jsonString = JSON.stringify(json);
+    return jsonString;
```
BeksOmega: nit: `return JSON.stringify(json);`

plugins/workspace-backpack/src/backpack.js
```diff
@@ -457,16 +461,16 @@ export class Backpack extends Blockly.DragTarget {
    *     provided block.
    */
   containsBlock(block) {
-    const cleanedBlockXml = this.blockToCleanXmlString_(block);
-    return this.contents_.indexOf(cleanedBlockXml) !== -1;
+    const blockJSONString = this.blockToJSONString_(block);
+    return this.contents_.indexOf(blockJSONString) !== -1;
```
BeksOmega: Can you remove the unnecessary variable assignments (here and below)? I know they predated your change :P but would be nice to clean them up while you're in the area!

plugins/workspace-backpack/src/backpack.js
```diff
@@ -59,7 +59,7 @@ export class Backpack extends Blockly.DragTarget {
     this.flyout_ = null;
 
     /**
-     * A list of XML (stored as strings) representing blocks in the backpack.
+     * A list of JSON (stored as strings) representing blocks in the backpack.
```
BeksOmega, line 62: Thank you for all of the lovely doc updates :D

plugins/workspace-backpack/src/backpack.js
```diff
@@ -817,3 +821,59 @@ Blockly.Css.register(`
   opacity: 0.8;
 }
 `);
+
+
+/**
+ * Custom serializer so that the backpack can save and later recall which
+ * blocks have been saved in a workspace.
+ */
+class BackpackSerializer {
```
BeksOmega, line 849: This serializer implementation is beautiful =)

BeksOmega (changes requested):
General fix LGTM! Thank you for digging into this when it required more changes than we first expected :D

A few comments about formatting/style and then one major one about cleaning the JSON representation of the blocks. Then I'll give this a re-review!

BeksOmega:
Ope one other thing! Since this is a breaking change, you'll need to change the commit/PR title to reflect that.

It should be prefixed `feat!: <your current description>` And you'll want to add a "Breaking changes"  section to the PR description to tell folks what they'll need to update =)

plugins/workspace-backpack/src/backpack.js
```diff
@@ -440,14 +440,20 @@ export class Backpack extends Blockly.DragTarget {
...
    * @private
    */
-  blockToCleanXmlString_(block) {
-    return cleanBlockXML(Blockly.Xml.blockToDom(block));
+  blockToJsonString(block) {
+    const json = Blockly.serialization.blocks.save(block);
+    // Add a BLOCK parameter so that the flyout can categorize it
+    json.kind = 'BLOCK';
+    // Does not allow the same block to be added to the backpack
+    delete json['id'];
```
BeksOmega: Removing the ID looks good! But I think the `blockToJsonString` method still needs two fixes:

1) It should also strip out any of the extraneous info about comments (`width`, `height`, and `pinned`).
2) The IDs & comment data need to be recursively removed from all child blocks as well as the top-level block. So you'll need to do a bit of tree traversing :P

[Edit: Disabled attributes should also be stripped]

BeksOmega (changes requested):
Thanks for the fixes :D I think the cleanup still needs a bit of cleaning up! but otherwise lookin good =)

plugins/workspace-backpack/src/backpack.js
```diff
@@ -442,17 +442,36 @@ export class Backpack extends Blockly.DragTarget {
...
    */
   blockToJsonString(block) {
     const json = Blockly.serialization.blocks.save(block);
-    // Add a BLOCK parameter so that the flyout can categorize it
-    json.kind = 'BLOCK';
-    // Does not allow the same block to be added to the backpack
-    delete json['id'];
+
+    // The keys to remove
+    const keys = ['id', 'height', 'width', 'pinned', 'enabled'];
```
BeksOmega, line 458: Love the addition of the keys! Very elegant =)

plugins/workspace-backpack/src/backpack.js
```diff
@@ -442,17 +442,36 @@ export class Backpack extends Blockly.DragTarget {
...
    * @private
    */
   blockToJsonString(block) {
     const json = Blockly.serialization.blocks.save(block);
-    // Add a BLOCK parameter so that the flyout can categorize it
-    json.kind = 'BLOCK';
-    // Does not allow the same block to be added to the backpack
-    delete json['id'];
+
+    // The keys to remove
```
BeksOmega: Nit: comments should end w/ a period. (here and below) as per our style guide:
https://developers.google.com/blockly/guides/contribute/core/style_guide#do

plugins/workspace-backpack/src/backpack.js
```diff
@@ -442,17 +442,36 @@ export class Backpack extends Blockly.DragTarget {
...
+
+    // Traverse the JSON recursively
+    const traverseJson = function(json, keys) {
+      for (const key in json) {
+        if (key) {
+          const val = json[key];
+          if (keys.includes(key)) delete json[key];
+
+          // Add a 'kind' key so the flyout can recognize it as a block
+          else if (key == 'type') json.kind = 'BLOCK';
```
BeksOmega: Multiline if-statements should have braces as per the style guide: https://google.github.io/styleguide/jsguide.html#formatting-braces

This makes it easier to refactor in the future without breaking things!

---

Also, while you're in the area, the 'kind' property does only need to be added at the top level! So you can break this out of the tree walk.

plugins/workspace-backpack/src/backpack.js
```diff
@@ -440,14 +440,39 @@ export class Backpack extends Blockly.DragTarget {
...
+
+          if (val != null && typeof val == 'object') {
```
BeksOmega: Nit: use strict equality checks as per the style guide:
https://google.github.io/styleguide/jsguide.html#features-equality-checks

(for the first part of the condition you can just do `val` instead of `val != null`

plugins/workspace-backpack/src/backpack.js
```diff
@@ -440,14 +440,39 @@ export class Backpack extends Blockly.DragTarget {
...
-  blockToCleanXmlString_(block) {
-    return cleanBlockXML(Blockly.Xml.blockToDom(block));
+  blockToJsonString(block) {
+    const json = Blockly.serialization.blocks.save(block);
+
+    // The keys to remove
+    const keys = ['id', 'height', 'width', 'pinned', 'enabled'];
+
+    // Traverse the JSON recursively
+    const traverseJson = function(json, keys) {
```
BeksOmega, line 461: Love the local function definition =) The fun of working in modern languages! (if JS still counts :P)

BeksOmega (changes requested):
Just a few nits and then this LGTM =)

plugins/workspace-backpack/src/backpack.js
```diff
@@ -440,14 +440,39 @@ export class Backpack extends Blockly.DragTarget {
...
+    const keys = ['id', 'height', 'width', 'pinned', 'enabled'];
+
+    // Traverse the JSON recursively.
+    const traverseJson = function(json, keys) {
+      for (const key in json) {
+        if (key) {
+          if (keys.includes(key)) {
+            delete json[key];
+          }
+          if (json[key] && typeof json[key] == 'object') {
```
BeksOmega:
```suggestion
          if (json[key] && typeof json[key] === 'object') {
```

Just one last fix! You should be able to commit this from the github UI.

BeksOmega (approved):
Wooot! LGTM :D Thanks for going back and forth with me on those changes!

PR #1719, opened 2023-06-17
chore: Migrate dynamic connections plugin to TypeScript.

johnnesky:
Addresses one of the plugins in: https://github.com/google/blockly-samples/issues/1647

This plugin involves monkey patching `Blockly.InsertionMarkerManager`. In particular, it tries to access private methods, which is generally not allowed by the TypeScript compiler. I did not attempt to change the logic, instead I used the array index syntax to force TypeScript to allow accessing private methods. Interestingly, the private methods don't have any type information in the exported d.ts files so as far as the TypeScript compiler is concerned the private methods have type `any` in this context.

This plugin defines blocks by assigning objects with methods to the `Blockly.Blocks` array. I added type information for the implicit `this` parameter to the methods by explicitly defining the parameter, although I did not add JSDoc for the `this` parameter, which confuses the linter a bit. It emits warnings about missing annotations for the `this` parameters, which I chose to ignore.


plugins/block-dynamic-connection/src/dynamic_if.ts
```diff
@@ -11,25 +11,20 @@
 
 import * as Blockly from 'blockly/core';
 
+type DynamicIfBlock = Blockly.Block | (typeof Blockly.Blocks)['dynamic_if'];
```
cpcallen: I don't think it's strictly correct: really it should be an intersection, not a union:
```suggestion
type DynamicIfBlock = Blockly.Block & typeof Blockly.Blocks['dynamic_if'];
```
but I think this is still not correct: the type the values of the `Blockly.Blocks` dictionary is `BlockDefinition`, which is just an alias for `any`, so this line just means `type DynamicIfBlock = any`.

This would be why `tsc` is not complaining about (e.g.) your calling `this.getInput` despite `.getInput` not existing on all members of the union (in particular, `Blockly.Blocks['dynamic_if']` does not declare a `.getInput` property).

Unfortunately when I tried getting this right in core (in `blocks/*.ts`) I had to split the code defining the block out from the code installation in the block dictionary, and using [this ~~triple-headed monster~~ three-line monstrosity](https://github.com/google/blockly/blob/5f2e143ab8821ad54331cbba231d31a834ea5c8f/blocks/math.ts#L433) to work around errors due to (alleged) circular definitions.

plugins/block-dynamic-connection/src/dynamic_if.ts
```diff
@@ -11,25 +11,20 @@
...
+  /** Minimum number of inputs for this block. */
   minInputs: 1,
 
   /**
    * Block for if/elseif/else statements. Must have one if input.
    * Can have any number of elseif inputs and optionally one else input.
-   * @this {Blockly.Block}
    */
-  init: function() {
+  init: function(this: DynamicIfBlock): void {
```
cpcallen: Curiously we don't get these lint warnings in core.  Can you see why we get them here and, if due to an intentional configuration difference we don't want to rectify, add suitable `eslint-disable <appropriate rule>` directives?  (Since this is a common case, turning it off at the top of the file is probably reasonable.)

(Here and various places below.)

plugins/block-dynamic-connection/src/dynamic_if.ts
```diff
@@ -109,18 +102,17 @@ Blockly.Blocks['dynamic_if'] = {
...
-   * @param {!Element} xmlElement XML storage element.
-   * @this {Blockly.Block}
+   * @param xmlElement XML storage element.
    */
-  deserializeCounts_: function(xmlElement) {
-    const elseifCount = parseInt(xmlElement.getAttribute('elseif'), 10) || 0;
-    const elseCount = parseInt(xmlElement.getAttribute('else'), 10) || 0;
+  deserializeCounts: function(this: DynamicIfBlock, xmlElement: Element): void {
+    const elseifCount = parseInt(xmlElement.getAttribute('elseif')!, 10) || 0;
+    const elseCount = parseInt(xmlElement.getAttribute('else')!, 10) || 0;
```
cpcallen: Unless you're planning a second pass to improve typing (as I am for core `blocks/`—see google/blockly#6920) then this lint warning should be fixed.

Might be worth discussing the return-value-of-`getAttribute` situation in team chat, since this is a very common situation where a method might return `null` and it might be good to see if there is (or should be) an established convention for how we deal with it, but in this case it would seem reasonable to me to supply the default value as a string to parseInt:
```suggestion
    const elseifCount = parseInt(xmlElement.getAttribute('elseif') ?? '0', 10) || 0;
    const elseCount = parseInt(xmlElement.getAttribute('else') ?? '0', 10) || 0;
```
(Note that you still need to supply the default as a number in case `parseInt` returns `NaN`, and you still have to use `||` for this since `NaN ?? x` returns `NaN`, not `x`.  Alas `parseInt(-0, 10) || 0` returns `0`, not `-0`, but I guess we can live with that.)

Alternatively, you  could an an explicit check, which is arguably safer / more defensive but would be a breaking change:
```suggestion
    const elseifAttribute = xmlElement.getAttribute('elseif');
    const elseAttribute = xmlElement.getAttribute('else');
    if (elseIfAttribute === null) throw new TypeError(/*...*/);
    if (elseAttribute === null) throw new TypeError(/*...*/);
    const elseifCount = parseInt(elseIfAttribute, 10) || 0;
    const elseCount = parseInt(elseAttribute, 10) || 0;
```

plugins/block-dynamic-connection/src/dynamic_if.ts
```diff
@@ -168,11 +161,13 @@ Blockly.Blocks['dynamic_if'] = {
   },
 
   /**
-   * Called when a block is dragged over one of the connections on this block.
-   * @param {!Blockly.Connection} connection The connection on this block that
-   * has a pending connection.
+   * Called by a monkey-patched version of InsertionMarkerManager when a block
+   * is dragged over one of the connections on this block.
```
cpcallen: An observation: in a nod to [usenet netiquette](https://www.ietf.org/rfc/rfc1855.txt#:~:text=%2D%20Limit%20line%20length%20to%20fewer%20than%2065%20characters%20and%20end%20a%20line%0A%20%20%20%20%20%20with%20a%20carriage%20return.) I have emacs configured to wrap block comments at 70 columns, so that it leaves a little room in case a future edit lengthens one of the lines slightly (i.e., reduces the likelihood of needing to reflow them).  I'd certainly make exceptions where doing so lets me save a line (especially if I can fit it all on one line), but keeping lines roughly balanced in places like this seems like good practice:
```suggestion
   * Called by a monkey-patched version of InsertionMarkerManager when
   *  a block is dragged over one of the connections on this block.
```

plugins/block-dynamic-connection/src/dynamic_list_create.ts
```diff
@@ -11,24 +11,18 @@
 
 import * as Blockly from 'blockly/core';
 
+type DynamicListCreateBlock =
+    Blockly.Block | (typeof Blockly.Blocks)['dynamic_list_create'];
```
cpcallen: See comment for `DynamicIfBlock`.

plugins/block-dynamic-connection/src/dynamic_if.ts
cpcallen: I think you can use method syntax here (and many places below):
```suggestion
  init(this: DynamicIfBlock): void {
```
(Note that I did not do that in my first pass on core `blocks/`, but I'm intending to do so when I go back to tidy things up.)

plugins/block-dynamic-connection/src/dynamic_list_create.ts
```diff
@@ -78,18 +72,18 @@ Blockly.Blocks['dynamic_list_create'] = {
       this.inputList[0]
           .appendField(Blockly.Msg['LISTS_CREATE_WITH_INPUT_WITH']);
     }
-    const next = parseInt(xmlElement.getAttribute('next'));
+    const next = parseInt(xmlElement.getAttribute('next')!);
```
cpcallen: Why doesn't eslint complain about this `!`?

Though in any case: see comment above about `getAttribute`/`parseInt`

(And several places below.)

plugins/block-dynamic-connection/src/dynamic_if.ts
```diff
@@ -109,18 +102,17 @@ Blockly.Blocks['dynamic_if'] = {
       this.appendStatementInput('ELSE')
           .appendField(Blockly.Msg['CONTROLS_IF_MSG_ELSE'], 'else');
     }
-    const next = parseInt(xmlElement.getAttribute('next'));
+    const next = parseInt(xmlElement.getAttribute('next')!);
```
cpcallen: In addition to the non-null assertion lint, I think this should probably have a radix specified (and perhaps also default values; see earlier comment about `.getAttribute`/`parseInt`):
```suggestion
    const next = parseInt(xmlElement.getAttribute('next')!, 10);
```

plugins/block-dynamic-connection/src/dynamic_list_create.ts
```diff
@@ -78,18 +72,18 @@ Blockly.Blocks['dynamic_list_create'] = {
...
-   * @param {!Element} xmlElement XML storage element.
-   * @this {Blockly.Block}
+   * @param xmlElement XML storage element.
    */
-  deserializeCounts_: function(xmlElement) {
+  deserializeCounts: function(
+      this: DynamicListCreateBlock, xmlElement: Element): void {
     const itemCount = Math.max(
-        parseInt(xmlElement.getAttribute('items'), 10), this.minInputs);
+        parseInt(xmlElement.getAttribute('items')!, 10), this.minInputs);
```
cpcallen: Ditto here.

cpcallen (changes requested):
> This plugin involves monkey patching `Blockly.InsertionMarkerManager`. In particular, it tries to access private methods, which is generally not allowed by the TypeScript compiler. I did not attempt to change the logic, instead I used the array index syntax to force TypeScript to allow accessing private methods. Interestingly, the private methods don't have any type information in the exported d.ts files so as far as the TypeScript compiler is concerned the private methods have type `any` in this context.

That seems reasonable, but please open a bug in the blockly repo to add facilities to support this plugin (you don't need to propose a design, just note that we need to come up with one), and another in this repo to refactor this code to use those facilities once provided.

> This plugin defines blocks by assigning objects with methods to the `Blockly.Blocks` array. I added type information for the implicit `this` parameter to the methods by explicitly defining the parameter, although I did not add JSDoc for the `this` parameter, which confuses the linter a bit. It emits warnings about missing annotations for the `this` parameters, which I chose to ignore.

See comments below both about the type of `this` and also about the linter warnings, which should variously be fixed (where avoidable) or suppressed (where not).

johnnesky:
I created an issue on Blockly to provide better support for this plugin: https://github.com/google/blockly/issues/7204

plugins/block-dynamic-connection/src/dynamic_if.ts
johnnesky: Done, I imitated your 3 line monstrosity, thanks!

plugins/block-dynamic-connection/src/dynamic_list_create.ts
johnnesky: Done.

plugins/block-dynamic-connection/src/dynamic_if.ts
johnnesky: Done.

plugins/block-dynamic-connection/src/dynamic_if.ts
johnnesky: Done.

plugins/block-dynamic-connection/src/dynamic_if.ts
johnnesky: Done, I went with providing default values.

plugins/block-dynamic-connection/src/dynamic_if.ts
johnnesky: Done.

johnnesky:
Should I also add some variable type annotations that the TypeScript compiler is able to figure out but might not be immediately obvious to the human reader?

plugins/block-dynamic-connection/src/dynamic_if.ts
johnnesky: Looks like the blockly-samples eslint configuration is here: https://github.com/google/blockly-samples/blob/master/plugins/eslint-config/index.js

It extends `'eslint:recommended'`, which turns on `'jsdoc/require-param'` which is the source of our lint warning and is documented here:
https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-param.md

That documentation includes a code snippet demonstrating including `this` in the jsdoc params, and no other reference to `this`, so I guess that's what we're expected to do:
image
EDIT: Actually one of those examples in the snippet does NOT add `this to the jsdoc params, now I'm even more confused.

Meanwhile, the Blockly core repo seems to be using this eslint configuration: https://github.com/google/eslint-config-google/blob/master/index.js

Which does not extend anything as far as I can tell, and does not explicitly enable `jsdoc/require-param` although it looks like it does enable `jsdoc/require-param-description` or something like it.

EDIT: Aha, eslint was changed 8 months ago to make the this param optional, we're probably using an old version: 
image
Yep, the version of eslint-plugin-jsdoc in my node_modules in 37.9.7, and the fix is in: https://github.com/gajus/eslint-plugin-jsdoc/releases/tag/v39.3.16
I'm not sure what the best way to fix that is.

johnnesky: Maribeth suggested I update the installed version of eslint-plugin-jsdoc. I created https://github.com/google/blockly-samples/pull/1740 for that.

plugins/block-dynamic-connection/src/dynamic_if.ts
```diff
@@ -11,9 +11,15 @@
...
-Blockly.Blocks['dynamic_if'] = {
+/** Type of a block that has the dynamicIfMixin. */
+type DynamicIfBlock = Blockly.Block & DynamicIfMixin;
+/* eslint-disable @typescript-eslint/no-empty-interface */
+/** This interface avoids a "circular reference" compile error. */
+interface DynamicIfMixin extends DynamicIfMixinType {}
+/* eslint-enable @typescript-eslint/no-empty-interface */
+type DynamicIfMixinType = typeof dynamicIfMixin;
+
+const dynamicIfMixin = {
```
cpcallen: In core I've used `CONSTANT_CASE` for these mixins [because they're not supposed to be mutated](go/tsjs-style#identifiers-constants).  I don't know if there are other named block mixins in samples; if there are then it probably is best to be consistent with them, but if not then I recommend using the same convention as core.

plugins/block-dynamic-connection/src/dynamic_if.ts
cpcallen: Try merging master into this branch to get the updated version; if that is not sufficient to quell these warnings then I would suggest disabling that rule in the section of this file that uses `this:` declarations.

plugins/block-dynamic-connection/src/dynamic_if.ts
```diff
@@ -198,14 +206,16 @@ Blockly.Blocks['dynamic_if'] = {
...
     // Remove Else If inputs if neither the if nor the do has a connected block.
     for (let i = 2; i < this.inputList.length - 1; i += 2) {
       const ifConnection = this.inputList[i];
       const doConnection = this.inputList[i + 1];
-      if (!ifConnection.connection.targetConnection &&
-          !doConnection.connection.targetConnection) {
+      if ((!ifConnection.connection ||
+              !ifConnection.connection.targetConnection) &&
+          (!doConnection.connection ||
+              !doConnection.connection.targetConnection)) {
```
cpcallen: Hmm.  Why not use `?.`—e.g. `if(!ifConnection.connection?.targetConnection && …`?

I think that should be equivalent (but maybe double-check that I am not confused!)

(Here and several places below where you have added `||` clauses.)

plugins/block-dynamic-connection/src/dynamic_list_create.ts
```diff
@@ -11,18 +11,23 @@
...
-Blockly.Blocks['dynamic_list_create'] = {
+/** Type of a block that has the dynamicListCreateMixin. */
+type DynamicListCreateBlock = Blockly.Block & DynamicListCreateMixin;
+/* eslint-disable @typescript-eslint/no-empty-interface */
+/** This interface avoids a "circular reference" compile error. */
+interface DynamicListCreateMixin extends DynamicListCreateMixinType {}
+/* eslint-enable @typescript-eslint/no-empty-interface */
+type DynamicListCreateMixinType = typeof dynamicListCreateMixin;
+
+const dynamicListCreateMixin = {
```
cpcallen: See comment above re: `CONSTANT_CASE`.

plugins/block-dynamic-connection/src/dynamic_list_create.ts
```diff
@@ -119,8 +121,12 @@ Blockly.Blocks['dynamic_list_create'] = {
     }
 
     const nextInput = this.inputList[connectionIndex + 1];
-    const nextConnection = nextInput && nextInput.connection.targetConnection;
-    if (nextConnection && !nextConnection.sourceBlock_.isInsertionMarker()) {
+    const nextConnection =
+        nextInput &&
+        nextInput.connection &&
+        nextInput.connection.targetConnection;
```
cpcallen:
```suggestion
    const nextConnection = nextInput?.connection?.targetConnection;
```

plugins/block-dynamic-connection/src/dynamic_list_create.ts
```diff
@@ -148,12 +154,11 @@ Blockly.Blocks['dynamic_list_create'] = {
...
    * drag ends if the dragged block had a pending connection with this block.
    */
-  finalizeConnections: function(this: DynamicListCreateBlock): void {
+  finalizeConnections(this: DynamicListCreateBlock): void {
     if (this.inputList.length > this.minInputs) {
       let toRemove: string[] = [];
       this.inputList.forEach((input: Blockly.Input) => {
-        const targetConnection = input.connection!.targetConnection;
-        if (!targetConnection) {
+        if (!input.connection || !input.connection.targetConnection) {
```
cpcallen: `?.`

plugins/block-dynamic-connection/src/dynamic_text_join.ts
```diff
@@ -11,18 +11,23 @@
...
-Blockly.Blocks['dynamic_text_join'] = {
+/** Type of a block that has the dynamicTextJoinMixin. */
+type DynamicTextJoinBlock = Blockly.Block & DynamicTextJoinMixin;
+/* eslint-disable @typescript-eslint/no-empty-interface */
+/** This interface avoids a "circular reference" compile error. */
+interface DynamicTextJoinMixin extends DynamicTextJoinMixinType {}
+/* eslint-enable @typescript-eslint/no-empty-interface */
+type DynamicTextJoinMixinType = typeof dynamicTextJoinMixin;
+
+const dynamicTextJoinMixin = {
```
cpcallen: `CONSTANT_CASE`?

plugins/block-dynamic-connection/src/dynamic_text_join.ts
```diff
@@ -119,8 +121,12 @@ Blockly.Blocks['dynamic_text_join'] = {
...
 
     const nextInput = this.inputList[connectionIndex + 1];
-    const nextConnection = nextInput && nextInput.connection.targetConnection;
-    if (nextConnection && !nextConnection.sourceBlock_.isInsertionMarker()) {
+    const nextConnection =
+        nextInput &&
+        nextInput.connection &&
+        nextInput.connection.targetConnection;
+    if (nextConnection &&
+        !nextConnection.getSourceBlock().isInsertionMarker()) {
```
cpcallen: `?.`

plugins/block-dynamic-connection/src/dynamic_text_join.ts
```diff
@@ -148,12 +154,11 @@ Blockly.Blocks['dynamic_text_join'] = {
...
    * drag ends if the dragged block had a pending connection with this block.
    */
-  finalizeConnections: function(this: DynamicTextJoinBlock): void {
+  finalizeConnections(this: DynamicTextJoinBlock): void {
     if (this.inputList.length > this.minInputs) {
       let toRemove: string[] = [];
       this.inputList.forEach((input: Blockly.Input) => {
-        const targetConnection = input.connection!.targetConnection;
-        if (!targetConnection) {
+        if (!input.connection || !input.connection.targetConnection) {
```
cpcallen: `?.`

plugins/block-dynamic-connection/src/dynamic_if.ts
cpcallen: Additionally: I'm puzzled about why these warnings seem to be generated for only one of the at least three files that has `this:` declarations.

cpcallen (approved):
LGTM but see below for a few nits.  Let me know when you're happy for it to be merged.

plugins/block-dynamic-connection/src/dynamic_if.ts
johnnesky: Done.

plugins/block-dynamic-connection/src/dynamic_if.ts
johnnesky: Done.

plugins/block-dynamic-connection/src/dynamic_list_create.ts
johnnesky: Done.

johnnesky:
Thanks for the review, I updated the PR! 

Maribeth walked me through getting the lint thing fixed, we need to update another dependency but I'll do that in another PR, for the meantime I do get the same lint warnings on all three files locally.

johnnesky:
https://github.com/google/blockly-samples/pull/1753 should hopefully fix the lint warnings.

PR #1672, opened 2023-04-21
fix: Fix compatibility with Node 18

gonfunko:
This PR fixes compatibility with Node 18. [eslint-loader](https://www.npmjs.com/package/eslint-loader) was deprecated, and never updated for the SSL-related changes in Node 18. This PR replaces eslint-loader with eslint-webpack-plugin, and fixes #1512.

PR #1723, opened 2023-06-21
fix: put eslint cache files in a better spot

maribethb:
Lately when you run `npm run test` there's a bunch of `.eslintcache` files that appear.

They appear now because in #1672 we updated from eslint-loader to eslint-webpack-plugin. eslint-loader was nice enough to put the cache in a [fairly standard](https://github.com/eslint/eslint/issues/11757#issuecomment-496945984) location, hidden away in `node_modules`. Unfortunately eslint itself doesn't do that and just puts them in the main directory, which is pretty unfortunate :/

so this PR updates to use a `node_modules/.cache/eslint` directory (which is already ignored by gitignore) so we don't have to look at the thing

also enables caching when running `npm run lint` because I tried that first before I realized the problem was in the webpack config.

The alternative is to just gitignore the `.eslintcache` file, but I'm really tired of seeing them tbh so I didn't want to keep them in each plugin's root level. 

(in the second commit I realized I don't need to create the directory, just need to pass a location that ends in the OS-specific path separator)

### DIFF
--- a/plugins/dev-scripts/config/webpack.config.js
+++ b/plugins/dev-scripts/config/webpack.config.js
@@ -20,12 +20,6 @@ const packageJson = require(resolveApp('package.json'));
 
 const ESLintPlugin = require('eslint-webpack-plugin');
 
-// Create the eslint cache directory if it doesn't exist.
-const eslintCacheLocation = path.join('node_modules', '.cache', '.eslint');
-if (!fs.existsSync(eslintCacheLocation)) {
-  fs.mkdirSync(eslintCacheLocation, {recursive: true});
-}
-
 module.exports = (env) => {
   const mode = env.mode;
   const isDevelopment = mode === 'development';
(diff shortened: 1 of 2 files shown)


plugins/dev-scripts/config/webpack.config.js
```diff
@@ -120,6 +120,7 @@ module.exports = (env) => {
       // Run the linter.
       !env.skipLint && new ESLintPlugin({
         cache: true,
+        cacheLocation: path.join('node_modules/.cache/eslint/'),
```
BeksOmega, line 123: Isn't the more standard way to do this to pass each directory as a different parameter? E.g:

```suggestion
        cacheLocation: path.join('node_modules', '.cache', 'eslint'),
```

maribethb: You can do that too, but you don't need to. It will automatically normalize the path segments and switch them to the appropriate operating-system-specific separator even if they're in one string. So I didn't see a reason to split them up; that makes more sense when parts of the path are variables as opposed to string literals.

PR #1725, opened 2023-06-22
fix: continuous flyout calling wrong private method

BeksOmega:
The continous flyout is overriding the entire `position` function to modify one value (as documented in its JSDoc). This means it has to call a private function, which is less than ideal.

I'm going to file a bug in core to fix that, but for now this makes it so the plugin continues to work with v10.

[Edit: found google/blockly/issues/4396 just after posting, I'm going to update the plugin to use the better API]

### DIFF
--- a/plugins/continuous-toolbox/src/ContinuousFlyout.js
+++ b/plugins/continuous-toolbox/src/ContinuousFlyout.js
@@ -190,69 +190,52 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
     return 0;
   }
 
-  /**
-   * Overrides the position function solely to change the x coord in RTL mode.
-   * The base function allows the workspace to go "under" the flyout, so
-   * to calculate the left edge of the flyout in RTL you would just subtract
-   * the flyout width from the total viewWidth to get x. However, in our
-   * flyout, the workspace already starts at the left edge of the flyout, so
-   * we don't need to subtract the flyout width again.
-   * Ideally there would be a smaller method for us to override instead,
-   * but for now we copy/paste this method and make our fixes.
-   * @override
-   */
-  position() {
+  /** @override */
+  getX() {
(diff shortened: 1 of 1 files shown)


plugins/continuous-toolbox/src/ContinuousFlyout.js
```diff
@@ -190,69 +190,52 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
...
+          x = toolboxMetrics.width;
         } else {
-          // TODO(https://github.com/google/blockly/issues/4396): Use a better
-          // API to adjust this value.
-          // This is the only line that changed from the original.
-          x = targetWorkspaceMetrics.viewWidth;
+          // This is the one thing that has been changed from the base
+          // implementation, to account for RTL. This makes it so blocks cannot
+          // go under the flyout in RTL mode.
+          x = viewMetrics.width;
```
cpcallen: If this is the only thing that is changed, couldn't the function be rewritten as:

```JS
  getX() {
    if (this.isVisible() &&
        this.targetWorkspace.toolboxPosition === this.toolboxPosition_ &&
        this.targetWorkspace.getToolbox() &&
        this.toolboxPosition_ !== Blockly.utils.toolbox.Position.LEFT) {
      // This is the one thing that has been changed from the base
      // implementation, to account for RTL. This makes it so blocks cannot
      // go under the flyout in RTL mode.
      return viewMetrics.width;
    }
    return super.getX();
}
```

(I'm not certain whether the call to `getToolbox()` is strictly necessary, but have preserved it for compatibility.)

That would seem to be at least a little more robust against getting out of sync with changes to the superclass implementation.

BeksOmega: Yep that's much better!

cpcallen (approved):

PR #1726, opened 2023-06-22
chore: update ci to drop node 14 and use node 18 and 20

maribethb:
Drop node 14, add testing for node 18 and 20


maribethb:
FYI if this passes on master, I'll then open another PR to merge master into the q3 branch.

BeksOmega (approved):

PR #1734, opened 2023-06-22
fix: update all usages of generators in script tags for blockly v10

maribethb:
Updated the usage of generators in all examples and codelabs that load Blockly from script tags, except graph demo (done in #1732) and the devsite demo page (which does something slightly different involving dropdowns so I want to test separately).

Generally,
`Blockly.JavaScript` -> `javascript.javascriptGenerator`
`Blockly.JavaScript['block_name']` -> `javascript.javascriptGenerator.forBlock['block_name']`
`Blockly.JavaScript.ORDER_FOO` -> `javascript.Order.FOO`

I tested each of the example pages, and loaded the codelabs complete code page (after locally updating to point to blockly@beta)

Also updated the README of the hsv-colour-sliders plugin which still used old style generator names with import statements, which definitely doesn't work anymore.

N.B. I don't think this change should have been strictly required, because Blockly.JavaScript is supposed to remain backwards-compatible but deprecated for Blockly v10. I think the wrong name is used in https://github.com/google/blockly/pull/7169 and I'm following up with that.

### DIFF
--- a/examples/interpreter-demo/wait_block.js
+++ b/examples/interpreter-demo/wait_block.js
@@ -13,25 +13,25 @@
  */
 
 Blockly.defineBlocksWithJsonArray([{
-  "type": "wait_seconds",
-  "message0": " wait %1 seconds",
-  "args0": [{
-    "type": "field_number",
-    "name": "SECONDS",
-    "min": 0,
-    "max": 600,
-    "value": 1,
+  'type': 'wait_seconds',
+  'message0': ' wait %1 seconds',
+  'args0': [{
+    'type': 'field_number',
+    'name': 'SECONDS',
+    'min': 0,
+    'max': 600,
+    'value': 1,
   }],
-  "previousStatement": null,
-  "nextStatement": null,
-  "colour": "%{BKY_LOOPS_HUE}",
+  'previousStatement': null,
+  'nextStatement': null,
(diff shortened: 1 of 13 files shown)


examples/interpreter-demo/wait_block.js
```diff
@@ -13,25 +13,25 @@
...
+    'min': 0,
+    'max': 600,
+    'value': 1,
   }],
-  "previousStatement": null,
-  "nextStatement": null,
-  "colour": "%{BKY_LOOPS_HUE}",
+  'previousStatement': null,
+  'nextStatement': null,
+  'colour': '%{BKY_LOOPS_HUE}',
```
cpcallen, line 27: Unless this file gets passed through Closure Compiler in `ADVANCED_OPTIMIZATIONS` mode the quotes on the keys here are not required.

cpcallen (approved):
> I think the wrong name is used in [google/blockly#7169](https://github.com/google/blockly/pull/7169) and I'm following up with that.

Indeed the wrong name was used in that PR; thanks for catching that!

examples/interpreter-demo/wait_block.js
maribethb: We quote tons of stuff that doesn't need to be, all over the place :/ these were only updated to single quotes because i format my files on save

PR #1737, opened 2023-06-22
chore: remove serialize disabled interactions demo

BeksOmega:
Removes the serialize disabled interactions demo since this plugin is deprecated.

### DIFF
--- a/gh-pages/index.md
+++ b/gh-pages/index.md
@@ -30,7 +30,6 @@
 * [Zoom to Fit](plugins/zoom-to-fit/test/index.html)
 * [Keyboard Navigation](plugins/keyboard-navigation/test/index.html)
 * [Scroll Options](plugins/scroll-options/test/index.html)
-* [Serialize Disabled Interactions](plugins/serialize-disabled-interactions/test/index.html)
 * [Shadow Block Converter](plugins/shadow-block-converter/test/index.html)
 * [Suggested Blocks](plugins/suggested-blocks/test/index.html)
 * [Cross Tab Copy Paste](plugins/cross-tab-copy-paste/test/index.html)


cpcallen (commented):
Did you intend to remove the demo, or just the TOC entry for it?

BeksOmega:
> Did you intend to remove the demo, or just the TOC entry for it?

The TOC entry controls what gets uploaded to github pages, so no one will be able to access the demo anymore. (At least that's my understanding). The rest of the plugin will be deleted at some later point once the downloads have decreased!

cpcallen:
> The TOC entry controls what gets uploaded to github pages, so no one will be able to access the demo anymore. (At least that's my understanding). The rest of the plugin will be deleted at some later point once the downloads have decreased!

I have had to revise this reply several.  What I knew to start with:

* GitHub Pages will server everything from the `gh-pages` branch, provided it includes a `.nojekyll` file in the root of that branch.  (Absent that file it will not serve anything in `node_modules/` or `vendor/`, and will presumably pass some of what it does serve through Jekyll.)
* We use the `gh-pages` NPM package to update the `gh-pages` (but only in -samples; in the core repo we update that branch directly).

What I have learned trying understand whether your claim was true or not:

* Unlike in samples, the `gh-pages` branch is updated to contain only the contents of the `gh_pages/` directory.  (This is presumably why we use the `gh-pages` package; doing this with `git` + fileutils might be a bit of a faff.)
* `gh-pages/` has `plugins/` and `examples/` subdirectories that are gitignored and not included in the repo, and contain fewer subdirectories than the top-level `plugins/` and `examples/` directories.
* These subdirectories are created by the `predeploy` gulp task, which is run as the final step (of several) of the `deploy:prepare` npm script, which is itself the first step of the `deploy` and `deploy:upstream` tasks.
* The `predeploy` task runs `predeployAll` from `scripts/gh-predeploy.js`, which in turn uses `getPluginFolders` to decide what subdirectories to copy from `plugins/` to `gh-pages/plugins/`.
* It appears that `getPluginFolders` cares that the subdirectory have a `package.json` file [and a `tests/index.html` file](https://github.com/google/blockly-samples/blob/1442683f23a8b213ee16daf3bb23345cdf5a682e/scripts/gh-predeploy.js#L302); it does not appear to care whether the subdirectory or any of its contents are mentioned in `gh-pages/index.md`.

So AFAICT this change only remove `plugins/serialize-disabled-interactions/test/index.html` from the TOC; to remove the plugin from the set of published plugins you'd also need to remove `plugins/serialize-disabled-interactions/test/index.html`.

But this whole system is byzantine enough that I might have missed something; please do correct me if so.

cpcallen (approved):

BeksOmega:
Thanks for clarifying that it's not entirely deleted! but removing it from the demos page so that it's harder to find is good enough for me =) I don't want to delete the test page yet in case we need to test it for whatever reason!

PR #1741, opened 2023-06-26
feat: Initialized minimap plugin using the blockly-plugin template

cesarades:
Created the minimap plugin (and following the new plugin review checklist).
No code has been written so far, just got the template and made some tweaks per the guide.

### DIFF
--- a/plugins/workspace-minimap/package.json
+++ b/plugins/workspace-minimap/package.json
@@ -1,0 +1,58 @@
+{
+  "name": "@blockly/workspace-minimap",
+  "version": "0.0.0",
+  "description": "A Blockly plugin.",
+  "scripts": {
+    "audit:fix": "blockly-scripts auditFix",
+    "build": "blockly-scripts build",
+    "clean": "blockly-scripts clean",
+    "lint": "blockly-scripts lint",
+    "predeploy": "blockly-scripts predeploy",
+    "start": "blockly-scripts start",
+    "test": "blockly-scripts test"
+  },
+  "priavte": true,
+  "main": "./dist/index.js",
+  "module": "./src/index.js",
+  "unpkg": "./dist/index.js",
+  "author": "Blockly Team",
+  "keywords": [
+    "blockly",
+    "blockly-plugin",
+    "workspace-minimap"
+  ],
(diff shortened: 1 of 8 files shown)


plugins/workspace-minimap/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "scripts": {
+    "audit:fix": "blockly-scripts auditFix",
+    "build": "blockly-scripts build",
+    "clean": "blockly-scripts clean",
+    "lint": "blockly-scripts lint",
+    "predeploy": "blockly-scripts predeploy",
+    "start": "blockly-scripts start",
+    "test": "blockly-scripts test"
+  },
+  "priavte": true,
```
gonfunko: private

plugins/workspace-minimap/src/index.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+export class Minimap {
+  /** The workspace. */
+  protected workspace: Blockly.WorkspaceSvg;
+  /**
+   * Constructor for ...
+   * @param workspace The workspace that the plugin will
+   *     be added to.
+   */
+  constructor(workspace: Blockly.WorkspaceSvg) {
```
gonfunko, line 28: For simple cases where you're just directly assigning a constructor arg to an instance variable, you can use [parameter properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties) to simplify things a bit.

plugins/workspace-minimap/src/index.ts
gonfunko, line 1: I saw in a previous state of things you had this named minimap.ts; I generally prefer that, since it makes it clearer what this file contains. (index.ts is fine for the test directory though!)

gonfunko: Discussed offline that this will be resolved in a followup.

plugins/workspace-minimap/src/index.ts
gonfunko: Discussed offline that this will be resolved in a followup.

gonfunko (approved):

PR #1749, opened 2023-06-30
feat: Implement mirroring between the primary workspace and the minimap

cesarades:
**Description**
Successfully replicated the mirror demo

**Testing**
Created two separate divs, one for each workspace, and made sure the block events transferred between them

### DIFF
--- a/plugins/workspace-minimap/src/index.ts
+++ b/plugins/workspace-minimap/src/index.ts
@@ -10,29 +10,30 @@
  * appears on top of your main workspace. This gives you an overview
  * of what your code looks like, and how it is organized.
  */
-import * as Blockly from 'blockly/core';
+export {Minimap} from './minimap';
+export {PositionedMinimap} from './positioned_minimap';
 
 /**
  * A minimap is a miniature version of your blocks that appears on
  * top of your main workspace. This gives you an overview of what
  * your code looks like, and how it is organized.
  */
-export class Minimap {
-  /** The workspace. */
-  protected workspace: Blockly.WorkspaceSvg;
-  /**
-   * Constructor for ...
-   * @param workspace The workspace that the plugin will
-   *     be added to.
-   */
-  constructor(workspace: Blockly.WorkspaceSvg) {
-    this.workspace = workspace;
-  }
+// export class Minimap {
+//   /** The workspace. */
+//   protected workspace: Blockly.WorkspaceSvg;
+//   /**
+//    * Constructor for ...
+//    * @param workspace The workspace that the plugin will
+//    *     be added to.
+//    */
+//   constructor(workspace: Blockly.WorkspaceSvg) {
+//     this.workspace = workspace;
+//   }
 
-  /**
-   * Initialize.
-   */
-  init(): void {
-    return;
-  }
-}
+//   /**
+//    * Initialize.
+//    */
+//   init(): void {
+//     return;
+//   }
+// }
--- a/plugins/workspace-minimap/src/index.ts
+++ b/plugins/workspace-minimap/src/index.ts
@@ -12,28 +12,3 @@
  */
 export {Minimap} from './minimap';
 export {PositionedMinimap} from './positioned_minimap';
-
-/**
- * A minimap is a miniature version of your blocks that appears on
- * top of your main workspace. This gives you an overview of what
- * your code looks like, and how it is organized.
- */
-// export class Minimap {
-//   /** The workspace. */
-//   protected workspace: Blockly.WorkspaceSvg;
-//   /**
-//    * Constructor for ...
-//    * @param workspace The workspace that the plugin will
-//    *     be added to.
-//    */
-//   constructor(workspace: Blockly.WorkspaceSvg) {
-//     this.workspace = workspace;
-//   }
-
-//   /**
-//    * Initialize.
-//    */
-//   init(): void {
-//     return;
-//   }
-// }
--- a/plugins/workspace-minimap/src/minimap.ts
+++ b/plugins/workspace-minimap/src/minimap.ts
@@ -13,6 +13,14 @@
 
 import * as Blockly from 'blockly/core';
 
+// Events that should be send over to the minimap from the primary workspace
+const BLOCK_EVENTS_ = [
+    Blockly.Events.BLOCK_CHANGE,
+    Blockly.Events.BLOCK_CREATE,
+    Blockly.Events.BLOCK_DELETE,
+    Blockly.Events.BLOCK_DRAG,
+    Blockly.Events.BLOCK_MOVE];
+
 /**
  * A minimap is a miniature version of your blocks that appears on
  * top of your main workspace. This gives you an overview of what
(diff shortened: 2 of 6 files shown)


plugins/workspace-minimap/src/index.ts
```diff
@@ -10,29 +10,30 @@
  * appears on top of your main workspace. This gives you an overview
  * of what your code looks like, and how it is organized.
  */
-import * as Blockly from 'blockly/core';
+export {Minimap} from './minimap';
+export {PositionedMinimap} from './positioned_minimap';
```
BeksOmega, line 14: Export looks perfect =)

plugins/workspace-minimap/src/index.ts
```diff
@@ -10,29 +10,30 @@
...
-  protected workspace: Blockly.WorkspaceSvg;
-  /**
-   * Constructor for ...
-   * @param workspace The workspace that the plugin will
-   *     be added to.
-   */
-  constructor(workspace: Blockly.WorkspaceSvg) {
-    this.workspace = workspace;
-  }
+// export class Minimap {
```
BeksOmega: Nit: Remove commented out code.

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -0,0 +1,67 @@
...
+    Blockly.Events.BLOCK_CREATE,
+    Blockly.Events.BLOCK_DELETE,
+    Blockly.Events.BLOCK_DRAG,
+    Blockly.Events.BLOCK_MOVE];
+
+/**
+ * A minimap is a miniature version of your blocks that appears on
+ * top of your main workspace. This gives you an overview of what
+ * your code looks like, and how it is organized.
+ * @param {!Blockly.WorkspaceSvg} workspace The workspace to sit in.
```
BeksOmega: Nit: Remove param comment from class. Params are only applied to methods!

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+export class Minimap {
+    protected workspace_: Blockly.WorkspaceSvg;
+    protected minimap_: Blockly.WorkspaceSvg;
+    /**
+     * Constructor for a minimap
+     * @param {!Blockly.WorkspaceSvg} workspace The workspace to sit in.
```
```suggestion
     * @param workspace The workspace to mirror.
```

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * A minimap is a miniature version of your blocks that appears on
+ * top of your main workspace. This gives you an overview of what
+ * your code looks like, and how it is organized.
+ * @param {!Blockly.WorkspaceSvg} workspace The workspace to sit in.
+ */
+export class Minimap {
+    protected workspace_: Blockly.WorkspaceSvg;
+    protected minimap_: Blockly.WorkspaceSvg;
```
BeksOmega: I think the names here could be a bit clearer? E.g. is the `workspace` the primary workspace or the minimap workspace? Is the `minimap` an instance of `Minimap` or an instance of `Workspace`?

Maybe `primaryWorkspace` and `minimapWorkspace`? But whatever sounds clear to you!

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -0,0 +1,67 @@
...
+     * Constructor for a minimap
+     * @param {!Blockly.WorkspaceSvg} workspace The workspace to sit in.
+     */
+    constructor(workspace: Blockly.WorkspaceSvg) {
+        this.workspace_ = workspace;
+        const options = {
+            readOnly: true,
+        }
+        this.minimap_ = Blockly.inject('minimapDiv', options);
+        this.mirror = this.mirror.bind(this);
```
BeksOmega: Nit: You should be able to do this below in the `init` method, so I would recommend removing `this.mirror`.

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+  
+    /**
+     * Initialize.
+     */
+    init(): void {
+        this.workspace_.addChangeListener(this.mirror);
```
```suggestion
        this.workspace_.addChangeListener(() => { this.mirror(); });
```

Prefer arrow functions for binding, as per the [internal style guide](go/tsjs-style#rebinding-this).

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Creates the mirroring between workspaces. Passes on all desired events
+     * to the minimap from the primary worskpace.
```
```suggestion
     * to the minimap from the primary workspace.
```

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -0,0 +1,67 @@
...
+     * @param {Blockly.Events.Abstract} event The primary workspace event.
+     * @protected
```
BeksOmega: I would change this to `private mirror` unless you know of a use case for overriding this method! If it's `protected` that means we can never change it, because it's a binary breaking change (i.e. `protected` is equivalent to `public` from an API design standpoint).

So in the interest of keeping the API you have to maintain as small as possible, it's preferable to use `private` when possible!

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -0,0 +1,67 @@
...
+     */
+    mirror(event: Blockly.Events.Abstract): void {
+        if (BLOCK_EVENTS_.indexOf(event.type) > -1) {
```
BeksOmega: If you just want to see if the event type is one of the "approved" ones, you can also use a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set).

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -0,0 +1,67 @@
...
+            const json = event.toJson();
+            const duplciateEvent = Blockly.Events.fromJson(json, this.minimap_);
+            duplciateEvent.run(true);
```
```suggestion
            const duplicate = Blockly.Events.fromJson(json, this.minimap_);
            duplicate.run(true);
```

plugins/workspace-minimap/src/positioned_minimap.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @author cesarades@google.com (Cesar Ades)
+ */
+
+import * as Blockly from 'blockly/core';
+import {Minimap} from './minimap';
+
+/**
+ * A positionable version of minimap that implements IPositionable.
+ * @param {Minimap} minimap The minimap to position.
+ * @implements {Blockly.IPositionable}
```
BeksOmega, line 18: I would remove this, and then use `export class PositionedMinimap implements Blockly.IPositionable` when you get to that point.

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @fileoverview A minimap is a miniature version of your blocks that
+ * appears on top of your main workspace. This gives you an overview
+ * of what your code looks like, and how it is organized.
+ * @author cesarades@google.com (Cesar Ades)
+ */
+
+import * as Blockly from 'blockly/core';
+
+// Events that should be send over to the minimap from the primary workspace
+const BLOCK_EVENTS_ = [
```
BeksOmega: I really like the use of an array const here! Much better than a giant if statement :P

BeksOmega (changes requested):
I think this is basically just style nits! Overall approach looks perfect =)

plugins/workspace-minimap/src/minimap.ts
cesarades: Ok, the backpack plugin also does this fyi

BeksOmega: Haha good to know, I'll file an issue to clean it up.

plugins/workspace-minimap/src/minimap.ts
cesarades: For some reason this didn't give an error but it broke mirroring functionality

plugins/workspace-minimap/src/positioned_minimap.ts
cesarades: Yea will do when I get to work on that. If I add it now it gives me errors since I dont have the necessary methods.

plugins/workspace-minimap/src/minimap.ts
cesarades: The set was not working since it checks if the contents reference the same object

plugins/workspace-minimap/src/minimap.ts
cesarades: Yea I like your naming much better

plugins/workspace-minimap/src/minimap.ts
BeksOmega: Straaaange. Did you do:

```js
  init(): void {
      this.mirror = this.mirror.bind(this);
      this.primaryWorkspace.addChangeListener(() => { this.mirror(); });
    }
```

or:
```js
  init(): void {
      this.primaryWorkspace.addChangeListener(() => { this.mirror(); });
    }
```

I would expect the second example to work but not the first!

plugins/workspace-minimap/src/minimap.ts
BeksOmega: For ts it should be:

```ts
private mirror(event: Blockly.Events.Abstract): void { /* etc */ }
```

instead of using the `@private` annotation! Typescript has [actual access modifiers](https://www.typescripttutorial.net/typescript-tutorial/typescript-access-modifiers/) :D

plugins/workspace-minimap/src/minimap.ts
BeksOmega: Hmm it should work if you use the type strings (like you're currently doing). E.g: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has#using_the_has_method

cesarades: Yea I tried that idk

BeksOmega (approved):

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Creates the mirroring between workspaces. Passes on all desired events
+     * to the minimap from the primary workspace.
+     * @param event The primary workspace event.
+     */
+    private mirror(event: Blockly.Events.Abstract): void {
+      // TODO: shadow blocks get mirrored too (not supposed to happen)
+      if (BlockEvents.has(event.type)) {
+        const json = event.toJson();
+        console.log(json);
```
BeksOmega: Please remove the console log, and then feel free to merge =) Can also wait until after the long weekend!

PR #1750, opened 2023-06-30
chore: fix backpack docs

BeksOmega:
Fixup the backpack docs:
  * No need to apply `@param` to a class.
  * It says using `import("./options")` is the correct way to handle importing a type into JS. But we'll see what CI says about that :P

### DIFF
--- a/plugins/workspace-backpack/src/backpack.js
+++ b/plugins/workspace-backpack/src/backpack.js
@@ -12,12 +12,11 @@
 import * as Blockly from 'blockly/core';
 import {registerContextMenus} from './backpack_helpers';
 import {BackpackChange, BackpackOpen} from './ui_events';
-import {BackpackContextMenuOptions, BackpackOptions, parseOptions} from './options';
+import {parseOptions} from './options';
 
 /**
  * Class for backpack that can be used save blocks from the workspace for
  * future use.
- * @param {!Blockly.WorkspaceSvg} workspace The workspace to sit in.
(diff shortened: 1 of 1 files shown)


plugins/workspace-backpack/src/backpack.js
```diff
@@ -27,7 +26,8 @@ export class Backpack extends Blockly.DragTarget {
    * Constructor for a backpack.
    * @param {!Blockly.WorkspaceSvg} targetWorkspace The target workspace that
    *     the backpack will be added to.
-   * @param {!BackpackOptions=} backpackOptions The backpack options to use.
+   * @param {!import("./options").BackpackOptions=} backpackOptions The backpack
```
cpcallen: > It says using import("./options") is the correct way to handle importing a type into JS. But we'll see what CI says about that :P

Evidently the linter disagrees.  (Maybe try single quotes?)

* AFAIK there is nothing actually checking the types here, but if our doc generation tooling uses them then you should try to figure out how to make this work.
  * Maybe try single quotes?
  * ???
  * Falling back to unused imports (with suitable suppress annotation if required) would be reasonable.
* Otherwise—if docs are not being generated, or do not suffer from inaccurate type info here—then:
  *  If we're planning to convert this to TypeScript I'd suggest adding a `// import type {BackpackOptons} from './options'` comment at the top and just use `BackpackOptions` here in the interim (that's what I did with generators in core).
  * If not, then maybe just delete the type decl entirely?

PR #1751, opened 2023-06-30
feat: Toolbox Search Plugin

### LINKED ISSUE #242: Toolbox search
**Category**

Plugin

**Component**

@blockly/plugin-toolbox-search

It can be hard to find a block in the toolbox, even when you know which block you want.

A Search category in the toolbox with a text field.  The user types in the text field and the blocks in the flyout update in response.

The text field could be in either the category name or the flyout for the category.

**Additional context**

APIs needed in core Blockly:
- Ability to import a toolbox category and use it with your own toolbox
- (Possibly needed) Ability to add arbitrary UI elements in the toolbox category menu

See also: 
- https://github.com/google/blockly/issues/3851
- https://github.com/google/blockly/issues/766

gonfunko:
This PR adds a plugin that adds a new type of toolbox category. This category displays a search field in the toolbox, and filters the blocks it displays in the flyout based on the text entered into the search field. It filters blocks based on both their type/ID and the human-readable text they display, including the (default) values of fields on them. This fixes #242.

Only blocks included in another toolbox category will be filtered; if a block definition is loaded, but that block is not included in the toolbox, it will not be found when searching.

Because the text displayed on the blocks is searched, localization just works.

There are some limited unit tests; I manually verified that typing into the field shows the flyout and filters the blocks appropriately.

## Limitations and opportunities for future work
* Only the default values of fields are indexed, e.g. searching for `+` will match `math_arithmetic`, but searching for `-` won't because that isn't the default value of the dropdown field. Ideally it would match, and the block shown in the flyout would have the - option selected.
* I haven't tested this in deeply nested/collapsible toolboxes
* Non-block flyout items (e.g. the Create Variable button) are not indexed/shown
* The index is built when the plugin is initialized, so changes to the contents of the toolbox (if another category is loaded later, or a variable is created) are not reflected
* Function blocks are not indexed
* It might be nice to highlight the matching text on the block somehow
* Currently both the display text and type/ID of the block are searched. This should probably be made configurable, and it might be nice to index the tooltip text as well if present.
* Probably many other things!


plugins/toolbox-search/README.md
```diff
@@ -0,0 +1,41 @@
...
+yarn add blockly-toolbox-search
+```
+
+### npm
+```
+npm install blockly-toolbox-search --save
+```
+
+## Usage
+import * as Blockly from 'blockly';
```
rachel-fenichel, line 21: Insert ``` above the code block.

plugins/toolbox-search/README.md
```diff
@@ -0,0 +1,41 @@
+# blockly-toolbox-search [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
+
+A [Blockly](https://www.npmjs.com/package/blockly) plugin that adds a toolbox category for searching
```
rachel-fenichel, line 3: Somewhere in here (top level or under "Usage") add a link to the [toolbox definition documentation](https://developers.google.com/blockly/guides/configure/web/toolbox#categories).

plugins/toolbox-search/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "blockly": "^10.0.0",
+    "@blockly/dev-scripts": "^2.0.0",
+    "@blockly/dev-tools": "^6.0.0",
+    "typescript": "^5.0.4",
+    "chai": "^4.3.7"
+  },
+  "peerDependencies": {
+    "blockly": "^10.0.0"
+  },
+  "publishConfig": {},
```
rachel-fenichel: I think you also need this `publishConfig`:
```

  "publishConfig": {
    "access": "public",
  },
  ```

@maribethb to confirm.

File: plugins/toolbox-search/src/block_searcher.ts [hunk 4]
```diff
@@ -0,0 +1,88 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+
+/**
+ * A class that provides methods for indexing and searching blocks.
```
rachel-fenichel, line 10: Add a note that it functions by creating a headless workspace, instantiating blocks, indexing the strings on those blocks, and then cleaning up after itself.

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -0,0 +1,88 @@
...
+      const block = this.blockCreationWorkspace.newBlock(blockId);
+      this.indexBlockText(blockId.replaceAll('_', ' '), blockId);
+      block.inputList.forEach((input) => {
+        input.fieldRow.forEach((field) => {
+          this.indexBlockText(field.getText(), blockId);
+        });
+      });
+    });
+    // Clean up the temporary blocks that were created for indexing.
+    this.blockCreationWorkspace.clear();
```
rachel-fenichel: Can you just create and then delete the workspace in this function, instead of keeping it around?

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -0,0 +1,88 @@
...
+  private trigramsToBlocks = new Map<string, Set<string>>();
+
+  /**
+   * Populates the cached map of trigrams to the blocks they correspond to.
+   * @param blockIds A list of block IDs to index.
+   */
+  indexBlocks(blockIds: string[]) {
+    blockIds.forEach((blockId) => {
+      const block = this.blockCreationWorkspace.newBlock(blockId);
+      this.indexBlockText(blockId.replaceAll('_', ' '), blockId);
```
rachel-fenichel: Are the IDs guaranteed to be block types?

plugins/toolbox-search/src/toolbox_search.ts
```diff
@@ -0,0 +1,191 @@
...
+   */
+  private registerShortcut() {
+    const shortcut = Blockly.ShortcutRegistry.registry.createSerializedKey(
+        Blockly.utils.KeyCodes.B, [Blockly.utils.KeyCodes.CTRL]);
+    Blockly.ShortcutRegistry.registry.register({
+      name: ToolboxSearchCategory.START_SEARCH_SHORTCUT,
+      callback: () => {
+        const position = this.getPosition();
+        if (position < 0) return false;
+        this.parentToolbox_.selectItemByPosition(position);
```
rachel-fenichel, line 93: Did you have to do this because we don't have a way to select the toolbox category by name?

plugins/toolbox-search/test/index.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @param options The Blockly options.
+ * @returns The created workspace.
+ */
+function createWorkspace(blocklyDiv: HTMLElement,
+    options: Blockly.BlocklyOptions): Blockly.WorkspaceSvg {
+  return Blockly.inject(blocklyDiv, options);
+}
+
+document.addEventListener('DOMContentLoaded', function() {
+  toolboxCategories['contents'].push({
```
rachel-fenichel: Nit: make a new object for this instead of modifying the imported one.

plugins/toolbox-search/src/block_searcher.ts
gonfunko: The caller verifies this; is there a way we can check their validity here/is that what you're suggesting?

plugins/toolbox-search/src/toolbox_search.ts
gonfunko: Yes 😔

plugins/toolbox-search/src/block_searcher.ts
rachel-fenichel: Nope, checking on the naming because I expected them to be the UUIDs. Can you change the parameter/variable names to blockType and blockTypes instead of ID?

plugins/toolbox-search/README.md
gonfunko: Done.

plugins/toolbox-search/README.md
gonfunko: Done.

plugins/toolbox-search/src/block_searcher.ts
gonfunko: Done (on the indexing method).

plugins/toolbox-search/src/block_searcher.ts
gonfunko: Good call! Done.

rachel-fenichel: If you create the workspace in the function and then don't call `dispose` at the end of the function, but the workspace is headless, does it actually get cleaned up correctly? I think you still need to call `dispose`.

plugins/toolbox-search/test/index.ts
gonfunko: Done.

plugins/toolbox-search/src/block_searcher.ts
gonfunko: Done!

plugins/toolbox-search/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "blockly-toolbox-search",
```
maribethb: This is the wrong name format. The name must be `@blockly/toolbox-search` since this is a first-party plugin. Probably too late now but for future reference you can pass a --first-party flag to the plugin creation script to have it generate the correct names. Make sure you do a search for this string and change it everywhere e.g. in the README. I will send you the link to the rest of the plugin checklist.

maribethb (changes requested):

plugins/toolbox-search/package.json
gonfunko: Thanks for the catch, and good to know about the plugin creation script flag.

plugins/toolbox-search/package.json
gonfunko: Done.

rachel-fenichel (approved):

PR #1756, opened 2023-07-05
feat: Make the minimap responsive to the window and primaryWorkspace size

cesarades:
**Description**
As the primary workspace changes, the minimap will automatically adjust in scale (and it will center itself). It also now is responsive to window size.

**Tests**
I changed the window size in the advanced playground to make sure minimap would respond.
I replicated all the block events in the primary workspace and made sure the changes applied to the minimap.

### DIFF
--- a/plugins/workspace-minimap/src/minimap.ts
+++ b/plugins/workspace-minimap/src/minimap.ts
@@ -67,7 +67,6 @@ export class Minimap {
      */
     private mirror(event: Blockly.Events.Abstract): void {
       // TODO: shadow blocks get mirrored too (not supposed to happen)
-      // TODO: on connect the zoom cuts off the moved block
 
       if (!BlockEvents.has(event.type)) {
         return; // Filter out events.
@@ -78,6 +77,9 @@ export class Minimap {
       duplicate.run(true);
 
       // Resize and center the minimap.
-      this.minimapWorkspace.zoomToFit();
+      // We need to wait for the event to finish rendering to do the zoom.
+      Blockly.renderManagement.finishQueuedRenders().then(() => {
+        this.minimapWorkspace.zoomToFit();
+      });
     }
 }
--- a/plugins/workspace-minimap/src/minimap.ts
+++ b/plugins/workspace-minimap/src/minimap.ts
@@ -37,6 +37,14 @@ export class Minimap {
       this.primaryWorkspace = workspace;
       const options = {
         readOnly: true,
+        move: {
+          scrollbars: {
+            vertical: true,
+            horizontal: true,
+          },
+          drag: false,
+          wheel: false,
+        },
       };
       this.minimapWorkspace = Blockly.inject('minimapDiv', options);
     }
@@ -46,6 +54,7 @@ export class Minimap {
      */
     init(): void {
       this.primaryWorkspace.addChangeListener((e) => void this.mirror(e));
+      this.minimapWorkspace.scrollbar.setContainerVisible(false);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


plugins/workspace-minimap/src/minimap.ts
```diff
@@ -45,7 +53,11 @@ export class Minimap {
      * Initialize.
      */
     init(): void {
+      this.minimapWorkspace.scrollbar.setContainerVisible(false);
       this.primaryWorkspace.addChangeListener((e) => void this.mirror(e));
+      window.addEventListener('resize', (e) => {
```
gonfunko: It might be good to debounce this, e.g. only zoom to fit 250ms or so after the last resize event; given that a minimap is most useful when the workspace is large, there might be a noticeable performance penalty from doing this live if the workspace has hundreds of blocks since this callback gets invoked more or less continuously during a window resize. Could you see if there is a noticeable performance impact either way if the workspace has > 100 blocks?

cesarades: Sounds good, I implemented debounce. As for checking if there is a performance impact, I was planning on leaving that for a _next steps_. I tried to quickly create a script to auto-populate the workspace but could not replicate the 'airstrike' function from core.

gonfunko: Sorry, I should have been clearer - I think we need to verify the performance impact first, because if the non-debounced implementation is fine, that's preferable, since it updates the state of the minimap faster and requires less/simpler code. Feel free to ping me if you need help setting up something akin to the airstrike functionality!

cesarades: Ok, I was able to generate a script to auto populate the workspace. I tested with anywhere between 100 and 500 blocks at once, and the difference between live and debouncing was marginal. Not only was the performance basically the same, but with the live rendering it looked a lot smoother. I'm going to go ahead and revert to live rendering. Let me know if I should change something else!

plugins/workspace-minimap/test/index.ts
```diff
@@ -12,22 +12,23 @@ import * as Blockly from 'blockly';
...
+  block.initSvg();
+  block.render();
+}
+
+// Move the blocks to random locations.
+for (const block of ws.getAllBlocks(true)) {
+  const cord = new Blockly.utils.Coordinate(
+      Math.round(Math.random() * 450 + 40),
+      Math.round(Math.random() * 600 + 40));
+  block.moveTo(cord, ['drag']);
```
gonfunko: This isn't being triggered by a drag, I'd probably just omit the optional reason argument here.

plugins/workspace-minimap/test/index.ts
```diff
@@ -12,22 +12,23 @@ import * as Blockly from 'blockly';
...
-  const workspace = Blockly.inject(blocklyDiv, options);
-
-  const minimap = new Minimap(workspace);
-  minimap.init();
-
-  return workspace;
-}
-
+// Creates the primary workspace and adds the minimap.
 const ws = Blockly.inject('primaryDiv', {toolbox: toolboxCategories});
```
gonfunko: The prior names ('workspace' and 'minimap') are preferable - the styleguide generally discourages abbreviations for TS.

plugins/workspace-minimap/src/minimap.ts
gonfunko: Sounds good, thanks for checking and updating accordingly!

plugins/workspace-minimap/test/index.ts
```diff
@@ -12,22 +12,23 @@ import * as Blockly from 'blockly';
...
 const mp = new Minimap(ws);
 mp.init();
+
+// Creates 100 random blocks
+for (let i = 0; i < 100; i++) {
+  const prototype = 'controls_if';
+  const block = ws.newBlock(prototype);
```
gonfunko: Consider just inlining the string literal here vs stashing it in a variable first since this is the only place it's used. (Also re: the comment these are all if blocks, not random ones)

gonfunko (approved):

PR #1758, opened 2023-07-07
fix: update more broken / deprecated APIs for v10

BeksOmega:
Updates generators in all of our examples and plugins (which weren't previously updated).

Also updates references from `Blockly.Mutator` to `Blockly.icons.MutatorIcon` in test blocks.

### DIFF
--- a/examples/blockly-vue/src/blocks/stocks.js
+++ b/examples/blockly-vue/src/blocks/stocks.js
@@ -26,7 +26,7 @@
 
 
 import * as Blockly from 'blockly/core';
-import {javascriptGenerator} from 'blockly/javascript';
+import {javascriptGenerator, Order} from 'blockly/javascript';
 
 Blockly.Blocks["stock_buy_simple"] = {
   init: function() {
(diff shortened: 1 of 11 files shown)


examples/blockly-vue/src/blocks/stocks.js
```diff
@@ -46,14 +46,14 @@ Blockly.Blocks["stock_buy_simple"] = {
...
   var number_id = block.getFieldValue("ID");
   var number_amount = block.getFieldValue("Amount");
   var number_price = block.getFieldValue("Price");
-  var value_number = javascriptGenerator.valueToCode(
+  var value_number = generator.valueToCode(
     block,
     "Number",
-    javascriptGenerator.ORDER_ATOMIC
+    Order.ATOMIC
   );
```
cpcallen, line 57: I'm guessing that this is Prettier-approved, but can you check that it won't now fit on a single line?

(Here and various places below.)

cpcallen (approved):

examples/blockly-vue/src/blocks/stocks.js
BeksOmega: uhhh we don't have any formatter in samples afaik, and we only run the linter on plugins (not examples like this).

I'll be honest in that I'm not /really/ interested in manually reformatting all of these files :P

PR #1768, opened 2023-07-10
feat: Implement interactions with the minimap (click, drag)

cesarades:
**Description**
The minimap svg binds to browser events to translate interactions into primary viewport changes.

**Tests**
In the playground, I clicked everywhere on the minimap and confirmed that the click centered the primary viewport. The same was  done for dragging.

### DIFF
--- a/plugins/workspace-minimap/src/minimap.ts
+++ b/plugins/workspace-minimap/src/minimap.ts
@@ -59,7 +59,7 @@ export class Minimap {
           'mousedown', this, this.onClick);
       Blockly.browserEvents.bind(this.minimapWorkspace.svgGroup_,
           'mousemove', this, this.onMouseMove);
-      window.addEventListener('resize', (e) => {
+      window.addEventListener('resize', () => {
         this.minimapWorkspace.zoomToFit();
       });
     }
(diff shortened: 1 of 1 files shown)


plugins/workspace-minimap/src/minimap.ts
```diff
@@ -55,7 +55,11 @@ export class Minimap {
     init(): void {
       this.minimapWorkspace.scrollbar.setContainerVisible(false);
       this.primaryWorkspace.addChangeListener((e) => void this.mirror(e));
-      window.addEventListener('resize', () => {
+      Blockly.browserEvents.bind(this.minimapWorkspace.svgGroup_,
+          'mousedown', this, this.onClick);
+      Blockly.browserEvents.bind(this.minimapWorkspace.svgGroup_,
+          'mousemove', this, this.onMouseMove);
+      window.addEventListener('resize', (e) => {
```
BeksOmega: If a parameter is unused, I think you can also use `_` as a placeholder name! Not important though, just thought it might be interesting :P

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -55,7 +55,11 @@ export class Minimap {
     init(): void {
       this.minimapWorkspace.scrollbar.setContainerVisible(false);
       this.primaryWorkspace.addChangeListener((e) => void this.mirror(e));
-      window.addEventListener('resize', () => {
+      Blockly.browserEvents.bind(this.minimapWorkspace.svgGroup_,
+          'mousedown', this, this.onClick);
+      Blockly.browserEvents.bind(this.minimapWorkspace.svgGroup_,
+          'mousemove', this, this.onMouseMove);
```
BeksOmega: I think mousedown and mousemove were the correct choices!

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -82,4 +86,37 @@ export class Minimap {
         this.minimapWorkspace.zoomToFit();
       });
     }
+
+    /**
+     * Changes the primary workspace viewport when clicking on the minimap
```
BeksOmega: Nit: Period at the end of comments.

Good job using the correct sentence construction though!

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -82,4 +86,37 @@ export class Minimap {
...
+     * @param event The minimap browser event.
+     */
+    private onClick(event: PointerEvent): void {
+      // Calculates the scale between the minimap and primary workspace.
+      const primaryMetrics = this.primaryWorkspace.getMetrics();
+      const contentDiv = this.minimapWorkspace.svgBlockCanvas_.
+          getBoundingClientRect();
```
BeksOmega: Could you use `getMetrics` here as well? I'm pretty sure that's equivalent to what you're doing. If it's possible I think it's better to use consistent properties, so then we know we're actually comparing the different sizes of the same thing!

BeksOmega (commented):
Still working on review, but wanted to leave you some comments so you're not blocked =)

One larger point: I think this is a good opportunity to add some unit tests! If you could factor the coordinate conversion logic out of the event listeners, and then verify that it has the correct outputs for given coordinate inputs, I think that would be perfect!

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -82,4 +86,37 @@ export class Minimap {
...
+
+      // Calculates the location of the click in the minimap.
+      const wsDiv = this.minimapWorkspace.svgGroup_.
+          getBoundingClientRect();
+      const offsetX = event.offsetX - (wsDiv.width - contentDiv.width) / 2;
+      const offsetY = event.offsetY - (wsDiv.height - contentDiv.height) / 2;
+
+      // Converts the click from the minimap to the primary worskspace.
+      const x = -primaryMetrics.scrollLeft - offsetX * scale;
+      const y = -primaryMetrics.scrollTop - offsetY * scale;
```
BeksOmega: We talked through fixing up this step and splitting it out offline!

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -55,7 +55,11 @@ export class Minimap {
     init(): void {
       this.minimapWorkspace.scrollbar.setContainerVisible(false);
       this.primaryWorkspace.addChangeListener((e) => void this.mirror(e));
-      window.addEventListener('resize', () => {
+      Blockly.browserEvents.bind(this.minimapWorkspace.svgGroup_,
+          'mousedown', this, this.onClick);
+      Blockly.browserEvents.bind(this.minimapWorkspace.svgGroup_,
```
BeksOmega: One other thing is to make sure you're getting a `mousedown` event before you start responding to `mousemove` events! Otherwise someone can start a drag in one location, and then drag into the minimap, and the minimap will start updating the main workspace (which is unexpected).

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -124,21 +125,36 @@ export class Minimap {
     }
 
     /**
-     * Changes the primary workspace viewport when clicking on the minimap.
-     * @param event The minimap browser event.
+     * Scrolls the primary workspace viewport based on a minimap event.
+     * @param event The minimap broswer event.
```
```suggestion
     * @param event The minimap browser event.
```

BeksOmega (approved):
Looks great =) Just one nit then gtg!

PR #1775, opened 2023-07-13
chore: Unit tests for Minimap interactions

cesarades:
**Description**
I refactored the function in the minimap that converted a mouse event into scroll coordinates for the primary workspace. I made it so instead of an instance method taking in a mouse event, it was a static class method that took in workspace metrics and event coordinates. This way I could conduct unit tests on the method function.

**Tests**
The tests I made are broken down into three extensive categories: workspace shape, size, and location. Within the three different configurations I tested for clicks in the top left, center, and bottom right of the workspace.


plugins/workspace-minimap/src/minimap.ts
```diff
@@ -91,24 +91,25 @@ export class Minimap {
...
+        offsetY: number): [number, number] {
+      // Gets the click location relative to the
       // top left of the minimap content.
       offsetX -= (minimapMetrics.svgWidth - minimapMetrics.contentWidth) / 2;
       offsetY -= (minimapMetrics.svgHeight - minimapMetrics.contentHeight) / 2;
 
-      // Calculates the scale between the minimap and primary workspace
-      // and applies it to the offset.
-      const scale = primaryMetrics.contentWidth / minimapMetrics.contentWidth;
+      // Scales the click location into the primary workspace.
```
gonfunko: nit: Might be clearer as `coordinate` instead of `click` since presumably it could be some other type of non-click event.

plugins/workspace-minimap/test/coordinate_conversion.mocha.js
```diff
@@ -0,0 +1,160 @@
...
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+
+const chai = require('chai');
+const assert = chai.assert;
+const {Minimap} = require('../src/minimap');
+
+
```
gonfunko, line 12: It looks like each suite is broadly testing the same thing, just with different relative sizes; could you factor the variable bits (metrics, click positions, expected results) into an array, and then just loop through that and move the suite/test calls inside?

BeksOmega: I would actually recommend against doing this! While it works great with most test systems, I don't think it works well with Mocha, because there's no good way to isolate tests after you've separated the data out into a different structure.

I prefer using the AAA structure, even in cases where you have basic inputs/outputs.

But if you have strong opinions Aaron, happy to defer.

gonfunko: Beka raised a good point offline about not being able to .only/.skip the tests if they're structured this way, and she has a lot more experience with Mocha so I'm happy to defer to her. The existing test structure LGTM, sorry for the tangent here!

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -147,6 +152,8 @@ export class Minimap {
      * Unbinds the minimap mousemove when the mouse is not clicked.
      */
     private onClickUp(): void {
+      // TODO: If you start the click in the minimap and end it in the primary
+      //       the then this function is never unbinded
```
BeksOmega, line 154: Nice catch!

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -91,24 +91,25 @@ export class Minimap {
     /**
      * Converts a mouse event on the minimap into scroll coordinates for
      * the primary viewport.
+     * @param primaryMetrics The metrics from the primary workspace.
+     * @param minimapMetrics The metrics from the minimap workspace.
```
BeksOmega, line 95: Reorganization looks great =)

plugins/workspace-minimap/test/coordinate_conversion.mocha.js
```diff
@@ -0,0 +1,160 @@
...
+        contentHeight: 500, contentWidth: 500};
+    });
+
+    test('Top left click', function() {
+      const click = {x: 0, y: 0};
+      const converted = Minimap.minimapToPrimaryCoords(
+          this.primaryMetrics, this.minimapMetrics,
+          click.x, click.y);
+
+      assert.strictEqual(
```
BeksOmega: I think you should be able to call some assertion like `deepEqual` instead of converting these to strings.

plugins/workspace-minimap/test/coordinate_conversion.mocha.js
```diff
@@ -0,0 +1,160 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+
+const chai = require('chai');
+const assert = chai.assert;
+const {Minimap} = require('../src/minimap');
+
+
+suite('Interactions Suite (Shape, Size, Location)', function() {
```
BeksOmega: Test naming is super important, just like function and variable naming! If a test fails, you want to know immediately what behavior is broken.

For example, the below suite is called "Interactions Suite (Shape, Size, Location) > Square, Medium, Center". I know that you're testing the shape of the content! But someone else dealing with your failing test might not. 

I would probably update the top-level suite to be something like "Click coordinates are properly converted from the minimap to the primary workspace" and the lower-level suite to be something like "Square, medium, and centered content". But go through and figure out what you think works!

Here's a good article about test naming: https://testing.googleblog.com/2014/10/testing-on-toilet-writing-descriptive.html

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -91,24 +91,25 @@ export class Minimap {
...
      * @returns (x, y) primary workspace scroll coordinates.
      */
-    private minimapToPrimaryCoords(
-        offsetX: number, offsetY: number): [number, number] {
-      // Get the metrics from the workspaces
-      const primaryMetrics = this.primaryWorkspace.getMetrics();
-      const minimapMetrics = this.minimapWorkspace.getMetrics();
-
-      // Calculates the location of the click relative to the
+    static minimapToPrimaryCoords(
```
BeksOmega, line 100: I think the decision to make this static is great =)

plugins/workspace-minimap/test/coordinate_conversion.mocha.js
```diff
@@ -0,0 +1,160 @@
...
+    test('Top left click', function() {
+      const click = {x: 0, y: 0};
+      const converted = Minimap.minimapToPrimaryCoords(
+          this.primaryMetrics, this.minimapMetrics,
+          click.x, click.y);
+
+      assert.strictEqual(
+          converted.toString(),
+          [750, 750].toString(),
+          'Incorrect top left click');
```
BeksOmega: I believe when a test fails, the message is something like "Expected <message>" So this would read "Expected 'Incorret top left click'" which is probably not what you want :P but you can call `assert.fail()` here  to force a failure and double check me!

BeksOmega (commented):
Overal looks good to me! Love the tests you picked out =) Just some comments about structure!

plugins/workspace-minimap/test/coordinate_conversion.mocha.js
cesarades: Good call let me check that out

plugins/workspace-minimap/test/coordinate_conversion.mocha.js
cesarades: The following is a sample output if one of the tests fail:


 ```
 1) Interactions Suite (Shape, Size, Location)
       Square, Medium, Center
         Top left click:

Incorrect top left click
      + expected - actual

       [
      +  850
         750
      -  750
       ]
```

plugins/workspace-minimap/test/coordinate_conversion.mocha.js
cesarades: Yup, implemented your suggestion

plugins/workspace-minimap/test/coordinate_conversion.mocha.js
BeksOmega: Sweet sounds good then!

BeksOmega (approved):

PR #1776, opened 2023-07-13
feat: add a workflow that cleans up after dependabot

maribethb:
Adds a workflow that hopefully
- runs any time a dependabot-created PR is opened, reopened, or synchronized (more commits are added)
- runs `npm run boot` which runs `lerna bootstrap` in both plugins and examples
- commits any resulting changes to the same dependabot-created PR
- doesn't block dependabot from making further commits to the PR e.g. in response to a command to rebase

The reason this is needed is because when dependabot makes a PR it updates the package-lock for a plugin but doesn't run `lerna bootstrap`. The result is that plugins [keep having](https://github.com/google/blockly-samples/commit/4599a0e3fd01e46890c65bc487090b6ced5da221) local dependencies added to their package-locks by dependabot, and then later they might be removed when a human runs `lerna bootstrap`, cluttering up other PRs. These local dependencies should not be part of the `package-lock` files -- that's just how lerna works :/

No idea if it works as intended until we submit it and try to run it on a dependabot PR, but at the very least I'm pretty confident it shouldn't ruin anyone else's non-dependabot PRs.


.github/workflows/dependabot_update.yml
```diff
@@ -0,0 +1,58 @@
...
+      # https://github.com/actions/checkout/issues/13#issuecomment-724415212
+      - name: Git identity
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email '<41898282+github-actions[bot]@users.noreply.github.com'
+
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
BeksOmega: Should we use the latest?

.github/workflows/dependabot_update.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Lerna bootstrap
+        run: |
+          npm run boot
+          cd examples && npm run boot
+
+      # If any package-locks were updated by lerna bootstrap, commit them
+      # Using `[dependabot skip]` in the commit message allows dependabot
```
BeksOmega, line 47: Can you add a link to the docs for this?

BeksOmega (approved):
LGTM! This is so clever and lovely :D

.github/workflows/dependabot_update.yml
maribethb: yes but please don't ask me about the line with 200 length now

PR #1777, opened 2023-07-13
chore: update package locks and misc dependencies

maribethb:
- Updates `examples/blockly-vue` version of `eslint` to satisfy some of its peer dependency issues
- Updates `workspace-minimap` to use v7 of `@blockly/dev-tools` instead of v6. This was likely not updated because it was submitted after the major version bump to dev-tools, so it didn't get auto-bumped by lerna because it didn't exist yet. This can cause problems because lerna would be using the version from npm instead of locally linking it like everything else.
- Updates versions of other plugins similar to above used inside examples to latest versions, then tested that those plugins run correctly still (this resolves peer dependency version issues with blockly)
- Ran `npm install` at root and then `cd examples && npm install` to update all package-locks. Dependabot unfortunately messes up our package-locks and keeps adding back the local dependencies to the lockfiles.... see https://github.com/google/blockly-samples/pull/1776

This should fix CI failures caused by inconsistent package-lock states causing `npm ci` to fail. I tested by running `npm run clean:node` and `npm ci` at root, then `npx lerna clean --yes` and `npm ci` in examples after committing these changes and it passed / no pending changes / no peer dependency issues


maribethb:
unnamed

BeksOmega (approved):
Testing you outlined makes sense and these changes don't look crazy!

PR #1779, opened 2023-07-14
fix: run npm ci in lerna action

maribethb:
Updates the dependabot cleanup action to run `npm ci` and not just `npm run boot`. The previous attempt did not run an install at all so lerna was pulled from latest, which doesn't support `lerna bootstrap`... This attempt runs `npm ci` so the correct version of lerna will be installed first. hopefully.

### DIFF
--- a/.github/workflows/dependabot_update.yml
+++ b/.github/workflows/dependabot_update.yml
@@ -38,10 +38,12 @@ jobs:
         with:
           node-version: 20
 
-      - name: Lerna bootstrap
+      # Running npm ci also runs lerna bootstrap,
+      # which does the magic update we need to remove local dependencies.
+      - name: Npm clean install
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot_update.yml
```diff
@@ -38,10 +38,12 @@ jobs:
         with:
           node-version: 20
 
-      - name: Lerna bootstrap
+      # Running npm ci also runs lerna bootstrap,
+      # which does the magic update we need to remove local dependencies.
+      - name: Npm clean install
```
BeksOmega: I might just call this "fix package locks" cuz that's the important thing it's doing! The actual command being run doesn't really matter. But this is an unimportant change so feel free to ignore :P

BeksOmega (approved):

PR #1783, opened 2023-07-14
feat: add github workflow to check that examples works from a clean install

### LINKED ISSUE #1511: Add test that checks `npm ci` for examples
**Category**

 - Examples

**Component**

all

**Describe the bug**

Examples don't get tested or manually updated as frequently as plugins, especially those examples which do not appear on the samples site. This can lead to a situation where an example's dependencies are updated (such as by dependabot) in a way that causes `npm ci`/ `npm install` to fail for that example. Not only does this make the example less useful, it can also cause problems when updating the samples site on GitHub Pages.

If we had a test that ran `npm ci` for the examples (using lerna to run it for all examples) and failed if there were errors, then at the very least we'd have more confidence when merging dependabot PRs, and it can help us keep examples working and up to date.

maribethb:
Adds a GitHub Action workflow that only runs on PRs where `examples/` is affected.

Fixes #1511 

## Alternatives considered

I chose to make this a GH Action instead of a locally run test because:
- we don't have a way to run tests locally only if certain paths were changed, and this is an annoying thing to have to check if you are only updating plugins
- this test specifically checks the behavior of the true fresh clean install experience i.e. you have no `node_modules` directories present (npm behaves differently in the normal case where you do already have some things installed). it would be really annoying to clear out all of your node_modules directories locally when running tests.
- examples having outdated peer dependencies shouldn't block plugins from publishing (as they would if this were part of `npm run test`)

### DIFF
--- a/.github/workflows/clean_install_examples.yml
+++ b/.github/workflows/clean_install_examples.yml
@@ -23,6 +23,11 @@ jobs:
 
       - name: Check install status
         run: |
-          echo "If this step fails, reproduce locally by running `npm run clean:node && cd examples && npx lerna clean --yes` first"
           npm ci
           cd examples && npm ci
+
+      - name: Pro-tip in case of failure
+        if: ${{ failure() }}
+        run: |
+          echo "To reproduce locally, you need to replicate the fresh install process and get rid of all node_modules directories. Try running"
+          echo "npm run clean:node && cd examples && npx lerna clean --yes && npm ci"
--- a/.github/workflows/clean_install_examples.yml
+++ b/.github/workflows/clean_install_examples.yml
@@ -1,0 +1,28 @@
+# This workflow checks that examples can be installed cleanly.
+# It only runs on PRs that affect the `examples/` directory.
+
+name: Check examples install status
+
+on: 
+  pull_request:
+    paths:
+      - 'examples/**'
+
+jobs:
+  install:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 20
+
+      - name: Check install status
+        run: |
+          echo "If this step fails, reproduce locally by running `npm run clean:node && cd examples && npx lerna clean --yes` first"
+          npm ci
+          cd examples && npm ci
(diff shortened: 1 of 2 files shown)


.github/workflows/clean_install_examples.yml
```diff
@@ -0,0 +1,28 @@
+# This workflow checks that examples can be installed cleanly.
+# It only runs on PRs that affect the `examples/` directory.
+
+name: Check examples install status
+
+on: 
+  pull_request:
+    paths:
+      - 'examples/**'
```
BeksOmega, line 9: Can we update the dependabot workflow to also do this but for plugins?

.github/workflows/clean_install_examples.yml
```diff
@@ -0,0 +1,28 @@
...
+        uses: actions/checkout@v3
+
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 20
+
+      - name: Check install status
+        run: |
+          echo "If this step fails, reproduce locally by running `npm run clean:node && cd examples && npx lerna clean --yes` first"
```
BeksOmega:
```suggestion
          echo "If this step fails, reproduce locally by running `npm run clean:node && cd examples && npx lerna clean --yes && npm ci`"
```

I think it makes sense to just give the whole command if you're already giving most of it!

BeksOmega (approved):

.github/workflows/clean_install_examples.yml
maribethb: this already happens for plugins because doing a `npm ci` is part of the workflow for running the plugin tests

BeksOmega: Sorry I meant can we make the "fixup the package locks" workflow only run on PRs that are from dependabot and touch files in plugins?

.github/workflows/clean_install_examples.yml
maribethb: you're right. i changed this. also, the backticks were causing the commands to actually get executed lol so i fixed that, and moved the message to a step after the ci step. because otherwise that message is all the way at the beginning and easy to miss if you don't scroll up looking for it.

.github/workflows/clean_install_examples.yml
maribethb: oh yeah will do

Issue #777, opened 2021-06-22
Scaffolding for Third Party Plugins


alschmiedt:
**Category**

**Component**
gh-pages

**Is your feature request related to a problem? Please describe.**
We need a place to showcase third party plugins. We currently use [gh-pages](https://google.github.io/blockly-samples/) to display our plugins and examples. We would like to use this same site to display third party plugins. 

**Describe the solution you'd like**
For a third party plugin just display the [README](https://google.github.io/blockly-samples/plugins/block-dynamic-connection/README) page and add links to the github repository and npm page under "View code" and "View on npm". We should also display a banner or something on this page saying that it is a third party plugin.

**Describe alternatives you've considered**
- Add an "Explore plugins" page on the developer guides. 
- Add an example that uses each plugin. This would be nice, but requires more maintenance on our part. 

**Additional context**

rachel-fenichel:
Additional information:
- We use [Jekyll](https://jekyllrb.com/docs/) to make our githb pages content look good
- Jekyll [layouts](https://jekyllrb.com/docs/layouts/) determine what a page looks like
- ['layout' sections](https://github.com/google/blockly-samples/blob/master/gh-pages/_config.yml#L16) in `_config.yml` set which layout is used for each page
- We currently use a layout named [`default`](https://github.com/google/blockly-samples/blob/master/gh-pages/_layouts/default.html) for examples and plugins
- Layouts grab information from a `page` object and can use the information in the page ([for example](https://github.com/google/blockly-samples/blob/master/gh-pages/_layouts/default.html#L28), to add an npm link to the correct package)
- This object is called ["front matter"](https://jekyllrb.com/docs/front-matter/) in Jekyll documentation
- For our examples and plugins we [generate](https://github.com/google/blockly-samples/blob/master/gulpfile.js#L152) front matter during `predeploy`, in addition to running builds and copying files to the correct directories
  - Those end up in the examples and plugins folders that you can see if you switch to the [gh-pages branch](https://github.com/google/blockly-samples/tree/gh-pages)
  - For instance, here's the [index page](https://github.com/google/blockly-samples/blob/gh-pages/plugins/block-test/test/index.html) for the block-test plugin before processing by Jekyll


With all this, I think the steps are:
- Put another folder in the [gh-pages](https://github.com/google/blockly-samples/tree/master/gh-pages) directory, named something like `third-party-plugins`
- Add one page per third-party plugin that we want to showcase
[...]

PR #1806, opened 2023-07-20
feat: Improved the blockly-samples Github page UX.

### LINKED ISSUE #1754: Make demo website more skimmable + linkable
**Category**

Infrastructure

**Describe the bug**

The [current demo page](https://google.github.io/blockly-samples/) is incredibly difficult to navigate:

image

The number of links is overwhelming, it's difficult to find what you want even if you know what you're looking for. And if you don't know what you're looking for, there's no description to tell you what each plugin/demo does.

**Expected behavior**

1) The demo page should be arranged linearly instead of in a grid.
2) The demo page should include descriptions of each individual demo/plugin (ideally with a picture/thumbnail).
3) It should be possible to link to a section of the page, as well as an individual plugin description.

It would also be nice if the search bar was over on the right where people expect it to be (right now it's hard to notice).

**Additional context**

N/A

gonfunko:
This PR improves the appearance and usability of the Blockly Samples landing page on the Github Pages site. It improves the categorization of the plugins, and adds a written description and screen(shot|cast) of each to provide an easily skimmable presentation of the available plugins. A brief description of each of the demos/examples is also provided, to provide context without having to click through to each item.

I removed the search field, because it didn't do anything that command F doesn't and wasn't very visually prominent/discoverable.

This PR fixes #1754.

Screenshot 2023-07-20 at 1 24 50 PM


BeksOmega:
Once this is in I'm going to talk to a tech writer about how to we should organize our docs (devsite + this) to optimize for discoverability!

So wrt design suggestions, I'd rather punt on minor tweaks for now, then we can do another round of polish on it after we have feedback! But obviously if there are any major problems, we fix those now.

Also to state it frankly, this is such a vast improvement over the previous page that I'd love to get it to users sooner rather than later :P

gonfunko:
This also may address #777; I've included third-party plugins in this page, with an on-hover/focus banner indicating that it is a third party plugin.

gh-pages/_index.html
```diff
@@ -0,0 +1,472 @@
...
+<br>
+<h3 id="fields">
+  Fields
+</h3>
+<p>
+  Field plugins add new kinds of fields that can be incorporated into custom blocks, allowing your users to enter information in different ways. Learn more about how to <a href="https://developers.google.com/blockly/guides/create-custom-blocks/fields/customizing-fields/overview">create and customize fields</a>.
+</p>
+<div class="grid">
+  <a class="card" href="plugins/field-bitmap/test/index.html">
+    <img srcset="media/bitmap-field.png 2x" alt="">
```
rachel-fenichel, line 16: Add "a screenshot of the bitmap field plugin" here, and similar for others, as the alt text.

gh-pages/_index.html
```diff
@@ -0,0 +1,472 @@
...
+    </div>
+  </a>
+  <a class="card" href="plugins/field-colour/test/index.html">
+    <img srcset="media/colour-field.png 2x" alt="">
+    <div>
+      <h4>
+        Colour Field
+      </h4>
+      <p>
+        A field that lets users choose a colour. Although the core Blockly library includes this field, it will soon be removed, so developers should use this plugin instead.
```
rachel-fenichel: "soon be removed" -> "removed before the end of 2023"

(so that we don't have to quibble about "soon")

gh-pages/_index.html
```diff
@@ -0,0 +1,472 @@
...
+    </div>
+  </a>
+  <a class="card" href="plugins/block-shareable-procedures/test/index.html">
+    <img srcset="media/procedure-blocks.png 2x">
+    <div>
+      <h4>
+        Procedure Blocks
+      </h4>
+      <p>
+        These blocks allow users to define and call procedures. Although the core Blockly library includes procedure blocks that behave identically for end users, developers that don't need backwards compatibility should use these instead.
```
rachel-fenichel: @BeksOmega this phrasing sounds like the two sets of blocks are the same. Do you have suggestions on short ways to explain the differences?

scripts/gh-predeploy.js
```diff
@@ -576,15 +563,12 @@ function createIndexPage(isLocal) {
...
-
-  <script src="https://cdn.jsdelivr.net/npm/fuse.js/dist/fuse.js"></script>
-  <script src="js/index.js"></script>
 </body>
 
 </html>
 `;
 
-  let contents = injectHeader(indexBase, 'Plugins | blockly-samples', isLocal);
+  let contents = injectHeader(indexBase, 'Blockly Plugins & Demos', isLocal);
```
rachel-fenichel, line 571: Since you updated the page to be all HTML, is there any reason to inject the header and footer here vs just putting it all into the index page and not preprocessing index at all? 

Two reasons I could think of offhand:
- Making sure that the relative and absolute links work correctly.
- Being able to change headers and footers in one place (but we rarely change them so having to modify in two places seems like a reasonable trade-off to have the index page easy to open and edit in one place.

rachel-fenichel:
This looks great! It makes the page much more usable and much prettier.

gh-pages/_index.html
BeksOmega: How 'bouts:

"These blocks allow users to define and call procedures. These have more functionality than the equivalent blocks in core, which allows them to be shared between workspaces. The Blockly team recommends that developers use these instead."

It's not exactly shorter but I think it does explain the difference?

gh-pages/_index.html
gonfunko: Per https://www.w3.org/WAI/tutorials/images/decorative/, I think leaving it empty is generally the recommended approach in this case, since the adjacent title/description of the plugin provide information about it.

gh-pages/_index.html
gonfunko: Done.

gh-pages/_index.html
gonfunko: Done!

scripts/gh-predeploy.js
gonfunko: I think I prefer to leave it as-is for exactly those reasons, but happy to bundle everything into the page if you prefer!

rachel-fenichel (approved):

PR #1809, opened 2023-07-21
fix: Fix the positioning of the dropdown div relative to the bitmap field.

### LINKED ISSUE #1720: The DropDownDiv was off the right position of field-bitmap
**Category**
 - Plugins

**Component**

 The name of the component that has a bug (@blockly/field-bitmap Demo). 

**Describe the bug**

when add more than one field, then the DropDownDiv was off the right position.

**To Reproduce**

1. npm install @blockly/field-bitmap --save
2. the init block code as below:
```
 Blockly.Blocks["Nixie_tube"] = {
  init: function () {
    this.appendDummyInput()
      .appendField("D1: ")
      .appendField(new FieldBitmap(), "D1")
    this.appendDummyInput()
      .appendField("D2: ")
      .appendField(new FieldBitmap(), "D2")
    this.appendDummyInput()
      .appendField("D3: ")
      .appendField(new FieldBitmap(), "D3");
    this.setInputsInline(true);
    this.setPreviousStatement(true, null);
    this.setNextStatement(true, null);
  }
};

```
3. Click on D3 field to open DropDownDiv. 
4. See error "DropDownDiv" was off the right position.

**Expected behavior**
Fix this bug ASAP

**Screenshots**
bitmap_bug

gonfunko:
Fixes #1720. This PR adjusts the way that the bitmap field determines its bounding box, which is in turn used to determine the position of the drop down DIV when it's shown. Manually tested on Chrome, Safari and Firefox at different zoom levels.

### DIFF
--- a/plugins/field-bitmap/src/field-bitmap.js
+++ b/plugins/field-bitmap/src/field-bitmap.js
@@ -212,6 +212,13 @@ export class FieldBitmap extends Blockly.Field {
     return true;
   }
 
+  getScaledBBox() {
+    const boundingBox = this.fieldGroup_.getBoundingClientRect();
+    return new Blockly.utils.Rect(
+        boundingBox.top, boundingBox.bottom, boundingBox.left,
+        boundingBox.right);
+  }
+
   /**
    * Creates the bitmap editor and add event listeners.
    * @returns {!Element} The newly created dropdown menu.


BeksOmega (approved):
I'm kinda surprised this isn't what we're doing in core? It looks like it might have some logic in there to handle's zelo's full-block fields that is making everything weird =(

LGTM!

PR #1824, opened 2023-07-24
feat: Created a minimap that is positionable in the primary workspace

cesarades:
**Description**
I created a minimap that is positionable in the primary workspace. I did so by adding a sibling node to the primary workspace and then using absolute CSS positioning to place it correctly.

**Tests**
I updated the testing so that it includes a positioned and an unpositioned minimap use case.

### DIFF
--- a/plugins/workspace-minimap/src/minimap.ts
+++ b/plugins/workspace-minimap/src/minimap.ts
@@ -27,8 +27,8 @@ const BlockEvents = new Set([
  * your code looks like, and how it is organized.
  */
 export class Minimap {
-    private primaryWorkspace: Blockly.WorkspaceSvg;
-    private minimapWorkspace: Blockly.WorkspaceSvg;
+    protected primaryWorkspace: Blockly.WorkspaceSvg;
+    protected minimapWorkspace: Blockly.WorkspaceSvg;
     private onMouseMoveWrapper: Blockly.browserEvents.Data;
     /**
      * Constructor for a minimap.
@@ -48,36 +48,41 @@ export class Minimap {
       minimapWrapper.className = 'minimapWrapper';
 
       // Make the wrapper a sibling to the primary injection div.
-      const developerDiv = this.primaryWorkspace.getInjectionDiv().parentNode;
-      developerDiv.appendChild(minimapWrapper);
+      const primaryInjectParentDiv =
+        this.primaryWorkspace.getInjectionDiv().parentNode;
+      primaryInjectParentDiv.appendChild(minimapWrapper);
 
-      // Inject the minimap workspace
+      // Inject the minimap workspace.
       this.minimapWorkspace = Blockly.inject(minimapWrapper.id,
           {
-            readOnly: true,
             // Inherit the layout of the primary workspace.
             rtl: this.primaryWorkspace.RTL,
-            // Inlcude the scrollbars so that internal scrolling is enabled and
+            // Include the scrollbars so that internal scrolling is enabled and
             // remove direct interaction with the minimap workspace.
             move: {
               scrollbars: true,
               drag: false,
               wheel: false,
             },
-            // Removes the scale bounds of the minimap so that it can
-            // correctly resize.
+            // Remove the scale bounds of the minimap so that it can
+            // correctly zoomToFit.
             zoom: {
               maxScale: null,
               minScale: null,
             },
+            readOnly: true,
           });
 
       this.minimapWorkspace.scrollbar.setContainerVisible(false);
       this.primaryWorkspace.addChangeListener((e) => void this.mirror(e));
+
+      // The mouseup binds to the parent container div instead of the minimap
+      // because if a drag begins on the minimap and ends outside of it the
+      // mousemove should still unbind.
       Blockly.browserEvents.bind(
           this.minimapWorkspace.svgGroup_, 'mousedown', this, this.onClickDown);
       Blockly.browserEvents.bind(
-          developerDiv, 'mouseup', this, this.onClickUp);
+          primaryInjectParentDiv, 'mouseup', this, this.onClickUp);
     }
 
 
@@ -177,20 +182,5 @@ export class Minimap {
      */
     private onMouseMove(event: PointerEvent): void {
       this.primaryScroll(event);
-    }
-
-    /**
-     * Gets the injection div for the minimap workspace.
-     * @returns The injection div element for the minimap.
-     */
-    getMinimapInjectionDiv(): Element {
-      return this.minimapWorkspace.getInjectionDiv();
-    }
-
-    /**
-     * Resizes the minimap svg.
-     */
-    svgResize(): void {
-      Blockly.svgResize(this.minimapWorkspace);
     }
 }
--- a/plugins/workspace-minimap/src/minimap.ts
+++ b/plugins/workspace-minimap/src/minimap.ts
@@ -27,43 +27,54 @@ const BlockEvents = new Set([
  * your code looks like, and how it is organized.
  */
 export class Minimap {
-    protected primaryWorkspace: Blockly.WorkspaceSvg;
-    protected minimapWorkspace: Blockly.WorkspaceSvg;
+    private primaryWorkspace: Blockly.WorkspaceSvg;
+    minimapWorkspace: Blockly.WorkspaceSvg;
     private onMouseMoveWrapper: Blockly.browserEvents.Data;
     /**
-     * Constructor for a minimap
-     * @param workspace The workspace to mirror
+     * Constructor for a minimap.
+     * @param workspace The workspace to mirror.
      */
     constructor(workspace: Blockly.WorkspaceSvg) {
       this.primaryWorkspace = workspace;
-      const options = {
-        readOnly: true,
-        move: {
-          scrollbars: {
-            vertical: true,
-            horizontal: true,
-          },
-          drag: false,
-          wheel: false,
-        },
-      };
-      this.minimapWorkspace = Blockly.inject('minimapDiv', options);
     }
 
     /**
      * Initialize.
      */
     init(): void {
+      // Create a wrapper div for the minimap injection.
+      const minimapWrapper = document.createElement('div');
+      minimapWrapper.id = 'minimapWrapper_' + this.primaryWorkspace.id;
+      minimapWrapper.className = 'minimapWrapper';
+
+      // Make the wrapper a sibling to the primary injection div.
+      const developerDiv = this.primaryWorkspace.getInjectionDiv().parentNode;
+      developerDiv.appendChild(minimapWrapper);
+
+      // Inject the minimap workspace
+      this.minimapWorkspace = Blockly.inject(minimapWrapper.id,
+          {
+            readOnly: true,
+            move: {
+              scrollbars: true,
+              drag: false,
+              wheel: false,
+            },
+            zoom: {
+              maxScale: null,
+              minScale: null,
+            },
+          });
+
       this.minimapWorkspace.scrollbar.setContainerVisible(false);
       this.primaryWorkspace.addChangeListener((e) => void this.mirror(e));
-      window.addEventListener('resize', () => {
-        this.minimapWorkspace.zoomToFit();
-      });
+      this.minimapWorkspace.zoomToFit();
       Blockly.browserEvents.bind(
           this.minimapWorkspace.svgGroup_, 'mousedown', this, this.onClickDown);
       Blockly.browserEvents.bind(
-          this.minimapWorkspace.svgGroup_, 'mouseup', this, this.onClickUp);
+          developerDiv, 'mouseup', this, this.onClickUp);
     }
+
 
     /**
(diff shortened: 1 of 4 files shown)


plugins/workspace-minimap/src/minimap.ts
```diff
@@ -27,44 +27,55 @@ const BlockEvents = new Set([
...
+              minScale: null,
+            },
+          });
+
       this.minimapWorkspace.scrollbar.setContainerVisible(false);
       this.primaryWorkspace.addChangeListener((e) => void this.mirror(e));
-      window.addEventListener('resize', () => {
-        this.minimapWorkspace.zoomToFit();
-      });
+      this.minimapWorkspace.zoomToFit();
```
BeksOmega: Is this call necessary?

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -27,44 +27,55 @@ const BlockEvents = new Set([
  * your code looks like, and how it is organized.
  */
 export class Minimap {
-    protected primaryWorkspace: Blockly.WorkspaceSvg;
-    protected minimapWorkspace: Blockly.WorkspaceSvg;
+    private primaryWorkspace: Blockly.WorkspaceSvg;
+    minimapWorkspace: Blockly.WorkspaceSvg;
```
BeksOmega: Do you want the minimap workspace to be public?

It looks like the only things you're doing are setting the CSS of the injection div and triggering a resize! You can probably create methods for those on the minimap class and keep the `minimapWorkspace` `private` or `protected.

plugins/workspace-minimap/src/positioned_minimap.ts
```diff
@@ -14,23 +14,131 @@ import {Minimap} from './minimap';
...
     init(): void {
-    //   this.work
+      this.primaryWorkspace.getComponentManager().addComponent({
+        component: this,
+        weight: 3,
+        capabilities: [Blockly.ComponentManager.Capability.POSITIONABLE],
+      });
+      this.minimap.init();
+      this.setMinimapCss();
+      this.primaryWorkspace.resize();
```
BeksOmega, line 49: Is this call necessary?

plugins/workspace-minimap/src/positioned_minimap.ts
```diff
@@ -14,23 +14,131 @@ import {Minimap} from './minimap';
...
+            this.top = savedPositions[i].top - this.height - this.margin;
+          } else {
+            this.top = savedPositions[i].bottom + this.margin;
+          }
+          // Recheck other savedPositions.
+          boundingRect = this.getBoundingRectangle();
+          i = -1;
+        }
+      }
+      this.setMinimapCss();
```
BeksOmega, line 126: Hm you might want to trigger an `svgResize` here as well. Can you check when `position` is being called and make sure you're:
  1) Calling `svgResize` when you need to.
  2) Not calling `svgResize` when you don't.

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -27,44 +27,55 @@ const BlockEvents = new Set([
...
+            readOnly: true,
+            move: {
+              scrollbars: true,
+              drag: false,
+              wheel: false,
+            },
+            zoom: {
+              maxScale: null,
+              minScale: null,
+            },
```
BeksOmega: Can you add some two comments here, one about the move options and one about the zoom options? If you don't have context the reason these are set this way would be confusing!

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -27,44 +27,55 @@ const BlockEvents = new Set([
...
+      const minimapWrapper = document.createElement('div');
+      minimapWrapper.id = 'minimapWrapper_' + this.primaryWorkspace.id;
+      minimapWrapper.className = 'minimapWrapper';
+
+      // Make the wrapper a sibling to the primary injection div.
+      const developerDiv = this.primaryWorkspace.getInjectionDiv().parentNode;
+      developerDiv.appendChild(minimapWrapper);
+
+      // Inject the minimap workspace
+      this.minimapWorkspace = Blockly.inject(minimapWrapper.id,
```
BeksOmega, line 56: Also can you check that this works properly in RTL? I think you probably need to set the RTL value of the minimap to match the primary workspace. If it's not working that can be fixed in a follow up!

plugins/workspace-minimap/src/positioned_minimap.ts
```diff
@@ -14,23 +14,131 @@ import {Minimap} from './minimap';
...
+     * Positions the minimap.
+     * @param metrics The workspace metrics.
+     * @param savedPositions List of rectangles already on the workspace.
+     */
+    position(metrics: Blockly.MetricsManager.UiMetrics,
+        savedPositions: Blockly.utils.Rect[]): void {
+      const hasVerticalScrollbars = this.primaryWorkspace.scrollbar &&
+          this.primaryWorkspace.scrollbar.canScrollHorizontally();
+      const hasHorizontalScrollbars = this.primaryWorkspace.scrollbar &&
+          this.primaryWorkspace.scrollbar.canScrollVertically();
```
BeksOmega: Are these checking the correct scrollbars? I know this matches what the backpack is doing, but I am a bit confused haha.

BeksOmega (changes requested):
Looks great! So excited to see the minimap being mini :D

One more top-level comment: I think there should be some unit tests for the positioning logic, since it should be pretty easy to pass it dummy data =) Can be handled in a follow up PR though!

plugins/workspace-minimap/src/minimap.ts
cesarades: Nope, removing it now

plugins/workspace-minimap/src/minimap.ts
cesarades: Great idea will do that

plugins/workspace-minimap/src/minimap.ts
cesarades: I now have it inheriting the RTL value of the primary workspace and it works as intended.

plugins/workspace-minimap/src/positioned_minimap.ts
cesarades: The setMinimapCss is not, the other two are.

plugins/workspace-minimap/src/positioned_minimap.ts
cesarades: Ok, so position is called every time there's a window resize. That said, I think you can just embed the minimap svgResize into the this.setMinimapCss. This refactors the code and also makes it so that it is only being called when necessary.

plugins/workspace-minimap/src/positioned_minimap.ts
cesarades: Yes, they are checking the scrollbars in the primary workspace. It needs that to offset the location of the minimap in case they are present.

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -27,44 +27,60 @@ const BlockEvents = new Set([
...
      * Initialize.
      */
     init(): void {
+      // Create a wrapper div for the minimap injection.
+      const minimapWrapper = document.createElement('div');
+      minimapWrapper.id = 'minimapWrapper_' + this.primaryWorkspace.id;
+      minimapWrapper.className = 'minimapWrapper';
+
+      // Make the wrapper a sibling to the primary injection div.
+      const developerDiv = this.primaryWorkspace.getInjectionDiv().parentNode;
```
gonfunko: `workspaceParentDiv` might be a bit clearer

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -27,44 +27,60 @@ const BlockEvents = new Set([
...
       this.minimapWorkspace.scrollbar.setContainerVisible(false);
       this.primaryWorkspace.addChangeListener((e) => void this.mirror(e));
-      window.addEventListener('resize', () => {
-        this.minimapWorkspace.zoomToFit();
-      });
       Blockly.browserEvents.bind(
           this.minimapWorkspace.svgGroup_, 'mousedown', this, this.onClickDown);
       Blockly.browserEvents.bind(
-          this.minimapWorkspace.svgGroup_, 'mouseup', this, this.onClickUp);
+          developerDiv, 'mouseup', this, this.onClickUp);
```
gonfunko: A comment about why mousedown is bound to the minimap while mouseup is bound to its parent container div might be helpful here

plugins/workspace-minimap/src/positioned_minimap.ts
```diff
@@ -14,23 +14,130 @@ import {Minimap} from './minimap';
 
 /**
  * A positionable version of minimap that implements IPositionable.
- * @param {Minimap} minimap The minimap to position.
- * @implements {Blockly.IPositionable}
  */
-export class PositionedMinimap {
+export class PositionedMinimap implements Blockly.IPositionable {
```
gonfunko: It feels a bit odd that this has-a vs is-a `Minimap`; can you either add a comment explaining the reasoning for that, or make it a subclass of Minimap if possible?

plugins/workspace-minimap/src/positioned_minimap.ts
BeksOmega: I'm just confused why `hasVerticalScrollbars` is based on `canScrollHorizontally`. I'm pretty sure the vertical scrollbars affect the vertical scroll.

plugins/workspace-minimap/src/positioned_minimap.ts
BeksOmega: What is the `primaryWorkspace.resize()` doing? I wouldn't expect adding a minimap to affect the size of the primary workspace.

plugins/workspace-minimap/src/positioned_minimap.ts
gonfunko: I think Beka may have been referring to how hasVerticalScrollbars is checking canScroll*Horizontally*, and vice versa, vs checking the primary workspace scrollbars as opposed to the minimap scrollbars. That was (is) confusing to me at least - if intentional, a comment explaining the reasoning would be helpful!

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -27,44 +27,60 @@ const BlockEvents = new Set([
...
+      const developerDiv = this.primaryWorkspace.getInjectionDiv().parentNode;
+      developerDiv.appendChild(minimapWrapper);
+
+      // Inject the minimap workspace
+      this.minimapWorkspace = Blockly.inject(minimapWrapper.id,
+          {
+            readOnly: true,
+            // Inherit the layout of the primary workspace.
+            rtl: this.primaryWorkspace.RTL,
+            // Inlcude the scrollbars so that internal scrolling is enabled and
```
```suggestion
            // Include the scrollbars so that internal scrolling is enabled and
```

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -162,4 +178,19 @@ export class Minimap {
...
+    getMinimapInjectionDiv(): Element {
+      return this.minimapWorkspace.getInjectionDiv();
+    }
+
+    /**
+     * Resizes the minimap svg.
+     */
+    svgResize(): void {
+      Blockly.svgResize(this.minimapWorkspace);
+    }
```
BeksOmega: Sweet, love the reorganization =)

plugins/workspace-minimap/src/minimap.ts
BeksOmega: Comments look perfect!

plugins/workspace-minimap/src/positioned_minimap.ts
cesarades: You're right in that adding a minimap does not affect the size of the primary workspace, but the resize function doesn't only resize the workspace. It also deals with positioning. "This should be called when something changes that requires recalculating dimensions and positions of the trash, zoom, toolbox, etc."

Without this the minimap does not show up.

plugins/workspace-minimap/src/positioned_minimap.ts
cesarades: That's a great point! Will do.

plugins/workspace-minimap/src/positioned_minimap.ts
cesarades: Ok, so this is in fact an error. I've corrected it so that vertical goes with vertical and vise versa. Also, I added another use case that was no previously checked. If the scrollbar exists and scrolling is enabled, but the scrollbar is not visible, it should adjust for that.

BeksOmega (approved):
LGTM =)

gonfunko (approved):

plugins/workspace-minimap/src/positioned_minimap.ts
gonfunko: Please also file a bug to update the backpack plugin similarly!

PR #1839, opened 2023-08-02
feat: update codelabs to use blockly v10

maribethb:
- Updates custom-generator codelab to use v10, including codelab text
- Updates custom-renderer codelab to use v10; updates to generator file matches latest sample app file
- Updates keyboard-nav codelab to v10
    - this was partially updated previously, but the version wasn't bumped, and the index.js file wasn't updated, so the codelab code just didn't work at all in its previous state
    - also fixed a typo in the markdown
    - also removed spurious custom renderer option in the workspace options
- Updates tooltips demo to latest version of dev-tools which resolves a peer dependency issue when trying to install the demo
- No longer any traces of blockly v9 in samples!


codelabs/custom_generator/custom_generator.md
```diff
@@ -257,12 +257,12 @@ This error occurs because there has to be a block generator for each type of blo
 
 ## Block generator overview
 
-At its core, a block generator is a function that takes in a block, translates the block into code, and returns that code as a string.
+At its core, a block generator is a function that takes in a block and the language generator instance, translates the block into code, and returns that code as a string.
 
-Block generators are defined on the language generator object. For instance, here is the code to add a block generator for blocks of type `sample_block` on a language generator object (`sampleGenerator`).
+Each language generator has a property called `forBlock`, which is an object where all block generator functions must be placed. For instance, here is the code to add a block generator for blocks of type `sample_block` on a language generator object (`sampleGenerator`).
```
cpcallen: Maybe "dictionary object"?

codelabs/custom_generator/custom_generator.md
```diff
@@ -257,12 +257,12 @@ This error occurs because there has to be a block generator for each type of blo
 
 ## Block generator overview
 
-At its core, a block generator is a function that takes in a block, translates the block into code, and returns that code as a string.
+At its core, a block generator is a function that takes in a block and the language generator instance, translates the block into code, and returns that code as a string.
```
cpcallen: Maybe "(and optionally the language generator instance)", since there's no obligation to use the `CodeGenerator `instance (and indeed many generator functions won't need it).

codelabs/custom_generator/custom_generator.md
```diff
@@ -286,12 +286,13 @@ sampleGenerator['left_turn_block'] = function(block) {
 Value blocks represent code that returns a value.
 
 A value block's generator returns an array containing a string and a [precedence value](https://developers.google.com/blockly/guides/create-custom-blocks/operator-precedence).
+The built-in generators have predefined operator precedence values exported as an `Order` enum.
```
cpcallen: Should there be a paragraph break before this line?  Otherwise you could concatenate it onto the the previous line since we are apparently not doing 80-character line length limits.

codelabs/custom_generator/custom_generator.md
```diff
@@ -327,8 +328,8 @@ No matter what, it generates the code `'null'`. Notice that this is a string, be
 
 
 ```js
-jsonGenerator['logic_null'] = function(block) {
-  return ['null', jsonGenerator.PRECEDENCE];
+jsonGenerator.forBlock['logic_null'] = function(block, generator) {
+  return ['null', generator.PRECEDENCE];
```
cpcallen: Attaching this comment here, but it pertains also to the the "### Operator precedence" section: I think the codelab should declare a `const Order {ATOMIC: 0};` and then use `Order.ATOMIC`, rather than creating and using a `jsonGenerator.PRECEDENCE`.

examples/custom-generator-codelab/src/generators/json.js
```diff
@@ -24,53 +24,53 @@ jsonGenerator.scrub_ = function(block, code, thisOnly) {
   return code;
 };
 
-jsonGenerator['logic_null'] = function(block) {
-  return ['null', jsonGenerator.PRECEDENCE];
+jsonGenerator.forBlock['logic_null'] = function(block, generator) {
+  return ['null', generator.PRECEDENCE];
```
BeksOmega: I think we should probably switch this over to use an order enum like the other language generators? It seems like having the generator in this codelab work differently would get confusing if folks are looking at this + the code we have in core to help them build their generator.

BeksOmega (approved):
LGTM =) just one question!

codelabs/custom_generator/custom_generator.md
cpcallen, line 1: What do you think about updating this codelab to create a  `JsonGenerator` class and then a (mostly trivial) `jsonGenerator` instance, to reflect the way that the 1st party generators are created?  I think in particular such an example would be useful to TypeScript devs (though getting the 1p generators updated to TS would be even more so…)

codelabs/custom_generator/custom_generator.md
maribethb: Changed, but in general not interested in hand-formatting files unless it's really egregious :P

codelabs/custom_generator/custom_generator.md
maribethb: Eh, I'm torn, so definitely don't want to do this in this PR. I'm not sure we should do it in general, because I think most people never actually need to use both the class and a separate instance, so why do we need to complicate the codelab. But I could go either way on that so I'm not opposed to it if someone wants to do it as a follow up.

maribethb:
Could one of you take another look since I added more English? The sample code compiles, but unfortunately we don't have a prose compiler to yell at me about typos...

BeksOmega (approved):

codelabs/custom_generator/custom_generator.md
cpcallen: Filed #1858 for further discussion.

codelabs/custom_generator/custom_generator.md
cpcallen: The question was more about whether you had omitted the paragraph break, rather than requesting a formatting change—but I can see that I could have worded it more clearly.

PR #1842, opened 2023-08-03
feat: have backpack hide scrollbars when flyout opens

### LINKED ISSUE #1841: Have the backpack hide the workspace's scrollbars when the flyout opens
**Category**

 - Plugins

**Component**

workspace-backpack

Having multiple sets of overlapping scrollbars is confusing to users. 

Temporarily hide the scrollbars in the main workspace while the backpack flyout is open.

N/A

**Additional context**

Requested by App Inventor.

BeksOmega:
Fixes #1841 

### Testing

Opened and closed the backpack and observed that scrollbars where hidden and shown.

### Additional Info

Dependent on v10.1

### DIFF
--- a/plugins/workspace-backpack/src/backpack.js
+++ b/plugins/workspace-backpack/src/backpack.js
@@ -679,7 +679,12 @@ export class Backpack extends Blockly.DragTarget {
     }
     const jsons = this.contents_.map((text) => JSON.parse(text));
     this.flyout_.show(jsons);
-    this.workspace_.scrollbar?.setVisible(false);
+    // TODO: We can remove the setVisible check when updating from ^10.0.0 to
+    //    ^11.
+    if (this.workspace_.scrollbar &&
+    /** @type {*} */ (this.workspace_.scrollbar).setVisible) {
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):

BeksOmega:
@maribethb I need to make this dependent on `^10.1.0` do you think that's worth doing a breaking change for (I kinda don't)? I can also just check if `setVisible` exists before calling it. That makes this backwards compatible with `v10`.

BeksOmega:
> @maribethb I need to make this dependent on `^10.1.0` do you think that's worth doing a breaking change for (I kinda don't)? I can also just check if `setVisible` exists before calling it. That makes this backwards compatible with `v10`.

I decided to add a check for `setVisible`

maribethb:
> @maribethb I need to make this dependent on `^10.1.0` do you think that's worth doing a breaking change for (I kinda don't)? I can also just check if `setVisible` exists before calling it. That makes this backwards compatible with `v10`.

I think adding the check to make it backwards compatible is best option. Otherwise, yeah, it would have to be a breaking change, otherwise npm might update the plugin automatically to an incompatible version :/

Issue #1837, opened 2023-07-31
Upgrade the keyboard navigation plugin to use the new copy paste system


BeksOmega:
The keyboard navigation plugin uses the copy-paste system in core, which is being refactored. The plugin needs to be upgraded to use the new API.

The new API will be released at the end of Q3 2023.

PR #1844, opened 2023-08-03
fix: upgrade keyboard navigation to use new clipboard API

BeksOmega:
Closes https://github.com/google/blockly-samples/issues/1837

Upgrades keyboard navigation to use the new clipboard API.


plugins/keyboard-navigation/package.json
```diff
@@ -41,15 +41,15 @@
   "devDependencies": {
     "@blockly/dev-scripts": "^2.0.1",
     "@blockly/dev-tools": "^7.0.2",
-    "blockly": "^10.0.0",
+    "blockly": "^10.2.0-beta.0",
```
cpcallen, line 44: Looks like new API is expected in 10.2.  Perhaps update the bug, which says it will be in 11?

plugins/keyboard-navigation/test/navigation_modify_test.mocha.js
```diff
@@ -87,6 +87,7 @@ suite('Insert/Modify', function() {
         </xml>`;
 
     defineTestBlocks();
+    window.cancelAnimationFrame = function() {};
```
cpcallen: This would seem to warrant a comment.  Also: should it not be undone once the suite is over?  Failing to do so seems like it could cause problems if these were amalgamated into a larger test suite.

(Here and in `navigation_test.mocha.js`.)

cpcallen (approved):

plugins/keyboard-navigation/package.json
BeksOmega: I'll just change it to eoq3 since we don't know what the version is going to be.

Issue #1679, opened 2023-04-27
Remove less useful examples


maribethb:
**Category**

 - Examples

**Component**

- [x] blockly-node - basically shows importing blockly through node's `require`. This is not unique to blockly and just how you use packages in node. This one is arguably useful as a demo of headless mode, but it's really not doing anything interesting. (#1689)
- [x] blockly-requirejs - again, shows how to import blockly using AMD modules (requirejs). This is not specific to blockly and anyone wanting to use requirejs should look to them for examples.
- [x] blockly-umd - this one loads blockly in a script tag out of `node_modules` which is not a pattern we want people to replicate
- [x] blockly-vue - presumably outdated since we have a vue3 example. has several peer-dependency problems.
- [x] blockly-webpack - same argument as the first two. also, we have the sample-app now which uses webpack. (#1694)

Arguable ones:

- [ ] blockly-svelte - really basic example, basically just calls `inject` inside an `onMount` svelte hook. This hasn't been updated since it was added.
- [ ] blockly-vue3 - again, pretty basic example that isn't doing particularly blockly-specific.
- [ ] blockly-angular - this is using an old version of angular, there are a lot of peer-dependency issues with it, there is a lot of boilerplate code, and in the end basically it's just putting the normal blockly inject code into a `ngOnInit` call in a blockly component. so it's still not a very interesting example.
- [ ] blockly-react - similar arguments, and also maybe we should just tell people to use the `react-blockly` project instead of providing our own example.

Many of these examples are outdated and not that useful. We don't need an example of how to build a project using parcel. Parcel has their own examples and there is nothing unique to Blockly in the example. The same argument applies to many of these.

Keeping these examples around adds to our maintenance burden and I don't believe we are getting a lot of value out of them in return. Previously, this wasn't as much of an issue because we basically ignored `examples` in day-to-day maintenance of blockly-samples. Some of the examples don't even have a functioning dependency tree (e.g. the parcel example has a dependency that fails to build). This was ignored unless publishing examples or trying to run an individual example. If we move to using yarn workspaces,  then installing at the root of blockly-samples will require that at the very least all examples can install their own dependencies. 

If plugins were not in the same repository as examples, it would be easier to justify keeping the marginally-useful examples as the maintenance burden again drops to near zero (other than the guilt of having outdated examples).

maribethb:
We decided to remove the ones in the first list, and argue about the ones in the second list later.

pavi2410:
[...]

PR #1847, opened 2023-08-04
chore: Remove some less useful demos.

gonfunko:
This PR fixes part of #1679 by removing the blockly-requirejs, blockly-umd, and blockly-vue packages.


cpcallen (approved):
Can you check devsite to make sure that we do not link to any of the removed demos?

gonfunko:
> Can you check devsite to make sure that we do not link to any of the removed demos?

I grepped through everything and didn't find references to these but did find a reference to the previously-removed blockly-node and blockly-webpack. Sent you a CL fixing those.

PR #1848, opened 2023-08-04
feat: Implement focus mode

cesarades:
**Description**
Created a focus mode class that adds a masking svg to the minimap, highlighting the user's viewport.

**Tests**
Ran manual tests by scrolling in the primary workspace and ensuring the minimap updated, as well as interacting with the minimap and checking that the highlight was centered on the click.

### DIFF
--- a/plugins/workspace-minimap/src/focus_mode.ts
+++ b/plugins/workspace-minimap/src/focus_mode.ts
@@ -11,6 +11,13 @@
 
 import * as Blockly from 'blockly/core';
 
+const BlockEvents = new Set([
+  Blockly.Events.VIEWPORT_CHANGE,
+  Blockly.Events.BLOCK_CHANGE,
+  Blockly.Events.BLOCK_CREATE,
+  Blockly.Events.BLOCK_DELETE,
+  Blockly.Events.BLOCK_DRAG,
+  Blockly.Events.BLOCK_MOVE]);
 
 /**
  * A class that highlights the user's viewport on the minimap.
@@ -24,6 +31,7 @@ export class FocusMode {
     private background: SVGElement;
     private id: string;
 
+
     /**
      * Constructor for focus mode.
      * @param primary The primary workspaceSvg.
@@ -35,21 +43,22 @@ export class FocusMode {
       this.id = this.minimapWorkspace.id;
     }
 
+
     /**
      * Initializes focus mode.
      */
     init() {
-      // Make the group element.
+      // Make the svg group element.
       this.svgGroup = Blockly.utils.dom.createSvgElement(
           Blockly.utils.Svg.G, {'class': 'focusMode'}, null);
 
-      // Mask under group.
+      // Make the mask under the svg group.
       const mask = Blockly.utils.dom.createSvgElement(
           new Blockly.utils.Svg('mask'),
           {'id': 'focusModeMask' + this.id},
           this.svgGroup);
 
-      // Backround under group.
+      // Make the backround under the svg group.
       this.background = Blockly.utils.dom.createSvgElement(
           Blockly.utils.Svg.RECT, {
             'x': 0,
@@ -59,7 +68,7 @@ export class FocusMode {
             'mask': 'url(#focusModeMask' + this.id + ')',
           }, this.svgGroup);
 
-      // White layer under mask.
+      // Make the white layer under the svg mask.
       Blockly.utils.dom.createSvgElement(
           Blockly.utils.Svg.RECT, {
             'x': 0,
@@ -69,7 +78,7 @@ export class FocusMode {
             'fill': 'white',
           }, mask);
 
-      // Black layer under mask.
+      // Make the black layer under the mask.
       this.rect = Blockly.utils.dom.createSvgElement(
           Blockly.utils.Svg.RECT, {
             'x': 0,
@@ -81,10 +90,8 @@ export class FocusMode {
 
       // Theme.
       this.background.setAttribute('fill', '#e6e6e6');
-      this.svgGroup.setAttribute('opacity', '1');
-      this.svgGroup.style.transition = 'opacity ' + .25 + 's';
 
-      // Add the group to the minimap.
+      // Add the svg group to the minimap.
       const parentSvg = this.minimapWorkspace.getParentSvg();
       if (parentSvg.firstChild) {
         parentSvg.insertBefore(this.svgGroup, parentSvg.firstChild);
@@ -92,20 +99,13 @@ export class FocusMode {
         parentSvg.appendChild(this.svgGroup);
       }
 
+      window.addEventListener('resize', () => void this.update());
       this.onChangeWrapper = this.onChange.bind(this);
       this.primaryWorkspace.addChangeListener(this.onChangeWrapper);
 
-      const primaryMetrics = this.primaryWorkspace.getMetricsManager();
-      const minimapMetrics = this.minimapWorkspace.getMetricsManager();
-
-      this.position(primaryMetrics, minimapMetrics);
-      this.resize(primaryMetrics, minimapMetrics);
-
-      window.addEventListener('resize', () => {
-        this.resize(this.primaryWorkspace.getMetricsManager(),
-            this.minimapWorkspace.getMetricsManager());
-      });
+      this.update();
     }
+
 
     /**
      * Disposes of the focus mode.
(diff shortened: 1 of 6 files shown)


plugins/workspace-minimap/src/index.ts
```diff
@@ -12,3 +12,4 @@
  */
 export {Minimap} from './minimap';
 export {PositionedMinimap} from './positioned_minimap';
+export {FocusMode} from './focus_mode';
```
BeksOmega: Do you want folks to have access to `new FocusMode` externally? Or do you just want them to access it through `focus` and `unfocus`?

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -183,4 +190,18 @@ export class Minimap {
     private onMouseMove(event: PointerEvent): void {
       this.primaryScroll(event);
     }
+
+    /**
+     * Turns on focus mode; A highlight on the user's viewport in the minimap.
+     */
+    focus(): void {
+      this.focusMode.init();
+    }
```
BeksOmega: I think I'd prefer a different name since 'focus' means something specific in the browser world: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus

Maybe, `enableFocusMode` or `setFocusMode(enabled)`?

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,171 @@
...
+
+
+    /**
+     * Constructor for focus mode.
+     * @param primary The primary workspaceSvg.
+     * @param minimap The minimap workspaceSvg.
+     */
+    constructor(primary: Blockly.WorkspaceSvg, minimap: Blockly.WorkspaceSvg) {
+      this.primaryWorkspace = primary;
+      this.minimapWorkspace = minimap;
```
BeksOmega: You can also use parameter properties for these! https://medium.com/@daveford/parameter-properties-in-typescript-c99df6ba13f7

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,171 @@
...
+      this.id = this.minimapWorkspace.id;
```
BeksOmega: Change this to be a random number like discussed.

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,171 @@
...
+  Blockly.Events.BLOCK_CHANGE,
+  Blockly.Events.BLOCK_CREATE,
+  Blockly.Events.BLOCK_DELETE,
+  Blockly.Events.BLOCK_DRAG,
+  Blockly.Events.BLOCK_MOVE]);
+
+/**
+ * A class that highlights the user's viewport on the minimap.
+ */
+export class FocusMode {
```
BeksOmega: Love how you structured this class! Good separation of responsibilities, very elegant =)

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,171 @@
...
+        this.update();
+      }
+    }
+
+
+    /**
+     * Positions and sizes the highlight on the minimap
+     * based on the primary workspace.
+     */
+    private update(): void {
```
BeksOmega, line 146: Can you use a [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) in this method to aliase the `width`s `height`s etc? It may make it more messy :P but I'm hoping it makes this a bit easier to read!

Give it a shot and use your judgement :P

(I'm hoping it lets you conform to the [rectangle rule](https://github.com/google/google-java-format/wiki/The-Rectangle-Rule)).

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,171 @@
...
+            'fill': 'white',
+          }, mask);
+
+      // Make the black layer under the mask.
+      this.rect = Blockly.utils.dom.createSvgElement(
+          Blockly.utils.Svg.RECT, {
+            'x': 0,
+            'y': 0,
+            'rx': Blockly.Bubble.BORDER_WIDTH,
+            'ry': Blockly.Bubble.BORDER_WIDTH,
```
BeksOmega: Is this what the content highlight plugin is using? Either way, I would probably just define a constant in this class :P cuz this is not what that value was intended to be used for haha.

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,171 @@
...
+            'fill': 'black',
+          }, mask);
+
+      // Theme.
```
BeksOmega, line 92: Is this comment to remind yourself to update things to match the theme?

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,171 @@
...
+
+      // Add the svg group to the minimap.
+      const parentSvg = this.minimapWorkspace.getParentSvg();
+      if (parentSvg.firstChild) {
+        parentSvg.insertBefore(this.svgGroup, parentSvg.firstChild);
+      } else {
+        parentSvg.appendChild(this.svgGroup);
+      }
+
+      window.addEventListener('resize', () => void this.update());
```
BeksOmega, line 103: Nice catch!

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,171 @@
...
+
+
+    /**
+     * Disposes of the focus mode.
+     * Unlinks from all DOM elements and remove all event listeners
+     * to prevent memory leaks.
+     */
+    dispose() {
+      if (this.svgGroup) {
+        Blockly.utils.dom.removeNode(this.svgGroup);
```
BeksOmega: You should also remove any references to the svg elements (`svgGroup`, `background`, and `rect`). If there are still references to them, they won't get garbage collected even though they've been disconnected from the dom.

You can remove references by setting the properties to null:
```
this.svgGroup = null;
```

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,171 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview A positionable version of the minimap.
```
BeksOmega: Nit: update comment.

BeksOmega (commented):
Generally looks great! Just a few little comments =)

plugins/workspace-minimap/src/index.ts
cesarades: I only want them to access it through `focus` and `unfocus`. I'll remove that export.

plugins/workspace-minimap/src/focus_mode.ts
cesarades: Yea it is.
I'll make it a constant.

plugins/workspace-minimap/src/focus_mode.ts
cesarades: Yup

plugins/workspace-minimap/src/focus_mode.ts
cesarades: So I created aliases for each metric category but I couldn't find a way to use destructuring to really improve readability.

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -19,12 +20,12 @@ const BlockEvents = new Set([
   Blockly.Events.BLOCK_DRAG,
   Blockly.Events.BLOCK_MOVE]);
 
+const BorderRadius = 6;
```
BeksOmega: Nit: CONSTANT_CASE for constants.

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -183,4 +190,18 @@ export class Minimap {
...
+    enableFocusMode(): void {
+      this.focusMode.init();
+    }
+
+    /**
+     * Turns off focus mode.
+     */
+    disableFocusMode(): void {
+      this.focusMode.dispose();
+    }
```
BeksOmega: Thinking about this from an API design perspective, I think taking in a boolean to enable/disable the focus mode might be a better option. If a developer is modifying the focus mode, they're probably doing that based on another boolean they already have.

For example:
```js
const useFocusMode = shouldUseFocusMode();
if (useFocusMode) {
  myMinimap.enableFocusMode();
} else {
  myMinimap.disableFocusMode();
}
```

vs
```js
myMinimap.setFocusModEnabled(shouldUseFocusMode());
```

But I could see an argument for this being clearer as well! Give it a think and pick whichever one you prefer =)

plugins/workspace-minimap/src/focus_mode.ts
BeksOmega: Looks great!

BeksOmega (commented):
LGTM! But I'll let Aaron handle the approval.

plugins/workspace-minimap/src/minimap.ts
cesarades: Totally see the value of both the options. However, focus mode is on by default so there aren't many use cases for `enable.FocusMode()`. The developer would end up doing:
```
if (!useFocusMode) {
  myMinimap.disableFocusMode();
}
```

On another note, when it comes to API design, is there a reason not to provide both functionalities? That way they could use any of the two methods.

And, would it be good to also add a method that returns if it's enabled not, for example `myMinimap.isEnabled()`?

plugins/workspace-minimap/src/focus_mode.ts
cesarades: I get this eslit error:

Variable name `BORDER_RADIUS` must match one of the following formats: camelCase, PascalCase eslint@typescript-eslint/naming-convention 
const BORDER_RADIUS: 6

BeksOmega: Ha ok well I'm not surprised our eslint doesn't match the style guide :P I would change to camelCase then since PascalCase is used for classes.

plugins/workspace-minimap/src/minimap.ts
BeksOmega: > On another note, when it comes to API design, is there a reason not to provide both functionalities? That way they could use any of the two methods.

Adding multiple ways of doing things has a few draw backs:
  1) More maintainance for us (although in this case it would be small)
  2) More overhead for external developers (they have to decide what's right for their use case)
  3) More overhead for external developers (they have to read it both ways)

This is deifnitely something that people fall on either side of though! E.g. Perl (which Christopher can rave at you about :P ) is very much based around giving programmers the flexibility to do things in whichever way they want to. While Python does the opposite "There should be one-- and preferably only one --obvious way to do it."

I personally fall on the "less is better" side of things!

> And, would it be good to also add a method that returns if it's enabled not, for example myMinimap.isEnabled()?

I think that is a wonderful idea =)

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,186 @@
...
+      this.rect.setAttribute('height', height.toString());
+    }
+
+    /**
+     * Returns whether focus mode is initialized or not.
+     * @returns True if focus mode is initialized else false.
+     */
+    isEnabled(): boolean {
+      return this.initialized;
+    }
```
BeksOmega, line 184: You probably want to add something similar to the minimap as well (e.g. `isFocusEnabled`) because the focus mode is only accessible by subclasses. So if you just have a minimap, you can't tell if its focus mode is enabled or not atm.

cesarades: Yea I plan on doing that; however, I was leaving it for a separate PR.

BeksOmega: Sweet! sounds good =)

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -31,6 +31,7 @@ export class FocusMode {
     private rect: SVGElement;
     private background: SVGElement;
     private id: string;
+    private initialized: boolean;
```
gonfunko: Probably want to initialize this to false here.

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,186 @@
...
+  Blockly.Events.BLOCK_DELETE,
+  Blockly.Events.BLOCK_DRAG,
+  Blockly.Events.BLOCK_MOVE]);
+
+const borderRadius = 6;
+
+/**
+ * A class that highlights the user's viewport on the minimap.
+ */
+export class FocusMode {
```
gonfunko: Optional: I don't love FocusMode as the name of this class - it feels weird to write "new FocusMode()", I think because FocusMode sounds like a state or boolean as opposed to a class responsible for drawing a highlighted region. Maybe something like FocusRegion or FocusRect?

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,186 @@
...
+      this.svgGroup = Blockly.utils.dom.createSvgElement(
+          Blockly.utils.Svg.G, {'class': 'focusMode'}, null);
+
+      // Make the mask under the svg group.
+      const mask = Blockly.utils.dom.createSvgElement(
+          new Blockly.utils.Svg('mask'),
+          {'id': 'focusModeMask' + this.id},
+          this.svgGroup);
+
+      // Make the backround under the svg group.
```
gonfunko: Nit: background

plugins/workspace-minimap/src/focus_mode.ts
```diff
@@ -0,0 +1,186 @@
...
+      let left = (primaryView.left - primaryContent.left) * scale;
+      let top = (primaryView.top - primaryContent.top) * scale;
+
+      // Account for the padding outside the content on the minimap.
+      left += (minimapSvg.width - minimapContent.width) / 2;
+      top += (minimapSvg.height - minimapContent.height) / 2;
+
+      // Set the svg attributes.
+      this.rect.setAttribute('transform',
+          'translate(' + left + ',' + top + ')');
```
gonfunko: This might be a little more readable with backtick syntax: \`translate(${left}, ${top})\`

cesarades: Agreed

gonfunko (approved):

PR #1850, opened 2023-08-09
feat: Minimap responsive sizing

cesarades:
**Description**
Made the minimap size responsive to the window size.
I also did some code styling and added a minimap.dispose() method.

**Tests**
For the sizing I made sure to check if it was responding to the window size correctly. I also made sure the minimap minimum size was not exceeded. For the dispose method, I checked to see that all the relevant DOM elements were removed.


plugins/workspace-minimap/src/minimap.ts
```diff
@@ -31,7 +31,13 @@ export class Minimap {
     protected primaryWorkspace: Blockly.WorkspaceSvg;
     protected minimapWorkspace: Blockly.WorkspaceSvg;
     protected focusRegion: FocusRegion;
-    private onMouseMoveWrapper: Blockly.browserEvents.Data;
+    protected onMouseMoveWrapper: Blockly.browserEvents.Data;
+    protected onMouseDownWrapper: Blockly.browserEvents.Data;
+    protected onMouseUpWrapper: Blockly.browserEvents.Data;
+    protected focusEnabled = false;
```
BeksOmega: I think you can remove this and just return `!!this.focusRegion`

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -40,22 +46,24 @@ export class Minimap {
...
      * Initialize.
      */
     init(): void {
       // Create a wrapper div for the minimap injection.
-      const minimapWrapper = document.createElement('div');
-      minimapWrapper.id = 'minimapWrapper_' + this.primaryWorkspace.id;
-      minimapWrapper.className = 'minimapWrapper';
+      this.minimapWrapper = document.createElement('div');
+      this.minimapWrapper.id =
+        'minimapWrapper' + String(Math.random()).substring(2);
```
BeksOmega: You shouldn't need to add the random string here unless you're applying an svg filter. I think adding the primary workspace ID like before was good =)

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -191,17 +227,35 @@ export class Minimap {
       this.primaryScroll(event);
     }
 
+
     /**
      * Enables the focus region; A highlight of the viewport in the minimap.
      */
     enableFocusRegion(): void {
+      this.focusRegion = new FocusRegion(
```
BeksOmega: Curious why this is now constructing a new focus region each time.

plugins/workspace-minimap/src/positioned_minimap.ts
```diff
@@ -127,6 +135,17 @@ export class PositionedMinimap extends Minimap implements Blockly.IPositionable
       this.setMinimapCss();
     }
 
+
+    /**
+     * Sizes the minimap.
+     */
+    size(): void {
```
BeksOmega: I would probably rename this to `resize`. Technically `size` is a verb so this is fine :P but it sounds like it could also be a getter.

plugins/workspace-minimap/src/positioned_minimap.ts
```diff
@@ -127,6 +135,17 @@ export class PositionedMinimap extends Minimap implements Blockly.IPositionable
...
+      const viewWidth = this.primaryWorkspace.getMetrics().viewWidth;
+      this.width = Math.max(minWidth, viewWidth / 5);
+      this.height = this.width / 1.5;
```
BeksOmega: Could you change this to `this.width * 2/3`? I know they're equivalent but my brain doesn't like flip flopping them :P

plugins/workspace-minimap/src/minimap.ts
```diff
@@ -77,21 +85,44 @@ export class Minimap {
...
-          this.primaryWorkspace, this.minimapWorkspace);
       this.enableFocusRegion();
     }
 
+    /**
+     * Disposes the minimap.
+     * Unlinks from all DOM elements and remove all event listeners
+     * to prevent memory leaks.
+     */
+    dispose() {
```
BeksOmega, line 109: Great catch on this :D

plugins/workspace-minimap/src/positioned_minimap.ts
BeksOmega: Where are these values getting applied?

BeksOmega (commented):
Looks good! Just a few questions =)

plugins/workspace-minimap/src/positioned_minimap.ts
cesarades: In the `setMinimapCss` method (which is called in the position method and when the window resizes)

plugins/workspace-minimap/src/minimap.ts
cesarades: I guess I over-complicated the dispose... no need to do that looking back now

plugins/workspace-minimap/src/positioned_minimap.ts
BeksOmega: If `position` is already getting called when the window resizes, can we call `resize` at the beginning of position instead of adding it as a separate listener? I think that makes the control flow more consistent and less likely to trigger race conditions!

BeksOmega (approved):

PR #1851, opened 2023-08-10
fix: Focus region bug when primary workspace zoomed

cesarades:
**Description**
When the primary workspace zoomed in or out, the focus region would not be sized and positioned correctly. To fix this I used the viewMetrics in workspace units instead of pixel units. 

**Test**
I added the zoom buttons to the playground and ensured that with both zooming in and out, the focus region was sized and positioned correctly.

### DIFF
--- a/plugins/workspace-minimap/src/minimap.ts
+++ b/plugins/workspace-minimap/src/minimap.ts
@@ -215,6 +215,7 @@ export class Minimap {
     private onClickUp(): void {
       if (this.onMouseMoveWrapper) {
         Blockly.browserEvents.unbind(this.onMouseMoveWrapper);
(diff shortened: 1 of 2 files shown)


plugins/workspace-minimap/src/minimap.ts
```diff
@@ -215,6 +215,7 @@ export class Minimap {
     private onClickUp(): void {
       if (this.onMouseMoveWrapper) {
         Blockly.browserEvents.unbind(this.onMouseMoveWrapper);
+        this.onMouseMoveWrapper = null;
```
cesarades, line 218: This was just to fix a console error.

BeksOmega (approved):
LGTM =)

PR #1857, opened 2023-08-14
refactor: Convert the workspace-search plugin to Typescript

gonfunko:
This PR fixes part of #1647 by converting the workspace-search plugin to Typescript.


plugins/workspace-search/src/workspace_search.ts
```diff
@@ -0,0 +1,581 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Class responsible for workspace search.
```
maribethb: You can remove this part (fileoverview is optional, we removed authors from most places already)

plugins/workspace-search/src/workspace_search.ts
```diff
@@ -0,0 +1,581 @@
...
+   * @returns The component’s bounding box. Null in this
+   *     case since we don't need other elements to avoid the workspace search
+   *     field.
+   */
+  getBoundingRectangle(): Blockly.utils.Rect|null {
+    return null;
+  }
+
+  /**
+   * Positions the zoom-to-fit control.
```
maribethb: can you fix this text while you're here? it's wrong in both name and description

maribethb (approved):
There should be a devDependency on both `typescript` and `@typescript-eslint/parser` (see the `field-angle` plugin for an example; it's a mistake that the others don't list their dependency on ts. i can't find the issue where christopher and i debated about this though.)

plugins/workspace-search/src/workspace_search.ts
gonfunko: Done

plugins/workspace-search/src/workspace_search.ts
gonfunko: Done!

gonfunko:
> There should be a devDependency on both `typescript` and `@typescript-eslint/parser` (see the `field-angle` plugin for an example; it's a mistake that the others don't list their dependency on ts. i can't find the issue where christopher and i debated about this though.)

Done!

PR #1860, opened 2023-08-15
feat: Permits css capabilities for minimap

cesarades:
**Description**
The minimap workspace inherits the theme of the primary workspace. Now, however, developers can also edit the minimap with CSS. Namely the wrapper for the minimap (like the box shadow) and the focus region (the fill color).

### DIFF
--- a/plugins/workspace-minimap/src/focus_region.ts
+++ b/plugins/workspace-minimap/src/focus_region.ts
@@ -89,9 +89,6 @@ export class FocusRegion {
             'fill': 'black',
           }, mask);
 
-      // Theme.
-      this.background.setAttribute('fill', '#e6e6e6');
-
       // Add the svg group to the minimap.
       const parentSvg = this.minimapWorkspace.getParentSvg();
       if (parentSvg.firstChild) {
@@ -190,3 +187,9 @@ export class FocusRegion {
       return this.initialized;
     }
 }
+
+Blockly.Css.register(`
+.focusRegion {
+  fill: #e6e6e6;
+}
+`);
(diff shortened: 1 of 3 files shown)


plugins/workspace-minimap/src/minimap.ts
```diff
@@ -53,7 +53,7 @@ export class Minimap {
       // Create a wrapper div for the minimap injection.
       this.minimapWrapper = document.createElement('div');
       this.minimapWrapper.id = 'minimapWrapper' + this.primaryWorkspace.id;
-      this.minimapWrapper.className = 'minimapWrapper';
+      this.minimapWrapper.className = 'minimap';
```
BeksOmega: I think we want to add a prefix to ensure this doesn't interfere with external developer's css accidentally.

Style guide ref: https://google.github.io/styleguide/htmlcssguide.html#Prefixes

Maybe `blockly-minimap`? Similarly for the focus region.

plugins/workspace-minimap/src/focus_region.ts
```diff
@@ -190,3 +187,9 @@ export class FocusRegion {
       return this.initialized;
     }
 }
+
+Blockly.Css.register(`
+.focusRegion {
```
BeksOmega: Looking back at the style guide I think this should also be kebab-case and not camelCase.

Ref: https://google.github.io/styleguide/htmlcssguide.html#ID_and_Class_Name_Style

BeksOmega (approved):
LGTM! Just two little nits about class names =)

Issue #1838, opened 2023-07-31
Upgrade the shared procedures plugin to use the new copy paste system


BeksOmega:
A `doFullSerialization` parameter will be added in core for the 202e Q3 EOQ release: https://github.com/google/blockly/issues/7339

The shareable procedures plugin needs to be upgraded to use this new parameter so that blocks can be properly copy-pasted between workspaces.

PR #1863, opened 2023-08-16
fix: upgrade shareable procedures for new clipboard API

BeksOmega:
Closes https://github.com/google/blockly-samples/issues/1838
Closes https://github.com/google/blockly/issues/6846

Upgrades the shareable procedures plugin to use the new `doFullSerialization` parameter.

### DIFF
--- a/plugins/block-shareable-procedures/src/blocks.ts
+++ b/plugins/block-shareable-procedures/src/blocks.ts
@@ -419,23 +419,24 @@ const procedureDefMutator = {
    * Returns the state of this block as a JSON serializable object.
    * @returns The state of this block, eg the parameters and statements.
    */
-  saveExtraState: function() {
+  saveExtraState: function(doFullSerialization) {
     const state = Object.create(null);
     state['procedureId'] = this.getProcedureModel().getId();
 
-    const params = this.getProcedureModel().getParameters();
-    if (!params.length && this.hasStatements_) return state;
-
-    if (params.length) {
-      state['params'] = params.map((p) => {
-        return {
-          'name': p.getName(),
-          'id': p.getVariableModel().getId(),
-          // Ideally this would be id, and the other would be varId,
-          // but backwards compatibility :/
-          'paramId': p.getId(),
-        };
-      });
+    if (doFullSerialization) {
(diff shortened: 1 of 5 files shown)


plugins/block-shareable-procedures/src/blocks.ts
```diff
@@ -417,25 +417,29 @@ const procedureDefMutator = {
...
+            'name': p.getName(),
+            'id': p.getVariableModel().getId(),
+            // Ideally this would be id, and the other would be varId,
+            // but backwards compatibility :/
+            'paramId': p.getId(),
+          };
+        });
+      }
     }
     if (!this.hasStatements_) {
```
rachel-fenichel, line 444: Why do you only need to set this when `hasStatements_` is false, rather than always?

plugins/block-shareable-procedures/src/blocks.ts
```diff
@@ -450,20 +454,19 @@ const procedureDefMutator = {
    */
   loadExtraState: function(state) {
     const map = this.workspace.getProcedureMap();
+
     const procedureId = state['procedureId'];
-    if (procedureId && procedureId != this.model_.getId() &&
-      map.has(procedureId) &&
-      (this.isInsertionMarker() ||
```
rachel-fenichel, line 456: Why do you no longer have to check whether it's an insertion marker?

rachel-fenichel (commented):
Two questions about the intent of the code.

plugins/block-shareable-procedures/src/blocks.ts
BeksOmega: Default is true. We always skip serialization of defaults (or we try to anyways) because it keeps the save file size smaller. E.g. we also don't serialize a `'params'` array if the procedure has no params.

plugins/block-shareable-procedures/src/blocks.ts
BeksOmega: Insertion markers are saved with `doFullSerialization: false` so it is covered by the other check.

rachel-fenichel (approved):

PR #1867, opened 2023-08-17
fix: Shadow blocks replicated in minimap

cesarades:
**Description**
Shadow blocks were being mirrored in the minimap through the BLOCK_CREATE event. Now we check to see if their xml is of type shadow to filter them out.

**Tests**
To verify that it was working I manually checked in the playground. I added/removed a block from a shadow block's place and the shadow block was not mirrored over.


BeksOmega (approved):
Can you add something to the PR description about how you manually verified this change? Then LGTM!

PR #1869, opened 2023-08-18
chore: Convert zoom-to-fit plugin to TypeScript.

gonfunko:
This PR fixes part of https://github.com/google/blockly-samples/issues/1647 by converting the zoom-to-fit plugin to Typescript.


plugins/zoom-to-fit/src/index.ts
```diff
@@ -32,24 +32,24 @@ export class ZoomToFitControl implements Blockly.IPositionable {
   private top = 0;
   
   /**
-   * Width of the zoom-to-fit control.
+   * width of the zoom-to-fit control.
```
BeksOmega: Revert capitalization change, and in below docs for `height`

plugins/zoom-to-fit/tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+    "target": "es6",
+    "allowJs": true,
+    "sourceMap": true,
+    "moduleResolution": "nodenext",
+    "lib": ["es2021", "dom"]
+  },
+  "include": [
+    "src", "test"
+  ],
+}
```
BeksOmega, line 13: This does not match other tsconfigs in samples, (in particular the `target`, `module`, and `moduleResolution`). Not sure what it is supposed to look like though.

BeksOmega (commented):
Generally LGTM! But not sure how the tsconfig is supposed to be set up.

plugins/zoom-to-fit/tsconfig.json
gonfunko: It's the same as what I used for toolbox-search, although I don't remember exactly how I arrived at that.

plugins/zoom-to-fit/src/index.ts
gonfunko: Done!

plugins/zoom-to-fit/tsconfig.json
BeksOmega: Ok, could you file a bug for checking on this and updating the configs to be consistent?

BeksOmega (approved):

plugins/zoom-to-fit/tsconfig.json
gonfunko: Done, #1871.

PR #1875, opened 2023-08-23
fix: Test for minimap playground

cesarades:
**Description**
If you used the playground to change the workspace, it would create duplicate minimaps. To fix this I assigned the minimap to a module-local variable, and dispose of it if it already exists.

In this PR I also removed the private label from the package file.

**Tests**
To test that no minimaps were being duplicated I repeatedly altered the workspace attributes, and changed the window size.


BeksOmega (approved):

BeksOmega:
Double checked that everything in the package config looks good =) Such excite!

PR #1876, opened 2023-08-25
refactor: Convert the workspace-backpack plugin to Typescript.

gonfunko:
This PR fixes part of #1647 by converting the workspace-backpack plugin to Typescript.


plugins/workspace-backpack/src/backpack_helpers.ts
```diff
@@ -11,17 +11,17 @@
...
- * @param {!Blockly.WorkspaceSvg} workspace The workspace to register the
- *   context menu option on.
+ * @param workspace The workspace to register the context menu option on.
  */
-function registerEmptyBackpack(workspace) {
+function registerEmptyBackpack(workspace: Blockly.WorkspaceSvg) {
   const prevConfigureContextMenu = workspace.configureContextMenu;
-  workspace.configureContextMenu = (menuOptions, e) => {
-    const backpack = workspace.getComponentManager().getComponent('backpack');
+  workspace.configureContextMenu = (menuOptions, e: PointerEvent) => {
```
BeksOmega, line 25: Does `menuOptions` need a type as well?

BeksOmega, line 25: We probably want to update this to register the context menu item instead of assigning to `configureContextMenu` directly. Could you file a bug for that (if it doesn't already exist)?

 Although I'm a bit confused why this one is different when the other options to register the context menu item...

plugins/workspace-backpack/src/backpack_helpers.ts
```diff
@@ -208,12 +202,12 @@ function registerPasteAllBackpack() {
 
 /**
  * Register all context menu options.
- * @param {!BackpackContextMenuOptions} contextMenuOptions The backpack context
+ * @param contextMenuOptions The backpack context
  *    menu options.
```
BeksOmega: Nit: Fix line wrapping.

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -23,137 +23,101 @@ import {parseOptions} from './options';
  * @extends {Blockly.DragTarget}
  */
 export class Backpack extends Blockly.DragTarget {
+  
+  /**
+   * The unique id for this component.
+   */
```
BeksOmega: Nit: in general we keep these as one line if we can. But don't block merging on this - just for the future.

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -23,137 +23,101 @@ import {parseOptions} from './options';
...
+  id = 'backpack';
+  
+  /**
+   * The backpack options.
```
BeksOmega: Nit: maybe an example of what this can configure?

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -23,137 +23,101 @@ import {parseOptions} from './options';
...
+  protected contents_: string[] = [];
+  
+  /**
+   * Array holding info needed to unbind events.
+   * Used for disposing.
+   */
+  private boundEvents_: Blockly.browserEvents.Data[] = [];
+  
+  /**
+   * Left coordinate of the backpack.
```
BeksOmega: Nit: Units (for all coordinates and measurements)

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -23,137 +23,101 @@ import {parseOptions} from './options';
...
+   * Extent of hotspot on all sides beyond the size of the image.
+   */
+  protected readonly HOTSPOT_MARGIN_ = 10;
+  
+  /**
+   * The SVG group containing the backpack.
+   */
+  protected svgGroup_: SVGElement|null = null;
+  
+  protected svgImg_: SVGImageElement|null = null;
```
BeksOmega, line 70: Nit: docs (although I suspect the linter caught this).

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -899,47 +856,43 @@ Blockly.Css.register(`
...
 
   /**
    * Loads a serialized state into the target workspace.
-   * @param {object} state the serialized state JSON
-   * @param {Blockly.Workspace} workspace the workspace to load into
+   * @param state the serialized state JSON
+   * @param workspace the workspace to load into
    */
-  load(state, workspace) {
+  load(state: object[], workspace: Blockly.WorkspaceSvg) {
```
BeksOmega, line 852: Is `object` actually a valid typescript type?

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -32,92 +32,92 @@ export class Backpack extends Blockly.DragTarget {
   /**
    * The backpack options.
    */
-  private options_: BackpackOptions;
+  private options: BackpackOptions;
   
   /**
    * The backpack flyout. Initialized during init.
    */
-  protected flyout_: Blockly.IFlyout|null = null;
```
BeksOmega: We cannot modify the names of protected things unless we want to do a major release. Protected things are part of the public API :/

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -423,7 +423,7 @@ export class Backpack extends Blockly.DragTarget implements
     const json = Blockly.serialization.blocks.save(block);
 
     // Add a 'kind' key so the flyout can recognize it as a block.
-    json.kind = 'BLOCK';
+    (json as any).kind = 'BLOCK';
```
BeksOmega: I think you can cast this to a `Blockly.utils.toolbox.BlockInfo`

plugins/workspace-backpack/src/backpack_helpers.ts
rachel-fenichel: The original backpack code *may* predate good context menu APIs.

gonfunko: I think the reason for this is that the precondition function for context menu items doesn't provide the event that triggered the menu, so there's no way to check if the click was on the backpack icon as opposed to somewhere else in the workspace without using the older configureContextMenu approach. It seems pretty reasonable to pass that event in the precondition callback, so maybe that's the bug I should file?

BeksOmega: Uh yeah we should definitely take another look at this. We might want to reconsider how the context menu scope is defined. I think there should be an issue about "hard to define different context menu options for new workspace positionables".

plugins/workspace-backpack/src/backpack.ts
gonfunko: It is: https://www.typescripttutorial.net/typescript-tutorial/typescript-object-type/

BeksOmega:
@gonfunko is this ready for re-review?

gonfunko:
> @gonfunko is this ready for re-review?

Yes, PTAL when you get a chance!

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -21,87 +21,57 @@ import {BackpackChange, BackpackOpen} from './ui_events';
...
-   */
+  // The backpack options, such as which context menus to show, whether to
+  // allow opening the backpack when empty and whether to use a different
+  // image when the backpack contains blocks.
   private options: BackpackOptions;
 
-  /**
-   * The backpack flyout. Initialized during init.
-   */
+  // The backpack flyout. Initialized during init.
```
BeksOmega: Sorry I should have clarified: I meant like `/** the docs */` all on one line. I don't think the typescript compiler picks up these comments to add to the .d.ts

If it's multiple lines, you can use multiline `/** */` like you had before!

After this should be gtg!

BeksOmega (approved):

Issue #1843, opened 2023-08-03
Make the dynamic connections plugin serialize to the same format as blocks in core


BeksOmega:
**Category**

 - Plugins

**Component**

block-dynamic-connections

**Is your feature request related to a problem? Please describe.**

Currently the dynamic connections plugin serializes to a format that is not compatible with the blocks in core. This means that if you start using the dynamic connections plugin, you can never pull it out and go back to the default way of handling mutations.

**Describe the solution you'd like**

Have the dynamic connections plugin serialize to the same save format as blocks in core.

**Describe alternatives you've considered**

N/A

**Additional context**

App Inventor want to use this plugin, but they want to be able to quickly and easily roll it back if anything breaks. Plus they have a lot of analysis tools tha trely on the current save format. This would break those.

maribethb:
This is reasonable, but the plugin needs to be able to continue deserializing from its old format too, so that folks already using it are not broken.

PR #1880, opened 2023-08-29
fix: serialization of dynamic if blocks

BeksOmega:
Work on https://github.com/google/blockly-samples/issues/1843

Changes the serialization of the dynamic if blocks to match the serialization in core.

This means that we have to tear down the inputs every time and re-add them so that they have the correct indexing (strictly increasing from 0).

This just fixes the if blocks, I'll upgrade the other blocks in future PRs, which is why this is against a feature branch and not against the main branch directly.

### DIFF
--- a/plugins/block-dynamic-connection/src/dynamic_if.ts
+++ b/plugins/block-dynamic-connection/src/dynamic_if.ts
@@ -64,7 +64,12 @@ const DYNAMIC_IF_MIXIN = {
    * @returns XML storage element.
    */
   mutationToDom(this: DynamicIfBlock): Element|null {
+    // If we call finalizeConnections here without disabling events, we get into
+    // an event loop.
+    Blockly.Events.disable();
     this.finalizeConnections();
+    Blockly.Events.enable();
+
     if (!this.elseifCount && !this.elseCount) return null;
 
     const container = Blockly.utils.xml.createElement('mutation');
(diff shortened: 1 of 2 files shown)


plugins/block-dynamic-connection/src/dynamic_if.ts
```diff
@@ -188,7 +207,7 @@ const DYNAMIC_IF_MIXIN = {
     if (connection.targetConnection && input.name.includes('IF')) {
       const nextIfInput = this.inputList[inputIndex + 2];
       if (!nextIfInput || nextIfInput.name == 'ELSE') {
-        this.insertElseIf(inputIndex + 2);
+        this.insertElseIf(inputIndex + 2, Blockly.utils.idGenerator.genUid());
```
rachel-fenichel, line 210: This has to be a UUID to deal with reordering, right?

rachel-fenichel (approved):

plugins/block-dynamic-connection/src/dynamic_if.ts
BeksOmega: It has to be a UUID to make sure it doesn't conflict with any of the existing names. Any inputs created here are temporary and get deleted in `deleteInputs` so it doesn't really matter for ordering.

I could keep the `inputCount` tracker around to solve this as well. But that seemed confusing since it would only be used for differentiation in this specific case, and not keeping track of the input count in general. I felt this was a better reflection of what I'm actually trying to achieve (a temporary input whose name doesn't conflict with any existing ones).

PR #1885, opened 2023-08-31
fix: serialization of dynamic list blocks

BeksOmega:
Work on https://github.com/google/blockly-samples/issues/1843

Changes the serialization of the dynamic list blocks to match the serialization in core. Also makes it so the `minInputs` actually controls the minimum number of inputs.

This means that we have to tear down the inputs every time and re-add them so that they have the correct indexing (strictly increasing from 0).

This just fixes the list blocks, I'll upgrade the text blocks in another PR, which is why this is against a feature branch and not against the main branch directly.

Changes I implemented here that need to be backported to if-blocks (which I will do in my next PR):
- Making `minInputs` control the minimum number of inputs (or cases for if-blocks).
- Tearing down the entire block instead of leaving the top one.

### DIFF
--- a/plugins/block-dynamic-connection/src/dynamic_list_create.ts
+++ b/plugins/block-dynamic-connection/src/dynamic_list_create.ts
@@ -22,12 +22,10 @@ type DynamicListCreateMixinType = typeof DYNAMIC_LIST_CREATE_MIXIN;
 /* eslint-disable @typescript-eslint/naming-convention */
 const DYNAMIC_LIST_CREATE_MIXIN = {
   /* eslint-enable @typescript-eslint/naming-convention */
-  /** Counter for the next input to add to this block. */
-  inputCounter: 3,
-
   /** Minimum number of inputs for this block. */
   minInputs: 2,
 
+  /** Count of item inputs. */
   itemCount: 0,
 
   /** Block for concatenating any number of strings. */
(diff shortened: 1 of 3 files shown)


plugins/block-dynamic-connection/src/dynamic_list_create.ts
```diff
@@ -22,19 +22,20 @@ type DynamicListCreateMixinType = typeof DYNAMIC_LIST_CREATE_MIXIN;
...
   init(this: DynamicListCreateBlock): void {
+    this.itemCount = this.minInputs;
+
     this.setHelpUrl(Blockly.Msg['LISTS_CREATE_WITH_HELPURL']);
     this.setStyle('list_blocks');
-    this.appendValueInput('ADD0')
-        .appendField(Blockly.Msg['LISTS_CREATE_WITH_INPUT_WITH']);
-    this.appendValueInput('ADD1');
+    this.addFirstInput();
+    for (let i = 1; i < this.minInputs; i++) this.appendValueInput(`ADD${i}`);
```
NeilFraser: Are we doing single-line loops now?

NeilFraser (approved):

plugins/block-dynamic-connection/src/dynamic_list_create.ts
BeksOmega: Ope nope, only for `if`s! will fix

PR #1917, opened 2023-09-06
fix: deleting procedures with callers not being undoable

### LINKED ISSUE #1891: Undoing delete of procedure def causes crash when caller exists
**Category**

 - Plugins

**Component**

`block-shareable-procedures` `3.0.5`

**Describe the bug**

Deleting a procedure definition when a caller is present leads to a crash when undoing the delete.

**To Reproduce**

The easiest way is to add `block-shareable-procedures` to the playground.

(This bug not happen in the `block-shareable-procedures` demo where undo actions are set to 0.)

Steps:

```
cd plugins/dev-tools
npm install
npm add "@blockly/block-shareable-procedures"
```

Activate plugin in playground:
```diff
diff --git a/plugins/dev-tools/test/index.js b/plugins/dev-tools/test/index.js
index 15c40817..2a0ea1bd 100644
--- a/plugins/dev-tools/test/index.js
+++ b/plugins/dev-tools/test/index.js
@@ -12,6 +12,17 @@
 import * as Blockly from 'blockly';
 import {createPlayground, toolboxCategories} from '../src/index';
 
+const {
+  blocks,
+  unregisterProcedureBlocks,
+  registerProcedureSerializer,
+} = require('@blockly/block-shareable-procedures');
+
+unregisterProcedureBlocks();
+Blockly.common.defineBlocks(blocks);
+
+registerProcedureSerializer();
+
 /**
  * Create a workspace.
  * @param {HTMLElement} blocklyDiv The blockly container div.
```

See video below for the reproduction steps:
- Create a procedure with some arguments.
- Add a caller to the workspace.
- Delete the procedure definition (which will also delete caller)
- Press undo
- There is one error immediately, and more pup up each time it tries to access the missing procedure model.

**Expected behavior**

Restore procedure def/caller/model without crashing.

**Screenshots**

https://github.com/google/blockly-samples/assets/12326241/2f7027b1-0c30-4acf-97d3-c1bda5132ee7

**Additional context**

It appears that `destroy()` in one of the mixins will call `this.workspace.getProcedureMap().delete(this.getProcedureModel().getId());`, but it doesn't get restored on undo. (Or perhaps it is restored, but not in the right order?)

BeksOmega:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/1891

### Proposed Changes

Moves the recordUndo disabling to the parameter event level instead of the base procedure event level.

### Reason for Changes

We want to disable recording undo for parameter changes but not for all changes (because in some cases we actually do need to handle life cycle for the models separately, e.g. in the case of deletion).

### Test Coverage

Manually tested by disabling the secondary workspace in the test page, and re-enabling undo and redo.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/plugins/block-shareable-procedures/src/events_procedure_base.ts
+++ b/plugins/block-shareable-procedures/src/events_procedure_base.ts
@@ -27,7 +27,6 @@ export abstract class ProcedureBase extends Blockly.Events.Abstract {
       readonly procedure: Blockly.procedures.IProcedureModel) {
     super();
     this.workspaceId = workspace.id;
-    this.recordUndo = false;
   }
 
   /**
--- a/plugins/block-shareable-procedures/src/events_procedure_parameter_base.ts
+++ b/plugins/block-shareable-procedures/src/events_procedure_parameter_base.ts
@@ -29,6 +29,7 @@ export abstract class ProcedureParameterBase extends ProcedureBase {
       procedure: Blockly.procedures.IProcedureModel,
       readonly parameter: Blockly.procedures.IParameterModel) {
     super(workspace, procedure);
+    this.recordUndo = false;
   }
 
   /**


cpcallen (approved):
This seems like the thing that it would be nice to have a test for.

BeksOmega:
> This seems like the thing that it would be nice to have a test for.

Agree, but testing undo and redo is pretty difficult because of the event delay.

PR #1929, opened 2023-09-12
fix: serialization of dynamic blocks

BeksOmega:
Fixes were verified in individual PRs. This merges into the 2023-q3-eq branch so that the fixes can go out with the EOQ release.

Originally I thought they could just go out at once when they were done, but then I ran into connection highlights not being disposed (https://github.com/google/blockly/pull/7454) so I think it's best to release this fix with that one.

### DIFF
--- a/plugins/block-dynamic-connection/README.md
+++ b/plugins/block-dynamic-connection/README.md
@@ -18,17 +18,14 @@ import * as BlockDynamicConnection from '@blockly/block-dynamic-connection';
 ## API
 - `overrideOldBlockDefinitions`: Replaces the Blockly default blocks with the
   dynamic connection blocks. This enables projects to use the dynamic block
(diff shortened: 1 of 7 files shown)


cpcallen (approved):
Approved without detailed review (I trust our colleagues!), but subject to clarification:

This is changing the serialisation of these blocks, right?  How is that not a breaking change?  Does the deserialiser code accept both old and new (=== really old) forms?  Or have we just not yet published these blocks?

BeksOmega:
> Approved without detailed review (I trust our colleagues!), but subject to clarification:
> 
> This is changing the serialisation of these blocks, right? How is that not a breaking change? Does the deserialiser code accept both old and new (=== really old) forms? Or have we just not yet published these blocks?

The deserialization accepts all forms, and there are tests for all of them!

PR #1930, opened 2023-09-13
fix: toolbox search with XML toolboxes of kind BLOCK

### LINKED ISSUE #1925: Toolbox search doesn't work with XML toolboxes due to case sensitivity
**Category**

 - Plugins

**Component**

toolbox-search

**Describe the bug**

The toolbox search plugin doesn't work if the 'kind' is upper case (which occurs when using XML toolboxes).

**To Reproduce**

1. Go to the [toolbox search playground](https://github.com/google/blockly-samples/blob/master/plugins/toolbox-search/test/index.ts) file.
2. Replace the `DOMContentLoaded` listener with the below code:
```js
document.addEventListener('DOMContentLoaded', function() {
  const toolbox = {
    'kind': 'categoryToolbox',
    'contents': [
      {
        'kind': 'category',
        'name': 'Control',
        'contents': [
          {
            'kind': 'BLOCK',
            'type': 'controls_if',
          },
        ],
      },
      {
        'kind': 'category',
        'name': 'Logic',
        'contents': [
          {
            'kind': 'BLOCK',
            'type': 'logic_compare',
          },
          {
            'kind': 'BLOCK',
            'type': 'logic_operation',
          },
          {
            'kind': 'BLOCK',
            'type': 'logic_boolean',
          },
        ],
      },
      {
        'kind': 'search',
        'name': 'Search',
        'contents': [],
      },
    ],
  };
  const defaultOptions: Blockly.BlocklyOptions = {
    toolbox,
  };
  createPlayground(
      document.getElementById('root'), createWorkspace, defaultOptions);
});
```
4. Run `npm run start` from the toolbox-search directory to open the playground.
5. Type 'if' into the search.
6. Observe how no blocks appear in the flyout.

**Expected behavior**

Blocks should appear.

## To Fix

1) Run through the reproduction steps and make sure you can reproduce the error.
2) Modify [`getAvailableBlocks`](https://github.com/google/blockly-samples/blob/f06f64f8ab2785f587dda1875d5b2b194ed7e7f3/plugins/toolbox-search/src/toolbox_search.ts#L106) to be case insensitive. You probably want to use [toUpperCase](https://www.w3schools.com/jsref/jsref_touppercase.asp).
3) Run the reproduction steps again to make sure the error doesn't reproduce.

prasium:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/1925

### Proposed Changes

This PR adds an explicit lowerCase casting to schema.kind field which is being compared against 'block' literal.

### Reason for Changes

In the _toolbox_ object attached to the issue, the schema.kind field has value = 'BLOCK' in uppercase whereas in code we are comparing the field value with `schema.kind === 'block'`.  So I converted the check field to lowercase against the literal. Which will ensure case insensitivity of the expression. The expression now becomes `schema.kind.toLowercase() === 'block'`.
### Test Coverage
1. Run it using the steps mentioned in the issue. 
2. Search for blocks present in the test toolbox object present in the issue.
3. Here's the sample of working output - 
image

### DIFF
--- a/plugins/toolbox-search/src/toolbox_search.ts
+++ b/plugins/toolbox-search/src/toolbox_search.ts
@@ -109,7 +109,7 @@ export class ToolboxSearchCategory extends Blockly.ToolboxCategory {
       schema.contents.forEach((contents) => {
         this.getAvailableBlocks(contents, allBlocks);
       });
-    } else if (schema.kind === 'block') {
+    } else if (schema.kind.toLowerCase() === 'block') {
       if ('type' in schema && schema.type) {
         allBlocks.add(schema.type);
       }


BeksOmega:
Manually tested and this LGTM! Thank you @prasium =) If you'd like to work on another issue let me know and I can recommend something!

PR #1932, opened 2023-09-14
fix: zelos full block colour field rendering incorrectly

### LINKED ISSUE #7442: Zelos renders wrongly when dynamically adding fields
### Check for duplicates

### Description

when using the "Zelos" render adding dynamic fields causes the render to render completely white.

this doesn't happen with the other renders.

### Reproduction steps

1. set workspace render to "Zelos".
2. create a block with an text or number input.
3. add a field (e.g. dropdown) to it outside of the definition function.
4.

### Stack trace

_No response_

### Screenshots

https://github.com/google/blockly/assets/54947833/0839e2fb-00a6-4622-a90b-3b645066f4a4

### Browsers

Chrome desktop

BeksOmega:
## The basics

### Resolves

Fixes [#7442](https://github.com/google/blockly/issues/7442) but for the field in samples

### Proposed Changes

See https://github.com/google/blockly/pull/7490

### Reason for Changes

See https://github.com/google/blockly/pull/7490

### Test Coverage

See https://github.com/google/blockly/pull/7490

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/plugins/field-colour/src/field_colour.ts
+++ b/plugins/field-colour/src/field_colour.ts
@@ -199,12 +199,45 @@ export class FieldColour extends Blockly.Field<string> {
       constants.FIELD_COLOUR_DEFAULT_WIDTH,
       constants.FIELD_COLOUR_DEFAULT_HEIGHT,
     );
-    if (!constants.FIELD_COLOUR_FULL_BLOCK) {
-      this.createBorderRect_();
-      this.getBorderRect().style['fillOpacity'] = '1';
-    } else if (this.sourceBlock_ instanceof Blockly.BlockSvg) {
-      this.clickTarget_ = this.sourceBlock_.getSvgRoot();
+    this.createBorderRect_();
+    this.getBorderRect().style['fillOpacity'] = '1';
+    this.getBorderRect().setAttribute('stroke', '#fff');
+    if (this.isFullBlockField()) {
+      this.clickTarget_ = (this.sourceBlock_ as Blockly.BlockSvg).getSvgRoot();
     }
+  }
+
+  /**
+   * Defines whether this field should take up the full block or not.
+   */
+  protected isFullBlockField(): boolean {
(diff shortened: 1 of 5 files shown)


BeksOmega:
Pulled in changes from https://github.com/google/blockly/pull/7566 and https://github.com/google/blockly/pull/7584

Retested adding and removing fields in zelos. Tested that selecting th color updates the field, in both modes.

plugins/field-colour/src/field_colour.ts
```diff
@@ -137,32 +137,130 @@ export class FieldColour extends Blockly.Field<string> {
...
+  /**
+   * Defines whether this field should take up the full block or not.
+   * @returns True if this field should take up the full block. False otherwise.
+   */
+  protected isFullBlockField(): boolean {
+    const block = this.getSourceBlock();
+    if (!block) throw new Blockly.UnattachedFieldError();
+
+    const constants = this.getConstants();
+    return this.blockIsSimpleReporter() && !!constants?.FIELD_COLOUR_FULL_BLOCK;
```
cpcallen: I'm not 100% sure about hte order of precedence here, and think it might be clearer to write:
```suggestion
    return this.blockIsSimpleReporter() && Boolean(constants?.FIELD_COLOUR_FULL_BLOCK);
```
(Assuming that that is indeed equivalent.)

plugins/field-colour/src/field_colour.ts
```diff
@@ -137,32 +137,130 @@ export class FieldColour extends Blockly.Field<string> {
...
+  }
+
+  /**
+   * @returns True if this block is a value block with a single editable field.
```
cpcallen: `this` is a field, no?  Maybe:
```suggestion
   * @returns True if the enclosing block is a value block with a single editable field.
```

cpcallen (approved):

PR #1937, opened 2023-09-19
chore: Setting style property to make CSP less grumpy.

NeilFraser:
"Content-Security-Policy: The page’s settings blocked the loading of a resource at inline (“style-src”)."

The 'style' property should be set as an object, not as a string, according to CSP rules.

Back-ported from Blockly Games.

### DIFF
--- a/plugins/workspace-minimap/src/positioned_minimap.ts
+++ b/plugins/workspace-minimap/src/positioned_minimap.ts
@@ -161,13 +161,13 @@ export class PositionedMinimap extends Minimap implements Blockly.IPositionable 
      */
     private setAttributes(): void {
       const injectDiv = this.minimapWorkspace.getInjectionDiv();
-      injectDiv.parentElement.setAttribute('style',
-          `z-index: 2;
-          position: absolute;
-          width: ${this.width}px;
-          height: ${this.height}px;
-          top: ${this.top}px;
-          left: ${this.left}px;`);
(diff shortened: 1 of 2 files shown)


maribethb (approved):

maribethb:
This broke the minimap plugin: https://google.github.io/blockly-samples/plugins/workspace-minimap/test/index.html

The `left` value isn't being set, but I don't know why. I don't see anything wrong with it, but it's been a long day.

plugins/workspace-minimap/src/positioned_minimap.ts
```diff
@@ -161,13 +161,13 @@ export class PositionedMinimap extends Minimap implements Blockly.IPositionable
...
-          height: ${this.height}px;
-          top: ${this.top}px;
-          left: ${this.left}px;`);
+      const style = injectDiv.parentElement.style;
+      style.zIndex = '2';
+      style.position = 'absolute';
+      style.width = `${this.width}px`;
+      style.height = `${this.height}px`;
+      style.top = `${this.top}px`;
+      style.left = `${this.left}px;`;
```
BeksOmega, line 170: I think it's an extra semicolon! Should be
```
style.left = `${this.left}px`;
```

NeilFraser: I stared at it for several minutes this morning and didn't see that.  🙄

PR #1946, opened 2023-09-22
chore: add formatting with prettier

### LINKED ISSUE #962: Enable clang-format in samples
**Category**
Plugins and examples

**Component**
All of them 

Manually formatting code is a large pain. 

Using clang format to auto format all of our files similar to what we are doing in core.
I believe this would involve: 
- [ ] Running clang-format on all of our plugins and examples.
- [ ] Adding `npm run format` to dev-scripts.
- [ ] Adding `npm run format` in dev-create. 
- [ ] Adding a github action to check that code has been formatted properly. [Reference in core](https://github.com/google/blockly/blob/master/.github/workflows/check_clang_format.yml).

**Additional context**

### LINKED ISSUE #1695: improve eslint config and linting experience
**Category**

 - Plugins

**Component**

eslint-config

**Problem**

Our current eslint config is complex and probably not configured correctly. Packages that depend on it do not properly declare their dependencies on the plugins used in the config. Linting all plugins at once (e.g. in CI) is difficult because the error messages are hidden in a bunch of extraneous output.

**Proposed solution**

Migrate to the new eslint flat config. Set up the configuration at the top-level of blockly-samples so that the repository can be linted at once without needing to use lerna run.

**eslint cli**

Instead of using the eslint CLI, we have a custom script in `dev-scripts` that uses the javascript api. I don't know why this is necessary and I think it is probably simpler to use the CLI directly.

**eslint flat config**

Note that for the below paragraphs, I'll use "eslint plugin" to refer to plugins for eslint. I'll use "packages" to refer to the blockly-samples plugins, to avoid confusion.

The way eslint has historically been configured is via cascading layers of configuration. You can nest configuration files in different directory levels and they'll be merged into one config and run on the appropriate code. Because of this there are limitations in how eslint plugins can be required and shared for shared eslint configs. See [this issue](https://github.com/eslint/eslint/issues/3458) for more context. We have a shared eslint config that uses multiple eslint plugins. The blockly packages then declare their dev dependency on the shared eslint config, but they don't declare their dependency on the eslint plugins like they are supposed to. This means that likely the shared config does not work if you try to invoke lint outside of `dev-scripts` (unless you install the plugins, which we don't tell you about anywhere), and it causes problems because the version number you'll get of those plugins is essentially undefined behavior, since we have an undeclared dependency.

eslint has fixed this by introducing the flat config. Using the new style of configuration, some features like the cascading config have been removed, but in other ways the config has been simplified, such as allowing a shared config to just have a dependency on their plugins, and not require downstream consumers to also require those plugins.

The flat config simplifies deciding what files it applies to. It has a `files` property that supports globs. This is simpler than today where either you have to specify the current directory and lint everything in it, or carefully specify which files to invoke eslint on the command line.

**shared config**

Beyond this, I don't think we need a shared config that we publish as a plugin at all. We should be able to change the style of our packages without worrying about if we're making a breaking change to users of our shared config (which we do now, so anyone actually trying to use our config is probably not having a good time). There is not really any point to publishing the config as a blockly plugin because we don't intend for anyone outside of blockly-samples to use it.

If the config were just at the root level of blockly-samples, we could run lint from the root level of samples and lint the entire repository at once without using lerna. This would make lint warnings and errors more obvious on CI.

I believe we could still also lint an individual plugin because the config would be found in the enclosing directory structure, but this needs confirmation.

**plugins in core repo**

If we move plugins to the main google/blockly repo, this will make even more sense. We'd have one flat config at the root level that applies to both core and plugins. If needed, plugins can have an override with more strict rules than applies to core. This allows us to simplify the eslint setup, make sure the dependencies are correct and consistent, and see all lint messages at once without ever having to set up the shared config.

**references**

- https://eslint.org/blog/2022/08/new-config-system-part-2/
-

maribethb:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/1695
Fixes https://github.com/google/blockly-samples/issues/962 (with prettier instead of clang-format)

### Proposed Changes

Might be easier to review commit-wise. I've summarized here though not in the order of the commits. (sorry)

- Switches eslint config format to the new flat config style. See https://github.com/google/blockly-samples/issues/1695 for more background on this. This was a requirement to get Prettier working, due to the complex nature of the old shared config
    - Doing this exposed some files that weren't previously being linted. Some of these, I have no idea why they weren't. e.g. the `workspace-minimap` plugin had the wrong indentation. no idea why this wasn't fixed.
    - Largely kept the rules the same, other than rules that conflict with Prettier. I also changed some of the jsdoc rules to more closely match what we have in core (e.g. turned off auto-fixing for the rule that was adding blank jsdoc comments)
    - Disabled the `jsdoc/tag-lines` rule for now because it was super spammy. will turn it back on and autofix that in a separate PR.
- Installs Prettier and adds `format` and `format:check` commands.
- Removes the `blockly-scripts lint` script and associated machinery. We no longer need to invoke eslint through the javascript api. It's simpler to invoke it via command line.
- Removes linting from webpack. `eslint-webpack-plugin` doesn't support the new format. This was confusing to me anyway. It invoked the linter totally differently than how running `npm run lint` did which was pretty confusing. Now you won't see lint errors when you run `npm run test`. You need to run `npm run lint` separately.
- Auto formatted and auto-fixed lint problems in the whole repo
- Manually fixed two actual lint errors that were in a file not previously being linted

### Reason for Changes

fmt

Also, the old style of eslint config was just, not good. It caused us a lot of problems because technically anyone using a shared config is supposed to also depend on any plugins used by said config. We weren't doing that. That's why some of the typescript plugins had a random dependency on the typescript-eslint parser. Totally confusing that they needed that. Now they don't, because all of the configuration for linting is handled at the root level, while still giving you the option of linting a single plugin.

### Test Coverage

Tested that each of the following works:
- `npm run lint` at root
- `npm run lint` in each plugin
- `npm run format` at root
- `npm run build` at root and in each plugin
- `npm run test` at root and in each plugin

### Documentation

We can tell people they can run the formatter now

### Additional Information

The field-date tests might be failing? But they're failing on master for me locally too so unrelated.

Holding this until after OSD, then I'll rebase and re-run formatter and eslint fixer.

### Future work

- #1975 
- #1976
- #1977
- #1978
- #1979
- #1980
- #1981


eslint.config.js
```diff
@@ -0,0 +1,239 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
BeksOmega: Should this be 2023?

eslint.config.js
```diff
@@ -0,0 +1,239 @@
...
+import {FlatCompat} from '@eslint/eslintrc';
+import babelParser from '@babel/eslint-parser';
+import globals from 'globals';
+import tsParser from '@typescript-eslint/parser';
+import typescript from '@typescript-eslint/eslint-plugin';
+
+// Used to convert the Google eslint config into flat config format.
+const compat = new FlatCompat();
+
+export default [
```
BeksOmega: I purposefully didn't comment on any of the rules here because I don't want to bog this formatting PR down in an ESLint discussion :P

Seems to be the same as what we currently have, including the existing differences from core. So I think this is good for now! If there's anything you purposefully changed that you want to call out happy to discuss.

.prettierignore
```diff
@@ -0,0 +1,12 @@
+# All examples, codelabs, etc.
+examples/*
```
BeksOmega, line 2: Do we want to format examples? They're code too.

.prettierrc.js
```diff
@@ -0,0 +1,13 @@
+// This config attempts to match google-style code.
+
+module.exports = {
+  // Prefer single quotes, but minimize escaping.
+  singleQuote: true,
+  // Some properties must be quoted to preserve closure compiler behavior.
+  // Don't ever change whether properties are quoted.
+  quoteProps: 'preserve',
```
BeksOmega, line 8: Do we need this in samples?

eslint.config.js
```diff
@@ -139,7 +139,9 @@ module.exports = [
       'jsdoc/require-returns-type': 'off',
       // Ensure there is a blank line between the body and any @tags,
       // as required by the tsdoc spec.
-      'jsdoc/tag-lines': ['error', 'any', {'startLines': 1}],
+      // TODO: Re-enable soon.
```
BeksOmega, line 163: Do you want to file a separate issue for this puppy? (and maybe ones for the other future work you mentioned in the PR description)

plugins/dev-scripts/bin/blockly-scripts.js
```diff
@@ -18,7 +18,6 @@ const availableScripts = [
   'start',
   'clean',
   'predeploy',
-  'lint',
```
BeksOmega: Is this a breaking change for this plugin?

BeksOmega (approved):
LGTM!

eslint.config.js
maribethb: Yeah, the only things I think I changed were disabling the fixers for missing jsdoc because it kept creating empty comments, and disabling the `tag-lines` rule because there were a lot of violations and it was clogging my ability to make sure I had everything else correct.

.prettierignore
maribethb: Yes, but in the status quo they aren't being linted. I wanted to do that in a follow up PR. I'll add it to the future work section and file issues for those.

.prettierrc.js
maribethb: No, I don't think so. But keeping it as this value means prettier won't touch the quotes and I didn't want to do that in this PR either. I'll also add it to the future work section.

plugins/dev-scripts/bin/blockly-scripts.js
maribethb: Yeah, good call. I'll undo enough changes so that the script still can run lint, it just won't actually be invoked anywhere, and I'll add a deprecation message to the script.

maribethb:
After we merge osd branch into master, I'll rebase this, drop the commits that only did formatting and auto-fixing, and then re-run formatting and auto-fixing. Hoping that will avoid most merge conflicts.

maribethb:
@BeksOmega I need to kick the package-locks until they pass in ci (they pass on my machine running `npm ci` so I gotta poke it a bit until they work)

but in the meantime this is ready for re-review. i rebased and force pushed so most of the earlier commits are the same. you could view all the commits besides the two marked `autofix lint problems` and `format everything` to see all non-automated changes. one thing i added since last review is in the last commit, excluding the copy of sample-app that might end up in your version of `dev-create` if you've ever run `npm pack` in that plugin like i did today...

BeksOmega (approved):
Fixes from last round of comments lgtm!

Issue #1877, opened 2023-08-26
Unable to import plugin-workspace-search 8.0.4 normally.


simplenty:
**Category**

- Plugins

**Component**

- plugin-workspace-search

**Describe the bug**

In version 8.0.4, the library cannot be obtained normally through the package manager.

**To Reproduce**

```javascript
import { WorkspaceSearch } from '@blockly/plugin-workspace-search'
```

**Expected behavior**

In the newly released version 8.0.4, when importing the plugin, an error will be reported that the library does not exist.

**Screenshots**

![708803ff50b99535b15b299c3efa6be3](https://github.com/google/blockly-samples/assets/52663614/727e2844-1092-48de-ab7c-0505bf0d320a)

**Additional context**

In `node_modules/@blockly/plugin-workspace-search/package.json`, lines 14 to 16, there is:

```json
  "main": "./dist/index.js",
  "module": "./src/index.js",
  "unpkg": "./dist/index.js",
```

If changed to the following:

```json
  "main": "./dist/index.js",
  "module": "./dist/index.js",
  "unpkg": "./dist/index.js",
```

Then it can be used normally again.

maribethb:
@cpcallen I think I need your help with this one.

For background, all of the plugins specify the `module` property to point to `src/` instead of `dist`, and they have for a [long time](https://github.com/google/blockly-samples/pull/182). This plugin was recently converted to TS which probably exposed this behavior, but I'm not sure why nobody has ever reported this error before (with other TS plugins?) and why we don't see it ourselves when building plugins.

It seems like we are using the `module` keyword incorrectly in `package.json`. According to [this webpack guide](https://webpack.js.org/guides/author-libraries/#final-steps), it should point to an ecmascript module, but the file doesn't exist in ts plugins. In js plugins, it's unclear to me whether the file is actually an esmodule (it uses esmodule-style `import` as opposed to commonjs `require` but we don't specify the file extensions in relative imports which I thought was a requirement).

It's also worth noting that the `module` keyword was introduced and supported in webpack because it was expected to be a proposal accepted in Node.js, but node [went a different direction](https://nodejs.org/api/packages.html#packagejson-and-file-extensions) and uses `type: "module"` keyword alongside `exports` to determine which files should be interpreted as esmodules vs commonjs modules. So the `module` keyword is not actually supported by node, but does happen to still be supported by webpack and other build tools.

[...]

PR #1969, opened 2023-09-22
feat: add webpack asset loaders

### LINKED ISSUE #1145: dev-scripts webpack config does not support loading audio assets
**Category**

 - Plugins

**Component**

dev-scripts

**Describe the bug**

The current [webpack config](https://github.com/google/blockly-samples/blob/master/plugins/dev-scripts/config/webpack.config.js) (which is shared between all of the plugins in samples) does not support loading asset files (such as audio files).

**Expected behavior**

We should use the [`asset/resource`](https://webpack.js.org/guides/asset-modules/#resource-assets) loader to support loading audio files.

**Additional context**

I attempted to look into this, but my audio setup for my test environment isn't working, so I couldn't really test.

This is a blocker for #1072 

We [manually copy files](https://github.com/google/blockly-samples/blob/c74031e218fe0b01dc4d0d30cf07fd00f7e8504f/gulpfile.js#L207) used in examples so that the github pages work. This does not solve the problem of actually bundling assets with plugins.

alicialics:
## The basics

## The details

[for ghc-osd 2023 event]

This PR loads common image and audio assets inline so plugin developers can easily incorporate their own assets.
An example is included for workspace-backpack plugin.

### Resolves

Fixes  #1145 

### Proposed Changes

- Adds blockly-env.d.ts to create-package templates
- Add default blockly-env.d.ts to @blockly/dev-scripts + types field in package.json
- Fixes blockly-env.d.ts references in eslint
- Add asset/inline module loader in webpack.config.js blockly-scripts
- Add example for workspace-backpack

### Test Coverage

ran `npm run deploy:prepare:plugins`
build, lint, started devserver for workspace-backpack and verified svg was loaded correctly

### Documentation

TBD, please let me know.

### Additional Information

The reason I chose `asset/inline` over `asset/resource` is that as a library if some asset file is produced in the ./dist folder your host webapp need to pick it up and serve it themselves.  I don't think this approach would work.  So the images definitely needs to be inlined.  I am less familiar with audio files but I assume they work similarly.

Edit: I just tested `asset/resource` and it seems to work as well when I'm testing (`npm start`) workspace-backpack.. I believe it works because we use webpack to build a devServer that serves the svg at the right place.  However if the plugin is used in a different webapp (eg github pages) then unless webpack or another build system does some module magic and put the assets in a servable bundle I don't think having assets as a seperate file will work [ie have to inline them].  [from my testing github pages did not work as expected because test_bundle.js was looking for its svg files but they are not found]. If we can confirm this somehow works then it's best to use `'type': 'asset'` to let webpack auto determine whether the resource should be inlined or in a separate file.

I am not a huge fan of using webpack to bundle libraries.  I prefer this is moved to [tsup](https://github.com/egoist/tsup) (using esbuild) or rollup.

### DIFF
--- a/plugins/dev-create/templates/typescript-block/template/src/blockly-env.d.ts
+++ b/plugins/dev-create/templates/typescript-block/template/src/blockly-env.d.ts
@@ -1,0 +1,1 @@
+/// <reference types="@blockly/dev-scripts" />
--- a/plugins/dev-create/templates/typescript-field/template/src/blockly-env.d.ts
+++ b/plugins/dev-create/templates/typescript-field/template/src/blockly-env.d.ts
@@ -1,0 +1,1 @@
+/// <reference types="@blockly/dev-scripts" />
--- a/plugins/dev-create/templates/typescript-plugin/template/src/blockly-env.d.ts
+++ b/plugins/dev-create/templates/typescript-plugin/template/src/blockly-env.d.ts
@@ -1,0 +1,1 @@
+/// <reference types="@blockly/dev-scripts" />
--- a/plugins/dev-create/templates/typescript-theme/template/src/blockly-env.d.ts
+++ b/plugins/dev-create/templates/typescript-theme/template/src/blockly-env.d.ts
@@ -1,0 +1,1 @@
+/// <reference types="@blockly/dev-scripts" />
--- a/plugins/dev-scripts/config/blockly-env.d.ts
+++ b/plugins/dev-scripts/config/blockly-env.d.ts
@@ -1,0 +1,59 @@
+declare module '*.avif' {
+    const src: string;
+    export default src;
+  }
+
+  declare module '*.bmp' {
+    const src: string;
+    export default src;
+  }
+
+  declare module '*.cur' {
+    const src: string;
+    export default src;
+  }
+
+  declare module '*.gif' {
+    const src: string;
+    export default src;
+  }
+
+  declare module '*.jpg' {
+    const src: string;
+    export default src;
+  }
+
+  declare module '*.jpeg' {
+    const src: string;
+    export default src;
+  }
+
+  declare module '*.png' {
+    const src: string;
+    export default src;
+  }
+
+  declare module '*.svg' {
+    const src: string;
+    export default src;
+  }
+
+  declare module '*.webp' {
+      const src: string;
+      export default src;
+  }
+
+  declare module '*.mp3' {
+    const src: string;
+    export default src;
+  }
+
+  declare module '*.ogg' {
+    const src: string;
+    export default src;
+  }
+
+  declare module '*.wav' {
+    const src: string;
+    export default src;
+  }
--- a/plugins/dev-scripts/config/webpack.config.js
+++ b/plugins/dev-scripts/config/webpack.config.js
@@ -98,6 +98,10 @@ module.exports = (env) => {
           use: [require.resolve('source-map-loader')],
           enforce: 'pre',
         },
+        {
+          test: /\.(avif|bmp|cur|gif|jpe?g|png|svg|webp|mp3|ogg|wav)$/,
+          type: 'asset/inline',
+        },
         isTypescript && {
           test: /\.tsx?$/,
           loader: require.resolve('ts-loader'),
--- a/plugins/dev-scripts/package.json
+++ b/plugins/dev-scripts/package.json
@@ -51,6 +51,7 @@
     "access": "public",
     "registry": "https://wombat-dressing-room.appspot.com"
   },
+  "types": "./config/blockly-env.d.ts",
   "eslintConfig": {
     "extends": "@blockly/eslint-config"
   },
--- a/plugins/eslint-config/index.js
+++ b/plugins/eslint-config/index.js
@@ -219,6 +219,8 @@ module.exports = {
           {'accessibility': 'no-public'}],
         '@typescript-eslint/no-require-imports': 'error',
         '@typescript-eslint/semi': ['error', 'always'],
+        'spaced-comment': ['error', 'always',
+          {'markers': ['/']}],
       },
     },
   ],
--- a/plugins/workspace-backpack/src/backpack.ts
+++ b/plugins/workspace-backpack/src/backpack.ts
@@ -13,6 +13,9 @@
 
 import * as Blockly from 'blockly/core';
 
+import backpackSvgDataUri from '../media/backpack.svg';
+import backpackFilledSvgDataUri from '../media/filled_backpack.svg';
+
 import {registerContextMenus} from './backpack_helpers';
 import {BackpackOptions, parseOptions} from './options';
 import {BackpackChange, BackpackOpen} from './ui_events';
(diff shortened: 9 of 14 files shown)


BeksOmega:
@maribethb This isn't a change we planned on including as part of GHC, but looks reasonable to me! I'm going to assign this to you since it may need discussion. Happy to take some of your PRs in return!

maribethb:
Hi, thank you for opening this PR and looking into this issue. 

Unfortunately, I don't think it meets our goals here. For audio files, we wouldn't want to inline the entire file. I understand how that makes sense from the perspective of users of the plugin not having to host the media file themselves, but it would make the file size a lot bigger for an audio file.

I think instead, we need to configure webpack to simply copy over the asset in its entirety to the `dist` directory so that the audio assets are available to users of the plugin. Making the audio file available in their own webserver would be the responsibility of the plugin user. Then any plugins that use audio assets (n.b., none, as of today) should make the URL at which to look for the asset configurable by the user. That would mean someone could configure their own build process to copy the file from `node_modules/@blockly/workspace-backpack/dist/mySound.mp3` to be served from `public/assets/mySound.mp3` and then pass in `public/assets/` as the location for which the plugin should look for `mySound.mp3`. This is basically how Blockly core handles the same issue.

We may be able to set the default location of the file to use the new [`import.meta.url` feature](https://webpack.js.org/guides/asset-modules/#url-assets) but I think we should still let folks pass in a URL for anyone who wants to configure their server a certain way. I am not entirely sure what the code would look like for this to work, because I don't really understand what webpack is doing to these URLs under the hood.


-------

For svg files, inlining them does make sense. The files are already small and inlining them lets us avoid making users figure out how to host images. We were already inlining them by providing the dataUri for each SVG. The work done here makes it possible to inline the images automatically instead of having to provide that dataUri manually. However, it also requires adding the additional scaffolding as you've seen (e.g. configuring TypeScript types for the new file types, changing the eslint settings). Because we have such a small number of images, I don't think this extra scaffolding is worth it, and we should just continue to "manually" inline the images. If folks disagree, I'd be happy to take another look, but I think the way the new types are linked in would need another look in that case (to avoid using the triple slash directive)

--------


I hope this makes sense. Let me know how you'd like to proceed. I think we should probably close this pull request, but if you'd like to have a look at just configuring webpack to copy over mp3/other audio assets and want to use this same branch/PR that would be fine too. Let me know if you have any questions! Sorry we can't accept this PR in this form, but we do appreciate your work on this!

alicialics:
> Unfortunately, I don't think it meets our goals here. For audio files, we wouldn't want to inline the entire file. I understand how that makes sense from the perspective of users of the plugin not having to host the media file themselves, but it would make the file size a lot bigger for an audio file.

I agree with this assessment.  In addition, I'd like to make the following observations: no plugins uses audio files today, and plugin developers are not forced to inline their audio files using this approach.  They can always ask the users of their plugin to copy certain assets to a servable directory for the plugin to use.  Also my understanding is that most sound effects are [short and small](https://www.npmjs.com/package/blockly?activeTab=code)(like 3kb-7kb) while some plugin['s js payloads are 15kb+](https://www.npmjs.com/package/@blockly/field-colour-hsv-sliders?activeTab=code)

> I think instead, we need to configure webpack to simply copy over the asset in its entirety to the dist directory so that the audio assets are available to users of the plugin. 

This is easy, simply use `type: 'asset/resource'` or `type: 'asset'`.  However this generates a file with some kind of hash, it should probably be a stable name like `/dist/asset/foo.mp3` but this shouldn't be that hard to do.

> Making the audio file available in their own webserver would be the responsibility of the plugin user. Then any plugins that use audio assets (n.b., none, as of today) should make the URL at which to look for the asset configurable by the user. That would mean someone could configure their own build process to copy the file from node_modules/@blockly/workspace-backpack/dist/mySound.mp3 to be served from public/assets/mySound.mp3 and then pass in public/assets/ as the location for which the plugin should look for mySound.mp3. This is basically how Blockly core handles the same issue.

I think it's doable but it may open a can of worms.  Mainly, how would the user know which plugins needs this extra step?  I would not expect it if 90% of the plugins worked fine but there was one or two that required me to copy a few files to different directories (ideally each plugin should have its own asset directory).  Also we need to potentially consider the situation where we allow plugins that depends on other plugins..  

> We may be able to set the default location of the file to use the new [import.meta.url feature](https://webpack.js.org/guides/asset-modules/#url-assets) but I think we should still let folks pass in a URL for anyone who wants to configure their server a certain way. I am not entirely sure what the code would look like for this to work, because I don't really understand what webpack is doing to these URLs under the hood.

I'd be happy to test this but currently plugins only have examples of svg files.  But I guess I can just use svg files for testing if the blockly core team decide to try this approach.

> For svg files, inlining them does make sense. The files are already small and inlining them lets us avoid making users figure out how to host images. We were already inlining them by providing the dataUri for each SVG. The work done here makes it possible to inline the images automatically instead of having to provide that dataUri manually. However, it also requires adding the additional scaffolding as you've seen (e.g. configuring TypeScript types for the new file types, changing the eslint settings). Because we have such a small number of images, I don't think this extra scaffolding is worth it, and we should just continue to "manually" inline the images. If folks disagree, I'd be happy to take another look, but I think the way the new types are linked in would need another look in that case (to avoid using the triple slash directive)

additional scaffolding is fairly common, and since we are adding it to templates my personal opinion it won't effect new plugin developers (they may have issues if they are working on an existing plugin and bumping dev-scripts version however I assume they are probably just using dataUri and this won't affect them as well).   This is also standard practice among frameworks like Next, CRA: https://github.com/vercel/next.js/blob/canary/packages/create-next-app/templates/app-tw/ts/next-env.d.ts


> I hope this makes sense. Let me know how you'd like to proceed. 

Ultimately this is the blockly core team member's call but I am okay with whatever the team decides.

maribethb:
Requiring developers to host the audio assets themselves is certainly tricky, but I don't know of a way around that at the moment. Any plugin that wants to add audio assets would have to clearly document this requirement in the README. We do expect users of plugins to read the documentation for that plugin, so I believe this is a surmountable problem.

The piece we're missing right now is just bundling that asset with the published plugin. We may want to use `CopyWebpackPlugin` to simply copy the files instead of using an asset loader.

Regarding the inline SVGs and the additional TypeScript scaffolding: I don't like the magic of the triple-slash directives. It is confusing to me and it's not a clear way of following the dependency chain. I think we still don't need this feature right now, but if we did, we could upgrade to TS 5 and use [this new flag](https://www.typescriptlang.org/tsconfig#allowArbitraryExtensions). That would remove the need for magic module names; any plugin importing a `fileName.svg` file could also add a matching `fileName.d.svg.ts` file to satisfy the compiler.

For now, I'll close this PR. If you want to work on the audio bundling issue, you can either open a new PR or re-open this one. Thank you for the back and forth :)

alicialics:
> This is basically how Blockly core handles the same issue.

Can you share me how blockly asks developers to [handle its media assets](https://unpkg.com/browse/blockly@10.1.3/media/)? I don't see it in the [docs](https://developers.google.com/blockly/guides/get-started/web).

I find it difficult to find examples how one would use blockly and blockly-plugins as a developer.

> That would remove the need for magic module names; any plugin importing a fileName.svg file could also add a matching fileName.d.svg.ts file to satisfy the compiler.

In my experience, developers would generally prefer a single d.ts file that encapuslates all svg files rather than adding a ts file for each svg file they are exporting.

> The piece we're missing right now is just bundling that asset with the published plugin. We may want to use CopyWebpackPlugin to simply copy the files instead of using an asset loader.

This requires setting up a separate configuration file like blockly.config.ts and let people modify/inject webpack configuration. I am not convinced that `dev-scripts` should even be using webpack to create a library bundle in the first place.  There are much better options like `esbuild` and `rollup`, and once `blockly.config.ts` is introduced with webpack configuration its much harder to go back. 

> If you want to work on the audio bundling issue, you can either open a new PR or re-open this one.

I'd love to reach some kind of agreement on how best to approach this before attempting another PR.  There are related issues such as on blockly and blockly plugins should bundle eg #1877 and https://github.com/google/blockly/issues/7449 cc @cpcallen

BeksOmega:
> Can you share me how blockly asks developers to [handle its media assets](https://unpkg.com/browse/blockly@10.1.3/media/)? I don't see it in the [docs](https://developers.google.com/blockly/guides/get-started/web).

By default we point to `https://blockly-demo.appspot.com/static/media/` for our media assets, but external devs can configure it to point where ever. It is one of our [inject options](https://developers.google.com/blockly/guides/configure/web/configuration_struct) =)

> This requires setting up a separate configuration file like blockly.config.ts and let people modify/inject webpack configuration.

Could you explain a bit more about why this is required? I don't know much about the `ConfigureWebpackPlugin` but I was assuming you could have it copy based on a regexp! E.g. copy all mp4 files.

> There are much better options like `esbuild` and `rollup`

Would you be interested in writing up an issue about this? I think it's unlikely that we change the build system because there's a lot of infrastructure around it, and changing it would be a significant investment. But it would be interesting to hear your perspective on the pros and cons, since it seems like you've worked on other similar open source projects before!

> I'd love to reach some kind of agreement on how best to approach this before attempting another PR.

Makes sense! Happy to keep chatting about it more =)

PR #1971, opened 2023-09-22
feat: Index dropdown field option for toolbox-search

### LINKED ISSUE #1940: Update toolbox-search to search dropdown field options
### Check for duplicates

### Component

toolbox-search

### Problem

_No response_

### Request

The toolbox-search plugin should include the text of options in dropdown fields when filtering blocks. Right now, only the label text on the block (or the contents of text fields) are included in the index. Ideally, the matching option would also be shown pre-selected on the block in the toolbox.

### Alternatives considered

_No response_

### Additional context

_No response_

sammitako:
## The basics

### Resolves

Partially addresses 
https://github.com/google/blockly-samples/issues/1940

### Proposed Changes

When indexing blocks, check if a field is a dropdown, and index the text value and ImageProperties value of their options.

### Reason for Changes
When the block has a dropdown option, other than the default one, the hidden values in the options are not searchable.

### Test Coverage
Manually test only.

### Documentation
No.

### Additional Information
Collaborated with https://github.com/ludizhan

### DIFF
--- a/plugins/toolbox-search/src/block_searcher.ts
+++ b/plugins/toolbox-search/src/block_searcher.ts
@@ -26,7 +26,7 @@ export class BlockSearcher {
     const blockCreationWorkspace = new Blockly.Workspace();
     blockTypes.forEach((blockType) => {
       const block = blockCreationWorkspace.newBlock(blockType);
-      this.indexBlockText(blockType.replaceAll('_', ''), blockType);
+      this.indexBlockText(blockType.replaceAll('_', ' '), blockType);
       block.inputList.forEach((input) => {
         input.fieldRow.forEach((field) => {
           this.indexDropdownOption(field, blockType);
@@ -45,9 +45,9 @@ export class BlockSearcher {
     if (field instanceof Blockly.FieldDropdown) {
       field.getOptions(true).forEach((option) => {
         if (typeof option[0] === 'string') {
-          return this.indexBlockText(option[0], blockType);
+          this.indexBlockText(option[0], blockType);
         } else if ('alt' in option[0]) {
-          return this.indexBlockText(option[0].alt, blockType);
+          this.indexBlockText(option[0].alt, blockType);
         }
       });
     }
--- a/plugins/toolbox-search/src/block_searcher.ts
+++ b/plugins/toolbox-search/src/block_searcher.ts
@@ -26,13 +26,31 @@ export class BlockSearcher {
     const blockCreationWorkspace = new Blockly.Workspace();
     blockTypes.forEach((blockType) => {
       const block = blockCreationWorkspace.newBlock(blockType);
-      this.indexBlockText(blockType.replaceAll('_', ' '), blockType);
+      this.indexBlockText(blockType.replaceAll('_', ''), blockType);
       block.inputList.forEach((input) => {
         input.fieldRow.forEach((field) => {
+          this.indexDropdownOption(field, blockType);
           this.indexBlockText(field.getText(), blockType);
         });
       });
     });
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


humera811:
I faced a similar linting issue. You can try
` git commit --amend -m "fix: Update toolbox-search to search dropdown field options" `
` git push --force origin osd`

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -29,12 +29,30 @@ export class BlockSearcher {
...
+   * @param field We need to check the type of field
+   * @param blockType The block type to associate the trigrams with.
+   */
+  private indexDropdownOption(field: Blockly.Field, blockType: string) {
+    if (field instanceof Blockly.FieldDropdown) {
+      field.getOptions(true).forEach((option) => {
+        if (typeof option[0] === 'string') {
+          return this.indexBlockText(option[0], blockType);
+        } else if ('alt' in option[0]) {
+          return this.indexBlockText(option[0].alt, blockType);
```
humera811: I am not sure if this would call every option in dropdown field.
Maybe : remove the return statement  `this.indexBlockText(option[0].alt, blockType);`
but I will leave to @johnnesky to decide.

humera811:
> I faced a similar linting issue. You can try `git commit --amend -m "fix: Update toolbox-search to search dropdown field options"` ` git push --force origin osd`

fix instead of Fix

ludizhan:
@humera811 @sammitako There are 3 commits in this PR, the issue is probably caused by the other commits in the PR.

Can you try this?

- `git rebase -i HEAD~3` Enter interactive rebase mode for the 3 most recent commits
- Replace `pick` with `reword` for the commit that doesn't start with `fix:` or `feat:`
    - Basically replace the first line `pick f5822adf Index dropdown field option for toolbox-search` -> `reword f5822adf Index dropdown field option for toolbox-search`
    - The allows you to update the commit message
- Modify the commit message to `fix: Index dropdown field option for toolbox-search`
- After saving, you should exit the interactive rebase mode
- `git push --force origin osd`

plugins/toolbox-search/src/block_searcher.ts
ludizhan: We are calling `option[0]` because the MenuOption type is an array containing 2 elements, and we only care about the first element, which is the human-readable value (text or image) (defined here: https://github.com/google/blockly/blob/e4d8f160aa72649220aebc25e4eef6bd5ea267bf/core/field_dropdown.ts#L674)

The if / else statement here is handling cases when `option[0]` is a `string`, or a `ImageProperties`.

The actual logic looping through every option in the dropdown field is on line 46: `field.getOptions(true).forEach((option) => {...})`

ludizhan:
Note: This actually partially fixes https://github.com/google/blockly-samples/issues/1940

There were 2 steps stated in the original issue about how to fix it, and this completes and first step.

We intend to follow up with another PR later to complete the second step, as well as adding unit tests.

plugins/toolbox-search/src/block_searcher.ts
ludizhan: Good catch! You are right, we shouldn't be `return`-ing anything, and we just need to call `this.indexBlockText()` to generate trigrams. 

@sammitako can you remove the word `return` from line 48 and line 50? The rest should keep the same and do the work.

sammitako: Thank you for the catch @humera811. 
@ludizhan I just removed `return` on line 48 and line 50 and pushed.

johnnesky (approved):

johnnesky:
Thanks for the hard work everyone! I look forward to part 2 of this issue!

Issue #1531, opened 2023-01-26
Update minimum node engine value in plugins and template


maribethb:
The create-package templates create every plugin to say minimum node engine is 8; I think this needs to be updated though I'm actually not sure as of this writing what the minimum node version of Blockly is. Hopefully it's higher than 8 though...

We should find this value, update all existing plugins, and update the templates

cpcallen:
> Hopefully it's higher than 8 though...

Why would we hope that?  Ideally everything would work fine in the oldest versions of node (though I doubt that is the case!). 
 Do we know for sure that there is an incompatibility with a version >= 8?

On the other hand, we just say the minimum version is 14, since (at least in core) that is the oldest we currently test.

maribethb:
Ok, hopefully was the wrong word. But I am pretty sure we had to bump the minimum node version of Blockly, though that may have only been for test reasons.

We don't even specify this value in core blockly's package.json. I think we probably shouldn't specify it for every single plugin unless the plugin has an actual known problem or dependency on a specific node engine. If Blockly itself has a minimum version, we should document it there and not in the plugins. So I think we should just remove this value from the templates and from all the plugins as none of them (to my knowledge) actually require a specific node version.

cpcallen:
> So I think we should just remove this value from the templates and from all the plugins as none of them (to my knowledge) actually require a specific node version.

I concur.

alicialics:
minimum node version for [webpack 5](https://webpack.js.org/migrate/5/#preparations) which dev-script uses is 10.13 so IMO it should at least that.  But Node 16 is out of maintenance so I would prefer if we could just support [node >= 18](https://nodejs.dev/en/about/releases/)..

PR #1972, opened 2023-09-23
fix: progress for Have field-bitmap fire intermediate change events #1912

### LINKED ISSUE #1912: Have field-bitmap fire intermediate change events
🚧⚠️🚧 This issue is being saved for **Grace Hopper Open Source Day**. 🚧⚠️🚧 An event we're running to help girls and non-binary technologists make their first contributions to open source! Please see our list of [help wanted issues](https://github.com/google/blockly-samples/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22+-label%3A%22ghc-osd%22) if you'd like to contribute! 

---

**Category**

 - Plugins

**Component**

field-bitmap

The bitmap field fires undoable `BlockChange` events for every square that gets painted. This makes undoing a change time consuming.

**To reproduce**

1) `cd` into `plugins/field-bitmap`.
2) Run `npm run start` to open the playground.
3) In the right nav bar under "Actions" check "Log Events". Then open your browser console so you can see the logged events.
4) Drag a bitmap field onto the workspace.
5) Open the editor and edit the bitmap.
6) Observe how `BlockChange` events are fired for every change.
7) Undo the action using `ctrl + z` (or `cmd + z`), and observe how only one pixel is changed. Ideally the whole action should be undone.

Instead it should fire `IntermediateFieldChange` events for these changes, and only fire a `BlockChange` event when its editor is closed.

## To Fix

1) Run the reproduction steps and make sure you can reproduce the issue.
2) Add a property in the [constructor](https://github.com/google/blockly-samples/blob/11e2600f4b8391b82a67cb14dcf4f1c5e4c0466d/plugins/field-bitmap/src/field-bitmap.js#L34) that records what the value of the field is at the start of an edit.
3) Modify the [dropdownCreate](https://github.com/google/blockly-samples/blob/11e2600f4b8391b82a67cb14dcf4f1c5e4c0466d/plugins/field-bitmap/src/field-bitmap.js#L234) method to store the value of the field at the start of the edit.
4) Modify the [setPixel](https://github.com/google/blockly-samples/blob/11e2600f4b8391b82a67cb14dcf4f1c5e4c0466d/plugins/field-bitmap/src/field-bitmap.js#L461) method to fire `BLOCK_FIELD_INTERMEDIATE_CHANGE` events. Make sure to call `setValue` still, but pass `false` for the second parameter.
```js
if (this.getSourceBlock()) {
      Blockly.Events.fire(
          new (Blockly.Events.get(
              Blockly.Events.BLOCK_FIELD_INTERMEDIATE_CHANGE))(
              this.sourceBlock_,
              this.name || null,
              oldValue,
             this.value_
          )
      );
    }
```
5) Do the same for the `randomizePixels` and the `clearPixels` methods.
5) Modify the [dropdownDispose](https://github.com/google/blockly-samples/blob/11e2600f4b8391b82a67cb14dcf4f1c5e4c0466d/plugins/field-bitmap/src/field-bitmap.js#L367) method to fire a `BLOCK_CHANGE` event. It will look very similar to how you fired the other event, but it will use the property for the value at the start of the edit.
6) Run the reproduction steps again and make sure you can't reproduce the behavior.

wich229:
## The basics

### Resolves

Fixes #1912 

### Additional Information

Hi,

I've been working on fixing the undo issue for bitmap, and I've made some progress. However, I have a couple of questions and an issue that I'd like to discuss:

**dropdownDispose Implementation:** I've implemented changes related to dropdownDispose, but I'd appreciate a review to ensure it's correctly.

**Undo Behavior Issue**:  After making changes, I've noticed an issue where, when the last block is reached and all undo actions have been performed, it requires multiple presses of Ctrl+Z (or Cmd+Z) to clear the block. It seems like there are several intermediate steps involved in getting the block to delete. I'd appreciate some guidance on how to address this behavior.

As this is my first time contributing to an open-source project, I'm eager to learn and improve. Your feedback and assistance would be greatly appreciated.

Thank you for your time and support.

Best,
Yi-Chun

### DIFF
--- a/examples/rtl-demo/index.html
+++ b/examples/rtl-demo/index.html
@@ -23,147 +23,370 @@
   </style>
   <script>
     function init() {
-      var demoWorkspace = Blockly.inject('blocklyDiv',
-          {media: './node_modules/blockly/media/',
-           rtl: true,
-           toolbox: document.getElementById('toolbox')});
+      const toolbox = {
+        "kind": "categoryToolbox",
+        "contents": [
+          {
+            "kind": "category",
+            "name": "منطق",
+            "colour": "%{BKY_LOGIC_HUE}",
+            "contents": [
+              {
+                "kind": "block",
+                "type": "controls_if"
+              },
+              {
+                "kind": "block",
+                "type": "logic_compare"
+              },
+              {
+                "kind": "block",
+                "type": "logic_operation"
+              },
+              {
+                "kind": "block",
+                "type": "logic_negate"
+              },
+              {
+                "kind": "block",
+                "type": "logic_boolean"
+              },
+              {
+                "kind": "block",
+                "type": "logic_null"
+              },
+              {
+                "kind": "block",
+                "type": "logic_ternary"
+              },
+            ]
+          },
+          {
+            "kind": "category",
+            "name": "الحلقات",
+            "colour": "%{BKY_LOOPS_HUE}",
+            "contents": [
+              {
+                "kind": "block",
+                "type": "controls_repeat_ext",
+                "inputs": {
+                  "TIMES": {
+                    "block": {
+                      "type": "math_number",
+                      "fields": {
+                        "NUM": 10
+                      }
+                    }
+                  }
+                }
+              },
+              {
+                "kind": "block",
+                "type": "controls_whileUntil"
+              },
+              {
+                "kind": "block",
+                "type": "controls_for",
+                "inputs": {
+                  "FROM": {
+                    "block": {
+                      "type": "math_number",
+                      "fields": {
+                        "NUM": 1
+                      }
+                    }
+                  },
+                  "TO": {
+                    "block": {
+                      "type": "math_number",
+                      "fields": {
+                        "NUM": 10
+                      }
+                    }
+                  },
+                  "BY": {
+                    "block": {
+                      "type": "math_number",
+                      "fields": {
+                        "NUM": 1
+                      }
+                    }
+                  },
+                }
+              },
+              {
+                "kind": "block",
+                "type": "controls_forEach"
+              },
+              {
+                "kind": "block",
+                "type": "controls_flow_statements"
+              },
+            ]
+          },
+          {
+            "kind": "category",
+            "name": "رياضيات",
+            "colour": "%{BKY_MATH_HUE}",
+            "contents": [
+              {
+                "kind": "block",
+                "type": "math_number"
+              },
+              {
+                "kind": "block",
+                "type": "math_arithmetic"
+              },
+              {
+                "kind": "block",
+                "type": "math_single"
+              },
+              {
+                "kind": "block",
+                "type": "math_trig"
+              },
+              {
+                "kind": "block",
+                "type": "math_constant"
+              },
+              {
+                "kind": "block",
+                "type": "math_number_property"
+              },
+              {
+                "kind": "block",
+                "type": "math_round"
+              },
+              {
+                "kind": "block",
+                "type": "math_on_list"
+              },
+              {
+                "kind": "block",
+                "type": "math_modulo"
+              },
+              {
+                "kind": "block",
+                "type": "math_modulo"
+              },
+              {
+                "kind": "block",
+                "type": "math_constrain",
+                "inputs": {
+                  "LOW": {
+                    "block": {
+                      "type": "math_number",
+                      "fields": {
+                        "NUM": 1
+                      }
+                    }
+                  },
+                  "HIGH": {
+                    "block": {
+                      "type": "math_number",
+                      "fields": {
+                        "NUM": 100
+                      }
+                    }
+                  }
+                }
+              },
+              {
+                "kind": "block",
+                "type": "math_random_int",
+                "inputs": {
+                  "FROM": {
+                    "block": {
+                      "type": "math_number",
+                      "fields": {
+                        "NUM": 1
+                      }
+                    }
+                  },
+                  "TO": {
+                    "block": {
+                      "type": "math_number",
+                      "fields": {
+                        "NUM": 100
+                      }
+                    }
+                  }
+                }
+              },
+              {
+                "kind": "block",
+                "type": "math_random_float"
+              },
+              {
+                "kind": "block",
+                "type": "math_atan2"
+              },
+            ]
+          },
+          {
+            "kind": "category",
+            "name": "نص",
+            "colour": "%{BKY_TEXTS_HUE}",
+            "contents": [
+              {
+                "kind": "block",
+                "type": "text"
+              },
+              {
+                "kind": "block",
+                "type": "text_join"
+              },
+              {
+                "kind": "block",
+                "type": "text_append",
+                "inputs": {
+                  "TEXT": {
+                    "kind": "block",
+                     "type": "text"
+                  }
+                }
+              },
+              {
+                "kind": "block",
+                "type": "text_length"
+              },
+              {
+                "kind": "block",
+                "type": "text_isEmpty"
+              },
+              {
+                "kind": "block",
+                "type": "text_indexOf"
+              },
+              {
+                "kind": "block",
+                "type": "text_charAt"
+              },
+              {
+                "kind": "block",
+                "type": "text_changeCase"
+              },
+              {
+                "kind": "block",
+                "type": "text_trim"
+              },
+              {
+                "kind": "block",
+                "type": "text_print"
+              },
+              {
+                "kind": "block",
+                "type": "text_prompt_ext",
+                "inputs": {
+                  "TEXT": {
+                    "kind": "block",
+                     "type": "text"
+                  }
+                }
+              },
+            ]
+          },
+          {
+            "kind": "category",
+            "name": "قوائم",
+            "colour": "%{BKY_LISTS_HUE}",
+            "contents": [
+              {
+                "kind": "block",
+                "type": "lists_create_empty"
+              },
+              {
+                "kind": "block",
+                "type": "lists_create_with"
+              },
+              {
+                "kind": "block",
+                "type": "lists_repeat",
+                "inputs": {
+                  "NUM": {
+                    "block": {
+                      "type": "math_number",
+                      "fields": {
+                        "NUM": 5
+                      }
+                    }
+                  }
+                }
+              },
+              {
+                "kind": "block",
+                "type": "lists_length"
+              },
+              {
+                "kind": "block",
+                "type": "lists_isEmpty"
+              },
+              {
+                "kind": "block",
+                "type": "lists_indexOf"
+              },
+              {
+                "kind": "block",
+                "type": "lists_getIndex"
+              },
+              {
+                "kind": "block",
+                "type": "lists_setIndex"
+              },
+            ]
+          },
+          {
+            "kind": "category",
+            "name": "لون",
+            "colour": "%{BKY_COLOUR_HUE}",
+            "contents": [
+              {
+                "kind": "block",
+                "type": "colour_picker"
+              },
+              {
+                "kind": "block",
+                "type": "colour_rgb"
+              },
+              {
+                "kind": "block",
+                "type": "colour_blend"
+              },
+            ]
+          },
+          {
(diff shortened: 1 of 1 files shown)


BeksOmega:
Just wanted to let you know I'm traveling on monday @wich229 so I'll get back to you on Tuesday! From a brief glance changes generally look good =) 

If you could sign the CLA that would be great! That's necessary for us to be able to accept your changes =)

Thank you for your hard work on this!

wich229:
@BeksOmega Thank you for letting me know and for your help! I signed the CLA. Thanks again. :)

package.json
```diff
@@ -33,10 +33,10 @@
     "conventional-changelog-conventionalcommits": "^5.0.0",
     "eslint": "^7.15.0",
     "gh-pages": "^3.1.0",
-    "gulp": "^4.0.2",
```
debaraj-barua: Just wondering, are these changes to change the package version for `gulp` and `js-green-licenses` needed?

package-lock.json
debaraj-barua, line 1: there seems to be `15,895 additions, 10,475 deletions`, maybe unintentional?

plugins/field-colour-hsv-sliders/package.json
```diff
@@ -59,6 +59,6 @@
     "extends": "@blockly/eslint-config"
   },
   "engines": {
-    "node": ">=8.0.0"
```
debaraj-barua: Same question related to this change?

debaraj-barua (commented):
Just wondering if the changes to `package.json` and `package-lock` are intentional or side-effects

plugins/field-colour-hsv-sliders/package.json
wich229: Hi @debaraj-barua!
I updated the Node.js version to address issues on my local host. The suggestion was to upgrade to 14+, but I chose version 16 to resolve the problem, resulting in unintended package changes.

I apologize for any confusion. This is my first experience with an open-source project, and If you have any questions or suggestions, please feel free to let me know. Thx!

debaraj-barua: No need to apologize 😃 I was just wondering about the change. This seems like a good idea to upgrade node, but it might have been set to `8` for backwards compatibility. I am unsure of this.

One of the maintainers might be able to actually answer my question, if this change is required/desirable.

cc: @BeksOmega

package.json
debaraj-barua: @wich229 as for these change though, if you feel this was not something you meant to change, you could undo the changes. Again, I am not sure about that, just wanted to bring it up, since it seemed like a side-effect to me :)

wich229: Thank you for sharing your thoughts and suggestions. I'll await the maintainer's feedback and make any necessary adjustments. :)

package-lock.json
debaraj-barua: Additionally, if you are interested, here are some things about usage of `npm install` and `package-lock`s (which I wrote some time ago to explain to someone, maybe it's helpful for you too) You could also jump to the TL;DR;

## What is the `package-lock`?
`package-lock` is supposed to ensure exact versions of the packages that you and your team are using, so that you can ensure that your project works consistently across different environments. It is updated whenever you run `npm install` ([read more](https://stackoverflow.com/questions/44297803/what-is-the-role-of-the-package-lock-json))

## What happens when I use `npm install`?

From the name itself, it's evident that `npm install` installs the dependencies required to run the app. However, `npm install` is non-deterministic. When you run this command, it'll go through the `package.json` and install all compatible versions at that time. This might be different from the original file committed to git. 

Although projects might have different rules/practices around updating the `package-lock`, it is normally only advisable to update it if you are actually updating some package versions and/or you are aware that all changes in the `package-lock` is relevant. (since this file is generally quite big, GitHub often fails to show the diff)

## Any easy solutions?

use `npm ci` ([docs](https://docs.npmjs.com/cli/v8/commands/npm-ci)). This is actually meant to be used in automated environments like CI/CD pipelines. However, it works in our favor since it ensures that we only install the exact packages and versions as defined by the `package-lock`.  It ALWAYS creates the same `node_modules` tree, otherwise throws an error. ([Read More](https://medium.com/helpshift-engineering/package-lock-json-the-complete-guide-2ae40175ebdd))

## Problems while using `npm ci`

More often than not, you might also face some issues using `npm ci`; this is because, you might locally use a different `node.js` and `npm` major version as compared to the version which was used when the package-lock was committed. 

## So, back to square one? (TL;DR)

There are other solutions. You could use the flag `--no-save` when installing. i.e., `npm install --no-save`. This should ensure that the package-lock is not updated.

Or, just avoid committing the `package-lock` unless you feel that the change is relevant. You can just commit the relevant files.

### Bonus lecture

Probably you might prefer to use `git add -A` or `git add .` or `git commit -am "message"` to stage and then commit all the changes you are working with. However, while convenient, you should be cautious about overusing these, since you might accidentally add and commit things that aren’t ready or were never intended to be committed to the repo. ([taken from](https://www.gitkraken.com/learn/git/git-add))

It's a good idea to explicitly add files relevant to each commit. An IDE like VSCode now makes things way easier to do this :)  

## Apologies

Sorry for the long-winding lecture. I never know when to stop. Here's a doggo if you managed to read through my long wall of text

image

wich229: Thank you!

plugins/field-colour-hsv-sliders/package.json
BeksOmega: Hello! We don't want to bump the minimum node version at the moment.

We have a separate [issue](https://github.com/google/blockly-samples/issues/1531) filed for investigating this and possibly upgrading it. So we're planning to fix that separately =)

package.json
BeksOmega: Yes we would like to rever these changes. If they are necessary can you explain why? then we can get them upgraded as part of a separate change!

BeksOmega (changes requested):
I need to manually test the changes here, but wanted to get you feedback about the dependency updates!

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -37,6 +38,9 @@ export class FieldBitmap extends Blockly.Field {
     this.SERIALIZABLE = true;
     this.CURSOR = 'default';
 
+    // Add property in the constructor
```
BeksOmega: Nit: End all comments with a period. Here and below!

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -461,6 +517,25 @@ export class FieldBitmap extends Blockly.Field {
   setPixel_(r, c, newValue) {
     const newGrid = JSON.parse(JSON.stringify(this.getValue()));
     newGrid[r][c] = newValue;
+
+    /*
+    Modify the setPixel method to fire BLOCK_FIELD_INTERMEDIATE_CHANGE
+    events. Make sure to call setValue still, but pass false for the
+    second parameter.
+    */
```
BeksOmega: Can you delete these comments (here and above).

Writing code comments is great! But It's more useful when they explain /why/ something is happening, rather than /what/ is happening. The code should already tell you what is happening!

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -436,6 +458,23 @@ export class FieldBitmap extends Blockly.Field {
...
+    Do the same as setPixel for the randomizePixels method.
+    */
+    const oldValue = this.initialValue_;
+    if (this.getSourceBlock()) {
+      Blockly.Events.fire(
+          new (Blockly.Events.get(
+              Blockly.Events.BLOCK_FIELD_INTERMEDIATE_CHANGE))(
+              this.sourceBlock_,
+              this.name || null,
+              oldValue,
```
BeksOmega: This should be `this.getValue()`, I.e. the value before we randomize the pixels.

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -436,6 +458,23 @@ export class FieldBitmap extends Blockly.Field {
...
+              this.setValue(oldValue, false)
```
BeksOmega: This should be `newValue`. I.E. the value after we randomize the pixels.

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -436,6 +458,23 @@ export class FieldBitmap extends Blockly.Field {
...
+          )
+      );
+    }
+
     this.setValue(newVal);
```
BeksOmega: This is where you need to pass `false`.

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -448,6 +487,23 @@ export class FieldBitmap extends Blockly.Field {
...
+    const oldValue = this.initialValue_;
+    if (this.getSourceBlock()) {
+      Blockly.Events.fire(
+          new (Blockly.Events.get(
+              Blockly.Events.BLOCK_FIELD_INTERMEDIATE_CHANGE))(
+              this.sourceBlock_,
+              this.name || null,
+              oldValue,
+              this.setValue(oldValue, false)
+          )
```
BeksOmega: Apply the same changes as commented above!

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -461,6 +517,25 @@ export class FieldBitmap extends Blockly.Field {
...
+      );
```
BeksOmega: Apply the same changes as commented above!

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -365,13 +372,28 @@ export class FieldBitmap extends Blockly.Field {
...
   dropdownDispose_() {
+    // Fire BLOCK_CHANGE event with the initial and
+    // final values
+    if (this.getSourceBlock() && this.initialValue_ !== null) {
+      Blockly.Events.fire(
+          new (Blockly.Events.get(Blockly.Events.BLOCK_CHANGE))(
+              this.sourceBlock_,
+              'field',
+              this.name || null,
+              this.initialValue_,
```
BeksOmega: Can you set this back to null after firing for safety?

plugins/field-bitmap/src/field-bitmap.js
```diff
@@ -365,13 +372,28 @@ export class FieldBitmap extends Blockly.Field {
    * @private
    */
   dropdownDispose_() {
+    // Fire BLOCK_CHANGE event with the initial and
+    // final values
+    if (this.getSourceBlock() && this.initialValue_ !== null) {
```
BeksOmega: Can you also check that the current value isn't the same as the initial value? If it's the same there's really no reason to fire a change event!

BeksOmega (changes requested):
This change looks really great! Just a few fixes and then I'll do a retest!

wich229:
@BeksOmega Thank you for your feedback. As I am attending the event virtually, I will try to make the adjustments and provide you with an update by tomorrow night. :)

BeksOmega:
> @BeksOmega Thank you for your feedback. As I am attending the event virtually, I will try to make the adjustments and provide you with an update by tomorrow night. :)

All good, no rush! Hope you're enjoying the conference :D

wich229:
@BeksOmega I've just pushed the updates, and reverted the package.json files. Please take a look when you have a moment, and please let me know if I need to adjust anything. :)

plugins/field-bitmap/src/field-bitmap.js
BeksOmega: This still needs fixes! Here and the two unresolved comments below =)

wich229: Hello @BeksOmega, I've pushed the updates once more. I appreciate your patience in reviewing my changes.

BeksOmega:
Heylo! Sorry I haven't gotten to this yet :/ we're running a design sprint this week so I haven't had time to give this a good look! Just wanted to let you know I haven't forgotten about it!

wich229:
No need to worry at all! Thanks for letting me know. :)

PR #1973, opened 2023-09-25
chore: switch to new umd library output in webpack

### LINKED ISSUE #1849: Webpack libraryTarget config option is deprecated
**Category**

Plugins

**Component**

dev-scripts

**Describe the bug**

This is just a heads-up that it seems that webpack has deprecated the `output.libraryTarget` config option in favor of the `output.library.type` option.  You can see the deprecation warning in their config file documentation [here](https://webpack.js.org/configuration/output/#outputlibrarytarget).  The dev-scripts code that needs to be fixed is in [dev-scripts/config/webpack.config.js](../blob/master/plugins/dev-scripts/config/webpack.config.js#L79).  It looks like the fix is to replace:
```
      libraryTarget: 'umd',
```
with
```
      library: {
        type: 'umd',
      },
```

alicialics:
## The basics

## The details

`libraryTarget` is deprecated so we use the new configuration `library.type = 'umd'`.

### Resolves

Fixes #1849 

### Test Coverage

I built the library in the old way and the new way and verify the build output has no differences.

For those who are curious, if we add `name: 'Blockly'` to the library section, we get the following diff:

```
<   "object" == typeof exports && "object" == typeof module
<     ? (module.exports = e(require("blockly/core")))
<     : "function" == typeof define && define.amd
<     ? define(["blockly/core"], e)
<     : "object" == typeof exports
<     ? (exports.Blockly = e(require("blockly/core")))
<     : (t.Blockly = e(t.Blockly));
---
>   if ("object" == typeof exports && "object" == typeof module)
>     module.exports = e(require("blockly/core"));
>   else if ("function" == typeof define && define.amd)
>     define(["blockly/core"], e);
>   else {
>     var o =
>       "object" == typeof exports ? e(require("blockly/core")) : e(t.Blockly);
>     for (var s in o) ("object" == typeof exports ? exports : t)[s] = o[s];
>   }
```

### Additional Information

cc @rachel-fenichel

### DIFF
--- a/plugins/dev-scripts/config/webpack.config.js
+++ b/plugins/dev-scripts/config/webpack.config.js
@@ -77,7 +77,9 @@ module.exports = (env) => {
       path: isProduction ? resolveApp('dist') : resolveApp('build'),
       publicPath: isProduction ? '/dist/' : '/build/',
       filename: outputFile,
-      libraryTarget: 'umd',
+      library: {
+        type: 'umd',
+      },
       globalObject: 'this',
       clean: true,
     },


maribethb (approved):
The change as-is looks good to me.

We wouldn't want to set the name to `Blockly` as we don't want people to access plugin properties under the name `Blockly`. By not setting a name, everything exported from a plugin is just shoved into the root/window/global object, which is... maybe not ideal, but it's what we've been doing, so changing it would be a breaking change we'd have to figure out how to communicate.

maribethb:
Could you update this PR to be against master now? We've merged the OSD branch so the outstanding branches can go straight to master. Thanks!

alicialics:
done!

cpcallen (approved):
LGTM.  WIll leave to @maribethb to do the merge.

PR #1995, opened 2023-09-28
feat: basic block factory implementation

maribethb:
## The basics

### Proposed Changes

Can be reviewed commit-wise.

Basic implementation of developer tools. There is still some placeholder text and things that need to be cleaned up. More work is following this :)

- Copies over the block definitions from core/demos, and updates them to use TypeScript, es5 syntax, JSON serialization helpers where appropriate
    - Does not copy over the angle, colour, or multiline text field blocks since those will be used from the plugin. 
- Changes the `input` block to be a single block type with a dropdown for the input type. This makes it easier to change input styles.
-  Changes the `type` block to be a single block type with a dropdown for the type. This makes it easier to change types.
- Uses a CodeGenerator class to construct the JSON definition for the blocks
    - This is a copy of the JS generator that I commented a bunch of stuff out in. I plan to clean this up separately as I needed more investigation as to what was required and what wasn't.
    - I plan to follow a similar path for creating the JS definition, as well as preamble code like import statements needed for plugin fields, the generator stubs, etc.

Most of the code currently in `src/index.ts` will be rearranged and moved to different files. I have another branch branched from this one where I've started rearranging the logic to use controllers and separate out the dom manipulation code. I wanted to start by validating the CodeGenerator approach instead of starting by creating a bunch of application scaffolding.

Since this isn't being published anywhere I think it's fine to go directly into master but can also put it in a branch if that's preferred.

### Reason for Changes

Modernizing dev-tools.

### Test Coverage

No unit tests yet. We lack infrastructure for running tests in examples/

### Additional Information

desktop view
image

mobile view
image


examples/developer-tools/src/output-generators/fields/image.ts
```diff
@@ -0,0 +1,23 @@
...
+    block: Blockly.Block,
+    generator: JsonDefinitionGenerator
+): string {
+  const code = {
+    type: 'field_image',
+    src: block.getFieldValue('SRC'),
+    width: block.getFieldValue('WIDTH'),
+    height: block.getFieldValue('HEIGHT'),
+    alt: block.getFieldValue('ALT'),
+    flipRtl: block.getFieldValue('ALT_RTL'),
```
alicialics: FLIP_RTL

examples/developer-tools/src/blocks/factory_base.ts
```diff
@@ -0,0 +1,114 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly';
```
rachel-fenichel: Should these be `import * as Blockly from 'blockly/core';`? I remember you mentioning that in a different PR.

examples/developer-tools/src/blocks/factory_base.ts
```diff
@@ -0,0 +1,114 @@
...
+      ['inline inputs', 'INT']]);
+    this.appendDummyInput()
+        .appendField(inputsDropdown, 'INLINE');
+    const connectionsDropdown = new Blockly.FieldDropdown([
+      ['no connections', 'NONE'],
+      ['← left output', 'LEFT'],
+      ['↕ top+bottom connections', 'BOTH'],
+      ['↑ top connection', 'TOP'],
+      ['↓ bottom connection', 'BOTTOM']],
+    function(option: string): string {
```
rachel-fenichel: Fix indentation to make it clearer that this is the validator on the dropdown.

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,395 @@
...
+  }
+  const msg = (count > 1) ?
+      'There are ' + count + ' field blocks\n with this name.' : null;
+  referenceBlock.setWarningText(msg);
+};
+
+/**
+ * Label (non-serializable) field.
+ */
+export const fieldLabel = {
```
rachel-fenichel, line 41: A small opinion: block definitions that can be written only in JSON should be. JavaScript definitions should be used for more dynamic blocks.

rachel-fenichel (commented):
Overall looks reasonable. I prefer that you put it into a feature branch rather than master.

examples/developer-tools/src/blocks/fields.ts
maribethb: This gets complicated if I'm trying to export the block definitions so that they can be registered from a different file than they're defined in. I would have to know whether the thing I'm exporting is a JS definition or a JSON definition, and then call them differently in the register call. I think it's simpler and less prone to error to define and export them all consistently even though that means using JS when I could use JSON.

We have been trying to remove side effects from our files but in core we still register blocks in the file we declare them so I don't have a good pattern to follow. I'm open to suggestions if this is important to you.

maribethb:
@BeksOmega ready for review; I believe I addressed the feedback so far and I created a new feature branch for this. Thanks!

examples/developer-tools/tsconfig.json
BeksOmega, line 1: I know the tsconfigs have gone through a lot of changes recently. So just want to check this is up to date.

examples/developer-tools/src/blocks/factory_base.ts
```diff
@@ -0,0 +1,117 @@
+import * as Blockly from 'blockly';
+
+export const factoryBase = {
+  // Base of new block.
+  init: function() {
+    this.setColour(120);
```
BeksOmega, line 12: Do we care about using best practices? If so this should probably be using a theme. I won't comment on any other "best practices" things until I know if we care or not.

examples/developer-tools/src/blocks/factory_base.ts
```diff
@@ -0,0 +1,117 @@
...
+      case 'BOTH':
+        this.connectOutputShadow('TOPTYPE');
+        this.connectOutputShadow('BOTTOMTYPE');
+        break;
+    }
+  },
+  connectOutputShadow: function(outputType: string) {
+    // Helper method to create & connect shadow block.
+    const type = this.workspace.newBlock('type_null');
+    type.setShadow(true);
```
BeksOmega: `setShadow` is internal. This should use `setShadowState`.

examples/developer-tools/src/blocks/factory_base.ts
```diff
@@ -0,0 +1,117 @@
...
+  },
+  mutationToDom: function() {
+    const container = Blockly.utils.xml.createElement('mutation');
+    container.setAttribute('connections', this.getFieldValue('CONNECTIONS'));
+    return container;
+  },
+  domToMutation: function(xmlElement: Element) {
+    const connections = xmlElement.getAttribute('connections');
+    this.updateShape(connections);
+  },
```
BeksOmega: Why is this using XML hooks instead of JSON hooks?

examples/developer-tools/src/blocks/inputs.ts
```diff
@@ -0,0 +1,149 @@
+import * as Blockly from 'blockly';
+
+const fieldMessage = 'fields %1 %2';
```
BeksOmega: This confused me for a sec. It looks like this is just being separated out to remove code duplication? Personally I think it's better to duplicate code in block definitions to aid readability, since they're declarative and change so infrequently. Just like unit tests.

examples/developer-tools/src/blocks/inputs.ts
```diff
@@ -0,0 +1,149 @@
...
+  {
+    'type': 'input_value',
+    'name': 'TYPE',
+    'check': 'Type',
+    'align': 'RIGHT',
+  },
+];
+
+/**
+ * Check to see if more than one input has this name.
```
BeksOmega: Can you document the side effect that this method has? I.E. applying the warning.

examples/developer-tools/src/blocks/inputs.ts
```diff
@@ -0,0 +1,149 @@
...
+ * Highly inefficient (On^2), but n is small.
+ * @param referenceBlock Block to check.
+ */
+function inputNameCheck(referenceBlock: Blockly.Block) {
+  if (!referenceBlock.workspace) {
```
BeksOmega: Check `isDisposed` or `isDeadOrDying` instead.

examples/developer-tools/src/blocks/inputs.ts
```diff
@@ -0,0 +1,149 @@
...
+      'message0': 'value input %1 %2',
+      'args0': [
+        {
+          'type': 'field_input',
+          'name': 'INPUTNAME',
+          'text': 'NAME',
+        },
+        {
+          'type': 'input_dummy',
+        },
```
BeksOmega: I don't think the trailing dummy input is necessary. It should get added automatically. Although maybe that doesn't work with multiple message strings?

examples/developer-tools/src/toolbox.ts
```diff
@@ -0,0 +1,28 @@
+export const toolbox = {
+  'kind': 'categoryToolbox',
+  'contents': [
+    {
+      'kind': 'category',
+      'name': 'Input',
+      'categorystyle': 'logic_category',
```
BeksOmega: If you're going to use the category style might as well use themes on the blocks. That way they stay in sync.

examples/developer-tools/src/index.ts
```diff
@@ -33,9 +38,60 @@ const startBlocks = {
...
+
+// Whenever the workspace changes meaningfully, update the preview.
+mainWorkspace.addChangeListener((e: Blockly.Events.Abstract) => {
+  // Don't run the code when the workspace finishes loading; we're
+  // already running it once when the application starts.
+  // Don't run the code during drags; we might have invalid state.
+  if (e.isUiEvent || e.type == Blockly.Events.FINISHED_LOADING ||
+      mainWorkspace.isDragging()) {
+    return;
+  }
```
BeksOmega: Generally more predictable + future proof to use allow lists rather than deny lists for change listeners. E.g. https://github.com/google/blockly/pull/6337

examples/developer-tools/src/index.ts
```diff
@@ -33,9 +38,60 @@ const startBlocks = {
...
+    delete Blockly.Blocks[blockName];
+  }
+  const blockDefinitionString = jsonDefinitionGenerator.workspaceToCode(mainWorkspace);
+  definitionDiv.textContent = blockDefinitionString;
+  const blockDefinition = JSON.parse(blockDefinitionString);
+  blockName = blockDefinition.type;
+  Blockly.common.defineBlocks(Blockly.common.createBlockDefinitionsFromJsonArray([blockDefinition]));
+  const block = previewWorkspace.newBlock(blockName);
+  block.initSvg();
+  block.render();
```
BeksOmega, line 70: Can you add a TODO to remove this after v11?

BeksOmega (commented):
Left comments on the first five commits. Will continue in the morning.

examples/developer-tools/src/blocks/factory_base.ts
maribethb: I started out by copying the block definitions out of the existing project and then updating them. This was removed in a later commit because we don't need a mutation to save the dropdown value at all.

examples/developer-tools/src/index.ts
maribethb: I want pretty much all non-UI events though. I feel like this is a standard pattern we use elsewhere, and a list of every non-UI event would be harder to read and scan for accuracy (easy to forget one).

examples/developer-tools/src/blocks/factory_base.ts
maribethb: Using a theme sounds good, but I'd prefer to do that in a separate PR. I'll also update the toolbox to not just reuse `logic_category` etc. 

I do care about best practices, with the caveat that for some stuff this is just an initial PR to prove the feasibility of using the code generator approach instead of the manual approach used in the old tool, so some of the best practices will be done in later PRs

examples/developer-tools/tsconfig.json
maribethb: I think this is right. It's different than the plugins because we're not publishing this and it's not a library.

BeksOmega:
Heya @maribethb do you think you could split this up into a few different PRs? I don't think this is actually comit-wise reviewable because things in the earlier commits change in later commits.

BeksOmega:
@maribethb can I close this?

PR #1999, opened 2023-10-01
chore: add blockly-react-ts example

alicialics:
## The basics

### Proposed Changes

Add a vite + typescript example in blockly-react-ts

### Reason for Changes

This can be used to test build issues [related to vite](https://github.com/google/blockly-samples/issues/1877#issuecomment-1700051561).

Please also refer to #1679 to see if this example is useful to the team and whether it should replace the existing blockly-react example.

### Test Coverage

N/A.  We don't have a good way to unit test /examples.  In the future we can use a browser test via playwright or a similar solution.

### Documentation

N/A

### Additional Information

cc @maribethb 

According to https://github.com/nbudin/react-blockly the core maintainer has no longer uses blockly on a day-to-day basis, that being said a recent version of this library was just released a week ago so it's definitely not abandoned.

IMO these examples are useful in the short term to test things like different builds.  Longer term it might make sense to do some e2e tests with selenium/playwright.  Also things like codesandbox integration might be interesting:

For example this link opens up the codesandbox:
https://githubbox.com/alicialics/blockly-samples/tree/blockly-react-ts/examples/blockly-react-ts

this opens in stackblitz
https://stackblitz.com/github/alicialics/blockly-samples/tree/blockly-react-ts/examples/blockly-react-ts


maribethb:
Hey Zoey,

We've discussed adding this and unfortunately I don't think we want to right now. It's really tough for our team to maintain examples when none of us is familiar with the tools or framework that the example uses. This has been an ongoing problem with the current react example we have. If we have an example then folks expect us to be able to answer questions about it and keep it up to date and that's not something we can offer right now for this complex of an example. Plus, we do want to continue to point people over to Nat's react-blockly project, as it is still being maintained and we're still rooting for them to find another maintainer.

It would be useful as a test case for building applications using blockly + plugins under different build systems. But right now we don't have any kind of test harness set up for that and I don't want to maintain this as an example until we do. If we did have a test setup for this, we could maintain it as a test case in a simplified way as it would be more for testing than to be an example, so it wouldn't need e.g. full custom react components, an eslint config, etc. It could just build something basic that imports a plugin and is built with vite. So it's possible we could revisit this PR in the future if we do get that kind of test setup done, but for now this isn't something we have the capacity to maintain. Thank you for working on this, though. In the future if you have ideas for new examples, if you open an issue requesting them and explain your idea, we have a weekly triage meeting where we can decide if the example is something we'd be willing to maintain, that way we don't have to shoot you down after you've put in the work for it.

alicialics:
@maribethb no worries at all!  it's definitely a learning experience for me regardless of the PR status. I was using this example to test a bug where the user could not import one of the plugins in Vite.  I only found the other issue later where your team is trimming down the number of examples =)

PR #2016, opened 2023-10-15
fix: fix continuous toolbox not enabling blocks

### LINKED ISSUE #2002: ContinuousToolbox is not re-enabled after deleting blocks from workspace with maxBlocks option
### Check for duplicates

### Component

ContinuousToolbox

### Description

I'm creating a workspace with the `ContinuousToolbox`, and limiting the maximum number of blocks (`maxBlocks`):

```js
    var workspace = Blockly.inject('blocklyDiv', {
        toolbox: getToolboxJSON(window.workspaceConfig?.toolbox?.blocks),
        trashcan: true,
        renderer: 'zelos',
        maxBlocks: 10,
        plugins: {
            'toolbox': ContinuousToolbox,
            'flyoutsVerticalToolbox': ContinuousFlyout,
            'metricsManager': ContinuousMetrics,
        },  
    });
```

As expected, the toolbox is disabled when the number of blocks in the workspace reaches the maximum, but it is not re-enabled when blocks are deleted.

When I use the original toolbox, by removing the `plugins:` key, the toolbox works as expected.

### Reproduction steps

1. Add blocks to the workspace until `maxBlocks` is reached. The blocks in the toolbox are disabled, as expected.
2. Remove one block from the workspace. The blocks remain disabled; it should be expected that they are re-enabled.

### Stack trace

_No response_

### Screenshots

_No response_

alicialics:
## The basics

### Resolves

Fixes #2002

### Proposed Changes

do not recycle disabled blocks

### Reason for Changes

Blocks are being disabled if maxBlock constraint is met.  They are also being currently recycled and will be pushed to permanentlyDisabled list.  

Before this change:
If `maxBlocks` is set, when you add `maxBlocks` blocks the blocks will be disabled in the flyout and does not get renabled when the blocks are recycled.

If `maxInstances` is set on a block, after you add `maxInstances` of that block the block will be disabled.  Removing one will not renable it.

After:
As expected, when we recycle blocks for `maxBlocks` then we will enable blocks in the flyout.
For `maxInstances`, when we recycle the constrainted block you'll be able to add more blocks of that type.

### Test Coverage

tested in max-blocks-demo

### Additional Information

cc @gonfunko @BeksOmega

### DIFF
--- a/plugins/continuous-toolbox/src/ContinuousFlyout.js
+++ b/plugins/continuous-toolbox/src/ContinuousFlyout.js
@@ -229,6 +229,10 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
       return false;
     }
 
+    if (!block.isEnabled()) {
+      return false;
+    }
+
     for (const input of block.inputList) {
       for (const field of input.fieldRow) {
         // No variables.


BeksOmega:
Heya @alicialics thank you for taking a look at this! Were you able to reproduce the original issue? I was not so I can't tell if this fixes it :/

I am also still seeing the issue with `maxInstances`  :/

alicialics:
> Heya @alicialics thank you for taking a look at this! Were you able to reproduce the original issue? I was not so I can't tell if this fixes it :/
> 
> I am also still seeing the issue with `maxInstances` :/

yes I was able to reproduce and validate this fix with the max-blocks-demo modified with ContinuousToolbox.  let me check maxInstances again.

Edit: I've updated the PR description to reflect the before/after state.  I could not actually reproduce the maxInstances not getting disabled issue @BeksOmega[ you mentioned in your comment](https://github.com/google/blockly-samples/issues/2002#issuecomment-1751247078), but my change does fix the "re-enabling" a block with `maxInstances` after one of these blocks is recycled.

Here is my local changeset on max-blocks-demo:
```
diff --git a/examples/max-blocks-demo/index.html b/examples/max-blocks-demo/index.html
index c1e7f1c5..464abded 100644
--- a/examples/max-blocks-demo/index.html
+++ b/examples/max-blocks-demo/index.html
@@ -6,6 +6,7 @@
   <script src="./node_modules/blockly/blockly_compressed.js"></script>
   <script src="./node_modules/blockly/blocks_compressed.js"></script>
   <script src="./node_modules/blockly/msg/en.js"></script>
+  <script src="./node_modules/@blockly/continuous-toolbox/dist/index.js"></script>
   <style>
     body {
       background-color: #fff;
@@ -79,9 +80,19 @@
   </xml>

   <script>
+    var plugins = {
+      'toolbox': ContinuousToolbox,
+      'flyoutsVerticalToolbox': ContinuousFlyout,
+      'metricsManager': ContinuousMetrics,
+    };
+    // plugins = undefined;
     var demoWorkspace = Blockly.inject('blocklyDiv',
-        {media: './node_modules/blockly/media/',
+        {
+  plugins,
+
+media: './node_modules/blockly/media/',
          maxBlocks: 5,
+         maxInstances: {"math_number": 3},
          toolbox: document.getElementById('toolbox')});

     function onchange(event) {
(END)
```

BeksOmega (commented):
Ok tested it again, and I was able to reproduce the issue. Not sure why Aaron and I couldn't reproduce it earlier. Observed that your fix fixes the issue.

Also found that the `maxInstances` issue was because `maxInstances` isn't included in the playground config by default (I assumed it was because it is in core). So looks good there as well!

BeksOmega (approved):

PR #2039, opened 2023-10-26
feat: add reskinned devsite demo to samples

26damb:
## The basics

### Resolves

Chore

### Proposed Changes

Add reskinned devsite demo to samples

### Reason for Changes

To be used on new blockly developers page. Includes desktop and mobile examples.

Screenshot 2023-10-26 at 9 37 40 AM
Screenshot 2023-10-26 at 9 37 14 AM


examples/devsite-demo2/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "blockly-devsite-demo",
```
maribethb: This needs a unique name.

examples/devsite-demo2/script.js
```diff
@@ -0,0 +1,349 @@
...
+    rtl: LANGUAGE_RTL.includes(language),
+    zoom: {
+      controls: true,
+      maxScale: 1.8,
+      minScale: 0.6,
+      scaleSpeed: 1.2,
+      pinch: true,
+    },
+    trashcan: true,
+    theme: Blockly.Theme.defineTheme('modest', {
```
maribethb, line 110: The font name and size need to be part of the theme, not just in CSS. That's why the text sizing doesn't appear correct in the demo. https://developers.google.com/blockly/guides/configure/web/appearance/themes#font_styles

examples/devsite-demo2/script.js
```diff
@@ -0,0 +1,349 @@
...
+      blockStyles: {
+        logic_blocks: {
+          colourPrimary: '#AECBFA',
+        },
+        loop_blocks: {
+          colourPrimary: '#A8DAB5',
+        },
+        math_blocks: {
+          colourPrimary: '#4285f4',
```
maribethb: Can we avoid repeating colors where one is just a lighter version of another? Blockly already uses light vs dark shades to have meaning, so it's confusing. It also implies to me some kind of relationship between the math and logic blocks that doesn't make sense. There are more options in the [extended Material color palette](https://m2.material.io/design/color/the-color-system.html#tools-for-picking-colors). Or you could also look at the [zelos theme](https://blockly-demo.appspot.com/static/tests/playgrounds/advanced_playground.html#renderer=thrasos&debugEnabled=false&theme=zelos) for a brighter set of default colors, though they aren't Google-themed.

examples/devsite-demo2/script.js
```diff
@@ -0,0 +1,349 @@
...
+  let toolboxString = JSON.stringify(toolboxJson);
+  toolboxString = toolboxString.replace(
+      /%\{BKY_VARIABLES_DEFAULT_NAME\}/g,
+      Blockly.Msg.VARIABLES_DEFAULT_NAME
+  );
+  toolboxJson = JSON.parse(toolboxString);
+
+  // Inject Blockly.
+  const workspace = Blockly.inject('blocklyDiv', {
+    media: './node_modules/blockly/media/',
```
maribethb, line 99: In the options, you should use the thrasos renderer. https://developers.google.com/blockly/guides/create-custom-blocks/renderers/overview

You will need to adjust your css a little bit. This renderer only has one `blocklyPath` instead of a light and dark version. Setting that to white will make the edges of the blocks look less janky than under the default renderer.

Renderers are also how you can adjust things like the padding on blocks. If this is something we still want to add, and you have time, you could look into creating a custom renderer (a subclass of the thrasos renderer) that increases the block spacing constants. However, I'm not sure we will agree that the extra-wide block look works for us as it makes the code harder to read if every block has too much extra padding.

examples/devsite-demo2-mobile/custom_colors.js
```diff
@@ -0,0 +1,3 @@
+
+
+Blockly.WorkspaceSvg.zoomToFit();
```
maribethb: Why is this here?

examples/devsite-demo2-mobile/style.css
maribethb, line 1: We shouldn't need to maintain two entirely separate demos for mobile vs desktop if the only thing that's different is some css. We can use flex layout and media queries to change the layout on small screens. 

Having two separate demos increases the maintenance overhead for our team (we have to keep them in sync). It will also make the iframe code on devsite trickier because you'll have to load a different source depending on the platform, which is more involved than just using responsive css.

maribethb (commented):
nit: this is a `feat`, not a `chore`

Thanks for your work on this. I think it needs a bit more work before it's ready, but not that much. We are happy to meet or correspond over email/chat if you have follow-up questions.

examples/devsite-demo2/script.js
```diff
@@ -0,0 +1,349 @@
...
+    toolbox: toolboxJson,
+    rtl: LANGUAGE_RTL.includes(language),
+    zoom: {
+      controls: true,
+      maxScale: 1.8,
+      minScale: 0.6,
+      scaleSpeed: 1.2,
+      pinch: true,
+    },
+    trashcan: true,
```
maribethb: Set this to false if we decide to remove the trashcan.

26damb:
Screenshot 2023-10-27 at 4 01 09 PM
Screenshot 2023-10-27 at 4 01 30 PM

How they look w/ controls & w/o

maribethb (approved):

PR #2041, opened 2023-10-27
chore: update lerna to fix prettier

maribethb:
## The basics

### Resolves

Fixes publish process [breaking](https://github.com/google/blockly-samples/actions/runs/6657456809/job/18092209980) due to lerna issues

### Proposed Changes

- Updates lerna to v7
- Installs the legacy package management tools described [here](https://lerna.js.org/docs/legacy-package-management)

### Reason for Changes

- I recently installed Prettier v3 in this repository.
- Lerna has a [totally undocumented feature](https://github.com/lerna/lerna/issues/3806) where, if you have Prettier installed, it will run Prettier during `lerna version`.
- In Lerna v6, the syntax used to invoke Prettier is only compatible with prettier v2. We're using v3. The syntax is not compatible.
- Lerna v7 was updated to be compatible with prettier v3.
- Lerna v7 also removes the legacy package management features we use (i.e., lerna bootstrap) by default. However, you can reinstall them by installing the extra package which I did here.

### Test Coverage

I don't have a good way of testing if version will work because a dry run doesn't trigger the original problem. However, I did check that bootstrap still works after updating.

### Documentation

no

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@babel/eslint-parser": "^7.22.15",
         "@eslint/eslintrc": "^2.1.2",
         "@eslint/js": "^8.49.0",
+        "@lerna/legacy-package-management": "^7.4.1",
         "@typescript-eslint/eslint-plugin": "^6.7.2",
         "@typescript-eslint/parser": "^6.7.2",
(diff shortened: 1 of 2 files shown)


BeksOmega (approved):
Seems reasonable!

PR #2043, opened 2023-10-27
fix: proc call remain disabled as def changes name

### LINKED ISSUE #2035: Disabled procedure blocks are re-enabled after a change
### Check for duplicates

### Component

plugins/block-shareable-procedures

### Description

If a user disables procedure call blocks, they'll be re-enabled automatically when the matching definition block is modified (e.g. renamed or parameters added). This is undesirable.

### Reproduction steps

1. Open the block-shareable-procedures [test page](https://google.github.io/blockly-samples/plugins/block-shareable-procedures/test/index.html)
2. Drag a new procedure definition onto one of the workspaces
3. Drag a call block for this procedure onto either workspace (doesn't matter which one)
4. Disable the call block
5. Rename the procedure in the definition block
6. Observe how the block was re-enabled

### Stack trace

_No response_

### Screenshots

https://github.com/google/blockly-samples/assets/8573958/eea165e5-3e9b-41f1-8891-0f2a473e670e

alicialics:
## The basics

### Resolves

Fixes #2035 

### Proposed Changes

When a proc call changes enabled state independently, previousEnabledState_ should be kept in sync

### Reason for Changes

#2035

### Test Coverage

The unit test requires a setGroup call in order to simulate what happens in UI (ie when a proc call gets disabled)

### Additional Information

@BeksOmega I am not familiar with groups.  I checked that when a proc call gets disabled in the UI it has a group name but when it is disabled as part of proc def block update it does not have a group name.  So I might be totally wrong here =)

### DIFF
--- a/plugins/block-shareable-procedures/src/blocks.ts
+++ b/plugins/block-shareable-procedures/src/blocks.ts
@@ -1203,6 +1203,14 @@ const procedureCallerOnChangeMixin = {
     if (this.disposed || this.workspace.isFlyout) return;
     if (event.type === Blockly.Events.BLOCK_MOVE) this.updateArgsMap_(true);
     if (
+      event.type === Blockly.Events.BLOCK_CHANGE &&
+      event.blockId === this.id &&
+      event.element === 'disabled' &&
+      event.group
+    ) {
+      this.previousEnabledState_ = !event.newValue;
+    }
+    if (
       event.type !== Blockly.Events.FINISHED_LOADING &&
       !this.eventIsCreatingThisBlockDuringPaste_(event)
     )
--- a/plugins/block-shareable-procedures/test/procedure_blocks.mocha.js
+++ b/plugins/block-shareable-procedures/test/procedure_blocks.mocha.js
@@ -1249,6 +1249,26 @@ suite('Procedures', function () {
         );
       },
     );
+
+    test(
+      'if a procedure caller block was already disabled ' +
+        'its definition changes name, it is not reenabled',
+      function () {
+        const defBlock = createProcDefBlock(this.workspace);
+        const callBlock = createProcCallBlock(this.workspace);
+        globalThis.clock.runAll();
+        Blockly.Events.setGroup(true);
+        callBlock.setEnabled(false);
+        globalThis.clock.runAll();
+        defBlock.getProcedureModel().setName("other proc name");
+        globalThis.clock.runAll();
+
+        assert.isFalse(
+          callBlock.isEnabled(),
+          'Expected the caller block to continue to be disabled',
+        );
+      },
+    );
   });
 
   suite('creating procedure blocks', function () {


plugins/block-shareable-procedures/src/blocks.ts
```diff
@@ -1202,6 +1202,14 @@ const procedureCallerOnChangeMixin = {
   onchange: function (event) {
     if (this.disposed || this.workspace.isFlyout) return;
     if (event.type === Blockly.Events.BLOCK_MOVE) this.updateArgsMap_(true);
+    if (
+      event.type === Blockly.Events.BLOCK_CHANGE &&
+      event.blockId === this.id &&
+      event.element === 'disabled' &&
+      event.group
```
BeksOmega, line 1209: Groups are for handling undo and redo, so you shouldn't need to check for this explicitly! We don't care if it's being bundled with other events for undo/redo :P

You should also be able to remove it from your unit test.

plugins/block-shareable-procedures/src/blocks.ts
```diff
@@ -1202,6 +1202,14 @@ const procedureCallerOnChangeMixin = {
...
+    ) {
+      this.previousEnabledState_ = !event.newValue;
```
BeksOmega, line 1211: Nit: This is equivalent but I think it's a little clearer.
```suggestion
      this.previousEnabledState_ = event.oldValue;
```

BeksOmega (changes requested):
This seems to also be broken on load. (e.g. deserializing a disabled caller with an enabled def enables the caller) Could you figure out what's up with that and then I'll re-review?

If you're busy or that's too much load also totally cool! Just lmk!

plugins/block-shareable-procedures/src/blocks.ts
alicialics: Gotcha! makes sense.

Unfortunately if I remove event.group the patch won't work.  This is because if you disable a `proc def` it will trigger a `proc call` enable/disable call and I have no idea how to distinguish it with disabling it directly.  Adding `event.group` in the UI seem to work but it doesn't feel like the right fix so I will close this PR for now. =)

BeksOmega:
Hmm did you try checking for `recordUndo` instead of `group`? If undo is enabled on the direct call to the procedure caller, it shouldn't be, and we should fix that. Then you should be able to differentiate a user-made change vs a programmatic change.

alicialics:
@BeksOmega I double checked the events being outputted in these instances and the only difference I see is event.group is empty vs not. =(
> Then you should be able to differentiate a user-made change vs a programmatic change.

Totally agree, this is a little bit out of my league however so I'll leave it to the blockly team

plugins/block-shareable-procedures/src/blocks.ts
BeksOmega: Wanted to come back here and say that as I was rewriting this, you're right that `!event.newValue` is the correct way to do this :P

PR #2044, opened 2023-10-28
fix!: remove the lint script from blockly-scripts

### LINKED ISSUE #1976: Remove the `lint` script from `blockly-scripts`
### Check for duplicates

### Component

dev-scripts

### Problem

The lint script in dev-scripts runs eslint through the JS API. After https://github.com/google/blockly-samples/pull/1946 is merged, we'll no longer need this script because eslint can be invoked directly on the command line from any level of blockly-samples. This is an improvement because the eslint cli has better formatted output (especially when run at the entire repository level) and the flat config is easier to use and make changes to.

### Request

Remove the `lint` script from dev-scripts. This is a breaking change for the plugin. The lint script already logs a deprecation warning when run. I hope that nobody is using this script outside of blockly-samples.

### Alternatives considered

_No response_

### Additional context

If you are currently using `blockly-scripts lint`, switch to invoking eslint through its cli or write your own script that uses the eslint js api. Also, it would help us if you commented on this issue with your use case.

shivalipatel6:
Remove the lint script from blockly-scripts

## The basics

## The details
The lint script needed to be removed because once https://github.com/google/blockly-samples/pull/1946 is merged, the eslint used in the lint script can be run directly on the command line. 
### Resolves

Fixes #1976
A breaking change for the plug in. Allowing for a newer fix to replace a depreciating script. 
### Proposed Changes

Deletes the lint.js script from dev-scripts, as well as removing it from the "available scripts" in dev-scripts/bin/blockly-scripts.js.
### Reason for Changes
To remove a script that is depreciated and is in conflict with changes made in a new merged issue.


alicialics:
Thinking out loud and not actually asking for a change:

@maribethb 
I wonder if we should output a deprecation message saying`lint` command is removed.

BeksOmega:
> Thinking out loud and not actually asking for a change:
> 
> @maribethb I wonder if we should output a deprecation message saying`lint` command is removed.

There is [currently](https://github.com/google/blockly-samples/blob/c9d184fd045df234bf582d3ee52b77391c751a84/plugins/dev-scripts/scripts/lint.js#L48) a deprecation message. Do you mean leave that around isntead of deleting the script entirely?

alicialics:
oh ok, didn't realize that!

shivalipatel6:
So I have tried to change my commit message to fit in with conventional commits, but I can't get my changed commit message which is shown in my fork, to be displayed in my pull request.

BeksOmega:
> So I have tried to change my commit message to fit in with conventional commits, but I can't get my changed commit message which is shown in my fork, to be displayed in my pull request.

You probably need to do a `git push --force`! we have docs on that here: https://developers.google.com/blockly/guides/contribute/get-started/commits#fixing_non-conventional_commits

maribethb:
Hi @shivalipatel6 , since this is a breaking change, the commit message actually needs to be different. `chore` is only for things that plugin users wouldn't care about, but since this affects the functionality of the plugin it can't be a `chore`. And we use the `!` to show that it is a breaking change. This is important as our release process will use the commit messages to automatically change the major version number if there has been a commit message with a `!` in it.

So overall the commit message should be `fix!: remove the lint script from blockly-scripts`

You can follow the guide Beka linked above to fix your commit message and force-push. After that, this PR looks good and ready to merge. Thanks!

BeksOmega (approved):

Issue #1909, opened 2023-09-05
Publish types for workspace-minimap


BeksOmega:
**Category**

 - Plugins
 
**Component**

workspace-minimap

**Is your feature request related to a problem? Please describe.**

Currently types for this plugin are not being generated or exported :/

**Describe the solution you'd like**

1) Add the following config options to the "compilerOptions" section of the tsconfig.json
```
"declaration": true,
"declarationMap": true,
```
2) Add the following property to the package.json next to the "main" property:
```
"types": "./dist/index.d.ts"
```

**Additional context**

Related to: https://github.com/google/blockly-samples/issues/1892

scoopstroop92:
I would like to work on this
@BeksOmega

apatte:
Interested in working on this @BeksOmega

YashviDesaii:
I would like to work on this @BeksOmega

RadhikaRJ:
Hello, I'd like to contribute to this issue.

spiterwoman:
Hi! I am interested in working on this issue.

BeksOmega:
Hiya @ramyaram2092 Go for it =)

vprerana:
I would like to work on this

Annysah:
Hi I am interested in working on this issue

PrayagCodes:
@BeksOmega please assign me thses issue

vprerana:
Hey, I would like to work on this issue.

kapooraryan:
@BeksOmega Could I work on this?

BeksOmega:
Go for it @kapooraryan !

PR #2054, opened 2023-10-31
feat: enables strict mode in workspace-search

### LINKED ISSUE #2032: Enable strict mode for workspace-search
### Check for duplicates

### Component

workspace-search

### Problem

_No response_

### Request

The workspace-search plugin does not have "[strict mode](https://www.typescriptlang.org/tsconfig#strict)" enabled for TypeScript. We should enable it to have better type checking for our plugins, but doing so may show some TypeScript errors that need to be fixed first.

To work on this issue:

### Alternatives considered

_No response_

### Additional context

_No response_

ghoshvidip26:
## The basics

### Resolves

Fixes #2032 

### Proposed Changes

I've enabled the strict mode for workspace_search.ts and there were some type after applying strict mode. So I've solved them by adding `!` where I got `Object is possibly null`


ghoshvidip26:
@BeksOmega please review the changes.

plugins/workspace-search/src/workspace_search.ts
```diff
@@ -132,9 +132,10 @@ export class WorkspaceSearch implements Blockly.IPositionable {
      * </div>
      */
     const injectionDiv = this.workspace.getInjectionDiv();
-    this.addEvent(injectionDiv, 'keydown', this, (evt: KeyboardEvent) =>
-      this.onWorkspaceKeyDown(evt),
-    );
+    this.addEvent(injectionDiv, 'keydown', this, (evt: Event) => {
+      const keyboardEvent = evt as KeyboardEvent;
```
BeksOmega, line 137: Instead of casting all of these individually, I think you can make `addEvent` generic and have this take the correct event from the get-go.

```ts
private addEvent<T extends Event>(
    node: Element,
    name: string,
    thisObject: object,
    func: (event: T) => void,
  ) { }
```

plugins/workspace-search/src/workspace_search.ts
```diff
@@ -149,13 +150,16 @@ export class WorkspaceSearch implements Blockly.IPositionable {
...
+    this.addEvent(this.inputElement, 'keydown', this, (evt: Event) => {
+      const keyboardEvent = evt as KeyboardEvent; // Type assertion
+      this.onKeyDown(keyboardEvent);
+    });    
+
     this.addEvent(this.inputElement, 'input', this, () => this.onInput());
     this.addEvent(this.inputElement, 'click', this, () => {
       this.searchAndHighlight(this.searchText, this.preserveSelected);
-      this.inputElement.select();
+      this.inputElement!.select();
```
BeksOmega: While `!` assertions do "work" they're not always the correct way to solve a nullability problem. They basically "turn off" the type checker, so that instead of getting errors at compile time, we get errors at runtime.

Instead, we'd like to check that the thing that's possibly null actually exists, and either throw an informative error if it doesn't or just skip the behavior.

I think throughout this file, we want to change code to skip the behavior if the thing is null. So instead of `!.` you can use `?.` or use `if` statements to check if the thing exists =)

plugins/workspace-search/src/workspace_search.ts
```diff
@@ -502,7 +507,7 @@ export class WorkspaceSearch implements Blockly.IPositionable {
       // Search the whole string for collapsed blocks.
       blockText = block.toString();
     } else {
-      const topBlockText = [];
+      const topBlockText:any = [];
```
BeksOmega: I think this can be a string array.

BeksOmega (approved):
Hiya thanks for your work on this! I think we want to take a slightly different approach to the nullability, but this looks like a good start =)

plugins/workspace-search/src/workspace_search.ts
ghoshvidip26: Screenshot 2023-11-01 at 14 11 29
Actually, I got the warning. That's why I added **any** here. 
"Variable 'topBlockText' implicitly has type 'any[]' in some locations where its type cannot be determined."

ghoshvidip26:
Please review the changes.

plugins/workspace-search/src/workspace_search.ts
BeksOmega: Hello! For this one, instead of adding a new local function, can you modify the method on the WorkspaceSearch class to be generic?

plugins/workspace-search/src/workspace_search.ts
BeksOmega: This still needs changes.

plugins/workspace-search/src/workspace_search.ts
BeksOmega: I think if you specificy it as `const topBLockText: string[] = [];` it will work! Can you give that a shot for me?

plugins/workspace-search/src/workspace_search.ts
ghoshvidip26: Screenshot 2023-11-01 at 21 41 08
Yes I've changed this one.

plugins/workspace-search/src/workspace_search.ts
ghoshvidip26: Yes it worked.

plugins/workspace-search/src/workspace_search.ts
ghoshvidip26: Screenshot 2023-11-01 at 21 51 56
So inside this class, shall I create a **`addEvent`** generic ?

ghoshvidip26:
@BeksOmega In generics one, do I need to create `private addEvent` inside the WorkSpaceSearch class?

plugins/workspace-search/src/workspace_search.ts
BeksOmega: It already exists you just need to modify it: https://github.com/google/blockly-samples/blob/39037d1e2113174c9d60c1f9389eddbe7df589e3/plugins/workspace-search/src/workspace_search.ts#L196

ghoshvidip26:
```
private addEvent <T extends Event> (
    node: Element,
    name: string,
    thisObject: object,
    func: (event: T) => void,
  ) {
    const event = Blockly.browserEvents.conditionalBind(
      node,
      name,
      thisObject,
      func,
    );
    this.boundEvents.push(event);
  }
```
  I've modified. Is this alright?

ghoshvidip26:
Please review the change. 
```
private addEvent <T extends Event>(
    node: Element,
    name: string,
    thisObject: object,
    func: (event: T) => void,
  ) {
    const event = Blockly.browserEvents.conditionalBind(
      node,
      name,
      thisObject,
      func,
    );
    this.boundEvents.push(event);
  }
```
  I hope this one is alright

plugins/workspace-search/src/workspace_search.ts
BeksOmega: Looks correct! Can you change the callers to use the type they expect? And then remove the casts you added?

plugins/workspace-search/src/workspace_search.ts
BeksOmega: Thanks! Can you change it throughout the file now?

plugins/workspace-search/src/workspace_search.ts
ghoshvidip26: Do I need to assign types to node, name, thisObject?

plugins/workspace-search/src/workspace_search.ts
ghoshvidip26: Okay alright.

plugins/workspace-search/src/workspace_search.ts
ghoshvidip26: > Looks correct! Can you change the callers to use the type they expect? And then remove the casts you added?

Can you please explain this one a bit?

plugins/workspace-search/src/workspace_search.ts
ghoshvidip26: Screenshot 2023-11-04 at 22 57 26
?. is throwing an error in this case @BeksOmega 
So should I keep it unchanged? That is to use `!`

ghoshvidip26:
@BeksOmega  can you please provide me an idea on `How to change the callers to use the type they expect`? I didn't get this one.

plugins/workspace-search/src/workspace_search.ts
BeksOmega: Yeah! When you're calling `this.addEvent` you should pass a callback that says the type it expects. E.g. `this.addEvent(injectionDiv, 'keydown', this, (evt: KeyboardEvent) => {})`

plugins/workspace-search/src/workspace_search.ts
BeksOmega: Ah in this case you can't use a `?.` operator because you can't assign to those. Instead you should check if the `htmlDiv` exists and return early if not. E.g. `if (!this.htmlDiv) return;`

plugins/workspace-search/src/workspace_search.ts
ghoshvidip26: > Yeah! When you're calling `this.addEvent` you should pass a callback that says the type it expects. E.g. `this.addEvent(injectionDiv, 'keydown', this, (evt: KeyboardEvent) => {})`

So shall I replace all the callers with injectionDiv whereever this.Element is used?? 
Screenshot 2023-11-09 at 22 56 05

BeksOmega: Nope, we're trying to fix the type of the event in the callback. I changed it to use `KeyboardEvent` instead of `Event`!

ghoshvidip26: > KeyboardEvent

Please explain where to make changes now. So that I could complete this at the earliest.

plugins/workspace-search/src/workspace_search.ts
ghoshvidip26: Yeah I have resolved these now.

plugins/workspace-search/src/workspace_search.ts
BeksOmega:
```
this.addEvent(this.inputElement, 'keydown', this, (evt: Event) => {
  const keyboardEvent = evt as KeyboardEvent;
  this.onKeyDown(keyboardEvent);
});
```

Becomes
```
this.addEvent(this.inputElement, 'keydown', this, (evt: KeyboardEvent) => {
  this.onKeyDown(evt);
});
```

We are now passing a callback that says the type it expects.

Please update for all calls to `this.addEvent`.

ghoshvidip26: I have resolved this one. 
For other calls like 
```
this.addEvent(this.inputElement, 'click', this, () => {
      this.searchAndHighlight(this.searchText, this.preserveSelected);
      this.inputElement?.select();
    });
    
    &&
    
    this.addEvent(this.inputElement, 'input', this, () => this.onInput());
```
that may not work.

ghoshvidip26:
Screenshot 2023-11-09 at 23 01 40
In few cases like this adding `?` gives warning. 
@BeksOmega please review now if everything is alright.

BeksOmega:
Heya @Vidip-Ghosh After going through the back and forth I think this is a bit of a large change for your first contribution. I think it's better to work on a smaller project like https://github.com/google/blockly-samples/issues/1909

I'm going to go ahead and close this for now. If you want to work on the other issue comment there and let me know!

Issue #2058, opened 2023-11-03
plugin-cross-tab-copy-paste 


bartbutenaers:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Component

plugin-cross-tab-copy-paste

### Description

Hi everybody,

I have used the [plugin-cross-tab-copy-paste](https://www.npmjs.com/package/@blockly/plugin-cross-tab-copy-paste) plugin for the first time, and it works very well.  Very useful plugin!

However in our application we create a ***new*** workspace when we show the old workspace in full screen.  When we do that now, the following exceptions is raised.  All my other plugins have no issue, but they all have a `dispose` function that does the cleanup of all stuff.  However the plugin-cross-tab-copy-paste plugin does 't offer such a function...

Thanks!
Bart

### Reproduction steps

1. Call ´Blockly.inject´
2. Initialize the plugin:
   ```
   node.copyPastePlugin = new CrossTabCopyPaste();
   node.copyPastePlugin.init(copyPastePluginOptions, function() {
      //console.log('Use this error callback to handle TypeError while pasting');
   });
   ```
3. For all other plugins I call now their `dispose` function, but this plugin doesn' t have it...
4. Call ´Blockly.inject´ again.  Now an exception occurs

### Stack trace

```shell
Uncaught Error: Menu item with ID "blockCopyToStorage" is already registered.
    at ContextMenuRegistry$$module$build$src$core$contextmenu_registry.register (contextmenu_registry.ts:45:13)
    at t.blockCopyToStorageContextMenu (index.js:90:5)
    at t.init (index.js:33:12)
    at createWorkspace (<anonymous>:455:34)
    at Object.open (<anonymous>:918:29)
    at n (red.min.js?v=:19:44660)
    at red.min.js?v=:19:45794
```

alicialics:
does it work if you call `Blockly.ContextMenuRegistry.registry.unregister('blockCopyToStorage');` in place of `plugin.dispose()`

bartbutenaers:
Hi @alicialics,

Thanks for joining (again)!

1. In some circumstances I arrive at the cleanup part before the plugins have been initialized, and then your statement results into this:
   ```
   Uncaught Error: Menu item with ID "blockCopyToStorage" not found.
       at ContextMenuRegistry$$module$build$src$core$contextmenu_registry.unregister (contextmenu_registry.ts:58:17)
       at cleanup_plugins (<anonymous>:549:50)
[...]

PR #2061, opened 2023-11-06
feat: dispose function for copy-paste plugin

bartbutenaers:
## The basics

### Resolves

Fixes issue #2058 

### Proposed Changes

Based on the tips from @alicialics in the above issue, I have added a `dispose` function to the plugin that executes the following actions:
1. It removes the 2 contextmenu items that the plugin has been added in the `init` function (if available).
2. It removes the 3 shortcut key actions, and replaces them by the 3 default shortcut key actions (cut/copy/paste).  

### Reason for Changes

In our application (i.e. a node for Node-RED) we re-create the workspace multiple times, which means we also need to dispose all the plugins that we have used.  

### Test Coverage

The plugin did not have any tests yet...

### Documentation

I have added the `dispose` function to the example code snippet on the plugin's readme page.

### DIFF
--- a/plugins/cross-tab-copy-paste/README.md
+++ b/plugins/cross-tab-copy-paste/README.md
@@ -44,7 +44,7 @@ Blockly.ContextMenuRegistry.registry.unregister('blockDuplicate');
 // optional: You can change the position of the menu added to the context menu.
 Blockly.ContextMenuRegistry.registry.getItem('blockCopyToStorage').weight = 2;
 Blockly.ContextMenuRegistry.registry.getItem(
-  'blockPasteFromStorage',
+  'blockPasteFromStorage'
 ).weight = 3;
 
 // optional: You can cleanup all resources used by this plugin
--- a/plugins/cross-tab-copy-paste/README.md
+++ b/plugins/cross-tab-copy-paste/README.md
@@ -46,6 +46,9 @@ Blockly.ContextMenuRegistry.registry.getItem('blockCopyToStorage').weight = 2;
 Blockly.ContextMenuRegistry.registry.getItem(
   'blockPasteFromStorage',
 ).weight = 3;
+
+// optional: You can cleanup all resources used by this plugin
+plugin.dispose();
 ```
 
 ## Options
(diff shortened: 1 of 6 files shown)


alicialics:
nice!  you should probably run `npm run format` for this change.

bartbutenaers:
@alicialics,
Thanks for the tip!  It is formatted now.
What a journey...

BeksOmega:
@bartbutenaers Do you need to dispose of the copy paste plugin when injecting your new workspace? The plugin works globally on the page, so I don't think you should need to tear it down.

bartbutenaers:
@BeksOmega,
I have no tested it in the way you describe.  So I cannot confirm at the moment if it works like that.

But personally I like to use all my plugins in the same way: I initialize all of them when I create my workspace, and I dispose all of them after stopping my workspace.  That keeps my code a bit readable.  Hope you understand.

bartbutenaers:
Perhaps I can add a comment to the code snippet on the readme page, to explain that it is not necessary to call the dispose function when the plugin is being initialized globally.  Then people can choose how they want to use this plugin.

BeksOmega:
> But personally I like to use all my plugins in the same way: I initialize all of them when I create my workspace, and I dispose all of them after stopping my workspace. That keeps my code a bit readable. Hope you understand.

I think it would be a bit misleading to add disposal for this plugin. Most plugins initialize themselves with a specific workspace (e.g. the backpack plugin), and therefore need to be disposed when the workspace is disposed. But this one is initialized globally, and therefore does not need to be disposed.

Similarly, this plugin only needs to be initialized once, and doesn't need to be initialized per-workspace.

If people think that the copy-paste plugin is a per-workspace thing like other plugins, it will actually cause them problems. For example, if you have two workspaces on a page, and then you tear down one of them + the copy paste plugin, the copy pasting will appear to break for the other workspace, because the copy-paste logic is actually global.

Does that make sense?

bartbutenaers:
Hi @BeksOmega,

Thanks for the explanation! 
Yes indeed that makes sense.
I have updated my code to load the plugin globally and then it indeed still works.
So I will close this pull request.

BTW It would be more clear to understand if these kind of plugins would have a short explanation on their readme page, to explain that they need to be initialized globally so no dispose is required.  Because now it looked - to me at least - that the developer of the node had forgotten to implement the dispose function...

Bart

PR #2065, opened 2023-11-07
fix: resolved 15 ESLint warnings

### LINKED ISSUE #1978: Fix lint warnings in blockly-samples
### Check for duplicates

### Component

multiple

### Description

After https://github.com/google/blockly-samples/pull/1946 is submitted:

There are several lint warnings in blockly-samples that we should fix or ignore the linter. Run the linter and fix the warnings shown.

Note that several of the warnings are about missing JsDoc comments, so you may need to write small amounts of documentation to fix this issue.

`unexpected any` warnings may need refactorings to avoid the `any` type, or may need to just add an ignore statement to silence the warning, if the use of `any` is appropriate.

### Reproduction steps

1. Run `npm run lint` from the root of blockly-samples
2. Note the multiple lint warnings

### Stack trace

_No response_

### Screenshots

_No response_

sdthaker:
## The basics

### Resolves

Fixes #1978  

### Proposed Changes

Fixes these warnings (_some of them were repeated warnings_):
* Unexpected any. Specify a different type - @typescript-eslint/no-explicit-any
* Missing JSDoc @param "type" type - jsdoc/require-param-type
* Missing JSDoc comment - jsdoc/require-jsdoc
* JSDoc @returns declaration present but return expression not available in function - jsdoc/require-returns-check
* Syntax error in type: MigrationData[]                                                                                                      
* Missing JSDoc @returns type  
* Missing JSDoc @param "identifier" description - jsdoc/require-param-description
* Missing JSDoc @param "identifier" type
* 'Minimap' is defined but never used - @typescript-eslint/no-unused-vars
* JSDoc @returns declaration present but return expression not available in function - jsdoc/require-returns-check

### Reason for Changes

These changes are done because aligning the code with the linting rules allows for the static analysis tool to catch any uncaught bugs at compile time.

### Test Coverage

`npm run test` results:
>image

`npm run build` results:
>image

### Additional Information

The only warning left to be fixed is,

>image

### DIFF
--- a/plugins/block-dynamic-connection/src/dynamic_list_create.ts
+++ b/plugins/block-dynamic-connection/src/dynamic_list_create.ts
@@ -131,7 +131,7 @@ const DYNAMIC_LIST_CREATE_MIXIN = {
    */
   loadExtraState: function (
     this: DynamicListCreateBlock,
-    state: { itemCount?: number; [x: string]: unknown} | string,
+    state: {itemCount?: number; [x: string]: unknown} | string,
   ) {
     if (typeof state === 'string') {
       this.domToMutation(Blockly.utils.xml.textToDom(state));
--- a/plugins/block-dynamic-connection/src/dynamic_list_create.ts
+++ b/plugins/block-dynamic-connection/src/dynamic_list_create.ts
@@ -131,14 +131,14 @@ const DYNAMIC_LIST_CREATE_MIXIN = {
    */
   loadExtraState: function (
     this: DynamicListCreateBlock,
-    state: {[x: string]: number} | string,
+    state: { itemCount?: number; [x: string]: unknown} | string,
   ) {
     if (typeof state === 'string') {
       this.domToMutation(Blockly.utils.xml.textToDom(state));
       return;
     }
 
-    this.itemCount = state['itemCount'];
+    this.itemCount = state['itemCount'] ?? 0;
     // minInputs are added automatically.
     for (let i = this.minInputs; i < this.itemCount; i++) {
       this.appendValueInput('ADD' + i);
--- a/plugins/block-dynamic-connection/src/dynamic_list_create.ts
+++ b/plugins/block-dynamic-connection/src/dynamic_list_create.ts
@@ -131,7 +131,7 @@ const DYNAMIC_LIST_CREATE_MIXIN = {
    */
   loadExtraState: function (
     this: DynamicListCreateBlock,
-    state: {[x: string]: any} | string,
+    state: {[x: string]: number} | string,
   ) {
     if (typeof state === 'string') {
       this.domToMutation(Blockly.utils.xml.textToDom(state));
--- a/plugins/migration/bin/fix-imports.js
+++ b/plugins/migration/bin/fix-imports.js
@@ -59,7 +59,7 @@ export const fixImports = createSubCommand(
 let MigrationData;
 
 // TODO (#1211): Make this database format more robust.
-/** @type {MigrationData[]} */
+/** @type {Array.<MigrationData>} */
 const database = [
   {
     import: 'blockly/dart',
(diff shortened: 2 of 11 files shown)


plugins/migration/bin/fix-imports.js
```diff
@@ -59,7 +59,7 @@ export const fixImports = createSubCommand(
 let MigrationData;
 
 // TODO (#1211): Make this database format more robust.
-/** @type {MigrationData[]} */
+/** @type {Array.<MigrationData>} */
```
maribethb: We don't use the `.` after `Array` so this should be `Array<MigrationData>`.

plugins/block-dynamic-connection/src/dynamic_list_create.ts
```diff
@@ -125,7 +125,7 @@ const DYNAMIC_LIST_CREATE_MIXIN = {
    */
   loadExtraState: function (
     this: DynamicListCreateBlock,
-    state: {[x: string]: any} | string,
+    state: {[x: string]: number} | string,
```
maribethb: I don't think this is quite right for the `state` types, here and below. The extraState could have any number of properties on it (we don't know) and it's not correct to assume all of those properties will be numbers (or strings, or whatever). We don't know the types. To be strict, we should type that as `unknown` rather than `any`.

I think the *most* correct would be

```
state: { itemCount?: number, [x: string]: unknown} | string,
```

This says the state can either be a string (when it's xml), or it can be an object with any number of properties whose type is `unknown`, and also it might have a property called `itemCount` that's a number. Then you'll get an error later because `unknown` is a stricter type than `any`. You'll need to change `this.itemCount = state['itemCount'];` to `this.itemCount = state['itemCount'] ?? 0;` to properly account for the case where `itemCount` does not exist in the extraState.

This might seem like a lot of work, but now the code will be truly type-safe instead of the current state which uses `any` to basically turn off type checking.

You would need a similar fix for the others below, but instead of `itemCount` it would be whatever property is actually used in the body of the function.

plugins/workspace-minimap/test/index.ts
```diff
@@ -37,7 +37,14 @@ function createWorkspace(
 }
 
 document.addEventListener('DOMContentLoaded', function () {
-  createPlayground(document.getElementById('root'), createWorkspace as any, {
-    toolbox: toolboxCategories,
-  });
+  createPlayground(
+    document.getElementById('root'),
+    createWorkspace as (
+      blocklyDiv: HTMLElement,
```
maribethb: I don't think you need this cast - `createWorkspace` already matches that type as far as I can tell. I think the original cast `as any` was maybe just unnecessary.

scripts/gh-predeploy.js
```diff
@@ -456,8 +456,7 @@ function prepareExample(exampleDir, isLocal, done) {
   // Cancel early if the package.json says this is not a demo.
   const {blocklyDemoConfig: demoConfig} = packageJson;
   if (!demoConfig) {
-    done();
-    return;
+    return done();
```
maribethb: I don't think this is correct; we don't want to return the value of this function (I don't even know what that value would be, it's just arcane gulp magic).

If we're returning early then we actually shouldn't return a gulp task. The types here are super unimportant. I think you can change the return type to be `{*}` or maybe `{Function | undefined}`

maribethb (commented):
Thanks for this PR! Unfortunately getting the types correct is a bit trickier than it looks, but I've left some pointers below. Let me know if you have questions!

sdthaker:
> Thanks for this PR! Unfortunately getting the types correct is a bit trickier than it looks, but I've left some pointers below. Let me know if you have questions!

@maribethb Wow! This is amazing feedback! I've never received detailed feedback of this sort, considering I'm quite early in this career and new to open source! Thanks so much for taking the time to explain in detail what should be the code changes and the impact of those changes. I've made changes and pushed the updated code! Please take a look and let me know if anything else needs to be added/removed from this PR.

maribethb (approved):
Awesome, this looks great. Thanks for your work on this!

PR #2067, opened 2023-11-09
fix: Enables strict mode for toolbox-search

### LINKED ISSUE #2029: Enable strict mode for toolbox-search
### Check for duplicates

### Component

toolbox-search

### Problem

_No response_

### Request

The toolbox-search plugin does not have "[strict mode](https://www.typescriptlang.org/tsconfig#strict)" enabled for TypeScript. We should enable it to have better type checking for our plugins, but doing so may show some TypeScript errors that need to be fixed first.

To work on this issue:

### Alternatives considered

_No response_

### Additional context

_No response_

kapooraryan:
## The basics

### Resolves

Closes #2029 

### Proposed Changes

Enables strict mode for the toolbox-search plugin by configuring "strict":true in tsconfig.json.

Typescript errors that arose as a result:
TS_Errors

Solved by making necessary changes to the toolbox_search.ts file:
Solved_Err_TS

### Reason for Changes

Enables strict mode and carries out checks for null or possibly undefined objects.

### Additional Information

Ran npm run lint to ensure no errors! Let me know if there are any changes required.

**Note**: Was using this StackOverflow [discussion](https://stackoverflow.com/a/72404011/22548352) as reference earlier but this seems to be a forbidden non-null assertion as we are telling typescript the object exists.

### DIFF
--- a/plugins/toolbox-search/src/toolbox_search.ts
+++ b/plugins/toolbox-search/src/toolbox_search.ts
@@ -131,7 +131,7 @@ export class ToolboxSearchCategory extends Blockly.ToolboxCategory {
    */
   private initBlockSearcher() {
     const availableBlocks = new Set<string>();
-    this.workspace_.options.languageTree?.contents?.map((item) =>
+    this.workspace_.options.languageTree?.contents?.forEach((item) =>
       this.getAvailableBlocks(item, availableBlocks),
     );
     this.blockSearcher.indexBlocks([...availableBlocks]);
@@ -156,15 +156,13 @@ export class ToolboxSearchCategory extends Blockly.ToolboxCategory {
    */
   override setSelected(isSelected: boolean) {
     super.setSelected(isSelected);
+    if (!this.searchField) return;
     if (isSelected) {
-      this.searchField?.focus();
+      this.searchField.focus();
       this.matchBlocks();
     } else {
-      const searchField = this.searchField
-      if(searchField){
-        searchField.value = '';
-      }
-      this.searchField?.blur();
+      this.searchField.value = '';
+      this.searchField.blur();
     }
   }
 
--- a/plugins/toolbox-search/src/toolbox_search.ts
+++ b/plugins/toolbox-search/src/toolbox_search.ts
@@ -63,7 +63,7 @@ export class ToolboxSearchCategory extends Blockly.ToolboxCategory {
 
       this.matchBlocks();
     });
-    this.rowContents_.replaceChildren(this.searchField);
+    this.rowContents_?.replaceChildren(this.searchField);
     return dom;
   }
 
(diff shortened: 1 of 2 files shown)


kapooraryan:
@BeksOmega Let me know if this is okay?

plugins/toolbox-search/src/toolbox_search.ts
```diff
@@ -131,7 +131,7 @@ export class ToolboxSearchCategory extends Blockly.ToolboxCategory {
    */
   private initBlockSearcher() {
     const availableBlocks = new Set<string>();
-    this.workspace_.options.languageTree.contents.map((item) =>
+    this.workspace_.options.languageTree?.contents?.map((item) =>
```
BeksOmega: This is unrelated, but can you switch this to `forEach` instead of `map`? We shouldn't be using `map` if we're not using the return value.

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#:~:text=Since%20map%20builds%20a%20new%20array%2C%20calling%20it%20without%20using%20the%20returned%20array%20is%20an%20anti%2Dpattern%3B%20use%20forEach%20or%20for...of%20instead.

plugins/toolbox-search/src/toolbox_search.ts
```diff
@@ -157,19 +157,22 @@ export class ToolboxSearchCategory extends Blockly.ToolboxCategory {
...
     } else {
-      this.searchField.value = '';
-      this.searchField.blur();
+      const searchField = this.searchField
+      if(searchField){
+        searchField.value = '';
+      }
+      this.searchField?.blur();
     }
   }
```
BeksOmega: I think in this case it's more readable to return early rather than using `?.` repeatedly.

```ts
super.setSelected(isSelected);
if (!this.searchField) return;
if (isSelected) {
// etc...
```

BeksOmega (commented):
This looks great! Thank you for the fixes :D Two nits (one unrelated to your changes) then I'll give this a manual test, and should be good to merge!

plugins/toolbox-search/src/toolbox_search.ts
kapooraryan: Went through the reference you provided and updated the same.

plugins/toolbox-search/src/toolbox_search.ts
kapooraryan: Yes, you're right. Added the check for the object before instead of using ? repeatedly. Let me know if there's anything else @BeksOmega .

BeksOmega:
Heya @kapooraryan, can you run the formatter (`npm run format` at root) and commit the changes? Then this should be good to go!

kapooraryan:
> Heya @kapooraryan, can you run the formatter (`npm run format` at root) and commit the changes? Then this should be good to go!

Hey @BeksOmega , pushed the commit to this PR. Lmk if it's okay?

BeksOmega:
> > Heya @kapooraryan, can you run the formatter (`npm run format` at root) and commit the changes? Then this should be good to go!
> 
> Hey @BeksOmega , pushed the commit to this PR. Lmk if it's okay?

Looks perfect! Thank you so much for your work on this :D

Issue #1981, opened 2023-09-26
Enable linting and formatting in examples


maribethb:
### Check for duplicates

- [ ] I have searched for similar issues before opening a new one.

### Component

all examples

### Problem

Current behavior: examples are not linted, nothing is formatted
After https://github.com/google/blockly-samples/pull/1946: examples are excluded from linting and formatting

We want to lint and format examples as well

### Request

1. Remove `examples` from the ignore list in `eslint.config.js`
2. Remove `examples` from the ignore list in `.prettierignore`
3. Potentially need to add new files to the ignore list(s) if there are any files in the examples that shouldn't be formatted. Commit all changes so far
4. Run `npm run lint:fix` and commit the changes.
5. Run `npm run format` and commit the changes.
6. Run `npm run lint`, manually fix any remaining lint errors or warnings, and commit the changes.

It is important that steps 4, 5, and 6 each be committed separately, as this will make reviewing and rebasing the PR easier.

Note that some examples specify their own lint command and I'm not sure how this will interact with the new flat config. If you attempt this process and find issues with linting any of the examples, let us know in the comments!

alicialics:
There were 1k+ eslint errors, but most of them were caused by a few select plugins (eg interpreter-demo).  Ignoring them brings down the number to 125 errors (and many more warnings).

It might be simpler to do it example by example to make it easier to review since not all of changes are trivial and its hard to guarantee a change is correct without actually running it and hopefully test what you've changed.

We can use the re-include pattern to ignore all examples but a project you are working on
https://eslint.org/docs/latest/use/configure/ignore
> Lines preceded by ! are negated patterns that re-include a pattern that was ignored by an earlier pattern.

BeksOmega:
Believe this was closed by #2070, correct me if I'm wrong @alicialics !

alicialics:
I think there are some work left over.  #2068 runs prettier on all example files as well as enabling eslint on *some* examples.  After #2068 is merged I think there are significant work in enabling eslint on the rest of examples.

PR #2068, opened 2023-11-09
chore: run prettier on examples

alicialics:
## The basics

Run eslint on the following examples, **this will result in no changes** since there are no errors detected by eslint.
Run prettier on **all** examples.  All changes in this PR comes from prettier.

```
examples/blockly-svelte
examples/custom-dialogs-demo
examples/custom-generator-codelab
examples/custom-renderer-codelab
examples/custom-tooltips-demo
examples/fixed-demo
examples/generator-demo
examples/graph-demo
examples/headless-demo
examples/keyboard-navigation-codelab
examples/max-blocks-demo
examples/mirror-demo
examples/resizable-demo
examples/rtl-demo
examples/sample-app
examples/single-direction-scroll-demo
examples/toolbox-demo
```

### Resolves

#1981 

### Proposed Changes

the ignored examples contains one or more eslint errors and so it's not tackled here

### Reason for Changes

#1981

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

Svelte example is broken.  I will open a separate PR
cc @maribethb


eslint.config.js
```diff
@@ -19,11 +19,26 @@ const compat = new FlatCompat();
 module.exports = [
   {
     ignores: [
-      'examples/',
-      'codelabs/',
-      'gh-pages/',
```
maribethb, line 24: I think we want to exclude `gh-pages` still as that is largely autogenerated. It may not error now but it would be annoying if something generated warnings later. I'd also probably ignore `codelabs` since they're still ignored by prettier.

maribethb (commented):
Thanks for looking at this!

maribethb:
I incorporated these commits into #2182 along with further updating examples based on current version of examples and prettier. Thanks for your work on this! Sorry this sat for a while.

PR #2072, opened 2023-11-10
chore: remove ng lint in angular example

### LINKED ISSUE #1679: Remove less useful examples

alicialics:
## The basics

ng lint was broken because tslint is deprecated, this PR removes tslint all together

### Resolves

#1679 

### Proposed Changes

ran automatic migrations

### Reason for Changes

fixes lint command

### Test Coverage

ran lint

### Documentation

N/A

### Additional Information

N/A


cpcallen (commented):
I'm struck by how many dev dependencies this adds; I don't know enough about linting agular projects but it is a surprisingly large number.

Mainly, though, I think I should defer to @maribethb on whether this is the right approach, so I'm going to reassign it to her.

maribethb (commented):
Is it possible to just remove the linting entirely, instead of updating it?

This is a great example of why I'm leaning towards not keeping the angular example at all. Unfortunately, nobody on our team has the expertise to give this a fair review or to keep it up to date. I don't want to pretend to folks who see this example that we know what we're doing and this is a good linting setup they should follow (it might be in this PR, but will it be 2 years from now?). Developers who want to understand how to lint angular projects should seek advice from angular docs, not our demos. Removing linting entirely if that's an option would focus the demo more clearly on the blockly part of it instead of being a full-fledged angular demo.

maribethb (approved):
Thank you! This should be a lot easier to maintain for us!

PR #2073, opened 2023-11-13
chore: clean up styling so it renders nicely in iframe

26damb:
## The basics

## The details
Just cleaning up styling a bit since it appears weird in iframe on devsite

### Resolves

Fixes

### DIFF
--- a/examples/devsite-landing-demo/style.css
+++ b/examples/devsite-landing-demo/style.css
@@ -12,8 +12,7 @@ body {
   --demo-height: 352px;
   --font-family: 'Google Sans';
   --output-max-height: 100%;
-  --output-max-width: 340px;
-  --output-min-width: 392px;
+  --output-max-width: 392px;
   --play-btn-right: 52px;
   --play-btn-top: 40px;
   --select-next-line: none;
@@ -128,7 +127,6 @@ option {
   margin: 28px;
   max-height: var(--output-max-height);
   max-width: var(--output-max-width);
-  min-width: var(--output-min-width);
   overflow: scroll;
   padding: 20px;
   width: 100%;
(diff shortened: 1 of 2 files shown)


examples/devsite-landing-demo/style.css
```diff
@@ -12,8 +13,10 @@ body {
   --font-family: 'Google Sans';
   --output-max-height: 100%;
   --output-max-width: 340px;
+  --output-min-width: 392px;
```
cpcallen: It seems surprising that `min-width` > `max-width`, but I'm not CSS expert.

cpcallen (commented):
I'm reluctant to approve any changes to devsite-landing-demo at the moment because I'm not sure how to test that they don't break our actual devsite landing page.

I'm going to assign this to @NeilFraser who created this in the first place and presumably knows how to verify the changes (and is also much more famillliar with CSS than I am).

examples/devsite-landing-demo/style.css
26damb: Oops good catch, let me fix that

26damb:
@cpcallen I was actually the one that created this specific sample, and it will only be utilized on one page. I'm also building out that page myself in staging so I'm seeing how it would look with these change I suggested here.

@cpcallen or @NeilFraser could you reassign to @maribethb that's who I worked with before

It seems I can't reassign myself

maribethb:
This is the new demo created by the marketing folks for the new landing site. This won't affect the demo currently on the site - apologies for the naming confusion, it was originally `devsite-demo2` but the name needed to end in just `demo` for lerna to work and I was fixing it on my day off :P

maribethb (approved):

PR #2074, opened 2023-11-14
fix: Fix TS errors in plugins/toolbox-search

### LINKED ISSUE #2029: Enable strict mode for toolbox-search

elvisvoer:
## The basics

### Resolves

This PR fixes the TypeScript errors shown when running the `plugins/toolbox-search` using tsc `5.0.4`.

### DIFF
--- a/plugins/toolbox-search/test/index.ts
+++ b/plugins/toolbox-search/test/index.ts
@@ -36,9 +36,11 @@ document.addEventListener('DOMContentLoaded', function () {
   const defaultOptions: Blockly.BlocklyOptions = {
     toolbox,
   };
-  createPlayground(
-    document.getElementById('root')!,
-    createWorkspace,
-    defaultOptions,
-  );
+
+  const root = document.getElementById('root');
+  if (!root) {
+    throw new Error('Root element is missing');
+  }
+
+  createPlayground(root, createWorkspace, defaultOptions);
 });
--- a/plugins/toolbox-search/test/index.ts
+++ b/plugins/toolbox-search/test/index.ts
@@ -27,7 +27,7 @@ function createWorkspace(
 }
 
 document.addEventListener('DOMContentLoaded', function () {
-  const toolbox = Object.assign({}, toolboxCategories);
+  const toolbox = Object.assign({}, toolboxCategories) as unknown as Blockly.utils.toolbox.ToolboxInfo;
   toolbox['contents'].push({
     'kind': 'search',
     'name': 'Search',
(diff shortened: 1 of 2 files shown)


plugins/toolbox-search/test/index.ts
```diff
@@ -27,7 +27,7 @@ function createWorkspace(
 }
 
 document.addEventListener('DOMContentLoaded', function () {
-  const toolbox = Object.assign({}, toolboxCategories);
+  const toolbox = Object.assign({}, toolboxCategories) as unknown as Blockly.utils.toolbox.ToolboxInfo;
```
BeksOmega: So it looks like the [type in the dev-tools plugin](https://github.com/google/blockly-samples/blob/6d1d5a57c78c2072f4874644a6eddca5f5764b35/plugins/dev-tools/src/index.d.ts#L112) is incorrect. Could you change `toolboxCategories` and `toolboxSimple` to both be of type `Blockly.utils.toolbox.ToolboxInfo`? Then you should be able to change this to:
```suggestion
  const toolbox = {...toolboxCategories};
```
And the types should work correctly!

plugins/toolbox-search/test/index.ts
```diff
@@ -37,7 +37,7 @@ document.addEventListener('DOMContentLoaded', function () {
     toolbox,
   };
   createPlayground(
-    document.getElementById('root'),
+    document.getElementById('root')!,
```
BeksOmega: Can you change this to check if `'root'` exists instead of asserting it exists?

BeksOmega (changes requested):
Thank you for your work on this! Looks like you uncovered a typing error in a different plugin too, so that's great :D

BeksOmega:
Heya @elvisvoer Are you still interested in working on this? If not I'll close it next Friday!

elvisvoer:
@BeksOmega yes, sorry didn't have the time to address your feedback yet. Hopefully I get a chance this weekend.

plugins/toolbox-search/test/index.ts
elvisvoer: done!

plugins/toolbox-search/test/index.ts
elvisvoer: done!

BeksOmega (approved):
This looks great thank you for the fixes!

PR #2089, opened 2023-12-01
fix: Prevent advanced playground OOMing when scrolling aggressively

### LINKED ISSUE #750: Advanced playground occasionally crashes with too much scrolling
**Category**

Plugins

**Component**

dev-tools

**Describe the bug**

Chrome occasionally crashes if you scroll too much/fast in the advanced playground.

**To Reproduce**

1. Open an advanced playground
2. Enable the "mousewheel move" option so that using the mousewheel will move the workspace
3. Drag some blocks onto the workspace, or do an airstrike (I reproduced with only 20 blocks)
4. Scroll rapidly using the mousewheel
5. Chrome will eventually crash, probably. If you have the console open, and you're lucky, Chrome will put a breakpoint just before the out-of-memory error.

**Additional context**

I reproduced this in several different plugins, so I don't believe the issue is with any particular plugin, other than the dev-tools advanced playground itself. I was able to get Chrome to stop once before crashing, and the call stack was somewhere in the [`generate` method](https://github.com/google/blockly-samples/blob/7097e15069d960bcf7eba0752370b145cfe7e95f/plugins/dev-tools/src/playground/index.js#L156). I think this is being called any time the viewport changes. You can see the text in the generator section flashing as you scroll. But I haven't looked any deeper than this yet.

gonfunko:
## The basics

### Resolves
Fixes #750

### Proposed Changes
Don't regenerate serialized workspace state in response to viewport change events.

### Reason for Changes
When scrolling aggressively, lots of viewport change events are fired, and the generation/display (with syntax highlighting) of the serialized workspace state can result in the tab crashing.

### DIFF
--- a/plugins/dev-tools/src/playground/index.js
+++ b/plugins/dev-tools/src/playground/index.js
@@ -371,7 +371,7 @@ export function createPlayground(
 
           updateEditor();
           workspace.addChangeListener((e) => {
-            if (e.type !== 'ui') {
+            if (e.type !== 'ui' && e.type !== 'viewport_change') {
               updateEditor();
             }
           });


maribethb (approved):
beautiful, thanks Aaron!

PR #2090, opened 2023-12-02
fix: Use setShadowState instead of setShadow in shadow-block-converter

### LINKED ISSUE #2064: shadow-block-converter should use `setShadowState` instead of `setShadow`
### Check for duplicates

### Component

shadow-block-converter

### Description

Currently the shadow-block-converter plugin uses `setShadow` on a block to swap it between being a shadow and being a real block. Instead it should use `setShadowState` or kill/create the real block.

`setShadow` is not a supported method of creating shadow blocks, because serialization looks at the connection's shadow state, not the state of a connected shadow block.

### Reproduction steps

1. 
2. 
3.

### Stack trace

_No response_

### Screenshots

_No response_

johnnesky:
## The basics

### Resolves

Fixes #2064

### Proposed Changes

The shadow-block-converter no longer uses the setShadow method of the edited shadow block. Instead, it serializes and re-instantiates any edited shadow block as a regular block with a different id, and attaches this block to the parent connection on top of the existing shadow block. It then uses the setShadowState method of the parent Connection to revert any changes made to the connection's persistent shadow state, preserving the original shadow state as a default value.

### Reason for Changes

The shadowState method on blocks is considered internal and is not intended to be used by external code such as plugins, because when used to convert a regular block to a shadow block that was not previously associated with a shadow state it can have confusing results. The plugin behavior was otherwise good, in fact it was nicer in some ways than the new behavior because it preserved the id and the instance of the edited block, and did not need to serialize or deserialize anything.

### Test Coverage

The existing shadow-block-converter tests were updated to accommodate the fact that the edited blocks became new instances with new ids. I also added tests to ensure that the connection's shadow state is consistent after edits, undos, and redos, so that we could rely on the shadow state acting as a default value.

### Documentation

I updated the readme.

### Additional Information

I attempted to at least preserve the id of the edited blocks, but ran into problems. Ideally, we would let the shadow block and the regular block have the same id, and in principle that ought to be possible since they don't need to coexist at the same time. However, we need to call setShadowState on the connection after attaching the regular block to revert any changes made by attaching the block, and the setShadowState implementation temporarily instantiates the provided shadow state, which can't have the same id as a block in the workspace. Changing the shadow state id won't be compatible with any existing undo history on the shadow block, such as the block change event that triggered this whole shadow conversion, so the only alternative is to change the id of the regular block so that the shadow state won't conflict with it.


plugins/shadow-block-converter/README.md
```diff
@@ -1,6 +1,6 @@
 # @blockly/shadow-block-converter [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
 
-A [Blockly](https://www.npmjs.com/package/blockly) plugin for automatically converting shadow blocks to real blocks when the user edits them.
+A [Blockly](https://www.npmjs.com/package/blockly) plugin for automatically converting shadow blocks to regular blocks when the user edits them.
```
cpcallen, line 3: How regular are they—have they been eating enough fibre??  😉

(No action required.)

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -11,48 +11,57 @@
...
+  shadowState: Blockly.serialization.blocks.State;
+}
 
 /**
- * A new blockly event class specifically for recording changes to the shadow
- * state of a block. This implementation is similar to and could be merged with
- * the implementation of Blockly.Events.BlockChange in Blockly core code.
+ * A Blockly event class to revert a parent block connection's shadow shadow
+ * state to the provided state after attaching a different block that would
+ * otherwise affect the connection's shadow state.
```
cpcallen: The parent block connection's "shadow shadow state"?  I'm not sure if this is a mistake or if I'm just not smart enough to understand this comment, but either way I feel like this could probably be revised for clarity.

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -11,48 +11,57 @@
...
-  static readonly EVENT_TYPE = 'block_shadow_change';
+  static readonly EVENT_TYPE = 'block_shadow_state_change';
   /* eslint-enable @typescript-eslint/naming-convention */
 
   /**
-   * The previous value of the field.
+   * The index of the connection in the parent block's list of connections.
    */
-  oldValue: unknown;
+  connectionIndex?: number;
```
cpcallen: This is fine, but I wonder if `indexInParent` might be clearer?  (Though it omits "connection", so maybe not.  Totally your call either way.)

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -11,48 +11,57 @@
...
+   * @param block The parent of the connection to modify.
+   * @param connectionIndex The index of the connection to modify.
+   * @param shadowState The intended shadow state of the connection.
    */
-  constructor(block?: Block, oldValue?: boolean, newValue?: boolean) {
+  constructor(
+    block?: Block,
+    connectionIndex?: number,
+    shadowState?: Blockly.serialization.blocks.State,
+  ) {
```
cpcallen: Should the parameters be optional if they're really needed?  (I see multiple checks below that they are set.)  Or is it necessary they be optional for deserialisation purposes?

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -122,17 +143,154 @@ export class BlockShadowChange extends Blockly.Events.BlockBase {
       );
     }
 
-    const value = forward ? this.newValue : this.oldValue;
-    block.setShadow(!!value);
+    const connections: Blockly.Connection[] = block.getConnections_(true);
+    if (
+      this.connectionIndex === undefined ||
+      this.connectionIndex == -1 ||
+      this.connectionIndex >= connections.length
```
cpcallen: I was going to comment on the use of `==` instead of `===` but I realised we could be more defensive: 
```suggestion
      typeof this.connectionIndex !== 'number' ||
      this.connectionIndex < 0 ||
      this.connectionIndex >= connections.length
```

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -122,17 +143,154 @@ export class BlockShadowChange extends Blockly.Events.BlockBase {
...
+ *     created.
+ */
+function reifyEditedShadowBlock(shadowBlock: Block): Blockly.Block | null {
+  // Determine how the shadow block is connected to the parent.
+  let parentConnection: Blockly.Connection | null = null;
+  let connectionIsThroughOutputConnection = false;
+  if (
+    shadowBlock.previousConnection &&
+    shadowBlock.previousConnection.isConnected()
+  ) {
```
cpcallen:
```suggestion
  if (shadowBlock.previousConnection?.isConnected()) {
```

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -122,17 +143,154 @@ export class BlockShadowChange extends Blockly.Events.BlockBase {
...
+    parentConnection = shadowBlock.previousConnection.targetConnection;
+  } else if (
+    shadowBlock.outputConnection &&
+    shadowBlock.outputConnection.isConnected()
+  ) {
```
cpcallen:
```suggestion
  } else if (shadowBlock.outputConnection?.isConnected()) {
```

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -122,17 +143,154 @@ export class BlockShadowChange extends Blockly.Events.BlockBase {
...
+    parentConnection = shadowBlock.outputConnection.targetConnection;
+    connectionIsThroughOutputConnection = true;
+  }
```
cpcallen, line 190: What if the shadow block has both a previous connection and an output connection?  🙃

Probably not _actually_ worth worrying too much about—none of our default blocks have both—but external devs have done some _interesting_ experiments over the years…

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -122,17 +143,154 @@ export class BlockShadowChange extends Blockly.Events.BlockBase {
       );
     }
 
-    const value = forward ? this.newValue : this.oldValue;
-    block.setShadow(!!value);
+    const connections: Blockly.Connection[] = block.getConnections_(true);
```
cpcallen: I don't think this type declaration is necessary—I believe it can be inferred from the return type of `.getConnections_`—and I'm not sure it really adds much in terms of readability.  Consider omitting variable type declarations like this when they are not required and are not particularly elucidating.  (Here and many places below.)

cpcallen: I note that `Blockly.prototype.getConnections_` is marked `@internal`.  Is there a way to do this avoiding `@internal` members?

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -122,17 +143,154 @@ export class BlockShadowChange extends Blockly.Events.BlockBase {
...
+    parentBlock = newParentBlock;
+
+    // The reference to the connection is obsolete. Find it from the new parent.
+    const parentConnections: Blockly.Connection[] =
+      parentBlock.getConnections_(true);
+    if (connectionIndex == -1 || connectionIndex >= parentConnections.length) {
+      // Couldn't find the corresponding connection on the new version of the
+      // parent block.
+      return null;
+    }
```
cpcallen: Would it be appropriate to throw an error here instead of returning `null`?

I note that this appears to be the only place this function can return `null`, and removing `null` from the return type could simplify things a bit (including removing the need for the check of `newParentBlock` above).

cpcallen (approved):
Approved subject to your consideration of the comments below.

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: Whoops, just a duplication mistake.

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: I renamed it to connectionIndexInParent.

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: To be honest I'm not totally clear on how Blockly's event serialization works and whether it's relevant here. I was just imitating how events in Blockly core are implemented.

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: Yeah I had asked about this the first time I implemented this plugin and I was told that there might be connections for both, but only one should actually be connected at any given time, and I think this implementation supports that.

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: It seems like I can use the parent block's input list to look up connections, so I'll use that instead.

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: Now I renamed it to inputIndexInParent.

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -32,9 +32,10 @@ export class BlockShadowStateChange extends Blockly.Events.BlockBase {
   /* eslint-enable @typescript-eslint/naming-convention */
 
   /**
-   * The index of the connection in the parent block's list of connections.
+   * The index of the connection in the parent block's list of connections. If
+   * null, then the nextConnection will be used instead.
    */
-  connectionIndex?: number;
+  inputIndexInParent?: number | null;
```
cpcallen: So I think this means inputIndexInParent can be a number, `null`, _or `undefined`.  That seems surprising.  Do `undefined` and `null` mean different things here?

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -45,12 +46,14 @@ export class BlockShadowStateChange extends Blockly.Events.BlockBase {
...
-   * @param connectionIndex The index of the connection to modify.
+   * @param inputIndexInParent The index of the input associated with the
+   *     connection to modify, if it is associated with one. Otherwise the
+   *     nextConnection will be used.
    * @param shadowState The intended shadow state of the connection.
    */
   constructor(
     block?: Block,
-    connectionIndex?: number,
+    inputIndexInParent?: number | null,
```
cpcallen: Same here.

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -72,7 +75,7 @@ export class BlockShadowStateChange extends Blockly.Events.BlockBase {
    */
   toJson(): BlockShadowStateChangeJson {
     const json = super.toJson() as BlockShadowStateChangeJson;
-    if (!this.connectionIndex) {
+    if (typeof this.inputIndexInParent === 'undefined') {
```
cpcallen:
```suggestion
    if (this.inputIndexInParent === undefined) {
```

(Unless you are worried that someone might try to assign a different value to the `undefined` global variable, but this has been disallowed since at least ES5.1!)

plugins/shadow-block-converter/src/shadow_block_converter.ts
```diff
@@ -228,24 +239,32 @@ function reifyEditedShadowBlock(shadowBlock: Block): Blockly.Block | null {
...
-    const parentConnections: Blockly.Connection[] =
-      parentBlock.getConnections_(true);
-    if (connectionIndex == -1 || connectionIndex >= parentConnections.length) {
-      // Couldn't find the corresponding connection on the new version of the
-      // parent block.
-      return null;
+    if (inputIndexInParent === null) {
+      parentConnection = parentBlock.nextConnection;
+    } else if (
+      inputIndexInParent == -1 ||
```
cpcallen:
```suggestion
      inputIndexInParent < 0 ||
```
(Or if that is not semantically correct, then at least change to `… === -1`.)

plugins/shadow-block-converter/src/shadow_block_converter.ts
cpcallen: Maybe @BeksOmega would offer her opinion about this.

cpcallen (commented):
Still LGTM but with a couple of minor tweaks suggested, and I'm still curious about the optional-but-not-really parameters to the constructor.

plugins/shadow-block-converter/src/shadow_block_converter.ts
BeksOmega: I believe we had to leave things as optional in core for backwards compatibility, but I don't think they need to be optional in newly created events.

However in this case, the first two parameters are already optional, which means any following parameters need to be optional as well.

cpcallen: (@johnnesky, probably:) Can all three parameters be made non-optional?

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: Since requiring the constructor arguments to be defined, there's no need for this property to allow undefined values. (But yes they meant different things, null means that the connection is not associated with an input, i.e. it's the nextConnection.)

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: Ditto.

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: This check isn't needed now that the properties and constructor arguments can't be undefined.

plugins/shadow-block-converter/src/shadow_block_converter.ts
johnnesky: Done.

plugins/shadow-block-converter/src/shadow_block_converter.ts
BeksOmega: Sorry didn't see this. @cpcallen @johnnesky That is a breaking change since this class constructor is exported.

johnnesky: Oops, I had already merged this PR. Reverted in https://github.com/google/blockly-samples/pull/2149 and submitted again with the breaking change indicator in https://github.com/google/blockly-samples/pull/2150

PR #2092, opened 2023-12-04
fix: Fix text for the disable top blocks plugin on the demo page #2081

### LINKED ISSUE #2081: Fix text for the disable top blocks plugin on the demo page
### Check for duplicates

### Component

https://github.com/google/blockly-samples/blob/a9ac5cb292620932128299d91a148f1be1f5a8b7/gh-pages/_index.html#L318

### Description

The current text states that the plugin disables top blocks, which is does not do. That behavior is handled by a change listener that is exported from core and can be optionally registered.

## To Fix

Replace the current text with `A plugin that hides the 'Enable' context menu option for orphan blocks. Works with the <code>Blockly.Events.disableOrphans</code> change listener in core.`

shivalipatel6:
Replaced text on line 318 with a better description for the plugin

## The basics

- [X ] I [validated my changes](https://developers.google.com/blockly/guides/contribute/samples#making_and_verifying_a_change)

### Resolves
Resolves an incorrect description of a plugin's capabilities in the demo page. 

Fixes #2081

### DIFF
--- a/gh-pages/_index.html
+++ b/gh-pages/_index.html
@@ -315,7 +315,7 @@
         Disable Top Blocks
       </h4>
       <p>
-        A plugin that disables all top-level blocks that have a previous connection. Useful for event-driven environments where all code execution must begin from top-level start/event blocks.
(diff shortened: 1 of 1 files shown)


BeksOmega (approved):
This looks perfect! Thank you for the fix =)

PR #2106, opened 2023-12-12
feat: create createBackingVariable method

### LINKED ISSUE #1861: Make it easier to subclass `ObservableParameterModel`
**Category**
 - Plugins

**Component**

block-shareable-procedures

It's difficult to add support for typed parameter models. Part of the problem is in core, but part of the problem is that the constructor for `ObservableParameterModel` always creates an untyped variable. Even after we fix https://github.com/google/blockly/issues/7397 you still wouldn't be able to subclass the `ObservableParameterModel` due to the requirement of calling `super` in the constructor, which would create an untyped variable. This is a pretty useful class and we should make it easy for developers to subclass it.

Move the [part of the constructor that creates a variable](https://github.com/google/blockly-samples/blob/cd570706cb21d31c6bbf64f3b6597402565a1330/plugins/block-shareable-procedures/src/observable_parameter_model.ts#L44) to a protected method called `createBackingVariable` (or similar). This method can be overridden by subclasses so that a typed variable could be created instead of untyped.

**Additional context**

shivalipatel6:
Created a protected method that creates the variable for observable_parameter_model to make the class easier to subclass

## The basics

- [X ] I [validated my changes](https://developers.google.com/blockly/guides/contribute/samples#making_and_verifying_a_change)

### Resolves

Fixes  https://github.com/google/blockly-samples/issues/1861 

### Proposed Changes
Creates a new method in the ObservableParameterModel class that will be used to create new variables instead of using the constructor. 

### Reason for Changes
Makes the class easier to subclass when a developer wants to create a typed variable. 

### Test Coverage
I ran the test class in the block-shareable-procedure files and it ran. 
createBackingVariableTest

### DIFF
--- a/plugins/block-shareable-procedures/src/observable_parameter_model.ts
+++ b/plugins/block-shareable-procedures/src/observable_parameter_model.ts
@@ -68,9 +68,8 @@ export class ObservableParameterModel
    *
    * @param name The string to set set the variable to.
    * @param varId The optional ID the backing variable should have.
-   * @returns This parameter model.
+   * @returns The created variable model.
    */
-
   protected createBackingVariable(
     name: string,
     varId?: string,
--- a/plugins/block-shareable-procedures/src/observable_parameter_model.ts
+++ b/plugins/block-shareable-procedures/src/observable_parameter_model.ts
@@ -63,21 +63,20 @@ export class ObservableParameterModel
     return this;
   }
 
-/**
-   * Creates a backing variable in a way that is subclassable. 
+  /**
+   * Creates a backing variable in a way that is subclassable.
    *
-   * @param name The string to set set the variable to. 
+   * @param name The string to set set the variable to.
    * @param varId The optional ID the backing variable should have.
    * @returns This parameter model.
    */
 
-protected createBackingVariable(name: string,varId?: string):this{
-
-  this.variable =
-  this.workspace.getVariable(name) ??
-  this.workspace.createVariable(name, '', varId);
-  return this;
-}
+  protected createBackingVariable(name: string, varId?: string): this {
+    this.variable =
+      this.workspace.getVariable(name) ??
+      this.workspace.createVariable(name, '', varId);
+    return this;
+  }
 
   /**
    * Unimplemented. The built-in ParameterModel does not support typing.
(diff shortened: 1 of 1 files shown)


BeksOmega:
@shivalipatel6 I think this looks great! Can you call the protected function from the constructor? I think if you do that it's no longer a breaking change.

plugins/block-shareable-procedures/src/observable_parameter_model.ts
```diff
@@ -65,6 +63,21 @@ export class ObservableParameterModel
...
 
+  /**
+   * Creates a backing variable in a way that is subclassable.
+   *
+   * @param name The string to set set the variable to.
+   * @param varId The optional ID the backing variable should have.
+   * @returns This parameter model.
+   */
+
+  protected createBackingVariable(name: string, varId?: string): this {
```
BeksOmega: Sorry I should have asked for this change in my last review :/

Could you actually change this to return the new variable instead of assigning it? Then you can assign it in the constructor. That way all of the side effects are localized to the constructor, and people overriding this function don't have to remember to do the side effect!

Then this should be good to go =)

plugins/block-shareable-procedures/src/observable_parameter_model.ts
```diff
@@ -65,6 +63,24 @@ export class ObservableParameterModel
...
   }
 
+  /**
+   * Creates a backing variable in a way that is subclassable.
+   *
+   * @param name The string to set set the variable to.
+   * @param varId The optional ID the backing variable should have.
+   * @returns This parameter model.
+   */
+
```
BeksOmega: Can you remove this blank line?

plugins/block-shareable-procedures/src/observable_parameter_model.ts
```diff
@@ -65,6 +63,24 @@ export class ObservableParameterModel
     return this;
   }
 
+  /**
+   * Creates a backing variable in a way that is subclassable.
+   *
+   * @param name The string to set set the variable to.
+   * @param varId The optional ID the backing variable should have.
+   * @returns This parameter model.
```
BeksOmega:
```suggestion
   * @returns The created variable model.
```

BeksOmega (approved):
Two nits then should be good to go!

BeksOmega:
This looks great! Thank you so much @shivalipatel6  =)

PR #2110, opened 2023-12-13
feat: Add `blocksToString()` function

cpcallen:
## The basics

### Resolves

Fixes https://github.com/orgs/google/projects/109/views/1?pane=issue&itemId=45223090

### Proposed Changes

* Add `src/blocks_to_string.js` exporting `blocksToString` function.
* Have `src/index.js` put both `Blockly` and `blocksToString` in the global scope, for console convenience.
* Hook up `getStringification` to `blocksToString`
* Whitespace removal.

### Test Coverage

As if I tested this.

(N.B. while not testing I notice that the "Generate Code Text" context menu is disabled for some reason.)

### DIFF
--- a/examples/sample-p5-app/src/block_svg_patch.js
+++ b/examples/sample-p5-app/src/block_svg_patch.js
@@ -27,7 +27,7 @@ Blockly.BlockSvg.prototype.toString = function() {
       console.log(reason);
     });
   }
-  
+
   return this.llmSummary ?? originalToString.call(this);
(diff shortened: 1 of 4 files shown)


BeksOmega:
The `Generate Code String` function is only enabled if you have a single contiguous stack of blocks selected!

And I think there's a bug with the multiselect plugin where if you right click on a block without having it selected first, it doesn't show up inthe selected set.

PR #2133, opened 2024-01-06
fix: Prevent workspace click handlers in minimap.

### LINKED ISSUE #1923: Minimap comments can be opened
**Category**

 - Plugins

**Component**

workspace-minimap

**Describe the bug**

If you click on a comment icon in the minimap, it will open (but it will be a readonly comment). We probably don't want people to be able to open comments at all because then the minimap won't match the main workspace.

**To Reproduce**

1. Run `npm run start` in the minimap directory.
2. Drag out a block and readd a comment. The comment will not be added to the block in the minimap because of https://github.com/google/blockly/issues/7480
3. Refresh the page so the minimap does get the comment.
4. Click on the comment icon in the minimap. Observe how the comment opens in read-only mode.

**Expected behavior**

Either:
1) Comments are never added.
2) Comments are not openable.

**Additional context**

See the original forum post: https://groups.google.com/g/blockly/c/8MxXhpFAR8c/m/3K4crVg6AgAJ

## To Fix

1) Run the reproduction steps to make sure you can reproduce the error.
2) Modify the minimap workspace config to set `comments: false`.
3) Rerun the reproduct steps. The comment icon should not be mirrored to the minimap workspace.

johnnesky:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/1923

### Proposed Changes

This attaches the pointerdown event in the event capture phase, instead of the (default) bubbling phase, which means it's handled before other event handlers that exist in the workspace. It also calls event.stopImmediatePropagation() so that the other event handlers will not be called.

### Reason for Changes

To prevent the event handler that opens comments from executing in the minimap.

### Test Coverage

I manually tested it with `npm run start`. The existing unit tests still pass.

### Documentation

N/A

### Additional Information

I had to reimplement part of Blockly.browserEvents.bind(), since this core method doesn't provide any way to attach event handlers in the capture phase.

### DIFF
--- a/plugins/workspace-minimap/src/minimap.ts
+++ b/plugins/workspace-minimap/src/minimap.ts
@@ -97,15 +97,24 @@ export class Minimap {
       }
     });
 
+    // The mousedown handler needs to take precedent over other mouse handlers
+    // in the workspace, such as the handler that opens comments, which means it
+    // needs to be attached in the capture phase. Blockly's built-in event
+    // binding does not let us use the capture phase so we reimplement it here.
+    const mouseDownFunc = (event: Event) =>
+      this.onClickDown(event as PointerEvent);
+    this.minimapWorkspace.svgGroup_.addEventListener(
+      'pointerdown',
+      mouseDownFunc,
+      /* usecapture */ true,
+    );
+    this.onMouseDownWrapper = [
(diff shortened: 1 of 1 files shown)


BeksOmega (approved):
LGTM!

PR #2151, opened 2024-01-17
feat: add basic setup for developer-tools

maribethb:
## The basics

### Proposed Changes

Adds basic setup & tooling for developer-tools


examples/developer-tools/package.json
```diff
@@ -0,0 +1,43 @@
+{
+    "name": "developer-tools",
```
BeksOmega, line 2: I think we normally prefix examples with `blockly-`

maribethb:
All of the commits except the last one are just updating the developer-tools branch to latest master. this fails CLA because of the github actions bot that updates package versions, which is expected/fine. the last commit is the change you should review... sorry -_-

examples/developer-tools/src/index.css
```diff
@@ -0,0 +1,55 @@
+body {
```
BeksOmega, line 1: Formatting.

examples/developer-tools/src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+import * as Blockly from 'blockly/core';
+import * as En from 'blockly/msg/en';
+import 'blockly/blocks';
+import './index.css';
+
+// Put Blockly in the global scope for easy debugging.
+(window as any).Blockly = Blockly;
+
+// Even though En should be loaded by default,
+// if you don't load it specifically, you'll get spurious message warnings.
```
BeksOmega, line 16: I think you get spurious warnings because you're loading from `blockly/core`.

examples/developer-tools/src/index.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
```
BeksOmega: Do you wanna update the licenses to 2024? I don't think it matters but not sure if it does.

examples/developer-tools/src/index.html
```diff
@@ -0,0 +1,23 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Blockly Developer Tools</title>
+  </head>
+  <body>
+    <div id="blockFactoryContainer">
```
BeksOmega: I'm no html expert, so I'm curious, is there a reason to wrap all of the html contents in a div instead of just applying the styling to the body?

examples/developer-tools/src/index.html
```diff
@@ -0,0 +1,23 @@
...
+      <div id="mainWorkspace"></div>
```
BeksOmega: Kebab case for id attributes: https://google.github.io/styleguide/htmlcssguide.html#id_Attributes

examples/developer-tools/package.json
```diff
@@ -0,0 +1,43 @@
...
+      "ts-loader": "^9.4.4",
+      "typescript": "^5.1.6",
+      "webpack": "^5.88.2",
+      "webpack-cli": "^5.1.4",
+      "webpack-dev-server": "^4.15.1"
+    },
+    "dependencies": {
+      "blockly": "^10.1.3"
+    }
+  }
```
BeksOmega: Newlines at end of file(s).

examples/developer-tools/package.json
```diff
@@ -0,0 +1,43 @@
...
+      "mini-css-extract-plugin": "^2.7.6",
+      "style-loader": "^3.3.3",
+      "ts-loader": "^9.4.4",
+      "typescript": "^5.1.6",
+      "webpack": "^5.88.2",
+      "webpack-cli": "^5.1.4",
+      "webpack-dev-server": "^4.15.1"
+    },
+    "dependencies": {
+      "blockly": "^10.1.3"
```
BeksOmega: Bump to latest? not sure that it really matters though.

BeksOmega (approved):

examples/developer-tools/src/index.ts
maribethb: Yes, that's true. I also load the English language files separately. If I don't call `setLocale` then the default behavior is that the English language files are loaded (and everything appears to work fine if I don't call it and rely on this default behavior) but I get the warning. Calling `setLocale` changes nothing about the behavior other than silences the warnings. so I'm just explaining why I call `setLocale` even though it appears to work if you don't. our documentation used to say you only needed to do that if you didn't want english, but i noticed it does tell you now to do it in english too.

examples/developer-tools/src/index.html
maribethb: Eventually there will be other "main" divs other than just block factory, e.g. to switch between tabs to do toolbox stuff too.

examples/developer-tools/package.json
maribethb: I don't want lerna to manage this right now so I explicitly left it out (this isn't really an "example" anyway, but it's not a plugin either so examples made more sense) but I can change it later if/when I do want lerna to manage it

PR #2155, opened 2024-01-18
feat: add block factory blocks and toolbox

maribethb:
## The basics

### Resolves

Works on new block factory.

### Proposed Changes

Adds the block factory blocks.

These are largely copied directly from the old block factory. They have been converted to JS where applicable. They have also been converted to TypeScript and they use JSON serialization hooks.
The input and type blocks have been converted into just one block type that uses a mutator to switch between different types or input styles. This makes it easier to modify a block instead of having to recreate whole inputs or drag new blocks out.

I also created a theme instead of manually setting block styles like the old blocks did. I used the same colors as legacy block factory for continuity.

Blocks for the angle, multiline, and colour fields do not yet exist. Because they will need to use and output code related to the plugin fields (which may change in the near future also) these have been saved for a later PR.


BeksOmega:
@maribethb did you mean to assign this to me on purpose?

BeksOmega:
oh sorry I thought it said you requested it, but nah it was assigned from the github bot.

examples/developer-tools/src/blocks/colour.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+
+/**
+ * A hue-picker to set the colour of a block.
+ * TODO: Use the new angle field.
```
BeksOmega: Issue number (unless it's in a follow up you've already got hanging)

examples/developer-tools/src/blocks/colour.ts
```diff
@@ -0,0 +1,29 @@
...
+export const colourHue = {
+  init: function () {
+    this.appendDummyInput()
+      .appendField('hue:')
+      .appendField(new Blockly.FieldAngle('0', this.validator), 'HUE');
+    this.setOutput(true, 'Colour');
+    this.setTooltip('Paint the block with this colour.');
+    this.setHelpUrl('https://www.youtube.com/watch?v=s2_xaEvcVI0#t=55');
+  },
+  validator: function (text: string) {
```
BeksOmega: The validator is passed a number not a string. So you should be able to remove this whole thing and use `this.setColour().bind(this)` instead.

examples/developer-tools/src/blocks/factory_base.ts
```diff
@@ -0,0 +1,109 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+
+export const factoryBase = {
+  // Base of new block.
```
BeksOmega: Change to jsdoc comment. Here and throughout all files.

examples/developer-tools/src/blocks/factory_base.ts
```diff
@@ -0,0 +1,109 @@
...
+        ['← left output', 'LEFT'],
+        ['↕ top+bottom connections', 'BOTH'],
+        ['↑ top connection', 'TOP'],
+        ['↓ bottom connection', 'BOTTOM'],
+      ],
+      function (option: string): string {
+        this.getSourceBlock().updateShape(option);
+        // Connect a shadow block to this new input.
+        this.getSourceBlock().spawnOutputShadow(option);
+        return option;
```
BeksOmega: You don't need to return the value if you're not modifying it.

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+
+import * as Blockly from 'blockly/core';
+
+/**
+ * Check to see if more than one field has this name.
+ * Highly inefficient (On^2), but n is small.
+ * @param referenceBlock Block to check.
+ */
+const fieldNameCheck = function (referenceBlock: Blockly.Block) {
+  if (!referenceBlock.workspace) {
```
BeksOmega: Check `isDisposed` or `isDeadOrDying` instead.

examples/developer-tools/src/blocks/input.ts
```diff
@@ -0,0 +1,110 @@
...
+            ['end-row', 'input_end_row'],
+          ],
+          inputTypeValidator.bind(this),
+        ),
+        'INPUT_TYPE',
+      )
+      .appendField('input')
+      .appendField(new Blockly.FieldTextInput('NAME'), 'INPUTNAME');
+    this.appendStatementInput('FIELDS')
+      .setCheck(null)
```
BeksOmega: I think this should have a check of 'Field'

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+ */
+
+import * as Blockly from 'blockly/core';
+
+/**
+ * Check to see if more than one field has this name.
+ * Highly inefficient (On^2), but n is small.
+ * @param referenceBlock Block to check.
+ */
+const fieldNameCheck = function (referenceBlock: Blockly.Block) {
```
BeksOmega: Method names are usually verbs https://google.github.io/styleguide/jsguide.html#naming-method-names

Maybe `checkForNameConflicts`?

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+      .appendField(new Blockly.FieldTextInput(''), 'TEXT')
+      .appendField(',')
+      .appendField(new Blockly.FieldTextInput('NAME'), 'FIELDNAME');
+    this.setPreviousStatement(true, 'Field');
+    this.setNextStatement(true, 'Field');
+    this.setTooltip(
+      'Static text that serves as a label, and is saved with' +
+        ' block data. Use only if you want to modify this label at runtime.',
+    );
+    this.setHelpUrl('https://www.youtube.com/watch?v=s2_xaEvcVI0#t=88');
```
BeksOmega, line 75: I don't think this help URL is accurate since this block wasn't included in the original video.

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+  init: function () {
+    this.setStyle('field');
+    this.appendDummyInput()
+      .appendField('checkbox')
+      .appendField(new Blockly.FieldCheckbox('TRUE'), 'CHECKED')
+      .appendField(',')
+      .appendField(new Blockly.FieldTextInput('NAME'), 'FIELDNAME');
+    this.setPreviousStatement(true, 'Field');
+    this.setNextStatement(true, 'Field');
+    this.setTooltip('Checkbox field.');
```
BeksOmega: Can you make this consistent with the other tooltips?

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+    this.setHelpUrl('https://www.youtube.com/watch?v=s2_xaEvcVI0#t=485');
+  },
+  onchange: function () {
+    fieldNameCheck(this);
+  },
+};
+
+export const fieldVariable = {
```
BeksOmega, line 155: Should this allow you to specify type information as well?

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+      .appendField(new Blockly.FieldTextInput(src), 'SRC');
+    this.appendDummyInput()
+      .appendField('width')
+      .appendField(new Blockly.FieldNumber('15', 0, NaN, 1), 'WIDTH')
+      .appendField('height')
+      .appendField(new Blockly.FieldNumber('15', 0, NaN, 1), 'HEIGHT')
+      .appendField('alt text')
+      .appendField(new Blockly.FieldTextInput('*'), 'ALT')
+      .appendField('flip RTL')
+      .appendField(new Blockly.FieldCheckbox('FALSE'), 'FLIP_RTL');
```
BeksOmega: It might be nice to add a tooltip to this field + label field explaining what it does.

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+    this.appendDummyInput()
+      .appendField('image')
+      .appendField(new Blockly.FieldTextInput(src), 'SRC');
+    this.appendDummyInput()
+      .appendField('width')
+      .appendField(new Blockly.FieldNumber('15', 0, NaN, 1), 'WIDTH')
+      .appendField('height')
+      .appendField(new Blockly.FieldNumber('15', 0, NaN, 1), 'HEIGHT')
+      .appendField('alt text')
+      .appendField(new Blockly.FieldTextInput('*'), 'ALT')
```
BeksOmega: Similarly, might be nice to have a tooltip explaining the alt text.

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+      width: number;
+      height: number;
+      alt: string;
+    };
+/**
+ * Type for a block that creates a dropdown field.
+ */
+export type FieldDropdownBlock = Blockly.BlockSvg & {
+  userData: DropdownOptionData;
+  cpuData: string;
```
BeksOmega, line 218: What is cpuData?

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+  loadExtraState: function (state: {options: string[]}) {
+    this.optionList = state.options;
+    this.updateShape();
+  },
+  decompose: function (workspace: Blockly.WorkspaceSvg) {
+    // Populate the mutator's dialog with this block's components.
+    const containerBlock = workspace.newBlock('field_dropdown_container');
+    containerBlock.initSvg();
+    let connection = containerBlock.getInput('STACK').connection;
+    for (let i = 0; i < this.optionList.length; i++) {
```
BeksOmega: I think you can use a `for...of` loop here.

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+  saveExtraState: function (): {options: string[]} {
+    return {
+      options: this.optionList,
+    };
+  },
+  loadExtraState: function (state: {options: string[]}) {
+    this.optionList = state.options;
+    this.updateShape();
+  },
+  decompose: function (workspace: Blockly.WorkspaceSvg) {
```
BeksOmega, line 255: qq, do we want to use the built-in mutator UI for this? or do we want to try to change it to a +/- UI?

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+  updateShape: function () {
+    // Delete everything.
+    let i = 0;
+    while (this.getInput('OPTION' + i)) {
+      this.removeInput('OPTION' + i);
+      this.removeInput('OPTION_IMAGE' + i, true);
+      i++;
+    }
+    // Rebuild block.
+    const src = 'https://www.gstatic.com/codesite/ph/images/star_on.gif';
```
BeksOmega: Maybe make this a module-level constant.

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+  // Image.
+  init: function () {
+    this.setStyle('field');
+    const src = 'https://www.gstatic.com/codesite/ph/images/star_on.gif';
+    this.appendDummyInput()
+      .appendField('image')
+      .appendField(new Blockly.FieldTextInput(src), 'SRC');
+    this.appendDummyInput()
+      .appendField('width')
+      .appendField(new Blockly.FieldNumber('15', 0, NaN, 1), 'WIDTH')
```
BeksOmega: Use `null` instead of `NaN`. Here and throughout file.

examples/developer-tools/src/blocks/colour.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export const colourHue = {
+  init: function () {
+    this.appendDummyInput()
+      .appendField('hue:')
+      .appendField(new Blockly.FieldAngle('0', this.validator), 'HUE');
+    this.setOutput(true, 'Colour');
+    this.setTooltip('Paint the block with this colour.');
+    this.setHelpUrl('https://www.youtube.com/watch?v=s2_xaEvcVI0#t=55');
+  },
```
BeksOmega, line 21: Whitespace in between methods in an object literal (here, below, and in all other files) https://google.github.io/styleguide/jsguide.html#formatting-vertical-whitespace

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+      connection = optionBlock.nextConnection;
+    }
+    return containerBlock;
+  },
+  compose: function (this: FieldDropdownBlock, containerBlock: Blockly.Block) {
+    // Reconfigure this block based on the mutator dialog's components.
+    let optionBlock = containerBlock.getInputTargetBlock(
+      'STACK',
+    ) as FieldDropdownBlock;
+    // Count number of inputs and save the existing data.
```
BeksOmega, line 275: Why do you need to save existing data twice? Is it in case things are modified while the mutator is open?

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+        this.optionList.push('image');
+      }
+      data.push([optionBlock.userData, optionBlock.cpuData]);
+      optionBlock =
+        optionBlock.nextConnection &&
+        (optionBlock.nextConnection.targetBlock() as FieldDropdownBlock);
+    }
+    this.updateShape();
+    // Restore any data.
+    for (let i = 0; i < this.optionList.length; i++) {
```
BeksOmega, line 291: Might be worth separating this + the saving above into subprocedures for readability.

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+          .appendField('image')
+          .appendField(new Blockly.FieldTextInput(src), 'SRC' + i);
+        this.appendDummyInput('OPTION_IMAGE' + i)
+          .appendField(' ')
+          .appendField('width')
+          .appendField(new Blockly.FieldNumber('15', 0, NaN, 1), 'WIDTH' + i)
+          .appendField('height')
+          .appendField(new Blockly.FieldNumber('15', 0, NaN, 1), 'HEIGHT' + i)
+          .appendField('alt text')
+          .appendField(new Blockly.FieldTextInput('*'), 'ALT' + i)
```
BeksOmega: Might be worth adding a tooltip as suggested on the other block.

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+          .appendField(',')
+          .appendField(new Blockly.FieldTextInput('OPTIONNAME'), 'CPU' + i);
+      }
+    }
+  },
+  onchange: function () {
+    if (this.workspace && this.optionList.length < 1) {
```
BeksOmega: Remove `this.workspace` check.

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+      this.setWarningText('Drop down menu must\nhave at least one option.');
+    } else {
+      fieldNameCheck(this);
```
BeksOmega: Is there any reason not to run both checks?

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+          .appendField('alt text')
+          .appendField(new Blockly.FieldTextInput('*'), 'ALT' + i)
+          .appendField(',')
+          .appendField(new Blockly.FieldTextInput('OPTIONNAME'), 'CPU' + i);
+      }
+    }
+  },
+  onchange: function () {
+    if (this.workspace && this.optionList.length < 1) {
+      this.setWarningText('Drop down menu must\nhave at least one option.');
```
BeksOmega: When does this warning get removed?

examples/developer-tools/src/blocks/fields.ts
```diff
@@ -0,0 +1,412 @@
...
+  init: function () {
+    this.setStyle('field');
+    this.appendDummyInput().appendField('add options');
+    this.appendStatementInput('STACK');
+    this.setTooltip(
+      'Add, remove, or reorder options\n' +
+        'to reconfigure this dropdown menu.',
+    );
+    this.setHelpUrl('https://www.youtube.com/watch?v=s2_xaEvcVI0#t=386');
+    this.contextMenu = false;
```
BeksOmega, line 399: Is this respected?

examples/developer-tools/src/blocks/input.ts
```diff
@@ -0,0 +1,110 @@
...
+ * If so, applies a warning to the block.
+ * 
+ * Highly inefficient (On^2), but n is small.
+ * @param referenceBlock Block to check.
+ */
+const inputNameCheck = function (referenceBlock: Blockly.Block) {
+  if (referenceBlock.isDeadOrDying()) {
+    // Block has been deleted.
+    return;
+  }
```
BeksOmega:
```suggestion
  if (referenceBlock.isDeadOrDying()) return;
```

examples/developer-tools/src/blocks/input.ts
```diff
@@ -0,0 +1,110 @@
...
+import * as Blockly from 'blockly/core';
+
+/**
+ * Check to see if more than one input has this name.
+ * If so, applies a warning to the block.
+ * 
+ * Highly inefficient (On^2), but n is small.
+ * @param referenceBlock Block to check.
+ */
+const inputNameCheck = function (referenceBlock: Blockly.Block) {
```
BeksOmega: Change to a verb as with the field check.

examples/developer-tools/src/blocks/input.ts
```diff
@@ -0,0 +1,110 @@
...
+  return value;
+};
+
+const tooltip: Record<string, string> = {
+  input_value: 'A value socket for horizontal connections.',
+  input_statement: 'A statement socket for enclosed vertical stacks.',
+  input_dummy:
+    'For adding fields without any block connections. Alignment options (left, right, centre) only affect multi-row blocks.',
+  input_end_row:
+    'For adding fields without any block connections that will be rendered on a separate row from any following inputs. Alignment options (left, right, centre) only affect multi-row blocks.',
```
BeksOmega: Line length (possibly for `input_dummy` as well).

examples/developer-tools/src/blocks/input.ts
```diff
@@ -0,0 +1,110 @@
...
+      this.appendValueInput('TYPE')
+        .setCheck(null)
+        .setAlign(Blockly.inputs.Align.RIGHT)
+        .appendField('type');
+    }
+  } else {
+    this.removeInput('TYPE', true);
+  }
+
+  return value;
```
BeksOmega: No need to return the value if you're not modifying it.

examples/developer-tools/src/blocks/input.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+  const msg =
+    count > 1 ? 'There are ' + count + ' input blocks\n with this name.' : null;
+  referenceBlock.setWarningText(msg);
+};
+
+// Inputs that should take a "type" input for connection checks
+const inputsWithTypeInputs = new Set(['input_value', 'input_statement']);
+
+const inputTypeValidator = function (
```
BeksOmega: Also change to a verb.

examples/developer-tools/src/blocks/input.ts
```diff
@@ -0,0 +1,110 @@
...
+  this: Blockly.Block,
+  value: string,
```
BeksOmega, line 42:
```suggestion
  inputType: string,
```

examples/developer-tools/src/blocks/input.ts
```diff
@@ -0,0 +1,110 @@
...
+      count++;
+    }
+  }
+  const msg =
+    count > 1 ? 'There are ' + count + ' input blocks\n with this name.' : null;
+  referenceBlock.setWarningText(msg);
+};
+
+// Inputs that should take a "type" input for connection checks
+const inputsWithTypeInputs = new Set(['input_value', 'input_statement']);
```
BeksOmega: Instead of having this (and below) use the "type" language, can you change it to use the term "connection check" consistantly? I think that reduces confusion.

examples/developer-tools/src/blocks/input.ts
```diff
@@ -0,0 +1,110 @@
...
+
+const inputTypeValidator = function (
+  this: Blockly.Block,
+  value: string,
+): string {
+  if (inputsWithTypeInputs.has(value)) {
+    if (!this.getInput('TYPE')) {
+      this.appendValueInput('TYPE')
+        .setCheck(null)
```
BeksOmega: Should this have a check of `['TypeArray', 'Type']`?

examples/developer-tools/src/blocks/input.ts
```diff
@@ -0,0 +1,110 @@
...
+      .appendField(
+        new Blockly.FieldDropdown([
+          ['left', 'LEFT'],
+          ['right', 'RIGHT'],
+          ['centre', 'CENTRE'],
+        ]),
+        'ALIGNMENT',
+      );
+    this.setPreviousStatement(true, null);
+    this.setNextStatement(true, null);
```
BeksOmega: Pretty sure these should have a check of 'Input'

examples/developer-tools/src/blocks/type.ts
```diff
@@ -0,0 +1,201 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+
+interface TypeState {
```
BeksOmega: As mentioned above, I think it would be less confusing if this changed to use the language of connection checks instead of types.

[Edit: Ah wait, realizing that's probably not possible because of data migration]

examples/developer-tools/src/blocks/type.ts
```diff
@@ -0,0 +1,201 @@
...
+  saveExtraState: function (): TypeState {
+    return {
+      typeCount: this.typeCount,
+    };
+  },
+  loadExtraState: function (extraState: TypeState) {
+    this.typeCount = extraState.typeCount;
+    this.updateShape();
+  },
+  decompose: function (workspace: Blockly.WorkspaceSvg) {
```
BeksOmega, line 35: Similarly to above, do we want this to use +/-?

examples/developer-tools/src/blocks/type.ts
```diff
@@ -0,0 +1,201 @@
...
+  },
+};
+
+export const typeGroupContainer = {
+  // Container.
+  init: function () {
+    this.jsonInit({
+      message0: 'add types %1 %2',
+      args0: [{type: 'input_dummy'}, {type: 'input_statement', name: 'STACK'}],
+      colour: 230,
```
BeksOmega: style? (here and below)

examples/developer-tools/src/blocks/type.ts
```diff
@@ -0,0 +1,201 @@
...
+      this.removeInput('TYPE' + i);
+      i++;
+    }
+  },
+};
+
+export const typeGroupContainer = {
+  // Container.
+  init: function () {
+    this.jsonInit({
```
BeksOmega: Why are these the only two blocks that use `jsonInit`?

examples/developer-tools/src/blocks/type.ts
```diff
@@ -0,0 +1,201 @@
...
+            if (!this.getField(customType)) {
+              this.getInput('TYPE').appendField(
+                new Blockly.FieldTextInput(this.customType ?? ''),
+                customType,
+              );
+            }
+          } else {
+            this.getInput('TYPE').removeField(customType, true);
+          }
+          return value;
```
BeksOmega: No need to return if not modifying.

examples/developer-tools/src/blocks/type.ts
```diff
@@ -0,0 +1,201 @@
...
+    this.setStyle('type');
+  },
+  saveExtraState: function () {
+    if (this.getField('CUSTOMTYPE')) {
+      return {customType: this.getFieldValue('CUSTOMTYPE')};
+    }
+  },
+  loadExtraState: function (state: any) {
+    this.customType = state?.customType;
+    if (this.getFieldValue('TYPEDROPDOWN') === 'CUSTOM') {
```
BeksOmega: Don't rely on checking for the value of the dropdown. Fields are deserialized in an arbitrary order and this may not be correct.

examples/developer-tools/src/blocks/type.ts
```diff
@@ -0,0 +1,201 @@
...
+          ['other...', 'CUSTOM'],
+        ],
+        (value: string): string => {
+          const customType = 'CUSTOMTYPE';
+          if (value === 'CUSTOM') {
+            if (!this.getField(customType)) {
+              this.getInput('TYPE').appendField(
+                new Blockly.FieldTextInput(this.customType ?? ''),
+                customType,
+              );
```
BeksOmega: Separate this out into a subprocedure.

BeksOmega (changes requested):

examples/developer-tools/src/blocks/colour.ts
maribethb: I'm not quite sure what you're suggesting with this. I could inline the function instead of having it be named, but I think that would be less readable. I can't just pass the literal string you suggested because I need the value from the validator to pass to `setColour`, and I still need to round to 10, etc.

examples/developer-tools/src/blocks/colour.ts
maribethb: IMO this falls under the category of formatting. If Prettier doesn't enforce this, we shouldn't enforce it manually. We've made this decision in other areas too. It might be possible to get eslint to do this instead, but generally if using prettier we should let it do all the formatting and use eslint for non-formatting violations.

examples/developer-tools/src/blocks/fields.ts
maribethb: Ehh, I'm torn. On the one hand, yes, and also soon the whole thing will look a little different so all of them will be somewhat wrong. On the other hand, some help is better than no help... This is the help url in the original block factory for this block, so I'm inclined to leave it.

examples/developer-tools/src/blocks/fields.ts
maribethb: Maybe done? I don't really know what else to say besides it's a field with a checkbox.

examples/developer-tools/src/blocks/fields.ts
maribethb: I think this could be follow-up work if we want to do it, but since the original block doesn't have this I don't want to add it now. (The other blocks that I changed were specific improvements I discussed with Rachel ahead of time / in the original design)

examples/developer-tools/src/blocks/fields.ts
maribethb: It's the computer-readable name for an option in a dropdown field, as opposed to the user-readable one.

examples/developer-tools/src/blocks/fields.ts
maribethb: I don't have particular feelings about this, but moving to +/- would be follow up work I don't want to do in this PR.

examples/developer-tools/src/blocks/fields.ts
maribethb: I'm not sure what you mean by this. This is copied from the way the old block works but it first gets the number and type of blocks connected in the mutator ws, then updates the shape of the block (which always deletes everything and re-adds it from scratch), then applies the existing data to the newly shaped block. This is how most of the built-in core blocks work too afaik. I'm not sure what you're proposing instead.

examples/developer-tools/src/blocks/fields.ts
maribethb: Previously it got removed because if it didn't apply anymore, then the duplicate name check would run, which would set the warning text to null if that didn't apply either. If I'm running both checks it needs a different approach.

examples/developer-tools/src/blocks/fields.ts
maribethb: replaced with dead or dying check

examples/developer-tools/src/blocks/fields.ts
maribethb: My guess is we didn't support multiple warnings at the time this code was originally written.

examples/developer-tools/src/blocks/fields.ts
maribethb: Yes.

examples/developer-tools/src/blocks/input.ts
maribethb: Ehh. I think the majority of the reason this is confusing it because of the use of the word "input", not the word "type". Adding inputs to an input block is confusing. The word type isn't overloaded here. If I changed it here, I'd also want to change the name of the "type" blocks, but I think it would be even more confusing to have the user-facing text say "type" (which matches all of our documentation) but have the block definition code use the language "connection check." So I don't think this would really reduce the confusion, and there just is bound to be a little bit when we're talking about making blocks out of blocks.

examples/developer-tools/src/blocks/input.ts
maribethb: I think this is more confusing personally. I'm used to validators dealing with the `value` and don't think the original is unclear. Adding the word "input" in more places makes it worse imo.

examples/developer-tools/src/blocks/input.ts
maribethb: Prettier doesn't change the line length of long strings, and again, if prettier doesn't format something I don't think we should do it manually.

examples/developer-tools/src/blocks/type.ts
maribethb: Same answer applies

maribethb:
btw this probs needs to be reformatted but i have prettier stuff in a different PR so I'll just reformat it then.

examples/developer-tools/src/blocks/colour.ts
BeksOmega: If you think the readability is important that's cool! But you can remove the `parseInt` and `isNaN` check. You're getting passed a number, and the 10 in `parseInt` is just setting the radix.

examples/developer-tools/src/blocks/fields.ts
BeksOmega: If you made a change I don't think it got pushed.

examples/developer-tools/src/blocks/fields.ts
BeksOmega: Ahhh ok I thought it was like some default data in the block as opposed to what the user specified in the mutator. Which seemed... weird. This makes way more sense!

Could you add an inline comment? I think it'll help with maintainance of this block.

examples/developer-tools/src/blocks/fields.ts
BeksOmega: Oh I see I was just reading this slightly wrong. I didn't realizing it was collecting the `userData` and the `cpuData` off of the blocks.

examples/developer-tools/src/blocks/input.ts
BeksOmega: I agree that we shouldn't do things that prettier doesn't do. But prettier is opposed to wrapping strings (see https://github.com/prettier/prettier/issues/4123). So we're making a decision that long strings are ok. I don't super love that because I think it'll make unit tests hard to review. But not the biggest deal to me either.

examples/developer-tools/src/blocks/input.ts
BeksOmega: sgtm

examples/developer-tools/src/blocks/input.ts
BeksOmega: Does "type" match all of our documentation? I've been working to change it over to use "connection check" or "check", since connection checks != types.

examples/developer-tools/src/blocks/fields.ts
maribethb: I did change it, I'm just not sure what else to say besides it's a field with a checkbox. I'm not sure in what way you find this one to be inconsistent to the other tooltips so I'm not sure what you're asking me to change. Open to suggestions!

examples/developer-tools/src/blocks/input.ts
maribethb: Sorry, I meant "type" matches the documentation & video for the block factory. I do think there is room for confusion by changing it since the other blocks more or less correlate to the old blocks pretty well. But I did change it anyways. I don't think it makes this code any less confusing tbh, but it will be more consistent with our other block definition documentation.

maribethb:
ready for re-review

BeksOmega (approved):

PR #2162, opened 2024-01-24
feat!: add block definitions to colour field plugin

### LINKED ISSUE #2152: Field colour: add block definitions
### Check for duplicates

### Component

Colour field

### Request

Copy block definitions into the plugin:
- Colour picker, random colour, RGB colour, blend

Copy in any tooltips, mutators, extensions, and styles
- Including `parent_tooltip_when_inline`

Provide `install` functions for each block, which install all fields, mutators, and extensions necessary to use that block.

Provide an `installAll` function for all blocks in the plugin
- Call each individual block function

Provide any functions needed for the language generator for each block, including reserved words

rachel-fenichel:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/2152

### Proposed Changes

- Adds block definitions and generators for four blocks associated with the colour field:
  - `colour_blend`
  - `colour_rgb`
  - `colour_picker`
  - `colour_random`
- Demonstrates a pattern for defining and exporting a library of blocks.

**BREAKING CHANGE:** The colour field no longer registers itself on load. The developer must either manually register the field or install blocks, which will install the  field. This is part of a move to have no side effects in field and block definitions, so that tree-shaking can remove unwanted fields and blocks.

### Reason for Changes

This is the next step in extracting block definitions from the core library and publishing them as smaller libraries. Developers can then choose which blocks and fields to install, and can tree-shake out unneeded ones.

### Test Coverage

- Updated mocha tests to call `registerColourField` to unregister previous field (from core).
- Updated test playground to delete/unregister blocks, generators, and the field from core, then install/register the new blocks, generators, and field.
- Stripped use of old `generateFieldTestBlocks` helper in favor of manually defining test blocks and the test toolbox.
  - I found the result easier to read and modify as I tried to test things, and it won't change frequently.

The test toolbox looks like this:
image

The first part of the toolbox tests the exported blocks. The later sections test the implementation of the colour field.

### Documentation
This PR includes documentation updates in the README.


plugins/field-colour/src/blocks/colourPicker.ts
```diff
@@ -0,0 +1,139 @@
...
+import * as JavaScript from 'blockly/javascript';
+import * as Dart from 'blockly/dart';
+import * as Lua from 'blockly/lua';
+import * as PHP from 'blockly/php';
+import * as Python from 'blockly/python';
+import { registerColourField } from '../field_colour';
+
+
+// Helper function to define a single block from a JSON definition.
+function defineBlockFromJson(blockJsonDef: any) {
```
maribethb: this is just `Blockly.defineBlocksWithJsonArray`, no?

rachel-fenichel: Yup. Replaced.

plugins/field-angle/src/field_angle.ts
```diff
@@ -658,12 +658,20 @@ export class FieldAngle extends Blockly.FieldNumber {
...
+  // core Blockly no longer defines field_angle.
+  // If field_angle is not defined in core, this generates a console warning.
+  Blockly.fieldRegistry.unregister('field_angle');
+
+  Blockly.fieldRegistry.register('field_angle', FieldAngle);
+}
 
-Blockly.fieldRegistry.register('field_angle', FieldAngle);
+// Immediately register the field.
+registerAngleField();
```
cpcallen: Is it planned that this will go away in a future version?

plugins/field-angle/src/field_angle.ts
```diff
@@ -658,12 +658,20 @@ export class FieldAngle extends Blockly.FieldNumber {
...
-// Unregister legacy field_angle that was in core.  Delete this once
-// core Blockly no longer defines field_angle.
-// If field_angle is not defined in core, this generates a console warning.
-Blockly.fieldRegistry.unregister('field_angle');
+/**
+ * Register the field and any dependencies.
+ */
+export function registerAngleField() {
+  // Unregister legacy field_angle that was in core.  Delete this once
+  // core Blockly no longer defines field_angle.
```
cpcallen: `TODO(#xxxx): …`

plugins/field-colour/src/blocks.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export function installAllBlocks(generators: Generators = {}) {
+    colourPicker.installBlock(generators);
+    colourRgb.installBlock(generators);
+    colourRandom.installBlock(generators);
+    colourBlend.installBlock(generators);
+}
+
+// Calling this installs blocks, which means it has side effects.
+installAllBlocks();
```
cpcallen: Again: is it planned that this will eventually be removed?

(Also: trailing newline.)

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,271 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly';
+import * as JavaScript from 'blockly/javascript';
```
cpcallen: I would recommend `javascript` rather than `JavaScript`—or better still save yourself some typing below by just getting the bits you need:

```suggestion
import {Block} form from 'blockly';
import {JavascriptGenerator, Order as JavascriptOrder} from 'blockly/javascript';
```

(& similarly for others)

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,271 @@
...
+};
+
+
+/**
+ * Javascript generator definition.
+ * @param block
+ * @param generator 
+ * @returns 
+ */
+export function jsGenerator(
```
cpcallen: I'd prefer "javascript" to "js", but that makes "Generator" even more potenatially confusing.  We really need a nice concise abbreviation for "block generator function…   :-/

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,271 @@
...
+import * as JavaScript from 'blockly/javascript';
+import * as Dart from 'blockly/dart';
+import * as Lua from 'blockly/lua';
+import * as PHP from 'blockly/php';
+import * as Python from 'blockly/python';
+import { registerColourField } from '../field_colour';
+import { Generators, installGenerators } from './generatorUtils';
+
+
+const blockName = 'colour_blend';
```
cpcallen:
```suggestion
const BLOCK_NAME = 'colour_blend';
```

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,271 @@
...
+
+// Block for blending two colours together.
+const jsonDef =
+{
```
cpcallen: I think you should export the block definition itself, not just the function that installs it:
```suggestion
export const blocks = Blockly.common.createBlockDefinitionsFromJsonArray({
{
```
It might alternatively be preferable to `export const colour_blend = blocks['colour_blend'];` (instead of all of `blocks`).  I'm not sure I have a strong preference one way or the other.  (The linter will complain about the name!)  Defining one block per file removes most of the issues around tree-shaking a `blocks` dictionary with multiple entries.

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,271 @@
...
+        functionName + '(' + colour1 + ', ' + colour2 + ', ' + ratio + ')';
+    return [code, Python.Order.FUNCTION_CALL];
+}
+
+/**
+ * Install the `colour_blend` block and all of its dependencies.
+ */
+export function installBlock(generators: Generators = {}) {
+    registerColourField();
+    Blockly.common.defineBlocksWithJsonArray([jsonDef]);
```
cpcallen:
```suggestion
    Blockly.common.defineBlocks(blocks);
```

plugins/field-colour/src/blocks/generatorUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+export type JsBlockCodeGenerator = (
+    block: Block,
+    generator: JavascriptGenerator,
+) => [string, number] | string | null;
+
+
+export type DartBlockCodeGenerator = (
+    block: Block,
+    generator: DartGenerator,
+) => [string, number] | string | null;
```
cpcallen: I _believe_ you can simplify this (and avoid type rot) by using [indexed access types](https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html):

```suggestion
export type DartBlockCodeGenerator = DartGenerator.forBlock[''];
```

plugins/field-colour/src/blocks/generatorUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+    block: Block,
+    generator: PythonGenerator,
+) => [string, number] | string | null;
+
+export function installGenerators(generators: Generators = {}, blockName: string,
+    jsGenerator?: JsBlockCodeGenerator,
+    dartGenerator?: DartBlockCodeGenerator,
+    luaGenerator?: LuaBlockCodeGenerator,
+    phpGenerator?: PhpBlockCodeGenerator,
+    pythonGenerator?: PythonBlockCodeGenerator) {
```
cpcallen: Why not make this an options object like `generators` is?: 
```suggestion
    functions: {
      javascript?: JsBlockCodeGenerator,
      dart?: DartBlockCodeGenerator,
      lua?: LuaBlockCodeGenerator,
      php?: PhpBlockCodeGenerator,
      python?: PythonBlockCodeGenerator
    }) {
```

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,271 @@
...
+    installGenerators(generators, blockName,
+        jsGenerator,
+        dartGenerator,
+        luaGenerator,
+        phpGenerator,
+        pythonGenerator);
+    // TODO: Should the reserved words be passed in at the same time as the generator
+    // for the installGenerators function?
+    generators.dart?.addReservedWords('Math');
```
cpcallen: I'd be tempted to put the reserved words in a property on the generator function and let `installGenerators` retrieve them from there, but I know that some of my colleagues have this funny idea that functions shouldn't be treated like normal objects…

I observe that doing that would make this install function less useful as a guide to how to install the required individual pieces—though in fact the `installGenerators` function already does that to large extent.

I think I might favour doing it inline:
```JS
  if(generators.dart) generators.dart.forBlock[blockName] = jsGenerator;
  if(generators.lua) generators.lua.forBlock[blockName] = luaGenerator;
  //..
```
(With judicious choice of variable names it may be possible to keep each of these on a single line.)

cpcallen (commented):
I've had a look at the first one, and it mostly looks good but I do have some… _opinions_.

plugins/field-angle/src/field_angle.ts
rachel-fenichel: Current plan: remove immediate field registration and block registration before merging this PR. Document that developers must call register. (Breaking change).

If devs want to use the blocks that are defined in this plugin, they should call `install` on the blocks. This will also install (register) the fields. If they want the field but not the blocks defined here, they must install (call register) for the field.

plugins/field-colour/src/blocks.ts
```diff
@@ -0,0 +1,31 @@
...
+import * as colourRandom from './blocks/colourRandom';
+import * as colourRgb from './blocks/colourRgb';
+import * as colourBlend from './blocks/colourBlend';
+import { Generators } from './blocks/generatorUtils';
+
+// Re-export all parts of the definition.
+export * as colourPicker from './blocks/colourPicker';
+export * as colourRandom from './blocks/colourRandom';
+export * as colourRgb from './blocks/colourRgb';
+export * as colourBlend from './blocks/colourBlend';
```
rachel-fenichel: Note that I will need to `export * from './blocks'` in `index.ts` (for all plugins).

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Plan: export the block definition (after creation from json, if necessary), but as a single object and not a dictionary.

Also export block name.

plugins/field-colour/src/blocks/generatorUtils.ts
rachel-fenichel: I get this error if I try that: "Exported type alias 'DartBlockCodeGenerator' has or is using private name 'DartGenerator'.ts(4081)"

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: If I do it inline for `colour_picker`, it looks like this:
```

export function installBlock(generators: Generators = {}) {
    registerColourField();
    Blockly.common.defineBlocks([blockDef]);
    if (generators.javascript) generators.javascript.forBlock[BLOCK_NAME] = jsGenerator;
    if (generators.dart) generators.dart.forBlock[BLOCK_NAME] = dartGenerator;
    if (generators.lua) generators.lua.forBlock[BLOCK_NAME] = luaGenerator;
    if (generators.php) generators.php.forBlock[BLOCK_NAME] = phpGenerator;
    if (generators.python) generators.python.forBlock[BLOCK_NAME] = pythonGenerator;
}
```

And if I extend that to a case where I *do* have reserved words, like `colour_blend`, I get this:
```
export function installBlock(generators: Generators = {}) {
    registerColourField();
    Blockly.common.defineBlocks([blockDef]);
    if (generators.javascript) generators.javascript.forBlock[BLOCK_NAME] = jsGenerator;
    if (generators.dart) {
        generators.dart.forBlock[BLOCK_NAME] = dartGenerator;
        generators.dart.addReservedWords('Math');
    }
    if (generators.lua) generators.lua.forBlock[BLOCK_NAME] = luaGenerator;
    if (generators.php) generators.php.forBlock[BLOCK_NAME] = phpGenerator;
    if (generators.python) generators.python.forBlock[BLOCK_NAME] = pythonGenerator;
}
```

It's verbose, but has several advantages:
- Keeps reserved words and block generator installation together.
- The `installBlock` function actually shows everything it's doing
  -  As you observed, `installGenerators` was obfuscating.

The disadvantages are:
- Lots of text to copy.
- Lots of duplicated code.

I think this is worth doing, even if it adds duplicated code. The types did help me avoid installing a dart block generator into python, and that makes me happy. I think it's worth the change. @maribethb do you have any thoughts or objections?

plugins/field-colour/src/blocks/generatorUtils.ts
cpcallen: I am not sure what that means.  (Is it complaining that "`DartGenerator`" is a private name?  Huh?  What's with "`'DartGenerator'.ts`"??)

In any case if you want to export a type like `DartBlockCodeGenerator` I think it would be best if it were extracted from `DartGenerator` rather than constructed manually.

rachel-fenichel: Resolved by moving the block code generator installation into the per-block files, which gets rid of the need for a function signature and explicit type declarations. TS will still type-check appropriately.

plugins/field-colour/src/blocks/generatorUtils.ts
rachel-fenichel: Removed in favor of installing generators in each block's file.

plugins/field-colour/README.md
```diff
@@ -30,11 +32,15 @@ This field accepts up to 4 parameters:
 - "columns" to specify the number of columns the colour dropdown should have.
   Defaults to 7.
 
-### JavaScript
+If you want to use only the field, you must install it by calling `registerFieldColour` before instantiating your blocks. If another field is registered under the same name, this field will overwrite it.
```
cpcallen: Wording: they must register it, but calling that function is not the only possible way to do so as the wording here implies.

plugins/field-colour/README.md
```diff
@@ -64,12 +72,55 @@ Blockly.defineBlocksWithJsonArray([
...
+```js
+import { javascriptGenerator } from 'blockly/javascript';
+import { dartGenerator } from 'blockly/dart';
+import { phpGenerator } from 'blockly/php';
+import { pythonGenerator } from 'blockly/python';
+import { luaGenerator } from 'blockly/lua';
+import { FieldColour } from '@blockly/field-colour';
+
+// Installs all four blocks, the colour field, and all of the language generators.
+FieldColour.installAllBlocks({
```
cpcallen: It appears you import the `FieldColour` constructor and then try to call a (static) method on it called `installAllBlocks`, which I don't think exists.

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,274 @@
...
+import { JavascriptGenerator, Order as JavascriptOrder } from 'blockly/javascript';
+import { DartGenerator, Order as DartOrder } from 'blockly/dart';
+import { LuaGenerator, Order as LuaOrder } from 'blockly/lua';
+import { PhpGenerator, Order as PhpOrder } from 'blockly/php';
+import { PythonGenerator, Order as PythonOrder } from 'blockly/python';
+import { registerFieldColour } from '../field_colour';
+import { Generators } from './generatorUtils';
+
+
+const BLOCK_NAME = 'colour_blend';
```
cpcallen: `export` (and maybe tell the linter to get stuffed).

plugins/field-colour/src/blocks/colourPicker.ts
```diff
@@ -0,0 +1,130 @@
...
+    // Colour picker.
+    const code = generator.quote_(block.getFieldValue('COLOUR'));
+    return [code, PythonOrder.ATOMIC];
+}
+
+
+const definitionMap = 
+BlocklyCommon.createBlockDefinitionsFromJsonArray([jsonDef]);
+
+export const blockDef = definitionMap[BLOCK_NAME];
```
cpcallen: Check styleguide about naming; I think it frowns on abbreviations.

plugins/field-colour/README.md
```diff
@@ -30,11 +32,15 @@ This field accepts up to 4 parameters:
...
   Defaults to 7.
 
-### JavaScript
+If you want to use only the field, you must install it by calling `registerFieldColour` before instantiating your blocks. If another field is registered under the same name, this field will overwrite it.
+
+#### JavaScript
 
 ```js
 import {FieldColour} from '@blockly/field-colour';
```
rachel-fenichel: Note: this sample code is wrong. I need to import `registerFieldColour` as well.

plugins/field-colour/README.md
```diff
@@ -1,6 +1,6 @@
 # @blockly/field-colour [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
 
-A [Blockly](https://www.npmjs.com/package/blockly) colour field.
+A [Blockly](https://www.npmjs.com/package/blockly) colour field and blocks that use it.
```
rachel-fenichel: Open question: how should I describe the plugin, given that it is no longer just the colour field?

plugins/field-colour/README.md
```diff
@@ -64,12 +72,55 @@ Blockly.defineBlocksWithJsonArray([
...
 
+### Blocks
+
+This package also provides four blocks related to the colour field. Each block has generators in JavaScript, Python, PHP, Lua, and Dart.
+- "colour_blend" takes in two colours and a ratio and outputs a single colour.
+- "colour_picker" is a simple block with just the colour field and an output.
+- "colour_random" generates a random colour.
+- "colour_rgb" generates a colour based on red, green, and blue values.
+
+You can install all four blocks by calling `FieldColour.installAllBlocks()`. This will install the blocks and all of their dependencies, including the colour field. You can choose which language generators to install by adding or removing items in the map of generators passed to `installAllBlocks`.
```
rachel-fenichel: Open question: how do I describe "choosing which language generators to install" to make it clear that they are per-block per-language generators?

plugins/field-colour/README.md
rachel-fenichel: Open question: where should I document the `generators` parameter in more detail? Where should the type live? Currently it's in `blocks/generatorUtils.ts` and needs to be duplicated into every plugin where I define blocks.

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,274 @@
...
+            'align': 'RIGHT',
+        },
+    ],
+    'output': 'Colour',
+    'helpUrl': '%{BKY_COLOUR_BLEND_HELPURL}',
+    'style': 'colour_blocks',
+    'tooltip': '%{BKY_COLOUR_BLEND_TOOLTIP}',
+};
+
+/**
```
rachel-fenichel, line 54: Open question: How should I describe these functions? I want to settle on a general-purpose description that I can copy into every plugin, or decide on a shorthand with reference to some other documentation.

rachel-fenichel (commented):
I have updated this PR to only change the colour field.

plugins/field-colour/README.md
cpcallen: perhaps something like "Fields and blocks for choosing and combining colours"?

plugins/field-colour/README.md
cpcallen: Hmm.   How about something like:

> When calling `installAllBlocks`—or any of the individual `installSomeBlock` functions—you can supply one or more `CodeGenerator` instances (e.g. `javascriptGenerator`), and the install function it will also install the correct generator function for each block for the corresponding language(s).

plugins/field-colour/src/blocks/colourBlend.ts
cpcallen: I'd suggest referencing https://developers.google.com/blockly/guides/create-custom-blocks/code-generation/overview#block-code_generators .  Beyond that… 🤔

I think "block code generators" or "block generator functions" are both OK.  The former we already use in some of our docs; the latter is I think slightly more clear (they are just a function, and not a kind of `CodeGenerator`).

plugins/field-colour/README.md
cpcallen: Hmm.  Do we auto-generate API documentation for plugins like we do for core?

In any case it seems like the best approach is probably to declare and explain the type in `GeneratorUtils.ts` (possibly renaming this file, if it no longer contains any functions), but have the individual `install` functions point to some examples in the README rather than trying to explain in detail (and repeatedly) exactly how this parameter works.

rachel-fenichel: My concern is that I will also need to declare this type in `field_angle` and `field_multilinetextinput` when I update those, because there's no shared place for this type to live.

rachel-fenichel:
This pull request is now (finally!) ready for review. Thanks for putting up with all the back and forth on this one.

plugins/field-colour/README.md
```diff
@@ -30,11 +32,15 @@ This field accepts up to 4 parameters:
 - "columns" to specify the number of columns the colour dropdown should have.
   Defaults to 7.
 
-### JavaScript
+If you want to use only the field, you must register it with Blockly. You can do this by calling `registerFieldColour` before instantiating your blocks. If another field is registered under the same name, this field will overwrite it.
```
cpcallen: Line length (here and several places below).  I'm slightly surprised that our lint config does not complain about this.  (I normally configure my editor to wrap at column 70, to allow for minor edits in future commits without having to reflow.)

plugins/field-colour/package.json
```diff
@@ -43,13 +43,13 @@
...
     "@typescript-eslint/parser": "^5.59.5",
-    "blockly": "^10.2.0",
+    "blockly": "^11.0.0-beta.3",
     "chai": "^4.2.0",
     "sinon": "^9.0.1",
     "typescript": "^5.0.4"
   },
   "peerDependencies": {
-    "blockly": "^10.2.0"
+    "blockly": "^11.0.0-beta.3"
```
cpcallen, line 52: This plugin should work with `-beta.1`  as well as the forthcoming `v10` release, no?

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,296 @@
...
+} from 'blockly/javascript';
+import {DartGenerator, Order as DartOrder} from 'blockly/dart';
+import {LuaGenerator, Order as LuaOrder} from 'blockly/lua';
+import {PhpGenerator, Order as PhpOrder} from 'blockly/php';
+import {PythonGenerator, Order as PythonOrder} from 'blockly/python';
+import {registerFieldColour} from '../field_colour';
+import {Generators} from './generatorUtils';
+
+/** The name this block is registered under. */
+// eslint-disable-next-line @typescript-eslint/naming-convention
```
cpcallen: I think this should be fixed in our eslint config, not via eslint-disable.

plugins/field-colour/src/blocks/colourBlend.ts
cpcallen: I am unresolving this because I am still unhappy with the exported names of the block generator functions being the same (or, in this case, nearly the same) as the names of the `CodeGenerator` instances exported by the main package.

What about `blockToJavascript`, `javascriptBlockGenerator` or even just `javascript`?

(And similarly for `dartGenerator` etc. below.)

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,296 @@
...
+  var g = Math.round(g1 * (1 - ratio) + g2 * ratio);
+  var b = Math.round(b1 * (1 - ratio) + b2 * ratio);
+  r = ('0' + (r || 0).toString(16)).slice(-2);
+  g = ('0' + (g || 0).toString(16)).slice(-2);
+  b = ('0' + (b || 0).toString(16)).slice(-2);
+  return '#' + r + g + b;
+}
+`,
+  );
+  const code = functionName + '(' + c1 + ', ' + c2 + ', ' + ratio + ')';
```
cpcallen:
```suggestion
  const code = `${functionName}(${c1}, ${c2}, ${ratio})`;
```

Here and similarly below.

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,296 @@
...
+  // Blend two colours together.
+  const c1 =
+    generator.valueToCode(block, 'COLOUR1', DartOrder.NONE) || "'#000000'";
+  const c2 =
+    generator.valueToCode(block, 'COLOUR2', DartOrder.NONE) || "'#000000'";
+  const ratio = generator.valueToCode(block, 'RATIO', DartOrder.NONE) || 0.5;
+
+  // TODO(#7600): find better approach than casting to any to override
+  // CodeGenerator declaring .definitions protected.
+  (generator as any).definitions_['import_dart_math'] =
```
cpcallen, line 121: Please suppress the lint warning until #7600 is fixed.

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,296 @@
...
+  rs = rs.substring(rs.length - 2);
+  gs = '0$gs';
+  gs = gs.substring(gs.length - 2);
+  bs = '0$bs';
+  bs = bs.substring(bs.length - 2);
+  return '#$rs$gs$bs';
+}
+`,
+  );
+  const code = functionName + '(' + c1 + ', ' + c2 + ', ' + ratio + ')';
```
cpcallen:
```suggestion
  const code = `${functionName}(${c1}, ${c2}, ${ratio})`;
```

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,296 @@
...
+end
+`,
+  );
+  const colour1 =
+    generator.valueToCode(block, 'COLOUR1', LuaOrder.NONE) || "'#000000'";
+  const colour2 =
+    generator.valueToCode(block, 'COLOUR2', LuaOrder.NONE) || "'#000000'";
+  const ratio = generator.valueToCode(block, 'RATIO', LuaOrder.NONE) || 0;
+  const code =
+    functionName + '(' + colour1 + ', ' + colour2 + ', ' + ratio + ')';
```
cpcallen, line 191:
```suggestion
  const code = `${functionName}(${colour1}, ${colour2}, ${ratio})`;
```

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,296 @@
...
+ * @param block The Block instance to generate code for.
+ * @param generator The JavascriptGenerator calling the function.
+ * @returns A tuple containing the code string and precedence.
+ */
+export function jsGenerator(
+  block: Block,
+  generator: JavascriptGenerator,
+): [string, JavascriptOrder] {
+  // Blend two colours together.
+  const c1 =
```
cpcallen: Per style guide I think `colour1` would be preferable, but at least be consistent between languages.

(Here and various places below.  I'm more concerned about variable names in our code than in the generated code, where there is some benefit in concision and less likelihood that a human is reading it.)

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,296 @@
...
+  $b = round($b1 * (1 - $ratio) + $b2 * $ratio);
+  $hex = '#';
+  $hex .= str_pad(dechex($r), 2, '0', STR_PAD_LEFT);
+  $hex .= str_pad(dechex($g), 2, '0', STR_PAD_LEFT);
+  $hex .= str_pad(dechex($b), 2, '0', STR_PAD_LEFT);
+  return $hex;
+}
+`,
+  );
+  const code = functionName + '(' + c1 + ', ' + c2 + ', ' + ratio + ')';
```
cpcallen:
```suggestion
  const code = `${functionName}(${c1}, ${c2}, ${ratio})`;
```

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,296 @@
...
+  return '#%02x%02x%02x' % (r, g, b)
+`,
+  );
+  const colour1 =
+    generator.valueToCode(block, 'COLOUR1', PythonOrder.NONE) || "'#000000'";
+  const colour2 =
+    generator.valueToCode(block, 'COLOUR2', PythonOrder.NONE) || "'#000000'";
+  const ratio = generator.valueToCode(block, 'RATIO', PythonOrder.NONE) || 0;
+  const code =
+    functionName + '(' + colour1 + ', ' + colour2 + ', ' + ratio + ')';
```
cpcallen, line 271:
```suggestion
  const code = `${functionName}(${colour1}, ${colour2}, ${ratio})`;
```

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,296 @@
...
+  return [code, PythonOrder.FUNCTION_CALL];
+}
+
+const definitionMap = BlocklyCommon.createBlockDefinitionsFromJsonArray([
```
cpcallen: I would avoid calling this `…Map` when it's not a `Map`, so I'd prefer `definitions` or even just `blocks` (which is what is currently used in `blocks/*` in core).

plugins/field-colour/src/blocks/colourPicker.ts
```diff
@@ -0,0 +1,142 @@
...
+export function pythonGenerator(
+  block: Block,
+  generator: PythonGenerator,
+): [string, PythonOrder] {
+  // Colour picker.
+  const code = generator.quote_(block.getFieldValue('COLOUR'));
+  return [code, PythonOrder.ATOMIC];
+}
+
+const definitionMap = BlocklyCommon.createBlockDefinitionsFromJsonArray([
```
cpcallen: See above.

plugins/field-colour/src/blocks/colourPicker.ts
```diff
@@ -0,0 +1,142 @@
...
+import {PythonGenerator, Order as PythonOrder} from 'blockly/python';
+import {registerFieldColour} from '../field_colour';
+import {Generators} from './generatorUtils';
+
+/** The name this block is registered under. */
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const BLOCK_NAME = 'colour_picker';
+
+// Block for colour picker.
+const jsonDef = {
```
cpcallen: See above re: naming.

plugins/field-colour/src/blocks/colourRandom.ts
```diff
@@ -0,0 +1,170 @@
...
+};
+
+/**
+ * Javascript block generator function.
+ *
+ * @param block The Block instance to generate code for.
+ * @param generator The JavascriptGenerator calling the function.
+ * @returns A tuple containing the code string and precedence.
+ */
+export function jsGenerator(
```
cpcallen: See above re: naming.

plugins/field-colour/src/blocks/colourRandom.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @returns A tuple containing the code string and precedence.
+ */
+export function dartGenerator(
+  block: Block,
+  generator: DartGenerator,
+): [string, DartOrder] {
+  // Generate a random colour.
+  // TODO(#7600): find better approach than casting to any to override
+  // CodeGenerator declaring .definitions protected.
+  (generator as any).definitions_['import_dart_math'] =
```
cpcallen, line 72: See above re: suppressing lint warning.

plugins/field-colour/src/blocks/colourRandom.ts
```diff
@@ -0,0 +1,170 @@
...
+export function jsGenerator(
+  block: Block,
+  generator: JavascriptGenerator,
+): [string, JavascriptOrder] {
+  // Generate a random colour.
+  const functionName = generator.provideFunction_(
+    'colourRandom',
+    `
+function ${generator.FUNCTION_NAME_PLACEHOLDER_}() {
+  var num = Math.floor(Math.random() * Math.pow(2, 24));
```
cpcallen, line 48: I think it would be preferable to use a constant rather than calling `Math.pow` every time this block is invoked:
```suggestion
  var num = Math.floor(Math.random() * 0x1000000);
```

plugins/field-colour/src/blocks/colourRandom.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @returns A tuple containing the code string and precedence.
+ */
+export function pythonGenerator(
+  block: Block,
+  generator: PythonGenerator,
+): [string, PythonOrder] {
+  // Generate a random colour.
+  // TODO(#7600): find better approach than casting to any to override
+  // CodeGenerator declaring .definitions protected.
+  (generator as any).definitions_['import_random'] = 'import random';
```
cpcallen, line 145: See above re: suppressing lint warning.

plugins/field-colour/src/blocks/colourRandom.ts
```diff
@@ -0,0 +1,170 @@
...
+  const code = "'#%06x' % random.randint(0, 2**24 - 1)";
+  return [code, PythonOrder.FUNCTION_CALL];
+}
+
+const definitionMap = BlocklyCommon.createBlockDefinitionsFromJsonArray([
```
cpcallen: See above re: naming.

plugins/field-colour/src/blocks/colourRgb.ts
```diff
@@ -0,0 +1,249 @@
...
+};
+
+/**
+ * Javascript block generator function.
+ *
+ * @param block The Block instance to generate code for.
+ * @param generator The JavascriptGenerator calling the function.
+ * @returns A tuple containing the code string and precedence.
+ */
+export function jsGenerator(
```
cpcallen: See above re: naming.

plugins/field-colour/src/blocks/colourRgb.ts
```diff
@@ -0,0 +1,249 @@
...
+  g = Math.max(Math.min(Number(g), 100), 0) * 2.55;
+  b = Math.max(Math.min(Number(b), 100), 0) * 2.55;
+  r = ('0' + (Math.round(r) || 0).toString(16)).slice(-2);
+  g = ('0' + (Math.round(g) || 0).toString(16)).slice(-2);
+  b = ('0' + (Math.round(b) || 0).toString(16)).slice(-2);
+  return '#' + r + g + b;
+}
+`,
+  );
+  const code = functionName + '(' + red + ', ' + green + ', ' + blue + ')';
```
cpcallen, line 83:
```suggestion
  const code = `${functionName}(${red}, ${green}, ${blue})`;
```

plugins/field-colour/src/blocks/colourRgb.ts
```diff
@@ -0,0 +1,249 @@
...
+  generator: DartGenerator,
+): [string, DartOrder] {
+  // Compose a colour from RGB components expressed as percentages.
+  const red = generator.valueToCode(block, 'RED', DartOrder.NONE) || 0;
+  const green = generator.valueToCode(block, 'GREEN', DartOrder.NONE) || 0;
+  const blue = generator.valueToCode(block, 'BLUE', DartOrder.NONE) || 0;
+
+  // TODO(#7600): find better approach than casting to any to override
+  // CodeGenerator declaring .definitions protected.
+  (generator as any).definitions_['import_dart_math'] =
```
cpcallen, line 106: See above re: suppressing warnings.

plugins/field-colour/src/blocks/colourRgb.ts
```diff
@@ -0,0 +1,249 @@
...
+  gs = gs.substring(gs.length - 2);
+  num bn = (Math.max(Math.min(b, 100), 0) * 2.55).round();
+  String bs = bn.toInt().toRadixString(16);
+  bs = '0$bs';
+  bs = bs.substring(bs.length - 2);
+  return '#$rs$gs$bs';
+}
+`,
+  );
+  const code = functionName + '(' + red + ', ' + green + ', ' + blue + ')';
```
cpcallen, line 128:
```suggestion
  const code = `${functionName}(${red}, ${green}, ${blue})`;
```

plugins/field-colour/src/blocks/colourRgb.ts
```diff
@@ -0,0 +1,249 @@
...
+  g = math.floor(math.min(100, math.max(0, g)) * 2.55 + .5)
+  b = math.floor(math.min(100, math.max(0, b)) * 2.55 + .5)
+  return string.format("#%02x%02x%02x", r, g, b)
+end
+`,
+  );
+  const r = generator.valueToCode(block, 'RED', LuaOrder.NONE) || 0;
+  const g = generator.valueToCode(block, 'GREEN', LuaOrder.NONE) || 0;
+  const b = generator.valueToCode(block, 'BLUE', LuaOrder.NONE) || 0;
+  const code = functionName + '(' + r + ', ' + g + ', ' + b + ')';
```
cpcallen:
```suggestion
  const code = `${functionName}(${red}, ${green}, ${blue})`;
```

plugins/field-colour/src/blocks/colourRgb.ts
```diff
@@ -0,0 +1,249 @@
...
+  $b = round(max(min($b, 100), 0) * 2.55);
+  $hex = '#';
+  $hex .= str_pad(dechex($r), 2, '0', STR_PAD_LEFT);
+  $hex .= str_pad(dechex($g), 2, '0', STR_PAD_LEFT);
+  $hex .= str_pad(dechex($b), 2, '0', STR_PAD_LEFT);
+  return $hex;
+}
+`,
+  );
+  const code = functionName + '(' + red + ', ' + green + ', ' + blue + ')';
```
cpcallen, line 192:
```suggestion
  const code = `${functionName}(${red}, ${green}, ${blue})`;
```

plugins/field-colour/src/blocks/colourRgb.ts
```diff
@@ -0,0 +1,249 @@
...
+`,
+  );
+  const r = generator.valueToCode(block, 'RED', PythonOrder.NONE) || 0;
+  const g = generator.valueToCode(block, 'GREEN', PythonOrder.NONE) || 0;
+  const b = generator.valueToCode(block, 'BLUE', PythonOrder.NONE) || 0;
+  const code = functionName + '(' + r + ', ' + g + ', ' + b + ')';
+  return [code, PythonOrder.FUNCTION_CALL];
+}
+
+const definitionMap = BlocklyCommon.createBlockDefinitionsFromJsonArray([
```
cpcallen: See above re: naming.

plugins/field-colour/src/blocks/colourRgb.ts
```diff
@@ -0,0 +1,249 @@
...
+  r = round(min(100, max(0, r)) * 2.55)
+  g = round(min(100, max(0, g)) * 2.55)
+  b = round(min(100, max(0, b)) * 2.55)
+  return '#%02x%02x%02x' % (r, g, b)
+`,
+  );
+  const r = generator.valueToCode(block, 'RED', PythonOrder.NONE) || 0;
+  const g = generator.valueToCode(block, 'GREEN', PythonOrder.NONE) || 0;
+  const b = generator.valueToCode(block, 'BLUE', PythonOrder.NONE) || 0;
+  const code = functionName + '(' + r + ', ' + g + ', ' + b + ')';
```
cpcallen, line 221:
```suggestion
  const code = `${functionName}(${r}, ${g}, ${b})`;
```
(but see above re: naming.)

plugins/field-colour/src/blocks/colourRgb.ts
```diff
@@ -0,0 +1,249 @@
...
+    'colour_rgb',
+    `
+def ${generator.FUNCTION_NAME_PLACEHOLDER_}(r, g, b):
+  r = round(min(100, max(0, r)) * 2.55)
+  g = round(min(100, max(0, g)) * 2.55)
+  b = round(min(100, max(0, b)) * 2.55)
+  return '#%02x%02x%02x' % (r, g, b)
+`,
+  );
+  const r = generator.valueToCode(block, 'RED', PythonOrder.NONE) || 0;
```
cpcallen, line 218: See above re: naming.

plugins/field-colour/src/blocks/generatorUtils.ts
cpcallen, line 1: Maybe `generator_types.ts` or just `generators.ts`?

plugins/field-colour/src/blocks/colourBlend.ts
```diff
@@ -0,0 +1,296 @@
...
+import {
+  JavascriptGenerator,
+  Order as JavascriptOrder,
+} from 'blockly/javascript';
+import {DartGenerator, Order as DartOrder} from 'blockly/dart';
+import {LuaGenerator, Order as LuaOrder} from 'blockly/lua';
+import {PhpGenerator, Order as PhpOrder} from 'blockly/php';
+import {PythonGenerator, Order as PythonOrder} from 'blockly/python';
+import {registerFieldColour} from '../field_colour';
+import {Generators} from './generatorUtils';
```
cpcallen: Per practice in the core repo, I think imports of files from the same package should use the full (compiled) filename:

```suggestion
import {registerFieldColour} from '../field_colour.js';
import {Generators} from './generatorUtils.js';
```

This doesn't matter so long as we are webpacking the file, but will matter if we start publishing unbundled packages (which we are intending to).

(Here and most files below.0

plugins/field-colour/src/index.ts
```diff
@@ -5,3 +5,4 @@
  */
 
 export * from './field_colour';
+export * from './blocks';
```
cpcallen: Same as for imports:

```suggestion
export * from './field_colour.js';
export * from './blocks.js';
```

… except, wait: there is no `blocks.ts` or `blocks.js` in this PR, as far as I can see.  Is there some magic going on here, where `export * from 'dirname'` creates namespaces for the exports from each file in `dirname`???

plugins/field-colour/test/index.ts
```diff
@@ -9,38 +9,261 @@
...
-import '../src/index';
+import {javascriptGenerator} from 'blockly/javascript';
+import {dartGenerator} from 'blockly/dart';
+import {phpGenerator} from 'blockly/php';
+import {pythonGenerator} from 'blockly/python';
+import {luaGenerator} from 'blockly/lua';
 
-const toolbox = generateFieldTestBlocks('field_colour', [
+import {createPlayground} from '@blockly/dev-tools';
+import {installAllBlocks as installColourBlocks} from '../src/index';
```
cpcallen, line 19: I'm not sure this rename is useful (unlike the similar one in the readme, which I think is useful.)

cpcallen (changes requested):
Generally looking good, but still a few nits and some dissatisfaction with naming.

- The local variable name suggestions are from my perspective just that, though I think mostly in line with the styleguide's rules.
- The export naming commens are about making sure we will be happy with these names in future / in other similar packages.

Also, one major item: there appear to be unittests for the colour block generators in core, and these should be reimplemented in samples.  I'd be fine with this being done in a separate PR, but I think they should be done (and passing) before this PR gets released.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: I am okay with `blockToJavascript`. 

I would be extremely unhappy with `javascript`, because even in context I don't think a variable named `javascript` or `dart` or `python` clearly indicates what it is and how it is used.

I'm neutral on `javascriptBlockGenerator`.

plugins/field-colour/src/index.ts
rachel-fenichel: I think there's magic where having an `index.ts` means you can access it through the directory name. The re-exports from each file in `blocks/` are done in `blocks/index.ts`, so there's no magic there.

plugins/field-colour/test/index.ts
rachel-fenichel: I found it useful since I have several sets of blocks that I'm working with here. Do you have a strong opinion about it?

plugins/field-colour/README.md
cpcallen: Are you proposing to move the declaration to Blockly itself?  I can see an argument in favour (it would be useful to have in a common location, and why not add it to the thing that all the client code will be importing anyway?), but on balance I am against adding a type to core that's not used anywhere in core.

There might be an argument for adding it to a utility package of field and block definition support code, but so far I think it's just this one type, and the type is straightforward enough that I don't think it's worth creating a separate package for it.  DRY is a useful principle but not IMHO a compelling reason to incur the expense of an extra dependency for this.

If we end up with a bunch of declarations that are actually complicated enough to be worth factoring out then I would think the utility package would be the way to go.

Of course your original question pertains to _documenting_ the type.  To that I feel like the examples usages in the readme are sufficient, especially in conjunction with the type's declaration and JSDoc comment.

plugins/field-colour/src/index.ts
cpcallen: Oh, OK.  I had not paid close enough attention to the file structure.

I think we should avoid relying on such magic, because it is almost certainly provided by webpack.  AFIAK browsers (absent an import map) only ever import the exact file that appears in the import statement.  That means that, at such point as we start publishing the `.js` versions of the source files the current layout will break.

The straightforward fix is to change the line to `export * from './blocks/index.js';`, but I think a better approach would be to fold the contents of `blocks/index.ts` directly into `src/index.ts`; AFAICT the only change needed would be to update the relative paths in the `import` and `export … from` statements.  This gives the top level `index.js` file something useful to do and eliminates at least one round-trip when loading the package unbundled.  (If you want to leave `installAllBlocks` in `blocks/index.js` then I would still recommend moving the `export … from` statements to the top-level `index.js` to improve loading parallelism.)

plugins/field-colour/test/index.ts
cpcallen: No, it is not objectionable.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Settled on `toJavascript`, `toDart`, etc.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Tracked in https://github.com/google/blockly-samples/issues/2198 for follow-up.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Changed to `colour1` and `colour2`, here and below.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Done.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Done.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Done.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Tracked in #2198 for follow-up.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Tracked in #2198 for follow-up.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Tracked in #2198 for follow-up.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Tracked in #2198 for follow-up.

plugins/field-colour/src/blocks/colourPicker.ts
rachel-fenichel: Renamed to `jsonDefinition`, here and in other files.

plugins/field-colour/src/blocks/colourPicker.ts
rachel-fenichel: Renamed.

plugins/field-colour/src/blocks/colourRandom.ts
rachel-fenichel: Done.

plugins/field-colour/src/blocks/colourRandom.ts
rachel-fenichel: Tracked in #2198 for follow-up.

plugins/field-colour/src/blocks/colourRandom.ts
rachel-fenichel: Done.

plugins/field-colour/src/blocks/colourRandom.ts
rachel-fenichel: Done.

plugins/field-colour/src/blocks/colourRandom.ts
rachel-fenichel: Done.

plugins/field-colour/src/blocks/colourRgb.ts
rachel-fenichel: Done.

plugins/field-colour/src/blocks/colourRgb.ts
rachel-fenichel: Tracked in #2198 for follow-up.

plugins/field-colour/src/blocks/colourRgb.ts
rachel-fenichel: Done.

plugins/field-colour/src/blocks/colourRgb.ts
rachel-fenichel: Tracked in #2198 for follow-up.

plugins/field-colour/src/blocks/colourRgb.ts
rachel-fenichel: Tracked in #2198 for follow-up.

plugins/field-colour/src/blocks/colourRgb.ts
rachel-fenichel: Tracked in #2198 for follow-up.

plugins/field-colour/src/blocks/colourRgb.ts
rachel-fenichel: Done.

plugins/field-colour/src/blocks/colourRgb.ts
rachel-fenichel: Tracked in #2198 for follow-up.

plugins/field-colour/src/blocks/colourRgb.ts
rachel-fenichel: Done.

plugins/field-colour/src/index.ts
rachel-fenichel: Exports moved for `src/index.ts` and file deleted.

plugins/field-colour/README.md
rachel-fenichel: Fixed. I also expected this to be done by the linter or formatter.

rachel-fenichel (commented):
I've addressed (or punted) all comments except for the file extensions, which I'm still digging into. 

After discussion with Christopher, I'm going to make a `blocks_for_fields` branch off `master` that I can merge successive field changes against, so that I can make smaller PRs to fix the follow-ups without accidentally releasing anything.

plugins/field-colour/src/blocks/colourBlend.ts
rachel-fenichel: Changing this to use `.js` extensions failed with this error:
```
ERROR in ./src/blocks/colourBlend.ts 12:0-57
Module not found: Error: Can't resolve '../field_colour.js' in '/home/fenichel/Documents/blockly-dev/blockly-samples/plugins/field-colour/src/blocks'
resolve '../field_colour.js' in '/home/fenichel/Documents/blockly-dev/blockly-samples/plugins/field-colour/src/blocks'
  using description file: /home/fenichel/Documents/blockly-dev/blockly-samples/plugins/field-colour/package.json (relative path: ./src/blocks)
    Field 'browser' doesn't contain a valid alias configuration
    using description file: /home/fenichel/Documents/blockly-dev/blockly-samples/plugins/field-colour/package.json (relative path: ./src/field_colour.js)
      no extension
        Field 'browser' doesn't contain a valid alias configuration
        /home/fenichel/Documents/blockly-dev/blockly-samples/plugins/field-colour/src/field_colour.js doesn't exist
      .ts
        Field 'browser' doesn't contain a valid alias configuration
        /home/fenichel/Documents/blockly-dev/blockly-samples/plugins/field-colour/src/field_colour.js.ts doesn't exist
      .js
        Field 'browser' doesn't contain a valid alias configuration
        /home/fenichel/Documents/blockly-dev/blockly-samples/plugins/field-colour/src/field_colour.js.js doesn't exist
      as directory
        /home/fenichel/Documents/blockly-dev/blockly-samples/plugins/field-colour/src/field_colour.js doesn't exist
 @ ./src/index.ts 10:0-52 17:0-54 18:0-40 30:4-28
 @ ./test/index.ts 16:0-71 272:4-23
 ```

rachel-fenichel (commented):
(Extra comment below that I forgot to include in the previous review).

cpcallen (approved):
Approved after a very cursory re-review.

PR #2174, opened 2024-01-27
fix: insertion markers destroying procedure models for block-shareable-procedures in v11

BeksOmega:
## The basics

### Resolves

Fixes N/A

### Proposed Changes

Found that with the new v11 beta the insertion marker managers for procedure definitions were destroying the procedure models when they are destroyed. This is because insertion markers now do the non-full erialization (since they should just reference the model of whatever block they're shadowing).

### Reason for Changes

Fix buggies. This was causing mirroring events into other workspaces to break.

### Test Coverage

Manually tested. Confirmed that tests pass.

### Documentation

I'll update the description of https://github.com/google/blockly/pull/7730 to reflect this break as well.

### Additional Information

N/A

### DIFF
--- a/plugins/block-shareable-procedures/src/blocks.ts
+++ b/plugins/block-shareable-procedures/src/blocks.ts
@@ -215,7 +215,11 @@ const procedureDefGetDefMixin = function () {
      * disposed.
      */
     destroy: function () {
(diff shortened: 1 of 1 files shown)


cpcallen (approved):
Please update the PR description (maybe title?) to explicitly mention that it is the block-shareable-procedures plugin that is being fixed.

PR #2188, opened 2024-01-31
feat: add json definition generator functions

maribethb:
## The basics

### Resolves

Works on Block Factory

### Proposed Changes

- Adds JSON Definition Generator functions for each of the block factory blocks
- Adds a change listener to the main workspace that shows the block preview using the json definition
- Renames the `blocks/type.ts` file to `blocks/connection_check.ts` to match the new name which was renamed in a previous PR

This PR depends on #2187 , but only barely. It's only the changes in `index.ts` and I'm adding to the change listener that was added in that PR. So you can still review this PR for content and I'll rebase when that PR goes in. The bulk of this PR is not dependent on that one.

Note that many parts of the `JsonDefinitionGenerator` are commented out. I plan to clean this up once I've written a) the other types of generators and b) implemented checking the block name doesn't conflict with a different block name and is a valid json identifier, which I am hoping to use the machinery for valid names in generators to do so. So I will need parts that are currently commented out and others I am not sure about yet. I got the bulk of this code from copying the `javascriptGenerator` from core and modifying as needed.

### Test Coverage

Manual testing. I've started working on an automated test that would exercise this generator but there's no mechanism for running tests in `examples/` right now so that's a bigger project and in a different PR.

### DIFF
--- a/examples/developer-tools/src/output-generators/factory_base.ts
+++ b/examples/developer-tools/src/output-generators/factory_base.ts
@@ -1,0 +1,125 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as Blockly from 'blockly/core';
+import {javascriptGenerator} from 'blockly/javascript';
+import {
+  JsonDefinitionGenerator,
+  jsonDefinitionGenerator,
+  Order,
+} from './json_definition_generator';
+
+/**
+ * Builds the 'message0' part of the JSON block definition.
+ * The message should have label fields' text inlined into the message.
+ * Doing so makes the message more translatable as fields can be moved around.
+ *
+ * @param argsList The list of fields and inputs generated from the input stack.
+ * @returns An object containing:
+ *    - a message string with one placeholder '%i`
+ *      for each field and input in the block
+ *    - the new args list, with label fields removed
+ */
+const buildMessageString = function (argsList: Array<Record<string, unknown>>) {
+  let i = 0;
+  let messageString = '';
+  const newArgs = [];
+  for (const arg of argsList) {
+    if (arg.type === 'field_label') {
+      // Label fields get added directly to the message string.
+      // They are removed from the arg list so they don't appear twice.
+      messageString += `${arg.text} `;
+    } else {
+      i++;
+      messageString += `%${i} `;
+      newArgs.push(arg);
+    }
+  }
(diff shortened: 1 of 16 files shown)


examples/developer-tools/src/output-generators/factory_base.ts
```diff
@@ -0,0 +1,125 @@
...
+  generator: JsonDefinitionGenerator,
+): string {
+  // TODO: Get a JSON-legal name for the block
+  const blockName = block.getFieldValue('NAME');
+  // Tooltip and Helpurl string blocks can't be removed, so we don't care what happens if the block doesn't exist
+  const tooltip = JSON.parse(
+    generator.valueToCode(block, 'TOOLTIP', Order.ATOMIC),
+  );
+  const helpUrl = JSON.parse(
+    generator.valueToCode(block, 'HELPURL', Order.ATOMIC),
```
NeilFraser, line 59: I've heard of trailing commas on object literals, but trailing commas on function parameters is new to me.

examples/developer-tools/src/output-generators/json_definition_generator.ts
```diff
@@ -0,0 +1,153 @@
...
+    // TODO: List any reserved words.
+    // this.addReservedWords();
+  }
+
+  /**
+   * Initialise the database of variable names.
+   *
+   * @param workspace Workspace to generate code from.
+   */
+  init(workspace: Blockly.Workspace) {
```
NeilFraser, line 40: Do we want commented out code here?  Over time the origin of this code will drift away from the commented out code.  I'm not sure it can be trusted in the future.

NeilFraser (approved):

examples/developer-tools/src/output-generators/factory_base.ts
maribethb: yeah, this was a relatively recent change by the formatter. I think they used the same logic for trailing commas elsewhere and applied it to multiline function definitions as well (smaller diffs when adding or removing parameters, etc).

examples/developer-tools/src/output-generators/json_definition_generator.ts
maribethb: No, we don't. I'll remove it all before I merge this branch into develop. I just haven't gotten far enough along to where I know which parts I want to delete and which parts I actually want to uncomment. But I won't do the final merge with this in here

PR #2189, opened 2024-02-02
fix: Include disabled block pattern in screenshot.

### LINKED ISSUE #2157: Disabled blocks transparent in downloaded screenshots
### Check for duplicates

### Component

dev-tools/screenshot

### Description

Downloaded screenshots render disabled blocks as transparent, which makes them hard to see.

Expected result: Downloaded file looking the same, with the same grid-like block background as visible in the browser.

### Reproduction steps

image

-----

Following the steps on the [advanced playground](https://blockly-demo.appspot.com/static/tests/playgrounds/advanced_playground.html#debugEnabled=false) above produces:

-----

screenshot (3)

### Stack trace

_No response_

### Screenshots

_No response_

johnnesky:
## The basics

### Resolves

Fixes #2157

### Proposed Changes

All SVG defs elements inside the workspace SVG are copied to the screenshot SVG. These elements contain, among other things, the pattern definition referenced by disabled blocks.

### Reason for Changes

Previously disabled blocks had an invisible background in screenshots.

### Test Coverage

I manually tested the screenshot feature.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/plugins/dev-tools/src/screenshot.js
+++ b/plugins/dev-tools/src/screenshot.js
@@ -68,12 +68,12 @@ function workspaceToSvg_(workspace, callback, customCss) {
   const height = bBox.height || bBox.bottom - y;
 
   const blockCanvas = workspace.getCanvas();
-  const clone = blockCanvas.cloneNode(true);
-  clone.removeAttribute('transform');
+  const blockCanvasClone = blockCanvas.cloneNode(true);
+  blockCanvasClone.removeAttribute('transform');
+
 
   const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
   svg.setAttribute('xmlns', 'http://www.w3.org/2000/svg');
-  svg.appendChild(clone);
   svg.setAttribute('viewBox', x + ' ' + y + ' ' + width + ' ' + height);
 
   svg.setAttribute(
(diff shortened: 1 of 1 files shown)


plugins/dev-tools/package.json
```diff
@@ -53,10 +53,10 @@
   "devDependencies": {
     "@blockly/dev-scripts": "^3.1.1",
     "@types/dat.gui": "^0.7.5",
-    "blockly": "^10.0.0"
+    "blockly": "^10.2.0"
```
BeksOmega: Why did the minimum version need to be bumped?

johnnesky: In order to access the relatively new accessor method: workspace.getSvgGroup()

BeksOmega: @maribethb does a major bump in dev tools (caused by the bump to the min version of Blockly) mean we have to do cascading major bumps in other plugins?

If so, we can just merge this into the `rc/v11.0.0` branch instead since that's going to require major bumps anyway.

maribethb: Sorry I had thought I replied to this. Yes it does, so let's put this in rc/v11 please. This should be marked as a breaking change due to bumping the minimum version

johnnesky: Thanks, I've added a "!" to the title now.

johnnesky: Okay I've adapted the changes to the rc/v11.0.0 branch and set that as the merge destination. From this branch, the plugin is already depending on a more recent version of blockly, so I don't need to make any changes to that, so I don't think it's a breaking change anymore in this context and I removed the "!" again.

BeksOmega: I believe that's fine. We'll do the actual `!` when we bump everything to v11.

BeksOmega (approved):

PR #2190, opened 2024-02-02
fix: change the block-dynamic-connections plugin to use a connection previewer instead of monkey patching

BeksOmega:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7204

### Proposed Changes + Reasons

Change the dynamic connections plugin to use a connection previewer so that we don't have to patch the insertion marker manager anymore.

### Test Coverage

Manually tested.

I did find one issue with deleting blocks (inputs don't get removed when you delete) which was exposed by the changes to `saveState` where we don't call `finalizeConnections` if the block is in the correct state. But this was only working because of how the advanced playground generates JSON on every workspace change. So I'm going to fix as a follow up.

### Documentation

N/A

### Additional Information

N/A


plugins/block-dynamic-connection/src/connection_previewer.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+import * as Blockly from 'blockly/core';
+
+interface ConnectionPreviewerConstructor {
+  new (draggedBlock: Blockly.BlockSvg): Blockly.IConnectionPreviewer;
+}
+
+interface DynamicBlock extends Blockly.BlockSvg {
+  onPendingConnection(connection: Blockly.Connection): void;
```
rachel-fenichel: Is this a Blockly.RenderedConnection?

plugins/block-dynamic-connection/src/connection_previewer.ts
```diff
@@ -0,0 +1,84 @@
...
+
+function blockIsDynamic(block: Blockly.BlockSvg): block is DynamicBlock {
+  return (
+    (block as any)['onPendingConnection'] !== undefined &&
+    (block as any)['finalizeConnections'] !== undefined
+  );
+}
+
+export function DynamicConnectionPreviewer(
+  // TODO: Make this optional before merging.
```
rachel-fenichel: Fix TODO

plugins/block-dynamic-connection/test/index.ts
```diff
@@ -85,6 +85,11 @@ document.addEventListener('DOMContentLoaded', function () {
 
   const defaultOptions: Blockly.BlocklyOptions = {
     toolbox,
+    plugins: {
+      connectionPreviewer: BlockDynamicConnection.DynamicConnectionPreviewer(
```
rachel-fenichel: Update the `usage` section of the readme to say the dev needs to add this stanza to the `plugins` section of their `options` struct.

BeksOmega:
@rachel-fenichel Ready for re-review =)

plugins/block-dynamic-connection/src/connection_previewer.ts
```diff
@@ -11,28 +11,38 @@ interface ConnectionPreviewerConstructor {
...
-  // TODO: Make this optional before merging.
-  basePreviewerConstructor: ConnectionPreviewerConstructor,
+/**
+ * Returns a connection previewer constructor that decorates the passed
+ * constructor to add connection previewing.
+ *
+ * @param BasePreviewerConstructor The constructor for the base connection
+ *     previewer class being decorated. If not provided, the default
+ *     InsertionMarkerPreviewer will be used.
+ * @return A decofrated connection previewer constructor.
```
rachel-fenichel: nit: decofrated -> decorated

plugins/block-dynamic-connection/src/connection_previewer.ts
```diff
@@ -72,6 +82,8 @@ export function DynamicConnectionPreviewer(
     /**
      * If the block is a dynamic block, calls onPendingConnection and
      * stores the block to be finalized later.
+     *
+     * @param conn The block to rtrigger onPendingConnection on.
```
rachel-fenichel: nit: rtrigger -> trigger

plugins/block-dynamic-connection/src/index.ts
```diff
@@ -13,9 +13,9 @@ import * as Blockly from 'blockly/core';
 import './dynamic_if';
 import './dynamic_text_join';
 import './dynamic_list_create';
-import {DynamicConnectionPreviewer} from './connection_previewer';
+import {decoratePreviewerWithDynamicConnections} from './connection_previewer';
```
rachel-fenichel: Naming question: how about `decoratePreviewer`, since it's already going to be called as `BlockDynamicConnection.decoratePreviewer(SomePreviewer)`?

rachel-fenichel (approved):
I have two nits and a naming question, but it's ready to go once you resolve those.

plugins/block-dynamic-connection/src/index.ts
BeksOmega: Love that good point!

PR #2192, opened 2024-02-05
fix: dynamic blocks finalizing connections when children are deleted

BeksOmega:
## The basics

### Resolves

Fixes N/A

### Proposed Changes + Reasons

While writing https://github.com/google/blockly-samples/pull/2190 I discovered that delete events were not removing connections on dynamic blocks.

Previously this bug was hidden in the playground because we were finalizing connections during serialization, which triggered on every event. So delete events /were/ kind of triggering finalization, in that they were triggering serialization which triggered finalization. But when I made it so that serialization only finalized events if the inputs weren't in escalating order, the bug was revealed

### Test Coverage

Manually tested that deletion triggers connections to be finalized (assuming you've properly added the change listener).

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/plugins/block-dynamic-connection/src/connection_previewer.ts
+++ b/plugins/block-dynamic-connection/src/connection_previewer.ts
@@ -15,6 +15,10 @@ interface DynamicBlock extends Blockly.BlockSvg {
   finalizeConnections(): void;
 }
 
+/**
+ * A type guard that checks if the given block fulfills the DynamicBlock
+ * interface.
+ */
 export function blockIsDynamic(block: Blockly.BlockSvg): block is DynamicBlock {
   return (
     (block as DynamicBlock)['onPendingConnection'] !== undefined &&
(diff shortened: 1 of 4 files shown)


BeksOmega:
I'm going to mark this as a draft. It's been so long since I sent it out for review that I've completely forgotten the context. I'll reopen this closer to when we release v11.

plugins/block-dynamic-connection/src/connection_previewer.ts
```diff
@@ -15,7 +15,7 @@ interface DynamicBlock extends Blockly.BlockSvg {
   finalizeConnections(): void;
 }
 
-function blockIsDynamic(block: Blockly.BlockSvg): block is DynamicBlock {
+export function blockIsDynamic(block: Blockly.BlockSvg): block is DynamicBlock {
```
NeilFraser, line 22: Since this is exported, a JSDoc wouldn't be amiss.

NeilFraser (approved):
Just add the JSDoc

PR #2196, opened 2024-02-09
feat: add javascript definition generators for blocks

maribethb:
## The basics

### Proposed Changes

- Adds JavaScript block definition code generators for block factory blocks
Here's an example of the code generated:

```
const block_type = {
  init: function() {
    // test comment
    this.appendEndRowInput('NAME')
      .appendField('neiou')
      .appendField(new Blockly.FieldLabelSerializable('hi'), 'NAME')
      .appendField(new Blockly.FieldTextInput('default'), 'NAME')
      .appendField(new Blockly.FieldNumber(0), 'NAME');
    this.appendEndRowInput('NAME')
      // dropdown comment
      .appendField(new Blockly.FieldDropdown([['option', 'OPTIONNAME'], ['option', 'OPTIONNAME'], ['option', 'OPTIONNAME'], ]), 'NAME')
      .appendField(new Blockly.FieldCheckbox('TRUE'), 'NAME')
      .appendField(new Blockly.FieldVariable('item'), 'NAME');
    this.setPreviousStatement(true, ['Boolean', 'String']);
    this.setNextStatement(true, null);
    this.setTooltip('tooltip text');
    this.setHelpUrl('');
    this.setColour(160);
  }
};
Blockly.common.defineBlocks({block_type: block_type});
```

### Reason for Changes

Works on block factory.

### Test Coverage

Manual testing for now, I'll add generator tests in a follow up PR because there is currently no test infrastructure for running tests in `examples/`

### Additional Information

- This code relies on an upcoming change in Blockly 10.4, so for now I've been using a locally linked build of blockly. CI doesn't test that examples run so this doesn't affect automated tests. I'll update the code to point at 10.4 when it's released.
- Right now there is no way to switch the block factory to show you the generated js code. That will be coming soon :)

### DIFF
--- a/examples/developer-tools/src/blocks/input.ts
+++ b/examples/developer-tools/src/blocks/input.ts
@@ -106,7 +106,7 @@ export const input = {
     this.setNextStatement(true, 'Input');
     this.setStyle('input');
     this.setTooltip((): string => {
-      const value = this.getFieldValue('INPUT_TYPE');
+      const value = this.getFieldValue('INPUTTYPE');
       return tooltip[value];
     });
     this.setHelpUrl('https://www.youtube.com/watch?v=s2_xaEvcVI0#t=71');
--- a/examples/developer-tools/src/output-generators/connection_check.ts
+++ b/examples/developer-tools/src/output-generators/connection_check.ts
@@ -7,9 +7,39 @@
 import * as Blockly from 'blockly/core';
 import {
   JsonDefinitionGenerator,
-  Order,
+  Order as JsonOrder,
   jsonDefinitionGenerator,
 } from './json_definition_generator';
+import {
+  javascriptDefinitionGenerator,
+  JavascriptDefinitionGenerator,
+} from './javascript_definition_generator';
+import {Order as JsOrder} from 'blockly/javascript';
+
+/**
+ * Gets the selected check from the block.
+ *
+ * @param selected Dropdown option selected.
+ * @param block Connection check block.
+ * @returns A custom option the user has typed in, the selected dropdown option,
+ * or null if `any` or a blank custom option is chosen.
+ */
+function getSelectedCheck(
+  selected: string,
+  block: Blockly.Block,
+): string | null {
+  switch (selected) {
+    case 'null': {
+      return null;
+    }
+    case 'CUSTOM': {
+      return block.getFieldValue('CUSTOMCHECK') || null;
+    }
+    default: {
+      return selected;
+    }
+  }
+}
 
 /**
  * JSON definition for a single "connection_check" block.
@@ -24,26 +54,13 @@ jsonDefinitionGenerator.forBlock['connection_check'] = function (
   generator: JsonDefinitionGenerator,
 ): [string, number] {
   const selected = block.getFieldValue('CHECKDROPDOWN');
-  let output = '';
-  switch (selected) {
-    case 'null': {
-      output = null;
-      break;
-    }
-    case 'CUSTOM': {
-      output = block.getFieldValue('CUSTOMCHECK') || null;
-      break;
-    }
-    default: {
-      output = selected;
-    }
-  }
+  const output = getSelectedCheck(selected, block);
 
-  return [JSON.stringify(output), Order.ATOMIC];
+  return [JSON.stringify(output), JsonOrder.ATOMIC];
 };
 
 /**
- * JSON Definition for the "any of" check block.
+ * JSON definition for the "any of" check block.
  *
  * @param block
  * @param generator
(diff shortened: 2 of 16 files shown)


examples/developer-tools/src/blocks/input.ts
```diff
@@ -106,7 +106,7 @@ export const input = {
     this.setNextStatement(true, 'Input');
     this.setStyle('input');
     this.setTooltip((): string => {
-      const value = this.getFieldValue('INPUT_TYPE');
+      const value = this.getFieldValue('INPUTTYPE');
```
cpcallen, line 109: Just out of interest: why not rename the field to `INPUT_TYPE`?

examples/developer-tools/src/output-generators/connection_check.ts
```diff
@@ -7,9 +7,39 @@
...
+    case 'null': {
+      return null;
+    }
+    case 'CUSTOM': {
+      return block.getFieldValue('CUSTOMCHECK') || null;
+    }
+    default: {
+      return selected;
+    }
+  }
```
cpcallen:
```suggestion
  switch (selected) {
    case 'null':
      return null;
    case 'CUSTOM':
      return block.getFieldValue('CUSTOMCHECK') || null;
    default:
      return selected;
  }
```

examples/developer-tools/src/output-generators/factory_base.ts
```diff
@@ -123,3 +127,75 @@ jsonDefinitionGenerator.forBlock['factory_base'] = function (
...
+
+  // Filter out empty code string pieces, join them all with newlines, and indent them twice.
+  // Inputs code is already indented once automatically by the generator
+  // since it's in a statement input, so it's not included here.
+  const codeStringPieces = generator.prefixLines(
+    [alignInputs, connections, tooltip, helpUrl, colour]
+      .filter((value) => value !== '')
+      .join('\n'),
+    generator.INDENT + generator.INDENT,
+  );
```
cpcallen, line 190: I wonder if this function might be easier to understand if all the thingies (`alignInputs`, `connections`, `tooltip`, etc.) were created as far as possible in a single template literal, which was then passed to `.prefixLines` directly (or after a split / filter / join, if the blank line filtering can't be done earlier).

e.g. something like:

```JS
  const stringCodePieces = generator.prefixLines(
    `this.setInputsInLine(${ /* .. */ });
${connections}
this.setTooltip(${ /* ... */ });
this.setHelpUrl(${ /* ... */ });`.split('\n').filter(/* ... */).join('\n'));
```

examples/developer-tools/src/output-generators/factory_base.ts
```diff
@@ -123,3 +127,75 @@ jsonDefinitionGenerator.forBlock['factory_base'] = function (
...
+  const code = `const ${blockName} = {
+  init: function() {
+${inputs}${codeStringPieces}
+  }
+};
+Blockly.common.defineBlocks({${blockName}: ${blockName}});`;
```
cpcallen, line 196: This can be shorthanded:
```suggestion
Blockly.common.defineBlocks({${blockName}});`;
```
…though it makes me wonder why we don't just do away with `const ${blockName}` and pass a literal directly to `defineBlocks`.

examples/developer-tools/src/output-generators/javascript_definition_generator.ts
```diff
@@ -0,0 +1,52 @@
...
+        if (block.inputList[i].type === Blockly.inputs.inputTypes.VALUE) {
+          const childBlock = block.inputList[i].connection.targetBlock();
+          if (childBlock) {
+            comment = this.allNestedComments(childBlock);
+            if (comment) {
+              commentCode += this.prefixLines(comment, '// ');
+            }
+          }
+        }
+      }
```
cpcallen, line 41: I think this can be rewritten as follows:

```suggestion
      for (let i = 0; i < block.inputList.length; i++) {
        if (block.inputList[i].type !== Blockly.inputs.inputTypes.VALUE) continue;
        const childBlock = block.inputList[i].connection.targetBlock();
        if (!childBlock) continue;
        comment = this.allNestedComments(childBlock);
        if (!comment) continue;
        commentCode += this.prefixLines(comment, '// ');
      }
```

(And indeed a similar tranformation can be applied WRT the enclosing `if (!block.outputConnection…` too.)

cpcallen (approved):
Some suggestions for readability but otherwise LGTM.

examples/developer-tools/src/output-generators/factory_base.ts
maribethb: I don't want to just pass to `defineBlocks` directly because I want to make it easy for people to grab the parts of the definition they need. Maybe they want to just create the block definition and export it, and then call `defineBlocks` in another file. I'm optimizing for copy/paste and modifiability, to an extent.

That's also why I didn't shorthand it, because for example in several places we use the actual block name with underscores, but name the block definition object a name that complies with our lint rules, so you end up with `defineBlocks('block_name': blockName)`. I again wanted to make this easy, even though this one is an easier fix than un-inlining the definition.

examples/developer-tools/src/output-generators/javascript_definition_generator.ts
maribethb: I agree, but I took this directly out of the `scrub_` function in core, and I think it's worth leaving them the same so it's easier to compare and see what changed. But I think I will play with adding the newlines directly to the code instead of in scrub, since I'm going to filter out empty lines later anyway. Then I won't need to override this function at all. So I'll come back to this!

examples/developer-tools/src/blocks/input.ts
maribethb: In a previous PR I renamed it from `INPUT_TYPE` to `INPUTTYPE` which matches how all of the other fields are named (no underscores). I just missed it here and it was breaking the tooltip function so this is just a followup fix. In most cases I took the field names directly from the old block factory, and no underscores was the style there. this particular field didn't exist in the old factory so I named it (badly, at first) myself

Issue #2194, opened 2024-02-06
Cleanup after deleting blocks and fields from core


rachel-fenichel:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Component

Angle, colour, and multiline input fields.

### Description

After #2162 there will be multiple pieces of code in field plugins that compensate for the blocks and fields already being installed as part of the core library. Once the fields are removed form core (https://github.com/google/blockly/issues/7226), the plugins will need to be updated.

Examples include
- In test pages, deleting blocks and generators for blocks provided by the plugin.
- In field registration functions, unregistering existing fields before registering the new ones.

BeksOmega:
Is this release blocking? If so, might be good to bump the priority. I think we might not get to some of the P1s this Q, and I don't want this to get lost.

rachel-fenichel:
No, this isn't release blocking. These are only changes in samples, and the existing code will not break when the fields are deleted from core--it'll just be useless clutter.

rachel-fenichel:
To find the remaining work, search for `TODO(2194)` in the codebase.

PR #2202, opened 2024-02-15
feat!: add block definition to multiline text field plugin

rachel-fenichel:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/2154

### Proposed Changes

- Adds a block definition and generators for one block associated with the multiline input field:
  - `text_multiline`
- Follows the previously established pattern for defining and exporting a library of blocks

BREAKING CHANGE: The multiline text input field no longer registers itself on load. The developer must either manually register the field or install blocks, which will install the field. This is part of a move to have no side effects in field and block definitions, so that tree-shaking can remove unwanted fields and blocks.

### Reason for Changes

This is the next step in extracting block definitions from the core library and publishing them as smaller libraries. Developers can then choose which blocks and fields to install, and can tree-shake out unneeded ones.

### Test Coverage
- Updated mocha tests to call `registerFieldMultilineInput`
- Updated test playground to delete/unregister the old block, generators, and field from core, then install/register the new versions.
- Stripped use of old `generateFieldTestBlocks` helper in favor of manually defining test blocks and the test toolbox.
  -  I found the result easier to read and modify, and it won't change frequently.

The test toolbox looks like this:
image

The first part of the toolbox tests the exported blocks. The later sections test the implementation of the field.

### Documentation

Updated README to describe the block and how to use it, and to note that users must install either the block or the field.

### Additional Information

Open question: this field only exports one block. I still also exported `installAllBlocks`. The only value there is consistency. Should I even bother for a single block? In the example code and README I call `textMultiline.installBlock()`.

As with google/blockly#2162, there is some follow-up work:
- Port generator tests from core, to test generated code against golden files.
  - For the purposes of this PR I tested by loading the block in the playground and checking the generated code in the playground.
- Other cleanup tracked in https://github.com/google/blockly-samples/issues/2194

### DIFF
--- a/plugins/field-multilineinput/README.md
+++ b/plugins/field-multilineinput/README.md
@@ -1,6 +1,6 @@
 # @blockly/field-multilineinput [![Built on Blockly](https://tinyurl.com/built-on-blockly)](https://github.com/google/blockly)
 
-A [Blockly](https://www.npmjs.com/package/blockly) multilineinput field.
+A [Blockly](https://www.npmjs.com/package/blockly) multilineinput field and associated block.
 
 ## Installation
 
@@ -18,6 +18,8 @@ npm install @blockly/field-multilineinput --save
 
 ## Usage
 
+### Field
+
 This field accepts up to 3 parameters:
 
 - "text" to specify the default text. Defaults to `""`.
@@ -28,12 +30,18 @@ This field accepts up to 3 parameters:
 
 The multiline input field is a subclass of Blockly.FieldInput
 
+If you want to use only the field, you must register it with Blockly. You can
+do this by calling `registerFieldMultilineInput` before instantiating your
+blocks. If another field is registered under the same name, this field will
+overwrite it.
+
 ### JavaScript
 
 ```js
 import * as Blockly from 'blockly';
-import {FieldMultilineInput} from '@blockly/field-multilineinput';
+import {registerFieldMultilineInput} from '@blockly/field-multilineinput';
 
+registerFieldMultilineInput();
 Blockly.Blocks['test_field_multilineinput'] = {
   init: function () {
     this.appendDummyInput()
@@ -50,8 +58,9 @@ Blockly.Blocks['test_field_multilineinput'] = {
 
 ```js
 import * as Blockly from 'blockly';
-import '@blockly/field-multilineinput';
+import {registerFieldMultilineInput} from '@blockly/field-multilineinput';
 
+registerFieldMultilineInput();
 Blockly.defineBlocksWithJsonArray([
     {
         "type": "test_field_multilineinput",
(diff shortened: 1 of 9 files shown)


cpcallen (commented):
> Open question: this field only exports one block. I still also exported `installAllBlocks`. The only value there is consistency. Should I even bother for a single block? In the example code and README I call `textMultiline.installBlock()`.

I think it's worth providing an `installAllBlocks` for consistency and in case we add more blocks in future—but it can just be an alias for the oen `installBlock` function for now.

Reciprocal question: I presume you have not attempted to address google/blockly#7209, correct?  I ask because the toolbox screenshot seems to be rendered better than I'd expect given that bug.

plugins/field-multilineinput/README.md
```diff
@@ -65,6 +74,36 @@ Blockly.defineBlocksWithJsonArray([
...
+import {javascriptGenerator} from 'blockly/javascript';
+import {dartGenerator} from 'blockly/dart';
+import {phpGenerator} from 'blockly/php';
+import {pythonGenerator} from 'blockly/python';
+import {luaGenerator} from 'blockly/lua';
+import {textMultiline} from '@blockly/field-multilineinput';
+
+// Installs the block, the field, and all of the language generators.
+textMultiline.installBlock({
+  javascript: javascriptGenerator,
```
cpcallen, line 99: Might it be worth adding an "all items optional" comment here?

plugins/field-multilineinput/src/index.ts
```diff
@@ -4,4 +4,21 @@
...
+/**
+ * Install all of the blocks defined in this file and all of their
+ * dependencies.
+ *
+ * @param generators The CodeGenerators to install per-block
+ *     generators on.
+ */
+export function installAllBlocks(generators: Generators = {}) {
+  textMultiline.installBlock(generators);
+}
```
cpcallen: Could this just an alias?:
```suggestion
export const installAllBlocks = textMultilline.installBlock;
}
```

plugins/field-multilineinput/src/blocks/textMultiline.ts
```diff
@@ -0,0 +1,169 @@
...
+ * @returns A tuple containing the code string and precedence.
+ */
+export function toJavascript(
+  block: Block,
+  generator: JavascriptGenerator,
+): [string, JavascriptOrder] {
+  // Text value.
+  const code = generator.multiline_quote_(block.getFieldValue('TEXT'));
+  const order =
+    code.indexOf('+') !== -1
```
cpcallen, line 70: Comment: this check seems kludgy but appears safe (and I see it's copied from the original).

plugins/field-multilineinput/src/field_multilineinput.ts
```diff
@@ -491,13 +491,18 @@ export class FieldMultilineInput extends Blockly.FieldTextInput {
...
-// this generates a console warning.
-Blockly.fieldRegistry.unregister('field_multilinetext');
-
-Blockly.fieldRegistry.register('field_multilinetext', FieldMultilineInput);
+/**
+ * Register the field and any dependencies.
+ */
+export function registerFieldMultilineInput() {
+  // Unregister legacy field_multilinetext that was in core.  Delete this once
+  // core Blockly no longer defines field_multilinetext.
```
cpcallen: Maybe make this a `TODO(#xxxx):`?

plugins/field-multilineinput/test/index.ts
```diff
@@ -9,30 +9,200 @@
...
+ * TODO(#2194): remove this when those blocks are removed from the core library.
+ */
+function uninstallBlock() {
+  delete Blockly.Blocks['text_multiline'];
+  delete javascriptGenerator.forBlock['text_multiline'];
+  delete dartGenerator.forBlock['text_multiline'];
+  delete luaGenerator.forBlock['text_multiline'];
+  delete pythonGenerator.forBlock['text_multiline'];
+  delete phpGenerator.forBlock['text_multiline'];
+}
```
cpcallen, line 205: Why  wrap this in a function that is called from `createWorkspace`, rather than just executing the contents directly a the top level any time before calling `document.addEventListener`?

cpcallen (approved):
(Duplicate comment body deleted.)

plugins/field-multilineinput/README.md
rachel-fenichel: The text above says "you can supply one or more `CodeGenerator` instances", which i think is sufficient.

plugins/field-multilineinput/src/index.ts
rachel-fenichel: Done.

plugins/field-multilineinput/src/field_multilineinput.ts
rachel-fenichel: Done, good catch.

plugins/field-multilineinput/test/index.ts
rachel-fenichel: I find it easy to lose track of where things that are specific to the plugin test are, so I like putting it all into `createWorkspace`. This is a slight personal preference.

rachel-fenichel:
As discussed offline, Blake has a fix for the multiline field issue in https://github.com/google/blockly-samples/pull/2208

PR #2211, opened 2024-02-20
feat!: Add registration function to field_angle and make it no longer install on file load

rachel-fenichel:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/2153

### Proposed Changes

- Update the angle field plugin to no longer install on file load.
- Instead, export a `registerFieldAngle` function.
- Update tests to guarantee that the correct field is being tested.
- Update usage section in README.
- 
BREAKING CHANGE: The angle field no longer registers itself on load. The developer must manually register the field. This is part of a move to have no side effects in field and block definitions, so that tree-shaking can remove unwanted fields and blocks.

### Reason for Changes

Match the behaviour of the colour and multiline text field plugins.
There are no blocks in the core library that use this field, so no blocks need to be ported. A useful follow-up would be to add the angle blocks that are used in Blockly Games.

### Test Coverage

Tested in the playground and in tests.
Confirmed that registration is actually happening by verifying failures if `registerFieldAngle` unregisters the old field but doesn't reregister the new field.

### Documentation

README updated in this PR.

### DIFF
--- a/plugins/field-angle/src/field_angle.ts
+++ b/plugins/field-angle/src/field_angle.ts
@@ -658,9 +658,7 @@ export class FieldAngle extends Blockly.FieldNumber {
   }
 }
 
-/**
- * Register the field and any dependencies.
- */
+/** Register the field and any dependencies. */
 export function registerFieldAngle() {
   // Unregister legacy field_angle that was in core.
   // TODO(#2194): Delete this once core Blockly no longer defines field_angle.
--- a/plugins/field-angle/src/field_angle.ts
+++ b/plugins/field-angle/src/field_angle.ts
@@ -658,12 +658,17 @@ export class FieldAngle extends Blockly.FieldNumber {
   }
 }
 
-// Unregister legacy field_angle that was in core.  Delete this once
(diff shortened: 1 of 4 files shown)


plugins/field-angle/src/field_angle.ts
```diff
@@ -658,12 +658,17 @@ export class FieldAngle extends Blockly.FieldNumber {
   }
 }
 
-// Unregister legacy field_angle that was in core.  Delete this once
-// core Blockly no longer defines field_angle.
-// If field_angle is not defined in core, this generates a console warning.
-Blockly.fieldRegistry.unregister('field_angle');
+/**
+ * Register the field and any dependencies.
+ */
```
cpcallen: If formatter + linter allow, consider:

```suggestion
/** Register the field and any dependencies. */
```

cpcallen (approved):

plugins/field-angle/src/field_angle.ts
rachel-fenichel: Done.

Issue #2198, opened 2024-02-09
Clean up code style in generators for field colour blocks


rachel-fenichel:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Request

Follow-up work from #2162.

Clean up the generator code (which is returned as a string) for blocks in field-colour.

For example, Christopher suggested this change in the Python code for the blend block:
 ```
-    functionName + '(' + colour1 + ', ' + colour2 + ', ' + ratio + ')';
+  const code = `${functionName}(${colour1}, ${colour2}, ${ratio})`;
```

PR #2220, opened 2024-02-23
feat: add tests for per-block generators in field-colour

rachel-fenichel:
## The basics

### Resolves

Pre-work for #2198, by making it possible to test that changes to generators are safe.

### Proposed Changes

Add generator tests to field colour.

The test:
- Uninstall old blocks and generators.
- Install new blocks and generators.
- Load a specific set of blocks (provided as JSON).
- Run each language generator in turn.
- Compare the generated code to a golden file for each language.
  - I created the golden file by running the old generators in the playground.

### Reason for Changes

Verify that generators are the same after moving them from core to samples. This is primarily change detection, and will help with later cleanup of generator functions (#2198).

### Test Coverage

Added tests in `plugins/field-colour/test/blocks-test.mocha

### Documentation

None


.prettierignore
```diff
@@ -21,3 +21,5 @@ PULL_REQUEST_TEMPLATE.md
 plugins/dev-create/templates/sample-app/*
 plugins/dev-create/templates/sample-app-ts/*
 
+# Golden test files
+plugins/**/golden/*
```
maribethb: i guess prettier doesn't format its own ignore file. could you add the missing newline here

plugins/field-colour/test/blocks_test.mocha.js
```diff
@@ -0,0 +1,272 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
```
maribethb: 2024

plugins/field-colour/test/blocks_test.mocha.js
```diff
@@ -0,0 +1,272 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import fs from 'fs';
+import * as Blockly from 'blockly';
```
maribethb: I think you actually want `from 'blockly/core'` here (sorry, i know i was dictating while you typed this) and then possibly also `import 'blockly/blocks'`

this is a relatively minor thing but it's what we recommend everyone use now so it helps with consistency and it also keeps your webpack bundles smaller

plugins/field-colour/test/blocks_test.mocha.js
```diff
@@ -0,0 +1,272 @@
...
+        },
+      },
+    ],
+  },
+};
+
+/**
+ * Uninstall old blocks and generators, since they come pre-installed
+ * with Blockly.
+ * TODO(2194): Delete this function and calls to it.
```
maribethb: nit: if you include the `#` here then tools are smart enough to link to it

plugins/field-colour/test/golden/golden.dart
```diff
@@ -0,0 +1,65 @@
...
+  colour_random();
+
+  colour_rgb(100, 50, 0);
+
+  colour_rgb(0, 1, 20);
+
+  colour_blend('#ff0000', '#3333ff', 0.5);
+
+  colour_blend('#000000', '#ffffff', 0.3);
+}
```
maribethb, line 65: manually add newline since this isn't formaTed

plugins/field-colour/test/blocks_test.mocha.js
```diff
@@ -0,0 +1,272 @@
...
+ * @param {string} suffix The suffix of the golden file.
+ * @param {string} generated The generated code to compare against the
+ *     golden file.
+ */
+function checkResult(suffix, generated) {
+  const fileName = `test/golden/golden.${suffix}`;
+  const goldenContents = fs.readFileSync(fileName);
+  // Normalize the line feeds.
+  const normalized = goldenContents.toString().replace(/(?:\r\n|\r|\n)/g, '\n');
+  assert.equal(generated, normalized);
```
maribethb, line 210: what does it look like if this fails? I think that's my only concern is that debugging this will be tricky, you'll get something like `Expected <giant string> to equal <other giant string>` which isn't super useful, but i'm not sure what i would suggest to fix that except maybe printing the results into a file that one could manually inspect later if needed. I don't think you should block on this, but if you have any ideas, that's where I think this could be improved.

plugins/field-colour/test/blocks_test.mocha.js
```diff
@@ -0,0 +1,272 @@
...
+import {assert} from 'chai';
+
+const blockJson = {
+  blocks: {
+    languageVersion: 0,
+    blocks: [
+      {
+        type: 'colour_picker',
+        id: '+g[iA}db.}w:CQV#k-p@',
+        x: 13,
```
maribethb, line 25: nit: sometimes in tests we recommend removing extraneous info from the serialization that isn't needed like IDs and coordinates, just to make things easier to read/maintain

maribethb (approved):

.prettierignore
rachel-fenichel: Done.

plugins/field-colour/test/blocks_test.mocha.js
rachel-fenichel: Done

plugins/field-colour/test/blocks_test.mocha.js
rachel-fenichel: Done.

plugins/field-colour/test/blocks_test.mocha.js
rachel-fenichel: I removed IDs but kept coordinates because it sets the order of the blocks, which determines the order of the generated code.

plugins/field-colour/test/blocks_test.mocha.js
rachel-fenichel: Done.

plugins/field-colour/test/blocks_test.mocha.js
rachel-fenichel: It actually gives a diff of the outputs, showing only the lines that are different + a few lines around them for context. I just ran into it while handling a change in block order and it was easy to see how to debug it.

plugins/field-colour/test/golden/golden.dart
rachel-fenichel: Ignored because adding a newline causes a test failure (newlines are part of the diff).

PR #2004, opened 2023-10-04
Add indexing of dropdown entries

koenvanwijk:
## The basics

- [x] I [validated my changes](https://developers.google.com/blockly/guides/contribute/samples#making_and_verifying_a_change)

### Resolves

Fixes #1940

### Proposed Changes
The search should find the block and select the dropdown item. 
image

### Reason for Changes
The search fails to find blocks with dropdown.

### Test Coverage
The manual test show the blocks are found and that the option is selected. 

### Documentation
No update of the documentation is required. Although an remark could be made.

PR #2222, opened 2024-02-26
fix: When using the toolbox-search, auto select the found dropdown item

### LINKED ISSUE #1940: Update toolbox-search to search dropdown field options

constantinIliescu:
Fixes: https://github.com/google/blockly-samples/issues/1940

Uses the Block State as suggested here:
https://github.com/google/blockly-samples/pull/2004#issuecomment-1832317907 

Fixes: bug that didn't indexed the last trigram from a word

## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/1940
### Proposed Changes

Use block state for efficient indexing

### Reason for Changes

Have the found drop down item automatically selected in the drop down

### Test Coverage

Unit test have been added to cover the false positive, false negative and duplicate cases

### DIFF
--- a/plugins/toolbox-search/CHANGELOG.md
+++ b/plugins/toolbox-search/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [1.2.7](https://github.com/google/blockly-samples/compare/@blockly/toolbox-search@1.2.6...@blockly/toolbox-search@1.2.7) (2024-02-23)
+### Bug Fixes
+ * Improved search for drop down
+ * Improved unit tests
+ * Last trigram from a word was never indexed
+
 ## [1.2.6](https://github.com/google/blockly-samples/compare/@blockly/toolbox-search@1.2.5...@blockly/toolbox-search@1.2.6) (2024-02-08)
 
 **Note:** Version bump only for package @blockly/toolbox-search
--- a/plugins/toolbox-search/src/block_searcher.ts
+++ b/plugins/toolbox-search/src/block_searcher.ts
@@ -27,13 +27,16 @@ export class BlockSearcher {
     const blockCreationWorkspace = new Blockly.Workspace();
     blockTypes.forEach((blockType) => {
       const block = blockCreationWorkspace.newBlock(blockType);
-      this.indexBlockText(blockType.replaceAll('_', ' '), blockType);
-      block.inputList.forEach((input) => {
-        input.fieldRow.forEach((field) => {
-          this.indexDropdownOption(field, blockType);
-          this.indexBlockText(field.getText(), blockType);
+      const blockState = Blockly.serialization.blocks.save(block);
+      if (blockState != null) {
+        this.indexBlockText(blockType.replaceAll('_', ' '), blockState);
+        block.inputList.forEach((input) => {
+          input.fieldRow.forEach((field) => {
+            this.indexDropdownOption(field, blockState);
+            this.indexBlockText(field.getText(), blockState);
+          });
         });
-      });
+      }
     });
   }
 
@@ -41,15 +44,27 @@ export class BlockSearcher {
    * Check if the field is a dropdown, and index every text in the option
    *
    * @param field We need to check the type of field
-   * @param blockType The block type to associate the trigrams with.
+   * @param blockState The block state to associate the trigrams with.
    */
-  private indexDropdownOption(field: Blockly.Field, blockType: string) {
+  private indexDropdownOption(
+    field: Blockly.Field,
+    blockState: Blockly.serialization.blocks.State,
+  ) {
     if (field instanceof Blockly.FieldDropdown) {
       field.getOptions(true).forEach((option) => {
+        const state = {...blockState};
+        state.fields = {...blockState.fields};
         if (typeof option[0] === 'string') {
-          this.indexBlockText(option[0], blockType);
+          if (state.fields == undefined) {
+            state.fields = {};
+          }
+          if (field.name) {
+            state.fields[field.name] = option[1];
+          }
+          this.indexBlockText(option[0], state);
+          this.indexBlockText(option[1], state);
         } else if ('alt' in option[0]) {
-          this.indexBlockText(option[0].alt, blockType);
+          this.indexBlockText(option[0].alt, state);
         }
       });
     }
@@ -59,10 +74,10 @@ export class BlockSearcher {
    * Filters the available blocks based on the current query string.
    *
    * @param query The text to use to match blocks against.
-   * @returns A list of block types matching the query.
+   * @returns A list of block states matching the query.
    */
-  blockTypesMatching(query: string): string[] {
-    return [
+  blockTypesMatching(query: string): Blockly.serialization.blocks.State[] {
+    const result = [
       ...this.generateTrigrams(query)
         .map((trigram) => {
           return this.trigramsToBlocks.get(trigram) ?? new Set<string>();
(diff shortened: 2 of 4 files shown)


plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -72,20 +87,33 @@ export class BlockSearcher {
...
    */
-  private indexBlockText(text: string, blockType: string) {
+  private indexBlockText(
+    text: string,
+    blockState: Blockly.serialization.blocks.State,
+  ) {
+    blockState.id = undefined;
+    blockState.extraState = undefined;
+    blockState.data = undefined;
+    const stateString = JSON.stringify(blockState);
```
BeksOmega, line 114: Is there a reason you need to stringify this? I think you should be able to add it to a `Set<Blockly.serialization.blocks.State>`. If the stringification is for deduplication or something, could you add a comment inline?

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -101,7 +129,7 @@ export class BlockSearcher {
     if (normalizedInput.length <= 3) return [normalizedInput];
 
     const trigrams: string[] = [];
-    for (let start = 0; start < normalizedInput.length - 3; start++) {
+    for (let start = 0; start <= normalizedInput.length - 3; start++) {
```
BeksOmega, line 132: Thanks for the catch!

plugins/toolbox-search/src/toolbox_search.ts
```diff
@@ -171,13 +172,19 @@ export class ToolboxSearchCategory extends Blockly.ToolboxCategory {
...
+            return {
+              kind: 'block',
+              type: blockState.type,
+              fields: blockState.fields,
+            };
+          } else
+            return {
+              kind: 'block',
+              type: blockState.type,
+            };
```
BeksOmega, line 187:
```suggestion
          return {kind: 'block', ...blockState};
```

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -27,29 +27,44 @@ export class BlockSearcher {
...
+      if (blockState != null) {
+        this.indexBlockText(blockType.replaceAll('_', ' '), blockState);
+        block.inputList.forEach((input) => {
+          input.fieldRow.forEach((field) => {
+            this.indexDropdownOption(field, blockState);
+            this.indexBlockText(field.getText(), blockState);
+          });
         });
-      });
+      }
```
BeksOmega, line 39: Instead of wrapping this could you use an early return?

```
if (!blockState) return;
this.indexBlockText // etc...
```

This function already has a lot of nesting and I'd prefer to avoid adding more if we can :P

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -27,29 +27,44 @@ export class BlockSearcher {
...
+  ) {
     if (field instanceof Blockly.FieldDropdown) {
       field.getOptions(true).forEach((option) => {
+        const state = {...blockState};
+        state.fields = {...blockState.fields};
         if (typeof option[0] === 'string') {
-          this.indexBlockText(option[0], blockType);
+          if (state.fields == undefined) {
+            state.fields = {};
+          }
```
BeksOmega, line 60: I think this case should be impossible since you assing to `state.fields` above. Was the typescript compiler giving you warnings about this?

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -27,29 +27,44 @@ export class BlockSearcher {
...
+          if (field.name) {
+            state.fields[field.name] = option[1];
+          }
+          this.indexBlockText(option[0], state);
+          this.indexBlockText(option[1], state);
```
BeksOmega, line 65: I don't think we should be indexing the language-neutral text.

BeksOmega (changes requested):
Generally LGTM! I'll give it a more thorough review after the first round of changes have been fixed.

plugins/toolbox-search/src/block_searcher.ts
constantinIliescu: Yes, the reason was duplication, without stringify we would have ended with a lot of duplication. I will put a comment in the code

plugins/toolbox-search/src/block_searcher.ts
constantinIliescu: Well, I added this on some blocks the language-neutral text was more informative than the block text. I don't remember now a clear example, but I take a look. BTW, sorry for replaying so late, but lately I am little busy with an internal project. I would try to find some time to fix your comments.

plugins/toolbox-search/src/block_searcher.ts
constantinIliescu: Agree, I will fix it

plugins/toolbox-search/src/block_searcher.ts
constantinIliescu: I think it was the compiler complaining. I don't exactly remember

plugins/toolbox-search/src/block_searcher.ts
BeksOmega: No rush! It took us three weeks to get back to you so we can't expect you to be more responsive :P

plugins/toolbox-search/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [1.2.7](https://github.com/google/blockly-samples/compare/@blockly/toolbox-search@1.2.6...@blockly/toolbox-search@1.2.7) (2024-02-23)
```
BeksOmega, line 6: Can you revert this? The CHANGELOG doesn't need to be edited manually (it gets updated automatically based on submitted PRs)

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -27,29 +27,44 @@ export class BlockSearcher {
...
+    field: Blockly.Field,
+    blockState: Blockly.serialization.blocks.State,
+  ) {
     if (field instanceof Blockly.FieldDropdown) {
       field.getOptions(true).forEach((option) => {
+        const state = {...blockState};
+        state.fields = {...blockState.fields};
         if (typeof option[0] === 'string') {
-          this.indexBlockText(option[0], blockType);
+          if (state.fields == undefined) {
```
BeksOmega, line 58: Throughout, can you replace `==` with `===` and `!=` with `!==`?

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -27,29 +27,44 @@ export class BlockSearcher {
...
    * @param field We need to check the type of field
-   * @param blockType The block type to associate the trigrams with.
+   * @param blockState The block state to associate the trigrams with.
    */
-  private indexDropdownOption(field: Blockly.Field, blockType: string) {
+  private indexDropdownOption(
+    field: Blockly.Field,
+    blockState: Blockly.serialization.blocks.State,
+  ) {
     if (field instanceof Blockly.FieldDropdown) {
```
BeksOmega, line 53: Can you switch this to use an early return as well? (not your code but I think it would make it more readable).

plugins/toolbox-search/src/block_searcher.ts
BeksOmega, line 1: Can you add a jsdoc comment to the `trigramsToBlocks` map that specifies that the `Set<string>` contains stringified JSON?

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -72,20 +87,33 @@ export class BlockSearcher {
         })
         .values(),
     ];
+    const resultState = result.map((item) => JSON.parse(item));
+    return resultState;
```
BeksOmega, line 91: You should be able to set this up to just `return` from the top. But I can't put a suggestion over this whole function b/c github :P
```suggestion
    ].map(JSON.parse);
```

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -72,20 +87,33 @@ export class BlockSearcher {
...
+   * @param blockState The block state to associate the trigrams with.
    */
-  private indexBlockText(text: string, blockType: string) {
+  private indexBlockText(
+    text: string,
+    blockState: Blockly.serialization.blocks.State,
+  ) {
+    blockState.id = undefined;
+    blockState.extraState = undefined;
+    blockState.data = undefined;
```
BeksOmega, line 113: Why are these being removed? It seems like they should probably be kept.

plugins/toolbox-search/src/block_searcher.ts
```diff
@@ -59,10 +74,10 @@ export class BlockSearcher {
    * Filters the available blocks based on the current query string.
    *
    * @param query The text to use to match blocks against.
-   * @returns A list of block types matching the query.
+   * @returns A list of block states matching the query.
    */
-  blockTypesMatching(query: string): string[] {
-    return [
+  blockTypesMatching(query: string): Blockly.serialization.blocks.State[] {
```
BeksOmega, line 79: Because this is public, changing the return type of this method is actually a breaking change. Could you create another method `blockStatesMatching` that has the current behavior here? And then modify this to only return the block /types/ not the full state. You'll need to update other places to use your new `blockStatesMatching` method.

BeksOmega (changes requested):
Added some more comments for when you have time to take a look!

BeksOmega:
@constantinIliescu are you still interested in working on this? If not I'll close this on Friday!

constantinIliescu:
Hi Beka,
Yes, I intend to finish this , but I still need some more time. I been caught in more urgent things lately.

BeksOmega:
> Hi Beka, Yes, I intend to finish this , but I still need some more time. I been caught in more urgent things lately.

All good no rush! Just wanted to check in in case you wanted to hand it over to someone else =)

PR #2227, opened 2024-02-27
chore: Clean up and convert field-bitmap to Typescript

gonfunko:
## The basics

### Resolves
Part of #1647

### Proposed Changes
This PR converts field-bitmap to Typescript. It also includes some light refactoring and cleanup of the code.


plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -0,0 +1,595 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Bitmap input field.
```
maribethb: This fileoverview isn't adding anything and we no longer include author annotations so I would remove this comment

plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -0,0 +1,595 @@
...
+ * @author gregoryc@outlook.com (Greg Cannon)
+ */
+
+import Blockly from 'blockly/core';
+
+export const DEFAULT_HEIGHT = 5;
```
maribethb, line 9: note on the lint errors: rachel has an update to the eslint config that allows constant case. it's bundled in another pr. I will poke her about submitting that when she is feeling better

plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -0,0 +1,595 @@
...
+      boundingBox.right,
+    );
+  }
+
+  /**
+   * Creates the bitmap editor and add event listeners.
+   *
+   * @returns The newly created dropdown menu.
+   */
+  private dropdownCreate_() {
```
maribethb: do you also want to rename this one to remove the underscore?

maribethb (approved):

plugins/field-bitmap/src/field-bitmap.ts
gonfunko: Done!

plugins/field-bitmap/src/field-bitmap.ts
gonfunko: Good catch, thanks! Renamed accordingly.

plugins/field-bitmap/src/field-bitmap.ts
gonfunko: Ah nice, and sounds good!

PR #2232, opened 2024-03-05
feat: add generator tests for the text_multiline block

rachel-fenichel:
## The basics

### Resolves
Part of #2154

### Proposed Changes

Clean up imports to import from `blockly/core` wherever possible, to reduce unnecessary imports and provide example code for writing plugins.

Add generator tests to field multiline input plugin.

The test:
- Uninstall old blocks and generators.
- Install new blocks and generators.
- Load a specific set of blocks (provided as JSON).
- Run each language generator in turn.
- Compare the generated code to a golden file for each language.
  - I created the golden file by running the old generators in the playground.

### Reason for Changes

Verify that generators are the same after moving them from core to samples. This is primarily change detection in case of later cleanup.

### Test Coverage

Added tests in `plugins/field-multilineinput/test/blocks-test.mocha`

### DIFF
--- a/plugins/field-multilineinput/src/blocks/textMultiline.ts
+++ b/plugins/field-multilineinput/src/blocks/textMultiline.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {Block, common as BlocklyCommon} from 'blockly';
+import {Block, common as BlocklyCommon} from 'blockly/core';
 import {
   JavascriptGenerator,
   Order as JavascriptOrder,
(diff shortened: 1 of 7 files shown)


NeilFraser (commented):
Just a question about the golden files, why are the line feeds as separate strings?  No problem, just unusual.

Much as I adore Jabberwocky, it might be a good idea to choose a test string which triggers escaping with both types of quotes.  Something like:

Picard said, "Beam me up!".
O'Brien made it so.

rachel-fenichel:
> Just a question about the golden files, why are the line feeds as separate strings? No problem, just unusual.

Not sure, but the golden files are generated from the fields currently in core Blockly, so it's consistent.

I updated the test string as you suggested, and updated goldens accordingly.

NeilFraser (approved):

PR #2236, opened 2024-03-08
fix: Align zoom-to-fit event handling with zoom_controls

elvisvoer:
## The basics

### Resolves

The zoom-to-fit plugin is not stopping the event propagation which results in 2 events being fired when clicking the reset zoom button: one with targetType `zoom_reset_control` and one with targetType `workspace`. The core zoom_controls implementation does however stop the event propagation.

### Proposed Changes

This PR aligns the event handling in zoom-to-fit plugin with the core zoom_controls implementation.
It does 2 things:
 - switches the event listener to `pointerdown`
 - adds `e.stopPropagation();` and `e.preventDefault();` to the event handler.

### Reason for Changes

The change aligns the behavior of zoom related controls making sure that users listening to workspace events get the same behavior across all zoom related actions.

### DIFF
--- a/plugins/zoom-to-fit/src/index.ts
+++ b/plugins/zoom-to-fit/src/index.ts
@@ -119,7 +119,7 @@ export class ZoomToFitControl implements Blockly.IPositionable {
     // Attach listener.
     this.onZoomToFitWrapper = Blockly.browserEvents.conditionalBind(
       this.svgGroup,
-      'mousedown',
+      'pointerdown',
       null,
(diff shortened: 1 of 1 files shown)


NeilFraser:
Verified that the old behaviour popped the context menu up on right-click, as well as performing the zoom-to-fit.  Whereas the new behaviour does not summon the context menu, with the zoom-to-fit executing as before.  This new behaviour is consistent with the zoom in/out/center buttons.

Looks good, thanks!

PR #2241, opened 2024-03-15
fix: remove references to colour fields and blocks

rachel-fenichel:
## The basics

### Resolves
Part of #2194

### Proposed Changes

Cleans up references to colour blocks and fields in other plugins and examples.

Updates the mutator test blocks to use text instead of colour.

### Reason for Changes

I removed these fields and blocks from core and don't want to have to cross-link plugins to get those fields and blocks back.

### Test Coverage

Mutator test blocks updated, and I opened every category in the blocks-test toolbox to confirm they all still worked + interacted with mutator test blocks.


plugins/dev-tools/src/toolboxCategories.js
```diff
@@ -771,84 +771,6 @@ export default {
         },
       ],
     },
-    {
-      kind: 'category',
```
BeksOmega, line 775: Oof this is making me realize I probably need to go an update some devsite screenshots... blehhh

examples/devsite-landing-demo/toolbox.js
```diff
@@ -649,71 +649,6 @@ let toolboxJson = {
       ],
     },
 
-    {
-      // Colour Category
-      kind: 'CATEGORY',
```
BeksOmega, line 654: We don't save any of the state of the devsite landing demo right?

plugins/block-test/src/mutators.js
```diff
@@ -50,23 +50,23 @@ const COLOR_CHANGE_MUTATOR = {
    * @this {Blockly.Block}
    */
   domToMutation: function (xmlElement) {
-    this.setColour(xmlElement.getAttribute('colour'));
+    this.setFieldValue('LABEL', xmlElement.getAttribute('arbitrary'));
```
BeksOmega: I think these params are reversed, should be value then name. https://developers.google.com/blockly/reference/js/blockly.block_class.setfieldvalue_1_method.md

BeksOmega (approved):
One fix, otherwise just Qs

Are we going to run into problems with the code demo in core not having these blocks anymore? I can't remember if we still update that demo or not.

Similarly for the demos site. I think that's all stored locally, so when we go to update those, anyone who dragged a color block into a demo is going to break. Maybe we should clear their local state if serialization errors instead of them having to clear their local state themselves? Otherwise I think we might get a bunch of bug reports about this.

examples/devsite-landing-demo/toolbox.js
rachel-fenichel: Correct

plugins/block-test/src/mutators.js
rachel-fenichel: Good catch. I went back to the test page and for each mutator test block I dragged into the workspace, opened the mutator, modified the block, closed the mutator, and then copy/pasted the block to confirm it was correct.

rachel-fenichel:
I have an issue tracking verifying the storage demo won't cause problems: 
https://github.com/google/blockly/issues/7826

Updating the code demo is now tracked in https://github.com/google/blockly/issues/7938

PR #2268, opened 2024-03-22
fix: build for the v11 branch

BeksOmega:
Random broken things, some of which were caused by the merge from master into v11, others of which were caused by bumping to the 11.0.0-beta.6


plugins/block-dynamic-connection/test/dynamic_if.mocha.js
```diff
@@ -6,11 +6,15 @@
 
 const chai = require('chai');
 const {testHelpers} = require('@blockly/dev-tools');
-const Blockly = require('blockly/node');
+const Blockly = require('blockly/core');
```
maribethb, line 9: Huh. This should have been done in #2229 and I'm surprised that not doing so didn't fail that build. You can probably just change it to 'blockly' and not add the rest of the lines importing blocks and en separately.

maribethb (approved):

plugins/block-dynamic-connection/test/dynamic_if.mocha.js
BeksOmega: That didn't fail because all of the plugins were still using a version of Blockly that didn't have the exports. Christopher was just linking every plugin locally instead of testing against a beta. In the future, I don't think we should test that way. It's too error prone (e.g. this + he lost a week of work to linking incorrectly).

I'm going to continue importing each thing individually, since that's how we [recommend people load the code](https://developers.google.com/blockly/guides/get-started/get-the-code#load_the_code).

PR #2271, opened 2024-03-25
fix: dynamic blocks erroring on workspace dispose

BeksOmega:
## The basics

### Resolves

Related to  https://github.com/google/blockly/issues/7864

### Proposed Changes

Makes it so that the dynamic connection blocks check whether they are being disposed before reconstructing the block during serialization.

### Reason for Changes

This is only necessary because of a long string of cleanup we're having to do because we originally gave the dynamic connection blocks a different serialization format than the blocks in core. The blocks had different input names in core vs in the plugin, so we had to normalize them. The only valid place to do that was during serialization. But normalizing them there caused errors if the workspace is being disposed.

Hopefully this is the last cleanup we need to do from that original bug.

### Test Coverage

Tested that you can dispose of the workspace while it has dynamic blocks in it by switching the theme in the advanced playground. (Any switching of options causes the workspace to be disposed and ~reinfected~ reinjected).

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/plugins/block-dynamic-connection/src/dynamic_if.ts
+++ b/plugins/block-dynamic-connection/src/dynamic_if.ts
@@ -69,12 +69,14 @@ const DYNAMIC_IF_MIXIN = {
    * @returns XML storage element.
    */
   mutationToDom(this: DynamicIfBlock): Element | null {
-    // If we call finalizeConnections here without disabling events, we get into
-    // an event loop.
-    Blockly.Events.disable();
-    this.finalizeConnections();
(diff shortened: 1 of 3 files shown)


rachel-fenichel:
>  (Any switching of options causes the workspace to be disposed and reinfected.

Please no reinfection.

BeksOmega:
> > (Any switching of options causes the workspace to be disposed and reinfected.
> 
> Please no reinfection.

Lol `reinjected` actually gets autocorrected to that.

maribethb (approved):

PR #2294, opened 2024-03-28
chore: Remove colour and multiline from devsite demo

NeilFraser:
Also update JS-Interpreter to latest version.

No idea why there are two devsite demos, so I just updated both.


BeksOmega (approved):
LGTM! @maribethb do you have context around why we have the two demos? Do we need the non-landing one anymore?

maribethb:
`devsite-demo` is the old one. It's not being used on devsite. It's not being used anywhere else to my knowledge.

PR #2295, opened 2024-03-29
feat: add generator stubs to block factory

### LINKED ISSUE #2288: Block factory: add generator stubs

maribethb:
## The basics

### Resolves

Fixes #2288

### Proposed Changes

- Adds generator stubs for each of the 5 blockly languages
- Adds the import or script headers for generators to the code headers section

The generated code is largely what is found in the old tool but it fixes several problems where the old tool:
- only showed you script-style `javascript.javascriptGenerator` which is incorrect if you're using imports
- used `Blockly.javascript.ORDER_NONE` in some places which is incorrect as of v10
- used `var` instead of const
- used `generator.nameDB_` instead of `generator.getVariableName`
- did not have newlines between input groups, which I found hard to read
- only had TODOs around fixing the `Order` value in one of the places `Order` is used

### Additional Information

Note to reviewer: if I manually added quotes around where I'm calling `generator.getFieldValue` such as 

```
`const code = blah('${generator.getFieldValue('FIELDNAME')}');`
```

 that is an error. I should be using `generator.quote_` to add quotes instead (only around field values specifically). I think I fixed them all but would appreciate you double-checking :)

### DIFF
--- a/examples/developer-tools/src/output-generators/factory_base.ts
+++ b/examples/developer-tools/src/output-generators/factory_base.ts
@@ -20,6 +20,10 @@ import {
   importHeaderGenerator,
   scriptHeaderGenerator,
 } from './code_header_generator';
+import {
+  GeneratorStubGenerator,
+  generatorStubGenerator,
+} from './generator_stub_generator';
 
 /**
  * Builds the 'message0' part of the JSON block definition.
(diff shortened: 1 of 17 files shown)


examples/developer-tools/src/output-generators/factory_base.ts
```diff
@@ -218,9 +226,35 @@ scriptHeaderGenerator.forBlock['factory_base'] = function (
...
+generatorStubGenerator.forBlock['factory_base'] = function (
+  block: Blockly.Block,
+  generator: GeneratorStubGenerator,
+): string {
+  const lang = generator.getLanguage();
+  const blockName = block.getFieldValue('NAME');
+  const inputs = generator.statementToCode(block, 'INPUTS');
+  const hasOutput = !!block.getInput('OUTPUTCHECK');
+  const returnIfOutput = `// TODO: Change Order.NONE to the correct operator precedence strength
+  return [code, ${generator.getScriptMode() ? lang + '.' : ''}Order.NONE];`;
```
rachel-fenichel: Can you pull `generator.getScriptMode() ? lang + '.' : ''` out into `const orderPrefix`? I find it hard to parse when it's smushed inside the `${}`

rachel-fenichel: You also use it on line 253.

rachel-fenichel (approved):
One code style comment, but LGTM after that.

Reading this is a bit mind-bending, even having recently written generators.

PR #2296, opened 2024-03-29
feat!: update blockly version to 10.4.3 for colour and multilineinput

rachel-fenichel:
## The basics

### Proposed Changes

Use Blockly 10.4.3 for field-colour and field-multilineinput instead of using a beta.

### Reason for Changes

Get ready to release the changes on the `blocks_for_fields` branch.


plugins/field-colour/package.json
```diff
@@ -43,7 +43,7 @@
     "@blockly/dev-scripts": "^3.1.1",
     "@blockly/dev-tools": "^7.1.6",
     "@typescript-eslint/parser": "^5.59.5",
-    "blockly": "^11.0.0-beta.3",
+    "blockly": "^10.4.3",
```
maribethb, line 46: You also need to update the `peerDependencies` for both of these

rachel-fenichel: Done.

maribethb (approved):
The peer dependencies aren't updated in package-lock.json so you also need to run `npm install` at root of samples so that those get updated too. After that, lgtm

PR #2297, opened 2024-03-29
feat: save block factory settings

maribethb:
## The basics

### Resolves

Fixes 

### Proposed Changes

- Saves the state of the various block factory output settings between page loads

### Reason for Changes

I was annoyed by clicking the same options every time the page refreshed

### DIFF
--- a/examples/developer-tools/src/storage.ts
+++ b/examples/developer-tools/src/storage.ts
@@ -4,12 +4,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+const settingsKey = 'blockFactorySettings';
 const lastEditedBlockKey = 'blockFactoryLastEditedBlock';
 const allBlocksKey = 'blockFactoryAllBlocks';
 const allBlocks: Set<string> = new Set(
   JSON.parse(window.localStorage?.getItem(allBlocksKey)) || [],
 );
 const prohibitedBlockNames = new Set([
+  settingsKey,
   lastEditedBlockKey,
   allBlocksKey,
   'blockly_block_factory_preview_block',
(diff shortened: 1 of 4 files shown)


examples/developer-tools/src/storage.ts
```diff
@@ -113,3 +115,22 @@ export const getAllSavedBlockNames = function (): Set<string> {
...
   return prohibitedBlockNames;
 };
+
+export interface BlockFactorySettings {
+  blockDefinitionFormat: string;
+  codeHeaderStyle: string;
+  codeGeneratorLanguage: string;
+}
+
+export const saveBlockFactorySettings = function (
```
BeksOmega:
```suggestion
export function saveBlockFactorySettings (
```
https://google.github.io/styleguide/tsguide.html#function-declarations

examples/developer-tools/src/storage.ts
```diff
@@ -113,3 +115,22 @@ export const getAllSavedBlockNames = function (): Set<string> {
...
+  settings: BlockFactorySettings,
+) {
+  localStorage.setItem(settingsKey, JSON.stringify(settings));
+};
+
+export const loadBlockFactorySettings =
+  function (): BlockFactorySettings | null {
```
BeksOmega:
```suggestion
export function loadBlockFactorySettings(): BlockFactorySettings | null {
```

examples/developer-tools/src/storage.ts
```diff
@@ -4,12 +4,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+const settingsKey = 'blockFactorySettings';
```
BeksOmega, line 7: Is this a constant? https://google.github.io/styleguide/tsguide.html#identifiers-constants

examples/developer-tools/src/storage.ts
```diff
@@ -113,3 +115,22 @@ export const getAllSavedBlockNames = function (): Set<string> {
 export const getProhibitedBlockNames = function (): Set<string> {
   return prohibitedBlockNames;
 };
+
+export interface BlockFactorySettings {
+  blockDefinitionFormat: string;
```
BeksOmega, line 120: You could make these enums to catch some buggies. But this code is simple enough I don't think it's really necessary.

BeksOmega (approved):
A few style nits. Logic LGTM!

examples/developer-tools/src/storage.ts
maribethb: Yeah, but our samples style guide prohibits CONSTANT_CASE haha.

I think Aaron was looking at fixing this recently so I'll check in on that, but will leave this for now

examples/developer-tools/src/storage.ts
maribethb: I started to, but since I get the values from the dom, it added a lot of boilerplate and some casting to convert the raw string out of the dom into the enum. It didn't feel worth the overhead but someone can link to this comment in the future when they discover a bug here :P

examples/developer-tools/src/storage.ts
maribethb: Fixed em all here and elsewhere.  Thanks!

PR #2304, opened 2024-04-03
feat: add ability to convert old block factory json to new

maribethb:
## The basics

### Resolves

Works on #2291 

### Proposed Changes

- Adds functions that can convert the JSON of a `factory_base` block from the old block factory on app engine into JSON that can be loaded into the new tool, given the changes to block definitions that were made

### Reason for Changes

- Several changes were made to the block definitions when converting them from the old tool to the new one. This includes using one type of input block with a dropdown to select value vs statement, one connection check block with a dropdown to specify the type, etc. Several fields and inputs were renamed as well.
- This means you can't just straight up load JSON from the old tool into the new one.

There are two options for fixing this:

#### Manual JSON editing
Pros: Relatively straightforward to edit JSON (as opposed to XML); pretty obvious exactly what changed between versions
Cons: Mildly complicated to explain what's happening (but hopefully I left good comments)

#### Alternate block definitions
meaning define blocks that match the old ones, and just not include them in the toolbox. e.g. define a `input_value` block.
Pros: This is the approach we recommend people use if they make breaking changes to their blocks.
Cons: The blocks have the same name in both tools, so I'd either have to rename some of them or manually edit the json anyway to change their names; I'd have to copy & paste the generators for each of the changed blocks and *very slightly* adjust them to deal with the different field names, etc. so it'd be difficult to tell what changed without diffing the generators (and there are multiple generators for each block); users importing old blocks will be forced to continue using the old blocks and can't use the new ones which are a bit easier to use

Both approaches have cons, but creating alternate block definitions had more cons so I went with the JSON approach. Not as much needed to be adjusted as I feared.

### Test Coverage

Made several very complicated blocks in the old tool that exercised every feature of block creation, then made sure the json could be loaded into the new tool and the block looked the exact same.

### Additional Information

This is only the actual converter code in this PR. I'm still working on the file upload part of the UI to allow you to actually easily import json, so I'll do that in a follow-up PR.

### DIFF
--- a/examples/developer-tools/src/backwards_compatibility.ts
+++ b/examples/developer-tools/src/backwards_compatibility.ts
@@ -1,0 +1,259 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * This file contains methods that can convert blocks saved in the old
+ * block factory tool that was hosted on app engine, into blocks that
+ * can be used with this new tool.
+ *
+ * Many of the blocks from the old tool are the same as the blocks
+ * in this tool. But in some cases, block, field, or input names have
+ * been changed for clarity. This file will edit block json so that
+ * the saved data from the old tool can be loaded into this tool.
+ */
+
+import * as Blockly from 'blockly/core';
+
+const connectionCheckShadow = {
+  type: 'connection_check',
+  fields: {
+    CHECKDROPDOWN: 'null',
+  },
+};
+
+/**
+ * The factory_base block is largely the same. However, the inputs that spawn
+ * if the block has top/bottom/left connectors have been renamed from
+ * `OUTPUTTYPE` to `OUTPUTCHECK`. The input blocks connected to this one
+ * also need to be converted.
+ *
+ * @param oldBlock The JSON for the factory_base as saved from old tool.
+ * @returns JSON that should be loaded instead.
+ */
+export function convertBaseBlock(oldBlock: any): object {
+  const newBlock = {...oldBlock};
+  // extraState from the old tool isn't relevant.
+  delete newBlock.extraState;
+
+  if (oldBlock.inputs?.INPUTS?.block) {
+    newBlock.inputs.INPUTS.block = convertInput(oldBlock.inputs.INPUTS.block);
+  }
+
+  if (oldBlock.inputs?.OUTPUTTYPE) {
+    newBlock.inputs.OUTPUTCHECK = {};
+    newBlock.inputs.OUTPUTCHECK.shadow = connectionCheckShadow;
+    if (oldBlock.inputs.OUTPUTTYPE.block) {
+      newBlock.inputs.OUTPUTCHECK.block = convertCheck(
+        oldBlock.inputs.OUTPUTTYPE.block,
+      );
+    }
+    delete newBlock.inputs.OUTPUTTYPE;
+  }
+  if (oldBlock.inputs?.TOPTYPE) {
+    newBlock.inputs.TOPCHECK = {};
+    newBlock.inputs.TOPCHECK.shadow = connectionCheckShadow;
+    if (oldBlock.inputs.TOPTYPE.block) {
+      newBlock.inputs.TOPCHECK.block = convertCheck(
+        oldBlock.inputs.TOPTYPE.block,
+      );
+    }
+    delete newBlock.inputs.TOPTYPE;
+  }
+  if (oldBlock.inputs?.BOTTOMTYPE) {
+    newBlock.inputs.BOTTOMCHECK = {};
+    newBlock.inputs.BOTTOMCHECK.shadow = connectionCheckShadow;
+    if (oldBlock.inputs.BOTTOMTYPE.block) {
+      newBlock.inputs.BOTTOMCHECK.block = convertCheck(
+        oldBlock.inputs.BOTTOMTYPE.block,
+      );
+    }
+    delete newBlock.inputs.BOTTOMTYPE;
+  }
+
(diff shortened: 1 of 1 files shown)


examples/developer-tools/src/backwards_compatibility.ts
```diff
@@ -0,0 +1,262 @@
...
+/**
+ * The factory_base block is largely the same. However, the inputs that spawn
+ * if the block has top/bottom/left connectors have been renamed from
+ * `OUTPUTTYPE` to `OUTPUTCHECK`. The input blocks connected to this one
+ * also need to be converted.
+ *
+ * @param oldBlock The JSON for the factory_base as saved from old tool.
+ * @returns JSON that should be loaded instead.
+ */
+export function convertBaseBlock(oldBlock: any): object {
```
BeksOmega: You can use the `Blockly.serialization.blocks.State` type here and throughout.

If you wanted, you could also define types for the specific blocks' state, which would ensure properties are getting properly created and deleted where necessary.

examples/developer-tools/src/backwards_compatibility.ts
```diff
@@ -0,0 +1,262 @@
...
+  if (oldBlock.inputs?.OUTPUTTYPE) {
+    newBlock.inputs.OUTPUTCHECK = {};
+    newBlock.inputs.OUTPUTCHECK.shadow = CONNECTION_CHECK_SHADOW;
+    if (oldBlock.inputs.OUTPUTTYPE.block) {
+      newBlock.inputs.OUTPUTCHECK.block = convertCheck(
+        oldBlock.inputs.OUTPUTTYPE.block,
+      );
+    }
+    delete newBlock.inputs.OUTPUTTYPE;
+  }
```
BeksOmega: Could you extract this into a subprocedure? It seems really easy to swap your TYPE vs CHECK inputs and create the wrong JSON.

examples/developer-tools/src/backwards_compatibility.ts
```diff
@@ -0,0 +1,262 @@
...
+  if (oldBlock.fields?.INPUTNAME) {
+    newBlock.fields.INPUTNAME = oldBlock.fields.INPUTNAME;
+  }
+
+  if (oldBlock.inputs?.TYPE) {
+    newBlock.inputs = {
+      CHECK: {
+        shadow: CONNECTION_CHECK_SHADOW,
+      },
+    };
```
BeksOmega: I think this should be defined regardless of whether a `TYPE` input exists.

examples/developer-tools/src/backwards_compatibility.ts
```diff
@@ -0,0 +1,262 @@
...
+    if (oldBlock.inputs.TYPE.block) {
+      newBlock.inputs.CHECK.block = convertCheck(oldBlock.inputs.TYPE.block);
+    }
+  }
+
+  if (oldBlock.inputs?.FIELDS?.block) {
+    if (!newBlock.inputs) newBlock.inputs = {};
```
BeksOmega: Maybe just add this to the original definition at the top?

examples/developer-tools/src/backwards_compatibility.ts
```diff
@@ -0,0 +1,262 @@
...
+ * The field blocks are all mostly the same, with a few exceptions:
+ * "field_static" in the old tool is called "field_label" here.
+ * "field_dropdown"'s extraState uses xml in the old tool and json in the new tool.
+ *
+ * TODO(#2290): Check for backwards-compatibility issues with plugin fields.
+ *
+ * @param oldBlock JSON for the "field_foo" block as saved from old tool.
+ * @returns JSON that should be used for the replacement field block.
+ */
+function convertField(oldBlock: any): object {
```
BeksOmega: Is this missing an `if(!oldBlock)` check? Would probably be helpful if this was at least `object | null` instead of `any`.

examples/developer-tools/src/backwards_compatibility.ts
```diff
@@ -0,0 +1,262 @@
...
+  if (!oldBlock) return {};
+  const oldName = oldBlock.type as string; // The block type i.e. name of block definition
+  if (!oldName.startsWith('type_')) {
+    throw Error(
+      `Found connection check block with unexpected block type ${oldName}`,
+    );
+  }
+  let connectionCheck = oldName.substring(5);
+  switch (connectionCheck) {
+    case 'null':
```
BeksOmega, line 188: So `null` continues to be lower case?

BeksOmega (approved):
LGTM! Mostly just type nits.

examples/developer-tools/src/backwards_compatibility.ts
maribethb: That would create data for "CHECK" inputs on the dummy and end-row blocks, which don't have connection checks because they don't have connections. I'm not sure if the serializer would throw an error if it reads data for an input that doesn't exist on the block, but even if it doesn't, we don't need it.

examples/developer-tools/src/backwards_compatibility.ts
maribethb: No, before calling this method I was always making sure there is a field block. But I was also doing that for most of the other methods too, so I also removed it from them so I don't have to accept/return null.

examples/developer-tools/src/backwards_compatibility.ts
maribethb: Correct, and I added a comment so you can tell it's intentional (the reason the others are capitalized is the dropdown value matches what blockly uses internally for those types; for `null` connection checks, it's just set to `null` in the json block definition and not `'null'` but we check for that exception)

BeksOmega (approved):

PR #2308, opened 2024-04-04
fix: remove code of removed colour toolbox

mobyw:
## The basics

### Resolves

Fixes error loading caused by applying changes in #2294.

### Proposed Changes

Remove JavasSript and CSS codes of the removed colour toolbox.

### DIFF
--- a/examples/devsite-demo/script.js
+++ b/examples/devsite-demo/script.js
@@ -71,10 +71,9 @@ function init() {
   toolboxJson['contents'][2].name = getMsg('Math');
   toolboxJson['contents'][3].name = getMsg('Text');
   toolboxJson['contents'][4].name = getMsg('Lists');
-  toolboxJson['contents'][5].name = getMsg('Colour');
   // Separator.
-  toolboxJson['contents'][7].name = getMsg('Variables');
-  toolboxJson['contents'][8].name = getMsg('Procedures');
+  toolboxJson['contents'][6].name = getMsg('Variables');
+  toolboxJson['contents'][7].name = getMsg('Procedures');
 
   // Inject default variable name.
   // https://github.com/google/blockly/issues/5238
--- a/examples/devsite-landing-demo/style.css
+++ b/examples/devsite-landing-demo/style.css
@@ -95,15 +95,11 @@ option {
   border-left: 15px solid #4db6ac !important;
 }
 
-#blockly-5 {
-  border-left: 15px solid #ffcdd2 !important;
-}
-
-#blockly-7 {
+#blockly-6 {
   border-left: 15px solid #ef9a9a !important;
 }
 
-#blockly-8 {
+#blockly-7 {
   border-left: 15px solid #d7ccc8 !important;
 }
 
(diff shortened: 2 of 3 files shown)


examples/devsite-demo/script.js
```diff
@@ -71,10 +71,9 @@ function init() {
   toolboxJson['contents'][2].name = getMsg('Math');
   toolboxJson['contents'][3].name = getMsg('Text');
   toolboxJson['contents'][4].name = getMsg('Lists');
-  toolboxJson['contents'][5].name = getMsg('Colour');
   // Separator.
-  toolboxJson['contents'][7].name = getMsg('Variables');
```
BeksOmega, line 76: @maribethb do you know why we're doing this instead of using Blockly's normal localization?

examples/devsite-landing-demo/style.css
```diff
@@ -95,15 +95,11 @@ option {
...
-
-#blockly-7 {
+#blockly-6 {
   border-left: 15px solid #ef9a9a !important;
 }
 
-#blockly-8 {
+#blockly-7 {
   border-left: 15px solid #d7ccc8 !important;
 }
```
BeksOmega, line 104: I don't think these CSS IDs are technically supported. We should be using a [custom toolbox](https://blocklycodelabs.dev/codelabs/custom-toolbox/index.html?index=..%2F..index#2), or [category CSS](https://developers.google.com/blockly/guides/configure/web/toolbox#category_css).

BeksOmega (approved):
I don't think this demo is using Blockly quite correctly, but it's fine to merge this to unblock other work.

examples/devsite-landing-demo/style.css
mobyw: I tried using category css and it works.

Should I update on this PR or open another one when this is done?

BeksOmega: You can make updates on this PR! Thanks for testing that out :D

examples/devsite-demo/script.js
maribethb: Toolboxes don't really have support for dynamic messages afaik. I could see a couple different ways of solving this but I'm not sure which method you're referring to as "normal."

If you want history though, you might have to check with Neil. This is the old demo that's been on devsite for forever and I'm not sure where it came from before we moved it to samples.

BeksOmega: They support [replacing message references in category names](https://github.com/BeksOmega/blockly/blob/df758b3c306b729b95161285f5f6886872d7c0fb/core/toolbox/category.ts#L177). Although I guess we don't document this anywhere?

@NeilFraser do you have context for this?

maribethb: TIL, I thought you had to do this manually. I thought one of the old demos was doing something similar but I can't find it.

examples/devsite-landing-demo/style.css
maribethb: Hey @mobyw since this fixes the issue with the live demo we'll actually go ahead and merge it now so we can make sure the demo is running. If you're still up for fixing the css to use the category CSS, that would be great and we'd love a follow up PR! Thanks!

PR #2309, opened 2024-04-04
fix(migration): Ignore non-renamed modules

### LINKED ISSUE #2171: Migration script includes 'Blockly' -> 'Blockly' rename, causing the script to exit early
### Check for duplicates

### Component

@blockly/migration

### Description

Something is causing the migration script to include a rename of `Blockly` -> `Blockly`. This causes any renames after that rename to be skipped, because the loop exits early [here](https://github.com/google/blockly-samples/blob/f89bb8b69c4d4917c482d2df06131949f8c7ef71/plugins/migration/bin/rename.js#L215). So some of the renamings in the database aren't getting run.

### Reproduction steps

1. Include `Blockly.Bubble` in the manual test file.
2. Run the migration script locally (as documented in the manual test file)
3. Observe how `Blockly.Bubble` doesn't get renamed to `Blockly.bubbles.Bubble` even though it exists in the database.

### Stack trace

_No response_

### Screenshots

_No response_

cpcallen:
## The basics

### Resolves

Fixes #2171.

### Proposed Changes

Don't create a renaming entry for modules that haven't been renamed.

Add  an entry to the test renamings that replicates the situation describe in #2171, and which causes multiple test failures without the aforementioned fix.

### Additional Information

We should _probably_ do a prefix sort on `.renamings_` (or even just sort by length descending), so that more specific entries precede less-specific ones, but after thinking quite hard about the possible consequences of this I wasn't 100% sure that this wouldn't have unforeseen consequences due to the (arguably excessive) complexity of how renamings are specified.

Since the renaming file (and therefore script) is due for a rethink anyway—since it is keyed on `goog.module` IDs that no longer actually exist in the Blockly sourcecode—I think doing the simplest possible thing for now is probably the best approach.

### DIFF
--- a/plugins/migration/bin/rename.js
+++ b/plugins/migration/bin/rename.js
@@ -192,8 +192,11 @@ class VersionRenamer {
       //   moduleA.exportA -> moduleC.exportB
       // And we performed the module rename first, we wouldn't be able to detect
       // the export rename.
-      this.renamings_.push({old: oldModulePath, new: newModulePath});
+      if (newModulePath !== oldModulePath) {
+        this.renamings_.push({old: oldModulePath, new: newModulePath});
+      }
     }
+    console.log(this.renamings_);
   }
 
   /**
--- a/plugins/migration/bin/rename.js
+++ b/plugins/migration/bin/rename.js
@@ -196,7 +196,6 @@ class VersionRenamer {
         this.renamings_.push({old: oldModulePath, new: newModulePath});
       }
     }
-    console.log(this.renamings_);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


plugins/migration/bin/rename.js
```diff
@@ -192,8 +192,11 @@ class VersionRenamer {
       //   moduleA.exportA -> moduleC.exportB
       // And we performed the module rename first, we wouldn't be able to detect
       // the export rename.
-      this.renamings_.push({old: oldModulePath, new: newModulePath});
+      if (newModulePath !== oldModulePath) {
+        this.renamings_.push({old: oldModulePath, new: newModulePath});
+      }
     }
+    console.log(this.renamings_);
```
maribethb: remove log statement

maribethb (approved):
also needs formatting, then lgtm

PR #2314, opened 2024-04-05
feat: move backpack to use interface for backpackability

### LINKED ISSUE #2311: Make it so that the backpack is compatible with other types besides blocks
### Check for duplicates

### Component

workspace-backpack

### Problem

Currently the backpack plugin is [hard coded](https://github.com/google/blockly-samples/blob/1f586d3200f3f120489d8efa77218b3b29906c2a/plugins/workspace-backpack/src/backpack.ts#L456) to only accept blocks into the backpack. This was fine back when the only objects that could be dragged and dropped were blocks. But now that we’ve [generalized the drag system](https://docs.google.com/document/d/1EdKxQafwO_j1QmoTQrhfE2fwPLXAy5mm0yapPjQ1Oeo/edit?usp=sharing) so that anything that fulfills the `ISelectable` and `IDraggable` interfaces can be dragged, it is too restrictive.

For example, this makes it impossible for the multiselect plugin to be compatible with the backpack plugin.

### Request

Add an interface to the backpack plugin that it can use to check if an IDraggable should be added to the backpack.

```
interface Backpackable {
  toFlyoutData(): FlyoutItemInfo[];
}
```

### Alternatives considered

N/A

### Additional context

This will also require the `BlockSvg` in core to implement `toFlyoutData`.

BeksOmega:
## The basics

### Resolves

Fixes #2311 

### Proposed Changes

Makes it so the backpack uses an interfaces for adding things to itself instead of hard coding to only work with blocks.

### Test Coverage

Manually tested that blocks still get added to the backpack.

### Documentation

Should be documented somewhere?? Would love opinions! Worst case I'll add it to the dragger docs?

### Additional Information

N/A

### DIFF
--- a/plugins/workspace-backpack/README.md
+++ b/plugins/workspace-backpack/README.md
@@ -153,6 +153,11 @@ Blockly.Msg['EMPTY_BACKPACK'] = 'Opróżnij plecak'; // Polish
 - `addBlock`: Adds Block to backpack.
 - `addBlocks`: Adds Blocks to backpack.
 - `removeBlock`: Removes Block to backpack.
+- `containsBlock`: Returns whether the Block is in the backpack.
+- `addBackpackable`: Adds Backpackable to backpack.
+- `addBackpackables`: Adds Backpackable to backpack.
+- `removeBackpackable`: Removes Backpackable to backpack.
+- `containsBackpackable`: Returns whether the Backpackable is in the backpack.
 - `addItem`: Adds item to backpack.
 - `removeItem`: Removes item from the backpack.
 - `setContents`: Sets backpack contents.
@@ -201,6 +206,17 @@ The Backpack Flyout uses the registered Flyout for either
 for `Blockly.Trashcan`. If a custom class is registered for either of these
 types, then the Backpack Flyout may need to be tested for compatibility.
 
+### Draggables
+
+If you have a custom draggable object, you can make it possible to add it to
+the backpack by implementing the [`Backpackable`][backpackable] interface.
+
+This interface requires the draggable to have a method that converts it into
+[`FlyoutItemInfo`][flyout-info]. As of 2024-04-08 flyouts only support displaying blocks, buttons, and labels.
+
 ## License
 
 Apache 2.0
+
+[flyout-info]: https://developers.google.com/blockly/reference/js/blockly.utils_namespace.toolbox_namespace.flyoutiteminfo_typealias.md
+[backpackable]: https://github.com/google/blockly-samples/blob/master/plugins/workspace-backpack/src/backpack.ts
--- a/plugins/workspace-backpack/README.md
+++ b/plugins/workspace-backpack/README.md
@@ -212,7 +212,7 @@ If you have a custom draggable object, you can make it possible to add it to
 the backpack by implementing the [`Backpackable`][backpackable] interface.
 
 This interface requires the draggable to have a method that converts it into
-[`FlyoutItemInfo`][flyout-info]. As of 2024-04-08 flyouts only support displaying blocks, buttons, and labels.
+[`FlyoutItemInfo`][flyout-info]. As of Blockly core v11 flyouts only support displaying blocks, buttons, and labels.
 
 ## License
 
--- a/plugins/workspace-backpack/src/backpack.ts
+++ b/plugins/workspace-backpack/src/backpack.ts
@@ -526,7 +526,11 @@ export class Backpack
    *     provided block.
    */
   containsBlock(block: Blockly.Block): boolean {
-    return isBackpackable(block) && this.containsBackpackable(block);
+    if (isBackpackable(block)) {
+      return this.containsBackpackable(block);
+    } else {
+      return this.contents_.indexOf(this.blockToJsonString(block)) !== -1;
+    }
   }
 
   /**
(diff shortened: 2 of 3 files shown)


plugins/workspace-backpack/src/backpackable.ts
```diff
@@ -0,0 +1,24 @@
...
+   * it can be displayed in the backpack.
+   *
+   * This method should remove any unique or useless state data (e.g. IDs or
+   * coordinates).
+   */
+  toFlyoutData(): Blockly.utils.toolbox.FlyoutItemInfo[];
+}
+
+/** Checks whether the given object conforms to the Backpackable interface. */
+export function isBackpackable(obj: any): obj is Backpackable {
```
maribethb, line 26: unknown might work here

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -553,7 +556,39 @@ export class Backpack
    * @param block The block to be removed from the backpack.
    */
   removeBlock(block: Blockly.Block) {
-    this.removeItem(this.blockToJsonString(block));
+    if (isBackpackable(block)) this.removeBackpackable(block);
+  }
+
+  /**
+   * Returns whether the backpack contains a duplicate of the provided
```
maribethb: please add the jsdoc. you can just add the parameter, i'm pretty sure we don't require it to have a description, but it being in jsdoc is important (this will matter for generated reference docs which will happen eventually)

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -534,7 +535,7 @@ export class Backpack
    * @param block The block to be added to the backpack.
    */
   addBlock(block: Blockly.Block) {
```
maribethb, line 541: should these methods be updated to take a `Backpackable` object instead of block? I wouldn't change the name, just the type of param and jsdoc

maribethb (approved):
I would document this in the readme for the plugin, you can add a section about making custom objects compatible with the backpack

plugins/workspace-backpack/src/backpack.ts
BeksOmega: Ah, we have it configured exactly the opposite way in core.

plugins/workspace-backpack/src/backpack.ts
BeksOmega: The `Blockly.Block` doesn't implement `Backpackable`, only `Blockly.BlockSvg` does, because backpacks don't exist in headless mode. So changing this to take in a `Backpackable` would be a breaking change.

Which actually means I implemented this incorrectly, whoops. If it's not a `Backpackable` we should use the old codepath.

plugins/workspace-backpack/src/backpack.ts
maribethb: oh weird, we should update it in core. It needs to be in the jsdoc to show up in the generated reference doc I'm pretty sure. will file an issue for that

BeksOmega: I think I've already filed an issue about it. So you might want to check. [Edit: like about the generated reference docs being kinda bad cuz they're missing info]

plugins/workspace-backpack/src/backpackable.ts
BeksOmega: Unknown means you can't do anything with the value until you've asserted what type the value is. Importantly for this case, you can't dereference properties on them. So for type guards you have to use `any`.

https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type

BeksOmega:
I also renamed `toFlyoutData` to `toFlyoutInfo` to match the actual type returned. This means that the plugin won't work with the .7 beta, but the plugin doesn't have tests so this should still be mergable.

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -553,7 +568,47 @@ export class Backpack
...
-    this.removeItem(this.blockToJsonString(block));
+    if (isBackpackable(block)) {
+      this.removeBackpackable(block);
+    } else {
+      this.removeItem(this.blockToJsonString(block));
+    }
+  }
+
+  /**
+   * @param backpackable The backpackable we want to check for existance within
```
maribethb: nit: s/existance/existence

plugins/workspace-backpack/src/backpack.ts
```diff
@@ -553,7 +568,47 @@ export class Backpack
...
+   *     the backpack.
+   * @return whether the backpack contains a duplicate of the provided
+   *     backpackable.
+   */
+  containsBackpackable(backpackable: Backpackable) {
```
maribethb: should this function say it returns boolean? it doesn't have a return type rn

maribethb (approved):

plugins/workspace-backpack/README.md
```diff
@@ -201,6 +206,17 @@ The Backpack Flyout uses the registered Flyout for either
 for `Blockly.Trashcan`. If a custom class is registered for either of these
 types, then the Backpack Flyout may need to be tested for compatibility.
 
+### Draggables
+
+If you have a custom draggable object, you can make it possible to add it to
+the backpack by implementing the [`Backpackable`][backpackable] interface.
+
+This interface requires the draggable to have a method that converts it into
+[`FlyoutItemInfo`][flyout-info]. As of 2024-04-08 flyouts only support displaying blocks, buttons, and labels.
```
maribethb: Use the version number of Blockly instead of the date.

maribethb (approved):

PR #2316, opened 2024-04-08
fix!: update the scroll options to work with v11 dragging

BeksOmega:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/2317

### Proposed Changes

Makes it so the scroll-options plugin works with the new dragging system.

### Test Coverage

Tested manually with changes in core. Dragging workspace comments to the edge of the workspace properly scrolls yay!

### Documentation

N/A

### Additional Information

Dependent on https://github.com/google/blockly/pull/8016, but we have no tests so we can still merge this.

## Breaking changes / To Fix

The `ScrollBlockDragger` now conforms to the new `IDraggable` interfaces in v11 of core Blockly.

When constructing an `AutoScroll` you now have to pass the `ScrollBlockDragger` as well.

### DIFF
--- a/plugins/scroll-options/src/AutoScroll.js
+++ b/plugins/scroll-options/src/AutoScroll.js
@@ -7,7 +7,7 @@
 import * as Blockly from 'blockly/core';
 import {ScrollMetricsManager} from './ScrollMetricsManager';
 import {getTranslation} from './utils';
-import {ScrollDragger} from './ScrollDragger';
+import {ScrollBlockDragger} from './ScrollBlockDragger';
 
 /**
  * AutoScroll is used to scroll/pan the workspace automatically. For example,
@@ -24,7 +24,7 @@ export class AutoScroll {
   /**
    * Creates an AutoScroll instance for a specified workspace.
    * @param {!Blockly.WorkspaceSvg} workspace Workspace to scroll.
-   * @param {!ScrollDragger} dragger The dragger that's currently dragging.
+   * @param {!ScrollBlockDragger} dragger The dragger that's currently dragging.
    * @constructor
    */
   constructor(workspace, dragger) {
--- a/plugins/scroll-options/src/AutoScroll.js
+++ b/plugins/scroll-options/src/AutoScroll.js
@@ -7,6 +7,7 @@
 import * as Blockly from 'blockly/core';
 import {ScrollMetricsManager} from './ScrollMetricsManager';
 import {getTranslation} from './utils';
+import {ScrollDragger} from './ScrollDragger';
 
 /**
  * AutoScroll is used to scroll/pan the workspace automatically. For example,
@@ -23,14 +24,18 @@ export class AutoScroll {
   /**
    * Creates an AutoScroll instance for a specified workspace.
    * @param {!Blockly.WorkspaceSvg} workspace Workspace to scroll.
+   * @param {!ScrollDragger} dragger The dragger that's currently dragging.
    * @constructor
    */
-  constructor(workspace) {
+  constructor(workspace, dragger) {
     /**
      * Workspace to scroll.
      * @protected {!Blockly.WorkspaceSvg}
      */
     this.workspace_ = workspace;
+
+    /** @private {!ScrollDragger} */
+    this.dragger = dragger;
 
     /**
      * Current active vector representing scroll velocity in pixels per
@@ -137,9 +142,7 @@ export class AutoScroll {
     // The dragger will update its values so that things like connection
     // markers will stay consistent.
     if (deltaX || deltaY) {
-      this.workspace_.currentGesture_
-        .getCurrentDragger()
-        .moveBlockWhileDragging(deltaX, deltaY);
+      this.dragger.moveBlockWhileDragging(deltaX, deltaY);
     }
   }
 
--- a/plugins/scroll-options/src/AutoScroll.js
+++ b/plugins/scroll-options/src/AutoScroll.js
@@ -34,7 +34,7 @@ export class AutoScroll {
      */
     this.workspace_ = workspace;
 
-    /** @private {!ScrollDragger} */
+    /** @private {!ScrollBlockDragger} */
     this.dragger = dragger;
 
     /**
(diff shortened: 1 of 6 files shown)


plugins/scroll-options/src/ScrollDragger.js
```diff
@@ -111,82 +111,48 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
...
-      this.scrollDelta_,
-      currentDragDeltaXY,
-    );
-    super.drag(e, totalDelta);
-    this.dragDelta_ = currentDragDeltaXY;
+    super.onDrag(e, totalDelta);
+    this.dragDelta_ = dragDelta;
 
-    if (ScrollBlockDragger.edgeScrollEnabled) {
+    console.log('here', ScrollDragger.edgeScrollEnabled);
```
maribethb: Remove extra log statements

plugins/scroll-options/src/ScrollDragger.js
```diff
@@ -250,7 +214,8 @@ export class ScrollBlockDragger extends Blockly.BlockDragger {
 
     // Update the autoscroll or start a new one.
     this.activeAutoScroll_ =
-      this.activeAutoScroll_ || new AutoScroll(this.workspace_);
+      this.activeAutoScroll_ || new AutoScroll(this.workspace, this);
+    console.log('autoscoll');
```
maribethb: and here

maribethb (approved):
LGTM, but can you update the PR description with a breaking change section? This is only a breaking change if you're subclassing elements of this plugin and if so then the class has been renamed. If you're simply installing and configuring this plugin, you don't need to take any action [edited].

BeksOmega:
> LGTM, but can you update the PR description with a breaking change section? This is only a breaking change if you're subclassing elements of this plugin and if so then the class has been renamed. If you're simply installing and configuring this plugin, you don't need to take any action [edited].

I think it's a breaking change either way because of the renaming. To pass it to the configuration structure you need to get it by name. Do you want me to not rename it?

maribethb:
> > LGTM, but can you update the PR description with a breaking change section? This is only a breaking change if you're subclassing elements of this plugin and if so then the class has been renamed. If you're simply installing and configuring this plugin, you don't need to take any action [edited].
> 
> I think it's a breaking change either way because of the renaming. To pass it to the configuration structure you need to get it by name. Do you want me to not rename it?

Hmm, yeah. I think the value gained in renaming it is lower than the annoyance of making developers rename their plugin struct. The rename is more accurate but there's little value in that to the plugin user imo. I'd probably keep the name and have a comment explaining it now can drag more than blocks.

I think renames that affect people subclassing these classes or doing more advanced work are ok. But we should try not to break people who are just using the plugin as-is unless we have good reason. Open to debate on this point though :)

BeksOmega:
> > > LGTM, but can you update the PR description with a breaking change section? This is only a breaking change if you're subclassing elements of this plugin and if so then the class has been renamed. If you're simply installing and configuring this plugin, you don't need to take any action [edited].
> > 
> > 
> > I think it's a breaking change either way because of the renaming. To pass it to the configuration structure you need to get it by name. Do you want me to not rename it?
> 
> Hmm, yeah. I think the value gained in renaming it is lower than the annoyance of making developers rename their plugin struct. The rename is more accurate but there's little value in that to the plugin user imo. I'd probably keep the name and have a comment explaining it now can drag more than blocks.
> 
> I think renames that affect people subclassing these classes or doing more advanced work are ok. But we should try not to break people who are just using the plugin as-is unless we have good reason. Open to debate on this point though :)

Ehh as a team, we have this same argument every time we want to rename something. And then it's like "do we want to create an alias?" "well aliases are also annoying" etc etc. And it goes on forever. I'd rather just get this merged.

maribethb:
Yes, but in the past we've also decided not to do renames that don't add a lot of value to the end-user. Most recently I recall were some changes in generators where we decided not to do renames and not to add aliases. So I don't think this decision is inconsistent.

BeksOmega:
> Yes, but in the past we've also decided not to do renames that don't add a lot of value to the end-user. Most recently I recall were some changes in generators where we decided not to do renames and not to add aliases. So I don't think this decision is inconsistent.

Oh to clarify I'm going to make the change you suggested! I was just saying I don't need to discuss it more and was just going to go with whatever you said haha.

BeksOmega:
@maribethb Can you give this another quick look since I made the naming changes?

plugins/scroll-options/src/ScrollBlockDragger.js
```diff
@@ -68,10 +68,10 @@ const defaultOptions = {
  * A block dragger that adds the functionality for a block to be moved while
  * someone is dragging it.
  */
-export class ScrollBlockDragger extends Blockly.BlockDragger {
+export class ScrollBlockDragger extends Blockly.Dragger {
```
maribethb, line 71: should this be BlockDragger again?

plugins/scroll-options/src/index.js
```diff
@@ -141,7 +141,7 @@ export class ScrollOptions {
     if (
       !canWheelMove ||
       !currentGesture ||
-      !(currentGesture.getCurrentDragger() instanceof Blockly.BlockDragger)
+      !(currentGesture.getCurrentDragger() instanceof Blockly.Dragger)
```
maribethb, line 144: same question here

plugins/scroll-options/src/AutoScroll.js
```diff
@@ -23,15 +24,19 @@ export class AutoScroll {
...
    */
-  constructor(workspace) {
+  constructor(workspace, dragger) {
     /**
      * Workspace to scroll.
      * @protected {!Blockly.WorkspaceSvg}
      */
     this.workspace_ = workspace;
 
+    /** @private {!ScrollDragger} */
```
maribethb: this name needs to be updated

maribethb (approved):
I think the 3 places I noted still need to be updated then lgtm

plugins/scroll-options/src/ScrollBlockDragger.js
maribethb: no it shouldn't

plugins/scroll-options/src/AutoScroll.js
BeksOmega: oh hm I guess rename symbol doesn't work for jsdoc tags?

PR #2320, opened 2024-04-10
feat: add file upload for block factory

maribethb:
## The basics

### Resolves

Works on #2291 

### Proposed Changes

- Adds the ability to upload a file (which is expected to be downloaded from the legacy block factory tool) to import it into this tool
- Compatible with drag and drop or file upload, and compatible with screenreaders
- Will automatically find an unused name so you don't overwrite any currently saved blocks

### Reason for Changes

migration from old to new tool

### Test Coverage

Manually tested:
- good file
- various bad files
- keyboard nav of buttons

### Documentation

This will need documentation. After both tools have been updated I'll update devsite and then put a "help" link somewhere in this modal explaining where to get the file from

### Additional Information

https://github.com/google/blockly-samples/assets/8573958/bfa0ac7b-1111-4bdc-99b5-ecbca03bc006

### DIFF
--- a/examples/developer-tools/src/controller.ts
+++ b/examples/developer-tools/src/controller.ts
@@ -241,7 +241,7 @@ export class Controller {
     if (e.target && e.target instanceof HTMLElement) {
       const blockName = e.target.getAttribute('data-id');
       if (blockName === IMPORT_BLOCK_FACTORY_ID) {
-        this.handleLoadFromBlockFactory();
+        this.handleLoadFromFile();
         return;
       }
       loadBlock(this.mainWorkspace, blockName);
@@ -274,7 +274,7 @@ export class Controller {
   }
 
   /** Shows the file upload modal when the user selects that option from the load menu. */
-  private handleLoadFromBlockFactory() {
+  private handleLoadFromFile() {
     this.viewModel.toggleFileUploadModal(true);
   }
 
--- a/examples/developer-tools/src/controller.ts
+++ b/examples/developer-tools/src/controller.ts
@@ -6,13 +6,16 @@
 
 import * as Blockly from 'blockly';
 import * as storage from './storage';
-import {createNewBlock, loadBlock} from './serialization';
+import {createNewBlock, loadBlock, loadBlockFromData} from './serialization';
 import {ViewModel} from './view_model';
 import {JavascriptDefinitionGenerator} from './output-generators/javascript_definition_generator';
 import {JsonDefinitionGenerator} from './output-generators/json_definition_generator';
 import {CodeHeaderGenerator} from './output-generators/code_header_generator';
 import {Menu} from '@material/web/menu/menu';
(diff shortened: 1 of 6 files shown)


examples/developer-tools/src/serialization.ts
```diff
@@ -59,6 +59,29 @@ export function loadBlock(workspace: Blockly.Workspace, blockName?: string) {
...
+  Blockly.Events.disable();
+  workspace.clear();
+  Blockly.Events.enable();
+
+  // There might be conflicts when loading a block from a file.
+  // If so, find a similar unused name so we don't overwrite.
+  const blockName = getNewUnusedName(blockJson.fields.NAME);
+  blockJson.fields.NAME = blockName;
+
+  Blockly.serialization.blocks.append(blockJson, workspace);
```
BeksOmega, line 83: I think you can change this to just call `load` instead of `clear` and `append`? Unless `getNewUnusedName` does something I don't know about.

examples/developer-tools/src/controller.ts
```diff
@@ -236,4 +272,73 @@ export class Controller {
...
+    // Prevent default behavior (Prevent file from being opened)
+    e.preventDefault();
+
+    this.viewModel.toggleFileUploadWarning(false);
+
+    this.viewModel.fileDropZone.classList.remove('isDragging');
+
+    if (!e.dataTransfer.items) return;
+
+    const firstItem = [...e.dataTransfer.items].find((item) => {
```
BeksOmega, line 298: Do we want to support uploading multiple items? My understanding is that right now if you wanted to migrate to the new tool, you'd have to download + upload each blocks individually.

examples/developer-tools/src/controller.ts
```diff
@@ -236,4 +272,73 @@ export class Controller {
     });
     menuEl.replaceChildren(...newItems);
   }
+
+  /** Shows the file upload modal when the user selects that option from the load menu. */
+  private handleLoadFromBlockFactory() {
```
BeksOmega: Can you change this to `handleLoadFromFile` or `handleLoadFromLegacyBlockFactory` ?

BeksOmega (commented):
This looks butiful :D

examples/developer-tools/src/serialization.ts
maribethb: Discussed offline since this is just one block data that no. I updated the tsdoc for the function to be more explicit it's just one block's data, lmk if that helps

examples/developer-tools/src/controller.ts
maribethb: Yeah, my goal is one file at a time. Supporting multiple file upload makes it harder to give feedback on what happens if one of them is successful and one isn't for example, and you can only look at one block at a time. So I'd prefer to keep it this way so I don't have to design that UI.

BeksOmega (approved):

PR #2322, opened 2024-04-14
fix: update screenshot.js DOM text reinterpreted as HTML

Shivam7-1:
By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML. Always be cautious when dealing with user input or dynamic content to prevent security risks.

### DIFF
--- a/plugins/dev-tools/src/screenshot.js
+++ b/plugins/dev-tools/src/screenshot.js
@@ -58,7 +58,7 @@ function workspaceToSvg_(workspace, callback, customCss) {
   // Go through all text areas and set their value.
   const textAreas = document.getElementsByTagName('textarea');
   for (let i = 0; i < textAreas.length; i++) {
-    textAreas[i].innerHTML = textAreas[i].value;
+    textAreas[i].innerText = textAreas[i].value;
   }
 
   const bBox = workspace.getBlocksBoundingBox();


Shivam7-1:
Hi @rachel-fenichel  Could You Please Review this PR
Thanks

BeksOmega (approved):
Honestly can't figure out what this code is currently achieving by copying all of the text area `value`s to the `innerHTML`. But can't hurt to change it over to `innerText` (especially when this should only be used for testing and not in production).

Shivam7-1:
Hi @BeksOmega  Thanks For Reviewing
Above PR was approved recently, and I wanted to check in to see if there's anything else needed from my side to proceed with the merge.

Thanks

BeksOmega:
Nope you're all good @Shivam7-1 =) Thank you for the fix!

NeilFraser:
I agree @BeksOmega, this code (before or after this change) makes no sense to me.  If the contents of these textareas are supposed to contain HTML code, then it was a vulnerability, and it is now broken.  But I can't figure out where these textareas are.  Anyone know how to test this?  There's no mention of screenshots in the readme for this plugin.

BeksOmega:
You can right click the workspace and hit "download screenshot". Even if you comment out this code, it appears the screenshot still works (all text is included). I dug back into the history, but Sam committed it whole cloth to core, and there were no comments on the PR about this code.

PR #2325, opened 2024-04-17
feat: add angle and colour fields to block factory

### LINKED ISSUE #2290: Block factory: import plugin fields

maribethb:
## The basics

### Resolves

Fixes google/blockly#2159 

### Proposed Changes

- Adds colour and angle blocks to field blocks
- Uses plugin angle field for the colour block used to color the block factory block
- Does not add the multiline text since that's not in the original block factory either

### DIFF
--- a/examples/developer-tools/src/blocks/index.ts
+++ b/examples/developer-tools/src/blocks/index.ts
@@ -4,6 +4,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {registerFieldAngle} from '@blockly/field-angle';
+import {registerFieldColour} from '@blockly/field-colour';
+
 import {factoryBase} from './factory_base';
 import * as Blockly from 'blockly/core';
 import {input} from './input';
@@ -14,7 +17,9 @@ import {
   connectionCheckItem,
 } from './connection_check';
 import {
+  fieldAngle,
   fieldCheckbox,
+  fieldColour,
   fieldDropdown,
   fieldDropdownContainer,
   fieldDropdownOptionImage,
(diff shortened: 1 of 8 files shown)


examples/developer-tools/src/blocks/index.ts
```diff
@@ -40,12 +45,18 @@ import '../output-generators/fields/label_serializable';
...
 import '../output-generators/fields/variable';
+import '../output-generators/fields/angle';
+import '../output-generators/fields/colour';
 import '../output-generators/colour';
 
 /* eslint-disable @typescript-eslint/naming-convention
  -- Blockly convention is to use snake_case for block names
 */
 export const registerAllBlocks = function () {
+  // Register the plugin fields before using them in blocks.
```
BeksOmega, line 56: Do they still need to be registered if we're constructing them explicitly?

examples/developer-tools/src/output-generators/fields/angle.ts
```diff
@@ -0,0 +1,77 @@
...
+  generator.addHeaderLine(`registerFieldAngle();`);
+  return '';
+};
+
+scriptHeaderGenerator.forBlock['field_angle'] = function (
+  block: Blockly.Block,
+  generator: CodeHeaderGenerator,
+): string {
+  generator.addHeaderLine(
+    `<script src="https://unpkg.com/@blockly/field-angle"></script>`,
```
BeksOmega, line 63: When we're importing from script tags does the field get registered automatically?

BeksOmega (approved):
Just so we don't lose it, we do have an open feature request for adding the mutiline field: https://github.com/google/blockly/issues/7206

examples/developer-tools/src/blocks/index.ts
maribethb: Yeah, they need to be registered to use them in json, and the block preview is generated from the json block definition.

examples/developer-tools/src/output-generators/fields/angle.ts
maribethb: No, thanks, fixed

maribethb:
> Just so we don't lose it, we do have an open feature request for adding the mutiline field: [google/blockly-samples#2326](https://github.com/google/blockly-samples/issues/2326)

Thanks! I'm planning to open follow up bugs to add all the plugin fields eventually. I'll move this issue to samples

PR #2328, opened 2024-04-19
Update screenshot.js DOM text reinterpreted as HTML 

Shivam7-1:
By using textContent, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.

### DIFF
--- a/plugins/dev-tools/src/screenshot.js
+++ b/plugins/dev-tools/src/screenshot.js
@@ -99,7 +99,7 @@ function workspaceToSvg_(workspace, callback, customCss) {
     })
     .join('\n');
   const style = document.createElement('style');
-  style.innerText = css + '\n' + customCss;
+  style.textContent = css + '\n' + customCss;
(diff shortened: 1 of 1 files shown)


Shivam7-1:
Hi @cpcallen  Could You Please Review this PR
Thanks

NeilFraser:
This one is actually a good catch, since there's dynamic content.

However, as mentioned in the previous PR, this should be `textContent`, not `innerText`.  Please revisit the other PRs you've made (particularly on other Google projects) and amend them accordingly.

Shivam7-1:
Hi @NeilFraser Thanks For Reviewing Above PR
As Per Suggestion I had Make changes and uses textContent instead of innertext 

Thanks

NeilFraser (approved):

PR #2329, opened 2024-04-19
fix: Update index.js DOM text reinterpreted as HTML

Shivam7-1:
By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.

### DIFF
--- a/examples/keyboard-navigation-codelab/src/index.js
+++ b/examples/keyboard-navigation-codelab/src/index.js
@@ -43,7 +43,7 @@ const runCode = () => {
   const code = javascriptGenerator.workspaceToCode(ws);
   codeDiv.innerText = code;
 
-  outputDiv.innerText = '';
+  outputDiv.textContent = '';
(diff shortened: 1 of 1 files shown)


Shivam7-1:
Hi @NeilFraser  Could You Please Review this PR
Thanks

NeilFraser:
Google bans use of assignment to innerHTML - with the exception of the empty string.  That said, there's no harm in changing this.

However, textContent is vastly preferred over innerText: https://kellegous.com/j/2013/02/27/innertext-vs-textcontent/

Shivam7-1:
Hi @NeilFraser Thanks for Reviewing 
As per Suggestion In had used textContent instead of innertext 

Regards

NeilFraser (approved):

PR #2331, opened 2024-04-19
fix: Update custom-dialog.js DOM text reinterpreted as HTML

Shivam7-1:
By using textContent, it will avoid the risk of Dynamic HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.

### DIFF
--- a/examples/custom-dialogs-demo/custom-dialog.js
+++ b/examples/custom-dialogs-demo/custom-dialog.js
@@ -102,7 +102,7 @@ CustomDialog.show = function (title, message, options) {
   backdropDiv.style.display = 'block';
   dialogDiv.style.display = 'block';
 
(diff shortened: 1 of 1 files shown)


NeilFraser:
You need to look at what the code is doing.  When projects set `.innerHTML` with HTML tags, one can't change this to either `.textContent` or `.innerText`.  Doing so would print raw HTML on the user's screen.  There are a dozen such changes you've proposed in your history.  Please don't keep proposing this breaking change.

Also (after looking at your other PRs), `.innerHTML` is absolutely safe when used as a getter.  There's never a need to change it unless it is being set.

PR #2332, opened 2024-04-19
fix: Update codelab/src/ index.js DOM text reinterpreted as HTML

Shivam7-1:
By using textContent, it will avoid the risk of Dynamic HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML


NeilFraser (approved):
Setting innerHTML as empty string is perfectly safe, but there's no issue with using textContent instead.

PR #2334, opened 2024-04-20
fix: Call blocks handle both manual disabling and disabled defs

johnnesky:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/2035

### Proposed Changes

https://github.com/google/blockly/pull/7958 makes the core Blockly system for disabling blocks more advanced, allowing a block to be simultaneously disabled for multiple independent reasons (such as manually disabling from the context menu, or being disabled because the block is not valid). Removing one of these reasons doesn't affect the others, and the block as a whole is only considered to be enabled if it doesn't have any disabled reasons. This PR takes advantage of that system to allow shared procedure call blocks to have a unique reason for being disabled when the corresponding procedure model is disabled.

Also, I updated all plugins to depend on the latest blockly beta: `11.0.0-beta.9`.

### Reason for Changes

Previously, if the user manually disabled a procedure call block, then edited the name of the procedure, the call block would be re-enabled (because the call block was updated to match the procedure model). Also, a call block that had been disabled by the model could be manually re-enabled.

### Test Coverage

I tested manually. (with `npm run start` in the plugin directory.) All bugs I'm aware of have been fixed.

There are existing tests to ensure that call blocks get disabled/enabled when the corresponding def is changed, and these still pass. I don't think there are any tests related to disabling from the context menu though.

### Documentation

N/A

### Additional Information

In `procedureDefUpdateShapeMixin.doProcedureUpdate()`, there's the following line:
```
    this.setEnabled(this.getProcedureModel().getEnabled());
```
This is using the deprecated Block method `setEnabled` which only updates the `MANUALLY_DISABLED` reason. If the def block is disabled for other reasons, then the procedure model can't fully enable the def block. This sounds like an unlikely scenario (what else would be disabling a def block? what would cause the model to become enabled when the def block is disabled?) but if we want to robustly handle this situation, then the procedure model should probably fully represent all of the procedure's disabled reasons, instead of just a boolean indicating whether there are any disabled reasons, so that it can properly sync the def with the model.

### DIFF
--- a/plugins/block-shareable-procedures/src/blocks.ts
+++ b/plugins/block-shareable-procedures/src/blocks.ts
@@ -745,7 +745,7 @@ const procedureDefOnChangeMixin = {
       e.blockId === this.id &&
       e.element === 'disabled'
     ) {
-      this.getProcedureModel().setEnabled(!e.newValue);
+      this.getProcedureModel().setEnabled(this.isEnabled());
     }
   },
 };
@@ -949,8 +949,6 @@ Blockly.Extensions.register(
 );
 
 const procedureCallerMutator = {
-  previousEnabledState_: true,
-
   paramsFromSerializedState_: [],
 
   /**
@@ -1053,6 +1051,8 @@ Blockly.Extensions.registerMutator(
   procedureCallerMutator,
 );
 
+const PROCEDURE_MODEL_DISABLED_REASON = 'PROCEDURE_MODEL_DISABLED';
+
 const procedureCallerUpdateShapeMixin = {
   /**
    * Renders the block for the first time based on the procedure model.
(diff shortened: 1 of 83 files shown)


cpcallen (approved):
Re:

> In `procedureDefUpdateShapeMixin.doProcedureUpdate()`, there's the following line:
> 
> ```
>     this.setEnabled(this.getProcedureModel().getEnabled());
> ```
> 
> This is using the deprecated Block method `setEnabled` which only updates the `MANUALLY_DISABLED` reason. If the def block is disabled for other reasons, then the procedure model can't fully enable the def block…

This seems reasonable enough to me but I don't think I have enough knowledge of this part of the codebase to be able to say for sure.  Perhaps @BeksOmega could weigh in?

plugins/block-shareable-procedures/src/blocks.ts
```diff
@@ -1094,12 +1094,10 @@ const procedureCallerUpdateShapeMixin = {
...
-    if (!this.getProcedureModel().getEnabled()) {
-      this.previousEnabledState_ = this.isEnabled();
-      this.setEnabled(false);
-    } else {
-      this.setEnabled(this.previousEnabledState_);
-    }
+    this.setDisabledReason(
+      !this.getProcedureModel().getEnabled(),
+      PROCEDURE_MODEL_DISABLED_REASON,
+    );
```
BeksOmega, line 1100: Yesss this code is so much more elegant, and it actually works!

BeksOmega (approved):

PR #2336, opened 2024-04-23
feat: support uploading file from old block factory

### LINKED ISSUE #2291: Block factory: backwards compatibility

maribethb:
## The basics

### Resolves

Fixes #2291 

### Proposed Changes

- Adds the ability to upload one or more files downloaded from the old block factory tool (json format)
- Will show a message with the results of the upload, then load the first successful block if possible

### Documentation

Coming soon!

### Additional Information

The corresponding change in the old block factory to let you download the required data is coming soon

https://github.com/google/blockly-samples/assets/8573958/ef70d4bc-5669-48c2-aa0b-f1b7a6983a59

### DIFF
--- a/examples/developer-tools/src/controller.ts
+++ b/examples/developer-tools/src/controller.ts
@@ -20,10 +20,11 @@ const IMPORT_BLOCK_FACTORY_ID = 'Import from block factory...';
 /**
  * A map where the keys are the names of a file that was uploaded,
  * and the value is an object with two properties:
- *     successes - an array of successfully saved block names
+ *     successes - an array of block names, which are represented by a
+ *        tuple as [newName, oldName] in case of block renames
  *     fails - the number of blocks that failed to parse from that file
  */
-type UploadResults = Map<string, {successes: string[]; fails: number}>;
+type UploadResults = Map<string, {successes: string[][]; fails: number}>;
 
 /**
  * This class handles updating the UI output, including refreshing the block preview,
@@ -311,17 +312,18 @@ export class Controller {
     const allResults: UploadResults = new Map();
     const allPromises: Array<Promise<void>> = [];
 
-    [...e.dataTransfer.items].forEach((item) => {
+    for (const item of e.dataTransfer.items) {
       // Only consider plain text files
-      if (item.kind === 'file' && item.type === 'text/plain') {
-        const file = item.getAsFile();
-        allPromises.push(
-          this.loadFromFile(file).then((results) => {
-            allResults.set(file.name, results);
-          }),
-        );
+      if (item.kind !== 'file' || item.type !== 'text/plain') {
+        continue;
       }
-    });
+      const file = item.getAsFile();
+      allPromises.push(
+        this.loadFromFile(file).then((results) => {
+          allResults.set(file.name, results);
+        }),
+      );
+    }
 
     // Wait for all the files to be processed, then display results
     Promise.all(allPromises).then(() => {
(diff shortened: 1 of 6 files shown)


examples/developer-tools/src/controller.ts
```diff
@@ -295,18 +308,25 @@ export class Controller {
 
     if (!e.dataTransfer.items) return;
 
-    const firstItem = [...e.dataTransfer.items].find((item) => {
-      // Get the first plain text file, in case the user uploaded multiple
+    const allResults: UploadResults = new Map();
+    const allPromises: Array<Promise<void>> = [];
+
+    [...e.dataTransfer.items].forEach((item) => {
```
cpcallen: Is `e.dataTransfer.items` not already iterable?
```suggestion
    e.dataTransfer.items.forEach((item) => {
```
(Though in fact I'd use a `for … of` loop rather than `.forEach`, since that obviates the need for the arrow function.  Alas GitHub won't let me suggest that change here.)

examples/developer-tools/src/controller.ts
```diff
@@ -295,18 +308,25 @@ export class Controller {
...
+      // Only consider plain text files
       if (item.kind === 'file' && item.type === 'text/plain') {
```
cpcallen: Prefer a guard clause: `if (/* not a plain text file */) continue` (or …`return` as appropriate).

examples/developer-tools/src/controller.ts
```diff
@@ -318,27 +338,117 @@ export class Controller {
     const input = this.viewModel.fileUploadInput as HTMLInputElement;
     const file = input.files[0];
     if (!file) return;
-    this.loadFromFile(file);
+    this.loadFromFile(file).then(({successes, fails}) => {
```
cpcallen: Any reason to destructure if you're only going to restructure it later?  (Maybe you need to make a shallow copy?  But if so then a comment noting this would be wise.)

examples/developer-tools/src/controller.ts
```diff
@@ -318,27 +338,117 @@ export class Controller {
...
+          fails++;
+          return {successes, fails};
+        }
+        for (const block of blocksData) {
+          const fixedBlockData = convertBaseBlock(block);
+
+          // The name from the old factory might be used in this tool.
+          // Get a fresh name so we don't overwrite anything.
+          const name = storage.getNewUnusedName(fixedBlockData.fields.NAME);
+          fixedBlockData.fields.NAME = name;
```
cpcallen: Possibly a bit fiddly to plumb back through to the results dialog, but might it be worth warning the user if the block(s) they are loading have been renamed?

examples/developer-tools/src/controller.ts
```diff
@@ -318,27 +338,117 @@ export class Controller {
...
+  ): Promise<{successes: string[]; fails: number}> {
+    const successes: string[] = [];
+    let fails = 0;
+    return file.text().then((contents) => {
+      try {
+        const blocksData = JSON.parse(contents);
+        if (!Array.isArray(blocksData) || blocksData.length === 0) {
+          // File did not contain a JSON array or was empty.
+          fails++;
+          return {successes, fails};
```
cpcallen: Perhaps just:
```suggestion
          return {[], 1};
```
This makes it clear that this exit point will definitely trigger the 'parse failed' error message, and would allow `successes` and `fails` declarations to appear just before the `for`…`of` loop.

examples/developer-tools/src/controller.ts
```diff
@@ -318,27 +338,117 @@ export class Controller {
...
-        this.viewModel.toggleFileUploadWarning(true);
-      });
+  private loadFromFile(
+    file: File,
+  ): Promise<{successes: string[]; fails: number}> {
+    const successes: string[] = [];
+    let fails = 0;
+    return file.text().then((contents) => {
+      try {
+        const blocksData = JSON.parse(contents);
```
cpcallen: I think it would be preferable to have a`catch` immediately after `JSON.parse`, with an immediate `return {[], 1};`—effectively a guard clause for parse failures, and then a separate `try` within the `for`…`of` loop, so that any errors thrown while loading a particular block will not derail attempting to finish loading the rest.

examples/developer-tools/src/index.html
```diff
@@ -45,21 +45,29 @@
...
-            <md-text-button
-              id="file-upload-close"
-              form="form-id"
-              value="cancel">
+            <md-text-button id="file-upload-close" value="cancel">
               Cancel
             </md-text-button>
           </div>
         </md-dialog>
+        <md-dialog id="upload-results-modal">
```
cpcallen, line 61: What is this I don't even.

cpcallen (commented):
I can't really comment on the material design stuff since I'm quite unfamiliar with this, but if it LGTY then it's probably fine.

One thing I note is that it looks like one can drag and drop multiple files at a time, but only load one file at a time via the file picker.  That seems like a surprising limitation.  Is it a limitation in the HTML living spec, in the `@material/web` package, or in your code?  If the latter it would seem worth removing, since most of the rest of the needed machinery is already present here.

examples/developer-tools/src/index.html
maribethb: Can't tell what part of this you are reacting to, is it the custom element? They're fun and also funfact you can't put blockly inside one :)

BeksOmega (approved):
I didn't read through the first commit, but fixes from the previous review all LGTM =) Let me know if you want me to go over the original changes, but Christophers review seemed detailed enough that I trust it!

PR #2341, opened 2024-05-06
fix: add boolean to dispose function call 

### LINKED ISSUE #2217: Disposing of a procedure definition disposes caller *and* next blocks
### Check for duplicates

### Component

block-sharable-procedures

### Description

When you dispose of a procedure definition, it disposes of the callers (this is correct). However, it also disposes of the callers next blocks, which is not correct.

### Reproduction steps

1. Create the following arrangement in the sharable procedures demo
image

3. Delete the procedure definition block
4. Observe how only the top level if block remains. Instead the two if blocks should be connected in a stack.

### Stack trace

_No response_

### Screenshots

_No response_

## To fix

1. Pass `true` [here](https://github.com/google/blockly-samples/blob/770d03dfdfb3bccf49bc74bf986b25f6bf4d32c3/plugins/block-shareable-procedures/src/blocks.ts#L1068). 
2. Repeat the reproduction steps and ensure only the procedure call block is deleted, not all of its children
3. Add a unit test as part of [this suite](https://github.com/google/blockly-samples/blob/770d03dfdfb3bccf49bc74bf986b25f6bf4d32c3/plugins/block-shareable-procedures/test/procedure_blocks.mocha.js#L1280) that tests to ensure the child blocks of the caller are not deleted.

The `dispose` function takes an optional parameter that controls whether to "heal" the stack of blocks when one is deleted. If true, only the one block will be deleted and the child blocks of it will be connected to the deleted block's parent, if possible. If false (or unset, as here) the block and its children will be deleted.

If you'd like to work on this issue and you're new to Blockly / blockly-samples, there are some guidelines [here](https://developers.google.com/blockly/guides/contribute/samples). Comment on this issue if you'd like to work on it so we can assign it to you.

sancodes:
## The basics

### Resolves
Fixes #2217

### Proposed Changes

- adds true to the dispose()
- Screenshot 2024-05-06 at 3 14 45 PM

### Reason for Changes
- adding `dispose(true)` makes sure the blocks that are not associated with the procedures block remain **undeleted**

- ### Test Coverage
- Test, I tried to recreate, the blocks being connected to each other, and mimicking the deletion process of procedures child, which should still keep the other blocks intact and undeleted
Screenshot 2024-05-06 at 3 17 33 PM

My testing logic was: 
- create the def procedure block`"type": "procedures_defnoreturn"`
Screenshot 2024-05-06 at 3 26 01 PM

- create the child block `"type": "procedures_callnoreturn"`
Screenshot 2024-05-06 at 3 26 51 PM

- then basically create two if blocks `"type": "controls_if"`
Screenshot 2024-05-06 at 3 27 28 PM

- then tried to connect those blocks
Screenshot 2024-05-06 at 3 29 02 PM
Screenshot 2024-05-06 at 3 29 29 PM

- I tried to pass in `dispose(true)`, so when the procedure block is deleted, the blocks not associated with procedure block remains intact.

### DIFF
--- a/plugins/block-shareable-procedures/test/procedure_blocks.mocha.js
+++ b/plugins/block-shareable-procedures/test/procedure_blocks.mocha.js
@@ -1330,8 +1330,8 @@ suite('Procedures', function () {
         const ifBlock1 = this.workspace.newBlock('controls_if');
         const ifBlock2 = this.workspace.newBlock('controls_if');
 
-        ifBlock1.nextConnection.outputConnection = callBlock1;
-        callBlock1.nextConnection.outputConnection = ifBlock2;
+        ifBlock1.nextConnection.connect(callBlock1.previousConnection);
+        callBlock1.nextConnection.connect(ifBlock2.previousConnection);
 
         defBlock.dispose(true);
         globalThis.clock.runAll();
--- a/plugins/block-shareable-procedures/test/procedure_blocks.mocha.js
+++ b/plugins/block-shareable-procedures/test/procedure_blocks.mocha.js
@@ -1320,6 +1320,36 @@ suite('Procedures', function () {
         );
       },
     );
+
+    test(
+      'when the procedure definition block is deleted, all of its ' +
+        'associated callers are deleted, but the other blocks remain',
+      function () {
+        const defBlock = createProcDefBlock(this.workspace);
+        const callBlock1 = createProcCallBlock(this.workspace);
+        const ifBlock1 = this.workspace.newBlock('controls_if');
+        const ifBlock2 = this.workspace.newBlock('controls_if');
+
+        ifBlock1.nextConnection.outputConnection = callBlock1;
+        callBlock1.nextConnection.outputConnection = ifBlock2;
+
+        defBlock.dispose(true);
+        globalThis.clock.runAll();
+
+        assert.isTrue(
+          callBlock1.disposed,
+          'Expected the first caller to be disposed',
+        );
+        assert.isFalse(
+          ifBlock1.disposed,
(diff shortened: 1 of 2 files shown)


plugins/block-shareable-procedures/test/procedure_blocks.mocha.js
```diff
@@ -1320,6 +1320,36 @@ suite('Procedures', function () {
...
+      'when the procedure definition block is deleted, all of its ' +
+        'associated callers are deleted, but the other blocks remain',
+      function () {
+        const defBlock = createProcDefBlock(this.workspace);
+        const callBlock1 = createProcCallBlock(this.workspace);
+        const ifBlock1 = this.workspace.newBlock('controls_if');
+        const ifBlock2 = this.workspace.newBlock('controls_if');
+
+        ifBlock1.nextConnection.outputConnection = callBlock1;
+        callBlock1.nextConnection.outputConnection = ifBlock2;
```
BeksOmega:
```suggestion
        ifBlock1.nextConnection.connect(callBlock1.previousConnection);
        callBlock1.nextConnection.connect(ifBlock2.previousConnection);
```

BeksOmega (changes requested):
Hello! Your testing logic is correct but your test code doesn't do what you want it to.

After that fix I'll pull it down to test manually. Then should be good to go =)

sancodes:
@BeksOmega applied your suggestions.

BeksOmega (approved):
Tested locally and this LGTM =)

Thank you for the fix!!

sancodes:
@BeksOmega Thank you :)

PR #2342, opened 2024-05-08
fix: update keyboard nav to use proper getContents

maribethb:
## The basics

### Resolves

Needed for https://github.com/google/blockly/pull/8064

### Proposed Changes

Updates the keyboard-navigation plugin to use the API provided by `getContents`

### Reason for Changes

Made a change to the type returned by `getContents`. This hasn't been released yet other than in betas so this isn't a breaking change.

### Test Coverage

Tested by linking blockly & blockly-samples together to test both PRs and verified you can navigate buttons and blocks.


maribethb:
The tests fail because navigation doesn't work with the current beta (7) because `getContents` returns something different there. They pass locally when the corresponding change to blockly is included. Will need to hold submitting this until we release another beta (likely Friday) but can be reviewed now.

plugins/keyboard-navigation/src/navigation.js
```diff
@@ -502,12 +502,14 @@ export class Navigation {
     if (flyout && flyout.getWorkspace()) {
       const flyoutContents = flyout.getContents();
       const firstFlyoutItem = flyoutContents[0];
-      if (firstFlyoutItem instanceof Blockly.FlyoutButton) {
-        const astNode = Blockly.ASTNode.createButtonNode(firstFlyoutItem);
+      if (!firstFlyoutItem) return;
+      if (firstFlyoutItem.type === 'button' && firstFlyoutItem.button) {
```
BeksOmega: Is the type guaranteed to be lowercase? would it be possible to just check the property instead?

BeksOmega (approved):

plugins/keyboard-navigation/src/navigation.js
maribethb: core uses an enum so yes it will be lowercase. the enum isn't exported so i can't use it here. it's probably fine to just check the property... we don't actually use the type property anywhere afaik. i don't know why we have it. will update

BeksOmega: Sweet feel free to merge after update! or if you decide to not change it. Lemme know if you'd prefer re-review.

PR #2351, opened 2024-05-13
feat: Allow configuration of buttons, pixel colors, and fixed field height for Bitmap field

### LINKED ISSUE #2353: Bitmap field is lacking configuration options
### Check for duplicates

### Component

field-bitmap

### Problem

Currently, some aspects of the bitmap field are not configurable that probably should be. Because the field's pixel size is fixed, blocks blow up in size for larger bitmaps. Pixel colors are also hard-coded. Most egregious for an international audience, the text on the editor's buttons is not translatable.

### Request

I request the addition of new configuration options that make it possible to customize the pixel colors and set a fixed field height. If the field height is fixed, pixels would be resized to fit automatically. The button label text "Randomize" and "Clear" should be made translatable.

### Alternatives considered

Code.org was able to work around the current limitations of the class by overriding many of its methods and providing new constants: 
- https://github.com/code-dot-org/code-dot-org/pull/57107
This approach isn't great because we needed to access and modify private properties. It also added a maintenance cost for this feature for us. For example, when the plugin was converted to TypeScript recently, we had to update our custom class to account for renamed properties - the same ones we probably shouldn't be using in the first place!

### Additional context

Screenshot shows an example of a new block we created. The field height has been made static (with smaller-than-normal pixels), and the randomize button has been removed.
image

mikeharv:
## The basics

### Resolves

Resolves https://github.com/google/blockly-samples/issues/2353

### Proposed Changes

1. Use the existing `Blockly.Msg` system to translate button text ("Randomize" and "Clear")
2. Provide additional configuration options for hidden buttons, field height, and pixel color

Examples

_Hiding the "Randomize" button:_
image

_Setting a static field height for a larger image:_
image

_Customizing pixel colors:_
image

### Reason for Changes

Button text should be translatable as a best practice for international and non-English users.

Buttons should be option in the event that "Randomize" or "Clear" are not desired (ex. Code.org).

Field height should be settable in order to prevent blocks from blowing up in size with larger bitmap images.

Pixel color should be configurable so that it can be aligned to an app's block colors, external branding guidelines, or other UI considerations.

### Test Coverage

The changes here affect UI and not functionality. The test blocks in the plugin playground have been updated to include a set with one hidden button, custom pixel colors, and a custom static field height.

### Documentation

I updated the plugin's ReadMe to explain the new configuration options. I'm not aware of other documentation specific to this plugin.

### Additional Information

This plug-in will be highlighted at the Blockly Summit in June as part of my presentation. It would be great if this could get published in time for me to update our app code to use the new options!


cpcallen:
> _Customizing pixel colors:_
> image

Ha ha, you know how to butter us up, don't you!  :-)

plugins/field-bitmap/README.md
```diff
@@ -20,14 +20,21 @@ npm install @blockly/field-bitmap --save
...
+This field accepts up to 6 parameters:
 
 - `"value"` to specify an initial value. Must be a 2D rectangular array of 1s and 0s.
   If not provided, the default is an empty grid of the specified size.
 - `"width"` to specify an initial width, if there is no initial value.
   If not provided, the default is a width of 5.
 - `"height"` to specify an initial height, if there is no initial value.
   If not provided, the default is a height of 5.
+- `fieldHeight"` to specify a static field height. If provided, the individual pixels
+  will be resized to fit inside the field. Good for larger images.
```
cpcallen: From the name I assume this is adjusting the height of the field-as-it-appears-in-a-block rather than the size of the pop-up editor, but perhaps the documentation could be reworded to make this crystal clear.

plugins/field-bitmap/README.md
```diff
@@ -20,14 +20,21 @@ npm install @blockly/field-bitmap --save
...
+- `"colors"` to override the default colors, Default values:  
+  `{filled: '#363d80', empty: '#fff'}`
```
cpcallen: Blockly and its plugins generally the spelling "colour" rather than "color".  Please be consistent unless there is a compelling reason to the contrary (e.g. it's fine to use `color` when needed on stylesheets.)

plugins/field-bitmap/README.md
```diff
@@ -20,14 +20,21 @@ npm install @blockly/field-bitmap --save
...
+- `"buttons"` to hide the "Randomize" and/or "Clear" buttons.
+  Default values:  
+  `{showRandomize: true, showClear: true}`
```
cpcallen: I think it might be better for this to be `showButtons: {randomize: true, clear: false}`, unless you foresee offering other configuration options in future, in which case I'd probably suggest `buttons: {randomize: true, clear: false}` with a view to supplying some non-boolean value to further configure each button.

Also, it would be useful to document what happens if one of the buttons is omitted from the options object.

plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -6,12 +6,20 @@
...
+Blockly.Msg['BUTTON_LABEL_CLEAR'] = 'Clear';
+
 export const DEFAULT_HEIGHT = 5;
 export const DEFAULT_WIDTH = 5;
-const PIXEL_SIZE = 15;
-const FILLED_PIXEL_COLOR = '#363d80';
-const EMPTY_PIXEL_COLOR = '#fff';
-
+const DEFAULT_PIXEL_SIZE = 15;
+const DEFAULT_PIXEL_COLORS: PixelColors = {
```
cpcallen: See above re: colours vs colors.  (Here and several places below.)

plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -60,6 +73,11 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
       // Set a default empty value
       this.setValue(this.getEmptyArray());
     }
+    if (config?.fieldHeight) {
+      this.pixelSize = config.fieldHeight / this.imgHeight;
```
cpcallen, line 77: I'm not not a CSS expert.  Do fractional sizes work?  Does the total height come out correctly, or might rounding result in the actual total height being quite different than `fieldHeight`?  It's OK if so but should probably be documented.

cpcallen (commented):
Thanks for sending us this.  Overall it looks great and we're looking forward to accepting your changes.  I have suggested a few tweaks that I think would improve things.

plugins/field-bitmap/README.md
mikeharv: That's correct. Will document!

plugins/field-bitmap/src/field-bitmap.ts
mikeharv: Great question! It seems to vary just a bit depending on the browser. Will document.

plugins/field-bitmap/README.md
mikeharv: I went with `buttons: {randomize: true, clear: false}`. Let me know what you think!

plugins/field-bitmap/README.md
```diff
@@ -20,14 +20,24 @@ npm install @blockly/field-bitmap --save
...
 
 - `"value"` to specify an initial value. Must be a 2D rectangular array of 1s and 0s.
   If not provided, the default is an empty grid of the specified size.
 - `"width"` to specify an initial width, if there is no initial value.
   If not provided, the default is a width of 5.
 - `"height"` to specify an initial height, if there is no initial value.
   If not provided, the default is a height of 5.
+- `fieldHeight"` to specify a static field height. If provided, the individual pixels
+  will be resized to fit inside the field. This only affects the field as it is
+  scene on a block and not the pop-up editor. Good for larger images. (_Note: If this
```
cpcallen:
```suggestion
  seen on a block and not the pop-up editor. Good for larger images. (_Note: If this
```

plugins/field-bitmap/README.md
```diff
@@ -20,14 +20,24 @@ npm install @blockly/field-bitmap --save
...
+  results in fractional pixel sizes, the overall field height may not exactly match
+  the specified value on all browsers._)
+- `"colours"` to override the default colours, Default values:  
+  `{filled: '#363d80', empty: '#fff'}`
+- `"showButtons"` to show or hide the "Randomize" and/or "Clear" buttons. If either is
+  omitted, the button will be shown. Default values:
```
cpcallen: Should this be `buttons`?:

```suggestion
- `"buttons"` to show or hide the "Randomize" and/or "Clear" buttons. If either is
  omitted, the button will be shown. Default values:  
```

plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -6,12 +6,20 @@
 
 import * as Blockly from 'blockly/core';
 
+Blockly.Msg['BUTTON_LABEL_RANDOMIZE'] = 'Randomize';
+Blockly.Msg['BUTTON_LABEL_CLEAR'] = 'Clear';
```
cpcallen, line 10: I hadn't noticed this previously.  I think it's fine as it is for the moment, but the team should probably make some decisions about how we will deal with localisation of plugins now that more formerly-core functionality has been moved out of Blockly.

cpcallen (changes requested):
Just a couple of minor doc corrections (plus I'll need to do some local manual testing) before merging.

plugins/field-bitmap/src/field-bitmap.ts
mikeharv: For this, I was following guidance from @maribethb who pointed out that this is how it’s done elsewhere:

https://google.github.io/blockly-samples/plugins/block-plus-minus/README

https://github.com/google/blockly-samples/blob/337ce75ab72a74b6c9c2551f92442cbf848a5e90/plugins/block-plus-minus/src/procedures.js#L15

Naturally, I’m open to changing this if there’s a better way. I had originally set up to take in the translation strings as configuration options (which is why there was ‘buttonOptions’ initially).

cpcallen: Her guidance is certainly good.  I've no issue with the code; my comment was more as a reminder to myself that we should probably think about whether we want to add messages like this to the set that get translated by translatewiki. Until now we've only had messages in core Blockly and Blockly Games translated, not messages for plugins—but with some formerly-core fields and blocks now being moved to plugins we should probably revisit that.

cpcallen (approved):
I'm going to go ahead and approve this and then merge it, because I'm satisfied that these changes do not create any significant issues, and are strictly an improvement that I know that we want to include in today's blockly-samples release.

In local testing with the supplied test blocks, however, I have noted what appears to be a bug with sizing the pop-up for the "Static field height, custom colors, no buttons" test block:

Screenshot 2024-05-16 at 13 09 10

(This is on Chrome v124 on MacOS Ventura.  It may be an issue with Blockly v11, which I had `npm link`ed at the time.)

Can you take a look and either submit a fix or (if that is not possible for some reason) file an issue to remind us to have a look again once the dust settles on the Blockly v11 release?

mikeharv:
@cpcallen I've prepared a follow-up PR that addresses the issue you found and one other. The pop-up sizing bug was present before this work was done as Core Blockly sets a maximum height on dropdown editors. I also proposed conditionally removing a bottom margin from the editor if no buttons are present.

- https://github.com/google/blockly-samples/pull/2360

cpcallen:
@mikeharv: wonderful, thanks!

Issue #2356, opened 2024-05-15
Issues with test blocks in field plugins


rachel-fenichel:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Component

Bitmap: “TypeError: Invalid block definition for type: 1test_field_bitmap_single”
Colour-hsv-slider: “TypeError: Invalid block definition for type: 1test_field_colour_hsv_sliders_single”
Date: “ TypeError: Invalid block definition for type: 1test_field_date_single”
Slider: “TypeError: Invalid block definition for type: 1test_field_slider_single”
Angle: “TypeError: Invalid block definition for type: 1test_field_angle_single”
Grid dropdown: “TypeError: Invalid block definition for type: 1test_field_grid_dropdown_single”

### Description

The field plugins call `generateFieldTestBlocks` to define blocks and a toolbox with the field under test on multiple blocks.

`generateFieldTestBlocks` avoids code duplication but also adds a level of indirection that makes it hard to debug the plugins. It generates XML instead of JSON for the block definition.

`generateFieldTestBlocks` currently adds the blocks to the wrong `Blockly.Blocks` object, probably due to the dual package hazard. Given the other issues with it, we're choosing to remove its uses in favor of explicitly defined test blocks.

### Reproduction steps

Open the test page for each plugin. Note that no Blockly workspace shows up, and there are errors in the console.

PR #2357, opened 2024-05-15
fix: remove some uses of generateFieldTextBlocks

rachel-fenichel:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/2356

### Proposed Changes

Removes uses of `generateFieldTextBlocks` for fields where it was breaking the test pages.

### Reason for Changes
Swap indirection for code duplication, to make it easier to read the test code and see what's being tested.

### Test Coverage
These are tests.

### Documentation
None

### DIFF
--- a/plugins/field-angle/test/index.ts
+++ b/plugins/field-angle/test/index.ts
@@ -9,61 +9,299 @@
  */
 
 import * as Blockly from 'blockly';
-import {generateFieldTestBlocks, createPlayground} from '@blockly/dev-tools';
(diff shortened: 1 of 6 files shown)


NeilFraser (approved):
Simpler, 'dumber' test code is good.
I thought we were supposed to merge into `develop` rather than `v11`?

BeksOmega:
> Simpler, 'dumber' test code is good. I thought we were supposed to merge into `develop` rather than `v11`?

Not for samples! See chat.

Issue #2362, opened 2024-05-16
blockly-rtc demo has webpack errors


rachel-fenichel:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Component

Examples

### Description

- [ ] blockly-rtc
  - [ ] "Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema."
  - [ ] Error also occurs with v10.4.3

maribethb:
I think the `enabled` deprecation warning should disappear since we decided in https://github.com/google/blockly/issues/8095 to remove it

rachel-fenichel:
> I think the `enabled` deprecation warning should disappear since we decided in [google/blockly#8095](https://github.com/google/blockly/issues/8095) to remove it

Removed the items for those warnings, which were:
 - [ ] Warning that "enabled was deprecated in v11 and will be deleted in v12. Use disabledReasons with the value ["MANUALLY_DISABLED"] instead."
    - [ ] blockly-angular
    - [ ] blockly-react
    - [ ] blockly-svelte
    - [ ] blockly-vue3
    - [ ] turtle demo

PR #2363, opened 2024-05-16
fix: update imports in examples for v11

rachel-fenichel:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/2364

### Proposed Changes

For each plugin that has issues with imports when updated to v11:
- Install v11.0.0.beta-12
- Update imports to stop using default imports

### Reason for Changes

Handle breaking change made in v11.

### Test Coverage

Tested manually

### DIFF
--- a/examples/blockly-react/package.json
+++ b/examples/blockly-react/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@blockly/field-date": "^8.0.0",
-    "blockly": "^10.0.0",
+    "blockly": "^11.0.0-beta.12",
     "react": "^17.0.1",
     "react-datepicker": "^3.3.0",
     "react-dom": "^17.0.1",
--- a/examples/blockly-react/package.json
+++ b/examples/blockly-react/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@blockly/field-date": "^8.0.0",
-    "blockly": "^11.0.0-beta.12",
+    "blockly": "^10.4.3",
     "react": "^17.0.1",
     "react-datepicker": "^3.3.0",
     "react-dom": "^17.0.1",
--- a/examples/blockly-svelte/src/components/BlocklyComponent.svelte
+++ b/examples/blockly-svelte/src/components/BlocklyComponent.svelte
@@ -1,6 +1,6 @@
 <script>
   import { onMount } from 'svelte';
-  import Blockly from 'blockly';
+  import * as Blockly from 'blockly';
 
   onMount(() => {
     Blockly.inject(document.getElementById("blocklyDiv"), {
--- a/examples/blockly-svelte/src/components/BlocklyComponent.svelte
+++ b/examples/blockly-svelte/src/components/BlocklyComponent.svelte
@@ -1,8 +1,11 @@
 <script>
   import { onMount } from 'svelte';
-  import * as Blockly from 'blockly';
+  import * as Blockly from 'blockly/core';
+  import * as En from 'blockly/msg/en';
+  import 'blockly/blocks';
 
   onMount(() => {
+    Blockly.setLocale(En);
     Blockly.inject(document.getElementById("blocklyDiv"), {
       toolbox: document.getElementById("toolbox")
     });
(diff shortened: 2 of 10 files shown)


BeksOmega:
Since #2362 covers three classes of issues, can you split it up into three issues? [Edit, that way we can close it, which is more visible in the github project]

examples/blockly-svelte/src/components/BlocklyComponent.svelte
```diff
@@ -1,6 +1,6 @@
 <script>
   import { onMount } from 'svelte';
-  import Blockly from 'blockly';
+  import * as Blockly from 'blockly';
```
BeksOmega: This should import each thing it needs individually as specified [here](https://developers.google.com/blockly/guides/get-started/get-the-code#imports). If you don't want to fix that now, feel free to file a bug to follow up.

Here and in Vue.

examples/blockly-react/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@blockly/field-date": "^8.0.0",
-    "blockly": "^10.0.0",
```
BeksOmega, line 7: Did you want to commit these changes? Slightly concerned that only updating a few of them will mean we miss updating them to the actual release when we do that. We've definitely missed things doing find-and-replace because of stuff like this.

BeksOmega (approved):

examples/blockly-react/package.json
rachel-fenichel: @maribethb said she usually searches for 'beta' to make sure she updates everything; I don't have a strong opinion, since I think the updated imports will work in either case.

examples/blockly-svelte/src/components/BlocklyComponent.svelte
rachel-fenichel: I don't need the generators, but I do need to have core, library blocks, and english langfiles.

I updated the imports to 
```
  import * as libraryBlocks from 'blockly/blocks';
  import * as En from 'blockly/msg/en';
  ```
and added a call to `Blockly.setLocale(En)`. This (and the `inject`) call fail because `Blockly` is `undefined`.

rachel-fenichel: I applied the same change to the Vue example and it worked fine. Do you know what the difference is? @cpcallen to comment as well.

examples/blockly-react/package.json
maribethb: Since they should work in v10 as well, it's fine either way imo. We definitely have missed some in the past but I do search the term beta when I do the rc branch merge so honestly I'm more likely to find this than if it were on some random 10.x version

examples/blockly-svelte/src/components/BlocklyComponent.svelte
maribethb: The Vue demo is built using Vite. The Svelte demo is built using rollup. Gonna guess some of the packaging changes aren't compatible with our rollup config, but our config is also pretty out of date (e.g. we're using some commonjs plugin whose git repo was archived 3 years ago) so who knows where the problem lies. This probably warrants Christopher's eyes.

I would probably start by trying a new rollup config from scratch if they have an updated hello world example.

I'll also note that rollup says it is the bundler behind Vite, so the fact that it works over there seems promising. More likely to point to a problem in our specific rollup config than a deeper incompatibility

rachel-fenichel: Fixed by updating the version of rollup and some of their plugins. Refreshingly, everything that needed to be replaced or updated was clearly labeled.

rachel-fenichel:
@BeksOmega this is ready for re-review.
Changes I made:
- Updated dependencies and import statements for vue example
- Went back to latest (instead of beta) in react demo, to avoid a peer dependency conflict with the date field
- Formatted

BeksOmega (approved):
Sweet! I don't know enough about any of these frameworks to approve in the truest sense. But sounds like you worked through a bunch of stuff. And this isn't load bearing code. So LGTM

PR #2377, opened 2024-05-28
chore: add test block that duplicates on drag

BeksOmega:
## The basics

### Resolves

Fixes N/A

### Proposed Changes

Adds a test blocks that duplicates on drag so we can ensure we don't regress in support for that functionality.

### Test Coverage

Tested that with my local version of Blockly the block duplicates both when it is a real block and when it is a shadow.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/plugins/block-test/src/drag.js
+++ b/plugins/block-test/src/drag.js
@@ -10,6 +10,47 @@
  * @author samelh@google.com (Sam El-Husseini)
  */
 
+import * as Blockly from 'blockly/core';
+
+/** Block that duplicates itself on drag. */
+Blockly.Blocks['drag_to_dupe'] = {
+  init: function () {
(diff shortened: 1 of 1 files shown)


cpcallen (approved):
Please fix lint warnings.

Is there (or will there be) a corresponding Blockly to add a test that exercises this functionality?  If so, please add a comment to this PR referencing it.

BeksOmega:
I'm not going to fix the lint warnings. imo the code is readable and sufficient for testing. I think for test code it's allowable to ignore the warning - that's why it's not an error.

ATM there's not going to be a test in core exercising this - it will just need to be manually tested. It might be possible to test this with a webdriver test? but dealing with the circular dependencies to get that working is not a high priority for me right now.

PR #2392, opened 2024-06-21
feat: use a js legal name for the block in code output

maribethb:
## The basics

### Resolves

Fixes 

### Proposed Changes

- Gets a JS-legal name for the block to use in output generators so that we generate valid json/js

### DIFF
--- a/examples/developer-tools/src/output-generators/factory_base.ts
+++ b/examples/developer-tools/src/output-generators/factory_base.ts
@@ -26,13 +26,21 @@ import {
 } from './generator_stub_generator';
 
 /**
- * Gets a JavaScript-legal name for the block.
+ * Gets a name for the block that is likely to be JavaScript-legal.
+ * Replaces any character that isn't a digit, ascii letter, underscore,
+ * or dollar sign with an underscore. Prepends names that start with
+ * digits with an underscore.
+ *
+ * This does not check for reserved words in JavaScript, so the name isn't
+ * guaranteed to be a legal identifier.
+ * This also disallows what would be legal identifiers that use unicode
+ * letters.
  *
  * @param name user-entered name.
- * @returns legal name with underscores replacing spaces.
+ * @returns legal-ish name.
  */
 const getLegalBlockName = function (name: string): string {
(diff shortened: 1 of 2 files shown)


examples/developer-tools/src/output-generators/factory_base.ts
```diff
@@ -25,6 +25,16 @@ import {
...
 } from './generator_stub_generator';
 
+/**
+ * Gets a JavaScript-legal name for the block.
+ *
+ * @param name user-entered name.
+ * @returns legal name with underscores replacing spaces.
+ */
+const getLegalBlockName = function (name: string): string {
+  return encodeURI(name.toLowerCase().replace(/ /g, '_').replace(/[^\w]/g, ''));
```
cpcallen: I'm not sure what restrictions might exist on the contents of `name`, but if it's an arbitrary string then:

1. This rejects certain characters that are legal in JS identifiers (in particular `$` and all the Unicode letters that are not in ASCII).
2. This does not reject leading digits.
3. This does not reject the empty string.

The first might at worst make some dev grumpy, but the latter two could still result in invalid code being generated.  (Of course there may be other validation code elsewhere that I've not seen that prevents that from happening.)

cpcallen (approved):

examples/developer-tools/src/output-generators/factory_base.ts
maribethb: OK, I had originally copied this from the old block factory but it's not enough here. The actual regex for determining if something is a legal js identifier is quite long, and can go out of date as reserved words are added to the es spec. So I compromised here by allowing only ascii letters, numbers, underscore, and `$`

1. allowing non-ascii unicode characters balloons the size of the regex, and i'm also not sure is really a good idea for identifiers anyway; if a developer wants them they can always change the name themselves
2. it does now, thanks!
3.  i made the validator for the field itself reject the empty string now, so the name can never be set to that in the first place

the net result is you can type almost any name into the name field of the block and keep it, but in the generated code we'll use much more boring names with a lot of underscores.

I didn't bother trying to reject reserved words because I don't want to maintain a list of them. at the end of the day the developer will have to take some responsibility for their name choices :)

PR #2417, opened 2024-07-11
fix: dev-tools only set monaco model and text when changing

HollowMan6:
## The basics

### Resolves

Fixes:
- https://github.com/mit-cml/workspace-multiselect/pull/62#issuecomment-2196819235
- https://github.com/mit-cml/workspace-multiselect/pull/62#issuecomment-2212561339
- https://github.com/mit-cml/workspace-multiselect/pull/62#issuecomment-2214203536

### Proposed Changes

This commit introduces a workaround, so that we only update the page when we have model or text updates in monaco.

### Reason for Changes

In multi-select plugin, a significant amount of workspace change events (for each block) can be triggered at the same time, which causes the garbage collection mechanism to fail, so that we eventually have JS heap overflow and the page crashed in Chromium.

### DIFF
--- a/plugins/dev-tools/src/playground/index.js
+++ b/plugins/dev-tools/src/playground/index.js
@@ -175,6 +175,12 @@ export function createPlayground(
               wordWrap: true,
             });
           }
+          if (
+            generateModel.getValue() === text &&
+            editor.getModel() === generateModel
+          ) {
+            return;
+          }
           generateModel.pushEditOperations(
             [],
             [{range: generateModel.getFullModelRange(), text}],


cpcallen (commented):
Hi HollowMan6.  Interesting situation.

Is there a bug open against Monaco and/or Chrome for this apparent garbage collection failure?

The workaround code seems perfectly reasonable but I would want it to be documented with reference to a bug so that it can be removed when the bug is fixed.

HollowMan6:
Hi @cpcallen! Thanks for taking a look at this PR.

> Is there a bug open against Monaco and/or Chrome for this apparent garbage collection failure?

No, I haven't filed a bug about this, and I guess this might not even be qualified as a "bug" on their side, as I saw that we update the editor with the same text almost 30 times at the same time when that garbage collection fails, which is unreasonable.

> The workaround code seems perfectly reasonable but I would want it to be documented with reference to a bug so that it can be removed when the bug is fixed.

Actually, I think it might be a good idea to have this workaround code even if the bug is fixed somehow, as we don't need to disturb the editor if the text and model are unchanged (so this might even improve the performance).

cpcallen:
> > Is there a bug open against Monaco and/or Chrome for this apparent garbage collection failure?
> 
> No, I haven't filed a bug about this, and I guess this might not even be qualified as a "bug" on their side, as I saw that we update the editor with the same text almost 30 times at the same time when that garbage collection fails, which is unreasonable.

I'd grant you that it is _foolish_, but the it is the crash which is unreasonable.  Either there is:

1. a GC failure in Chrome (bug!), or
2. a memory leak in Monaco (bug!), or
3. Monaco is for some reason keeping old copies of the text being edited around for some non-obvious reason (documentation bug?)

Especially if this does not occur in Firefox then it seems like it is probably 1 (and certainly not 3), so I think this deserves to be reported upstream.  Maybe reporting it to Monaco in the first instance, and let them report it upstream?

Nevertheless…

> Actually, I think it might be a good idea to have this workaround code even if the bug is fixed somehow, as we don't need to disturb the editor if the text and model are unchanged (so this might even improve the performance).

Upon reflection I agree with you about this.

cpcallen (approved):

HollowMan6:
Sure, I will file a bug report to monaco first.

> 2. a memory leak in Monaco (bug!), or
> 3. Monaco is for some reason keeping old copies of the text being edited around for some non-obvious reason (documentation bug?)

Edit: Monaco asks to reproduce the bug in the playground, but it seems like I can't reproduce it anyway in their playground (only see heap size surges but drops quickly), so it shouldn't be 2 and 3.

So I think it's more related to Chrome or even V8 JavaScript engine, about their memory management, when computing performance drops (e.g. with heavy rendering load), the garbage collection fails to start, file a Chrome issue should be a proper choice.

https://microsoft.github.io/monaco-editor/playground.html?source=v0.50.0#XQAAAAI9RwAAAAAAAABBqQkHQ5NjdMjwa-jY7SIQ9S7DNlzs5W-mwj0fe1ZCDRFc9ws9XQE0SJE1jc2VKxhaLFIw9vEWSxW3ysczeL4Lmawgw1sSReuQHxcPyDezRJO_M2pkpc0T117FhARU4R23lQUIo3WrrJThDpF_y9MWmzGjue3EOOefJj63Ut2VlSjohz4Kh2_HNGHc7QfiKWldBtILiZifrRqp0ABIkmTjqdC7mxJRGK6jVNZHZUqClvSFpkwcHTQAuhti3aNBwkJ5Sm3nZDQjuGmMyg9DG8ax2k0xUe6gOxW6t7AlF5s7tsP9HuR5e1vlEH21dCw_ZzpGdXikwqMUJzNEWtb8o24mwo0fnzI9kuKenFA0GKUzjpNPDqHmBcn276h_S0FBVdK1koLwAw6r320F95IWsXWoKznGPULdxEuGi1kSDl-xjSVpprgGbxzHDxeQCCFKSNGYRDatUktSTKGr7hCzJ_I4YGkHiwquqwQPbFifTitsOytba2cbJH6Hdau52MOEoUkddaWi7tmNGm1kswvQirnxQaEynnWerJozdGf0DjA4m1SFwvTKJPkX1KwFR-z4N9zLO5IapMk_ECFd_C-CXgZ6NdaGTIOa_wnpgyNb8oy3BLIN1dAZwO34GpVhoT0CZUbY6-j73aZgB8HnqJDeW6Cros9qH6xPstB9wsMZfl9tpmc2gtvgNVmC2QMBKFJgQ6XsKmExhNQoTV0-Md73hRlQCG2w5Yj5nm9Kn23kYoVGkoCiUBrkZe8mn-eEJ0YAxRJDR7gI-qJHiZNvmQC3ovLQHJ1qC83iZ0LpbheeVpzugBOSM7KosWsWMA9jx96hVyU-fAVIBRpDVCDZNCbZM3tEr5U9GftkNErbcU2-PWDDIRGCNzpydC0ELQS1qSnNufk1RJ4RfYUtIp08x4Hhqf9y5QeGVeF49ZoILpO_DM4u0wF8SSTolJ3MVjCbWfhdveybvG-SNxZaINiFXuvdfyUWvNK8q0-FD2-BdWv9e1rIKzJ-cMGatOeGLu5_8gAAPeuJjbnPuwEZzP86qX4gjquUDJjxx63nZ40JynKPoO0rjJ1CrerA3msAMx82OXqd3hO7FhTUUEGANBIXJWA49bhgnTWZcGNbVSyZt6bR0lAupNOAB_NOq8PlTYr7d_WgRJn_gqa1f7TFUzQnUXDij73OSdUQYSboXmaQXYGULbC6DnyIWBmlG3H4ljMOUCTpslmpXBo2Z2ZRvjFyCi3sfE_5-Uqectn-SIMTdi5J33a5w2tUfnds7a9AtrQMWqlqijFE_F3S72I9c5DfuphIZMpEFS-xyOlHijq-9NrVpyDySUqszEPA_ZlbTHkyA3vbVPkYPPdEZ5eC34DyuJV6TE9CVpA9OtWszSCxvHaH2z5aXLtqM7c1iJLj38xi4BOqUqrOeDH8FZ2h4OrH29OrTGzIVOT19CLEeggQTVLNFWpMV-cM_gRKc3pTPv7GU0TfrT2uMwlQ0UibX3A3ddEMdg8INhD0qPPehcV0f-DDcV5rSUS6cvGcmRWlE3qso0X_FrT2csIpjum8KhI7-sosCM5svA59OyYPUxyOHac2T45tfkDHQ3gpEh_u1iOUdWGfMooXHqKxRFVUqC_lHmF1ur4RbH-LjKXQR4QE6U89CMtojqkk84EbGiNBboPfn9Ew37CResyHzJxGgqYNVqHQkdRTeX2zMWMPQ80--qVFgxhEpb2tGInUXlJT4YwMrdjCHDZy5kISWpq4F_O7Ly5kW0SHEg2YxcAM9VogJYoSt-BtnXFwy5j6kcjLSVOM4wpjsbrw0_GNq1wFeyxTnUZKvcQAeGUe5FrthyT3YYZqhnefGD5iZCEWsaQBgllN3zEvON4fyfN7eJYzLVpHIEICTY2r6FB90twbfLTiL0qRKyA0vw4jLy3MwHpXjVS-NXq10skgHM-YP_QPc8enbSIsJuqzn1ubfl-oL5b1hJHezAjHra4LfNfNQccXbRu8zBLNoEN1eOyHbKAMyMQlY9QCnwKgLODB0J0Myn_UYXZ_IB2N9MAAZfXRL5EDhXikwpdZaZQhmqs1uvbP4AcHb6VvMqx34SXtZAt34p2oapQXZm3fAfUpxj7aGJi2XzzqXlyiqgZKhrSNwf0qCuH_JNuEGDYDBbAXoobJQ9rXW4iSI_X1zT6NdEraUlVmHFV8wX-zS-iK6gIN5GY_WWLM_MlVKRgEbaQQm2GpfPAc-GE951wT9jqIikB91EsJUJpgWozvEk4k3jt7tWfINGX93ZYqQ4huFPgB1RBbLyMhNUruPx0NC3ak7_a6bRfWPgJAf7wmNv0w_tCRfqGn5Ozc9F8ngdyEhWJtMxVZyLi7r3U4TrQIv-0LFc-iSXJ8aavIfq8M9Nl1SDcHK9nSmEOl48G-e5PkWnRILy_Un2Po-lglx2L-d9rK4i6r6_tcsnyJOXIPfyjzvyeCESLZ2Kfs93p9r5BbsHp7L0Zk7RWVdCJLUg-c0yqKhEi3uVPpHQ8KGVeo0S1CICXFF6Ni9eSiuZj3rzcCuqcz1JqK2eaHKU7yiFhhC9rdeQ4AwhgfON4Dqc77P8vBGoygd_WTHCeH0DZ7szpQ18-b6vHFlEkFw4P54jHGNETB1nZCPbvSjXZCkvZqSIqBb8CAvmE8wVuckEeQuH5CJdUVG24MPLj3F2dH1VbrSZhnTG7IMwewrb6ISgRZs2D5BxtnWP-YiB0yjLz7DLGaXob1fFNESg_RHl5gjPkTOP6xJwo2UAEYnsQSjycfsE1m32F6HtjtpPPCAeOKTXZyf0WgHBCn2we9tMnzaKi3xTFMon3XwA7lR1rZnJ04bPrAsC9RDVFhsMlfEanqkI08E4_3kHyr2dFLkWVJ6-guRNAjU05GGri0EsDP_uQ1cPs0TwfB29DB-GZCRoDrqnM0ZS6qRzHKKyT5brxk39hEGDT-TwSdJu4ovAG4ztfKGjHUDKrfcdSk5tH7gFqOuqYjCHWeTqSFyJu2nisGwxZcmoJIzJJOb9KnsjVyJWmiqoNEZWyB83f1G0hl9wVRWmDuKuAz4Z33XU1T2oSFzlBFV9CSZrUcwdyCrG-LuLns3ORp4AXrsPu80UTfp1MqxPjSpHjy4FmOro3rnbqdeBkGG8HaEiqfIk2g9d4ueH5iMUtMD9UpjQPe40a5R4ilcl5vpBIb8GFTHp1BCJ09k3Bi6E_yrRecDDxEnStuP4aQokOSgOxq3xtjKf1cUC_F3g4bzzdBw2hKXOEUdyts2D13wU_2HMmQCdx4WO4sz9i0NIMrWRpujTBkXwyj1_MkXapWjCcET815gAjj_sndhv2BbzETkJYzhxyD8I-EAyxkd2YSbKx5JkdUZc10LNEjErE5iAbxJYtU2NZ-a1PBHAG6-bjNYZm_C_-1mwJo6UWqLsRQUVDs86QKJyEtbdVh1e7KzW0GvskRzwGPJff_LwQJv0AGu8cSYpqauD_ByDwMnKDgeB13zxDwcYG_5gqmxxaYHGYAKWb8k50TYyBflQQoS5oap0-a5Rfsm2MsHGLlm_MPdZHJjUuic1apch5tQb5qM_F8HIEN4986qUaTKPTP1wTm-_YNo-Qr1-VPSPJMv4zbQ5uon8I8c4Sgdq_FLmiJTrneVr4x3sY_EXpiHE9Xjfm6HDNet4Chc0CiNs8YblIunu8EPc-WvtMHAE-iNza33CzLnP6_80ko6bKOuRnlERZMgMQ0NoffQb8Bz2YkdgGRHPgCRkdSfPXoq35OEnvSNJCOAQ5K2hDfEVpYG9r2wRceMoPXxJ_TdrHtirxgaf-rdCDz1cy0hhWpwVxfGzC_Zbtm_16oFWXKzwtnR_S03_95ot8A

PR #2453, opened 2024-08-30
fix: add lerna commands to plugins package.json

rachel-fenichel:
## The basics

### Resolves

Fixes an issue where lerna commands could not be run from the root `package.json` file. For instance, the `build` script is `cd plugins && lerna run build`, which was failing.

The issue was that `npm run` uses the contents of `node-modules/.bin` in the directory where it is run. The command was run at the root level, meaning that lerna was expected to be installed at the root level, but due to https://github.com/google/blockly-samples/pull/2446 lerna is only installed inside `examples/` and `plugins`.

Calling `cd plugins && npm run build`, with `build` defined as `lerna run build` in the plugin `package.json`, resolves this issue by changing which directory lerna is expected to be installed in.

Verified by running several of the changed commands from the top level, including `build`, `deploy:prepare`, and `clean`.

### Proposed Changes

- Add several commands to the plugins `package.json` file, all of which just call lerna commands.
- Change root `package.json` scripts to use `npm run <command>` instead of `lerna run <command>` after changing directories.

### Additional Information

I'm moderately concerned by other changes in `blockly-samples` passing PR checks when this was broken. What were they using for lerna? How was building and testing working?


cpcallen (approved):

plugins/package.json
```diff
@@ -3,8 +3,14 @@
   "private": true,
   "scripts": {
     "audit": "lerna run audit",
+    "audit:fix": "lerna run audit:fix",
     "boot": "lerna bootstrap",
-    "postinstall": "npm run boot"
+    "build": "lerna run build",
+    "clean": "lerna run clean",
+    "clean:node": "lerna run clean --yes",
```
cpcallen, line 10: This should have been `lerna clean --yes`.

PR #2465, opened 2024-10-19
fix: Modify Webpack sourcemap parameters to account for Windows path separators

Varriount:
## The basics

### Proposed Changes

This PR modifies the Webpack configurations throughout the repository such that:
  - The regular expression used to determine which sourcemaps to load/look for works on Windows systems.
  - The regular expression used to determine which sourcemap parsing errors to ignore only covers paths containing the term "blockly"

### Reason for Changes

Prior to this change, Blockly source maps wouldn't load on Windows systems, as Windows uses a backslash as a path separator, not a forward slash. The parsing error regex was modified as it occurred to me that it was overly broad, and might hide meaningful errors over the course of regular development.

### DIFF
--- a/examples/custom-generator-codelab/webpack.config.js
+++ b/examples/custom-generator-codelab/webpack.config.js
@@ -46,14 +46,14 @@ module.exports = (env, argv) => {
 
     // Include the source maps for Blockly for easier debugging Blockly code.
     config.module.rules.push({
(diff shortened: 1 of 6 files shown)


rachel-fenichel:
@cpcallen can you take this one since it hits the build system?

cpcallen (approved):

PR #2468, opened 2024-11-08
refactor!: Update and modernize the continuous toolbox plugin.

### LINKED ISSUE #857: continuous-toolbox not created function

### LINKED ISSUE #1041: Importing continuous-toolbox has side effects
Description:
I have my own Blockly workspace React component which has an attribute "isContinuous". When I set this to true, I would like to have the continuous toolbox displayed. When it's false, I would like to have a standard category toolbox.

The problem:
Regardless if I use the continuous-toolbox plugin or not, importing "@blockly/continuous-toolbox" into my jsx file has side effects and my toolbox categories are rendered differently even when I don't use the plugin in the workspace configuration.

Expected behaviour:
The toolbox categories should retain their standard visual representation even when I import "@blockly/continuous-toolbox" into the project.

### LINKED ISSUE #1922: Continuous toolbox flyout auto close
**Category**

 - Plugins

**Component**

Continuous toolbox

In normal toolbox, when dragging a block from flyout, it autoclose. But continous toolbox flyout stays always opened.

In API a function or parameter to define autoclose or not.

## To Fix

1) Add a `setAutoclose` method to the `ContinuousToolbox`.
2) Modify the [change listener](https://github.com/google/blockly-samples/blob/646fcbc309a86639d17b2f966e5d2d75e50c9f0d/plugins/continuous-toolbox/src/ContinuousToolbox.js#L35) to only `refreshSelection` if the flyout is visible.
3) Modify [`updateFlyout`](https://github.com/google/blockly-samples/blob/646fcbc309a86639d17b2f966e5d2d75e50c9f0d/plugins/continuous-toolbox/src/ContinuousToolbox.js#L84) to `refreshSelection`.

### LINKED ISSUE #2464: Update continuous toolbox plugin for flyout API changes
### Check for duplicates

### Component

Continuous Toolbox plugin

### Description

The flyout API is changing in Blockly v12 (see https://github.com/google/blockly/pull/8601), and the continuous toolbox plugin will need to be updated for compatibility. Most of the needed changes have already been made in https://github.com/gonfunko/scratch-blocks/pull/209#pullrequestreview-2352707135 and will just need to be backported, but we'll need to coordinate merging/publishing to coincide with the v12 release.

### Reproduction steps

_No response_

### Stack trace

_No response_

### Screenshots

_No response_

gonfunko:
## The basics

### Resolves

* Fixes #2464
* Fixes #1041
* Fixes #857
* Fixes #1922

### Proposed Changes
This PR updates the continuous toolbox plugin for compatibility with the revised flyout APIs introduced in Blockly v12. It also converts the plugin to Typescript.

Additionally, it fixes several existing issues:
* Importing the plugin is free of side effects; the plugin must be registered and injected to take effect. Previously, the CSS and some classes were unconditionally registered.
* The continuous flyout now supports autoclosing; users can simply call `Blockly.getMainWorkspace().getFlyout().setAutoClose(true)` to enable this behavior.
* Mutations to procedure blocks in the main workspace will be immediately reflected in the flyout.
* Extremely poor performance when adding/deleting large numbers of blocks to the workspace has been resolved by debouncing toolbox refreshes in response to block add/create/change events.
* The implementation has been slightly adjusted to make it more extensible in response to shortcomings identified by the Scratch modernization effort.

This PR should not be merged until Blockly v12 is released, as it depends upon APIs added in that version.


cpcallen:
> This PR should not be merged until Blockly v12 is released, as it depends upon APIs added in that version.

Might it be worth creating a v12 branch in this repo (if there is not already one) and retargeting this PR to it?

gonfunko:
> > This PR should not be merged until Blockly v12 is released, as it depends upon APIs added in that version.
> 
> Might it be worth creating a v12 branch in this repo (if there is not already one) and retargeting this PR to it?

Done!

gonfunko:
This also fixes #2471.

gonfunko:
And also fixes https://github.com/google/blockly/issues/8543

cwillisf:
Hello! Any chance this could be built into a beta release or something like that? I'm attempting to use `blockly@12.0.0-beta.0` but there's no corresponding compatible `continuous-toolbox@x.y.z-beta.q`. I can work around that for local dev, but I'm not sure I want to introduce `npm link` into our CI/CD pipeline even for staging :)

plugins/continuous-toolbox/src/ContinuousMetrics.ts
```diff
@@ -12,24 +12,26 @@ import * as Blockly from 'blockly/core';
...
-   * @override
+   *
+   * @param getWorkspaceCoordinates True to get the view metrics in workspace
+   *     coordinates, false to get them in pixel coordinates.
+   * @returns The width, height, top and left of the viewport in either
+   *     workspace coordinates or pixel coordinates.
    */
-  getViewMetrics(getWorkspaceCoordinates = undefined) {
+  override getViewMetrics(
+    getWorkspaceCoordinates?: boolean,
```
cpcallen: I think you could have changed this to 
```suggestion
    getWorkspaceCoordinates = false,
```
and effected the same type change while making the default clear.  Or do you want to be able to distinguish between "no value supplied" and "value supplied is `false`" cases?

plugins/continuous-toolbox/src/ContinuousMetrics.ts
```diff
@@ -52,11 +54,13 @@ export class ContinuousMetrics extends Blockly.MetricsManager {
   }
 
   /**
-   * Moves the absoluteLeft and absoluteTop so they no longer include the
-   * flyout.
-   * @override
+   * Gets the absolute left and absolute top in pixel coordinates.
+   * This is where the visible workspace starts in relation to the SVG
+   * container, shifted to not include the area behind the flyout.
```
cpcallen: Perhaps "adjusted" rather than "shifted"?  The latter might imply something is actually moving.

plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -59,7 +59,6 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
     super(value, validator, config);
 
     this.SERIALIZABLE = true;
-    this.CURSOR = 'default';
```
cpcallen, line 62: I'm not sure I understand why you are removing these.  Presumably they are no longer needed, but I'm not sure why not.

plugins/content-highlight/src/index.ts
```diff
@@ -13,7 +13,7 @@ import * as Blockly from 'blockly/core';
 /**
  * List of events that cause a change in content area size.
  */
-const contentChangeEvents = [
+const contentChangeEvents: string[] = [
```
cpcallen, line 16: This type change is to allow extensibility?

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+ */
+
+import * as Blockly from 'blockly/core';
+import {ContinuousToolbox} from './ContinuousToolbox';
+import {ContinuousFlyoutMetrics} from './ContinuousFlyoutMetrics';
+import {RecyclableBlockFlyoutInflater} from './RecyclableBlockFlyoutInflater';
+
+interface LabelFlyoutItem extends Blockly.FlyoutItem {
+  getElement(): Blockly.FlyoutButton;
+}
```
cpcallen: If a `LabelFlyoutItem` is a flyout item that is a label, it is surprising that its `getElement` method returns a `FlyoutButton`.  If that's not an mistake, perhaps a comment explaining why might be in order?  🤔

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+    super(workspaceOptions);
+
+    this.getWorkspace().setMetricsManager(
+      new ContinuousFlyoutMetrics(this.getWorkspace(), this),
+    );
+
+    this.getWorkspace().addChangeListener((e: Blockly.Events.Abstract) => {
+      if (e.type === Blockly.Events.VIEWPORT_CHANGE) {
+        this.selectCategoryByScrollPosition(-this.getWorkspace().scrollY);
+      }
```
cpcallen, line 63: I guess it should be safe to do this on any `VIEWPORT_CHANGE` rather than just on ones affecting the flyout's workspace.  (Reasoning: if calling `selectCategoryByScrollPosition` when triggered by a `VIEWPORT_CHANGE` on some other workspace, we should _already_ have had a `VIEWPORT_CHANGE` event for the flyout workspace.)

But `selectCategoryByScrollPosition` is not the cheapest operation, so it seems worthwhile to avoid calling it if the flyout was not affected.

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+    this.scrollPositions.clear();
+    this.getContents()
+      .filter(this.toolboxItemIsLabel.bind(this))
+      .map((item) => item.getElement())
+      .forEach((label) => {
+        this.scrollPositions.set(
+          label.getButtonText(),
+          label.getPosition().y - label.height,
+        );
+      });
```
cpcallen, line 94: No objection to this style of functional pipelining in general, but I think it's always worth considering whether a more straightforward imperative version would be more readable or not.  In this case I think there's not much in it one way or the other—but feel free to decide for yourself:

```suggestion
    for (const item of this.getContents()) {
      if (!this.toolboxItemIsLabel(item)) continue;
      const label = item.getElement();
      this.scrollPositions.set(
        label.getButtonText(),
        label.getPosition().y - label.height,
      );
    }
```

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+   * @returns True if the item represents a label in the flyout, and is a
+   *     Blockly.FlyoutButton.
+   */
+  protected toolboxItemIsLabel(
+    item: Blockly.FlyoutItem,
+  ): item is LabelFlyoutItem {
+    const element = item.getElement();
+    return !!(
+      item.getType() === 'label' &&
+      element instanceof Blockly.FlyoutButton &&
```
cpcallen, line 111: OK, so… labels _are_ a kind of button?  As noted above this is surprising and probably deserves some further explanation.

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+   * Selects an item in the toolbox based on the scroll position of the flyout.
+   *
+   * @param position Current scroll position of the workspace.
+   */
+  private selectCategoryByScrollPosition(position: number) {
+    // If we are currently auto-scrolling, due to selecting a category by
+    // clicking on it, do not update the category selection.
+    if (this.scrollTarget) {
+      return;
+    }
```
cpcallen: Better to keep these guard clauses to a single line:
```suggestion
    if (this.scrollTarget) return;
```

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+  /**
+   * Map from category name to its position in the flyout.
+   */
+  private scrollPositions = new Map<string, number>();
+
+  /**
+   * The percentage of the distance to the scrollTarget that should be
+   * scrolled at a time. Lower values will produce a smoother, slower scroll.
+   */
+  protected scrollAnimationFraction = 0.3;
```
cpcallen, line 41: 1. From reading `stepScrollAnimation`, it looks like this is a _fraction_, not a _percentage_ (it gets used without dividing it by 100).

2. If I have understood correctly, the default value chosen here results in the scroll being completed within four frames; that seems very fast (see below).

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+
+    const currentScrollPos = -this.getWorkspace().scrollY;
+    const diff = this.scrollTarget - currentScrollPos;
+    if (Math.abs(diff) < 1) {
+      this.getWorkspace().scrollbar?.setY(this.scrollTarget);
+      this.scrollTarget = undefined;
+      return;
+    }
+    this.getWorkspace().scrollbar?.setY(
+      currentScrollPos + diff * this.scrollAnimationFraction,
```
cpcallen, line 186: How hard would it be to adjust this for the display refresh rate?  Notably the default `scrollAnimationFraction` makes the animation finish in a brisk  0.06s on a 60Hz monitor, but a barely-perceptible 0.3s on my laptop's 120Hz display.  (I actually wondered if the animation was 

—perhaps not excessively fast on a typical 60Hz monitor, but on my laptop's built-in display the 120Hz refresh rate makes the animation so fast it's almost imperceptible.

(I think it would be fine to leave this to a later PR.)

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+   *
+   * @param e The mouse wheel event to handle.
+   */
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  protected override wheel_(e: WheelEvent) {
+    // Don't scroll in response to mouse wheel events if we're currently
+    // animating scrolling to a category.
+    if (!this.scrollTarget) {
+      super.wheel_(e);
+    }
```
cpcallen: Keep main flow outdented where possible by using guard clauses (especially appropriate here given wording of comment):

```suggestion
    // Don't scroll in response to mouse wheel events if we're currently
    // animating scrolling to a category.
    if (this.scrollTarget) return;
    super.wheel_(e);
```

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+    if (this.scrollPositions.size > 0) {
+      const lastPosition =
+        ([...this.scrollPositions.values()].pop() ?? 0) *
+        this.getWorkspace().scale;
+      const lastCategoryHeight = contentMetrics.height - lastPosition;
+      if (lastCategoryHeight < viewMetrics.height) {
+        return viewMetrics.height - lastCategoryHeight;
+      }
+    }
+    return 0;
```
cpcallen: Guard clauses?

```suggestion
    if (this.scrollPositions.size === 0) return 0;
    const lastPosition =
      ([...this.scrollPositions.values()].pop() ?? 0) *
      this.getWorkspace().scale;
    const lastCategoryHeight = contentMetrics.height - lastPosition;
    if (lastCategoryHeight > viewMetrics.height) return 0;
    return viewMetrics.height - lastCategoryHeight;
```

Also, is the `?? 0` needed on line 218?  Surely the check on `.size` means there is guaranteed to be at least one entry?  Or is that just to satisfy the type checker?

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  /**
+   * Returns the X coordinate for the flyout's position.
+   */
+  override getX(): number {
```
cpcallen, line 234: Despite the comment in the body I'm still not completely sure I understand what this method is doing when it opts not to make the `super` call.

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+    if (
+      this.isVisible() &&
+      this.targetWorkspace.toolboxPosition === this.toolboxPosition_ &&
+      this.targetWorkspace.getToolbox() &&
```
cpcallen, line 240: I am not clear on the purposes of these checks.

plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -0,0 +1,292 @@
...
+   *     flyout.
+   */
+  override show(flyoutDef: Blockly.utils.toolbox.FlyoutDefinition | string) {
+    super.show(flyoutDef);
+    this.recordScrollPositions();
+    this.getWorkspace().resizeContents();
+    if (!this.getParentToolbox().getSelectedItem()) {
+      this.selectCategoryByScrollPosition(0);
+    }
+    this.getRecyclableInflater().emptyRecycledBlocks();
```
cpcallen, line 263: I'm not clear on how this might interact with the rest of the code in this method, but I'm wondering why you don't empty the recycling bin first.

plugins/continuous-toolbox/src/ContinuousFlyoutMetrics.ts
```diff
@@ -0,0 +1,47 @@
...
+  ) {
+    const scrollMetrics = super.getScrollMetrics(
+      getWorkspaceCoordinates,
+      cachedViewMetrics,
+      cachedContentMetrics,
+    );
+    const contentMetrics =
+      cachedContentMetrics || this.getContentMetrics(getWorkspaceCoordinates);
+    const viewMetrics =
+      cachedViewMetrics || this.getViewMetrics(getWorkspaceCoordinates);
```
cpcallen, line 38: I started off merely suggesting use of `??` instead of `||`,  but then wondered if this alternative might be better: it's not semantically equivalent but might be more efficient as it avoids duplicate calls to `getContentMetrics` etc. at each stage of the super call chain:

```suggestion
  override getScrollMetrics(
    getWorkspaceCoordinates?: boolean,
    viewMetrics = this.getViewMetrics(getWorkspaceCoordinates),
    contentMetrics = this.getContentMetrics(getWorkspaceCoordinates),
  ) {
    const scrollMetrics = super.getScrollMetrics(
      getWorkspaceCoordinates,
      viewMetrics,
      contentMetrics,
    );
```

(We could refactor the superclasses similarly.)

plugins/continuous-toolbox/src/ContinuousToolbox.ts
```diff
@@ -0,0 +1,203 @@
...
+      let itemContents = toolboxItem.getContents();
+
+      // Handle custom categories (e.g. variables and functions)
+      if (typeof itemContents === 'string') {
+        itemContents = [
+          {
+            custom: itemContents,
+            kind: 'CATEGORY',
+          },
+        ];
```
cpcallen: Unless Prettier objects:
```suggestion
        itemContents = [{custom: itemContents, kind: 'CATEGORY'}];
```

plugins/continuous-toolbox/src/ContinuousToolbox.ts
```diff
@@ -0,0 +1,203 @@
...
+  /**
+   * Scrolls the flyout to display the newly selected category's contents.
+   *
+   * @param oldItem The previously selected toolbox category.
+   * @param newItem The newly selected toolbox category.
+   */
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  override updateFlyout_(
+    oldItem: Blockly.ISelectableToolboxItem | null,
+    newItem: Blockly.ISelectableToolboxItem | null,
```
cpcallen, line 112: Would `oldCategory` and `newCategory` be better names?  Or are you keeping them for consistency with the naming of the superclass's method's params?

(Here and for `shouldDeselectItem_`, `selectCategoryByName`, etc.)

plugins/continuous-toolbox/src/ContinuousToolbox.ts
```diff
@@ -0,0 +1,203 @@
...
+    const category = this.getToolboxItems().find(
+      (item) =>
+        item instanceof Blockly.ToolboxCategory &&
+        item.isSelectable() &&
+        name === item.getName(),
+    );
+    if (category) {
+      return category as Blockly.ISelectableToolboxItem;
+    }
+    return null;
```
cpcallen: Perhaps treat this as a guard?:
```suggestion
    if (!category) return null;
    return category as Blockly.ISelectableToolboxItem;
```

plugins/continuous-toolbox/src/ContinuousToolbox.ts
```diff
@@ -0,0 +1,203 @@
...
+   * because this is called while the flyout is being scrolled.
+   *
+   * @param name Name of category to select.
+   * @internal
+   */
+  selectCategoryByName(name: string) {
+    const newItem = this.getCategoryByName(name);
+    if (!newItem) {
+      return;
+    }
```
cpcallen:
```suggestion
    if (!newItem) return;
```

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
```diff
@@ -0,0 +1,193 @@
...
+import * as Blockly from 'blockly/core';
+
+/**
+ * A block inflater that caches and reuses blocks to improve performance.
+ */
+export class RecyclableBlockFlyoutInflater extends Blockly.BlockFlyoutInflater {
+  /**
+   * Whether or not block recycling is enabled.
+   */
+  private recyclingEnabled = false;
```
cpcallen: Similarly to previously: Can this not be `public`, obviating the need for `setRecyclingEnabled`?

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  /**
+   * Map from block type to block instance.
+   */
+  private recycledBlocks = new Map<string, Blockly.BlockSvg>();
+
+  /**
+   * Custom function to use for checking whether or not blocks can be recycled.
+   */
+  private recycleEligibilityChecker?: (block: Blockly.Block) => boolean;
```
cpcallen: Ditto: `public`?

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  /**
+   * Returns the type of a block from an XML or JSON block definition.
+   *
+   * @param blockDefinition The block definition to parse.
+   * @returns The block type.
+   */
+  private getTypeFromDefinition(
+    blockDefinition: Blockly.utils.toolbox.BlockInfo,
+  ): string {
```
cpcallen, line 54: I think this is confusingly worded.  To me a "block definition" is one of the things that gets turned into an entry in the `Blockly.Blocks` dictionary, and is therefore either a "JS block definition" or a "JSON block definition"—there are no "XML block definitions".

I gather we're talking here about block _serialisations_, rather than _definitions?

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    if (this.recycleEligibilityChecker) {
+      return this.recycleEligibilityChecker(block);
+    }
+
+    // If the block needs to parse mutations, never recycle.
+    if (block.mutationToDom && block.domToMutation) {
+      return false;
+    }
```
cpcallen: Should `saveExtraState` and `loadExtraState` not also be checked, for block definitions that don't support XML serialisation?

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    if (!block.isEnabled()) {
+      return false;
+    }
```
cpcallen:
```suggestion
    if (!block.isEnabled()) return false;
```

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    for (const input of block.inputList) {
+      for (const field of input.fieldRow) {
+        // No variables.
+        if (field.referencesVariables()) {
+          return false;
+        }
```
cpcallen:
```suggestion
        if (field.referencesVariables()) return false;
```

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
```diff
@@ -0,0 +1,193 @@
...
+        if (field instanceof Blockly.FieldDropdown) {
+          if (field.isOptionListDynamic()) {
+            return false;
+          }
```
cpcallen:
```suggestion
          if (field.isOptionListDynamic()) return false;
```

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    for (const input of block.inputList) {
+      for (const field of input.fieldRow) {
+        // No variables.
+        if (field.referencesVariables()) {
+          return false;
+        }
+        if (field instanceof Blockly.FieldDropdown) {
+          if (field.isOptionListDynamic()) {
```
cpcallen: I'm mildly concerned about this, since it's possible a `FieldDropdown` could have dynamically-updated options without actually using a menu generating function.  Hopefully it would save/load extra state and be caught that way—or the developer could override `isOptionListDynamic` to always return true, but I think perhaps this deserves:

* A general warning in the README (and release notes) about issues with dynamically-modified blocks and fields when using this plugin, and/or
* A general-purpose way to opt specific block types out of being recycled.

cpcallen (commented):
Overall this looks good and the preponderance of the comments/suggestions below are 'for your consideration' rather than requests for change.

Nevertheless, I find myself slightly confused about the recycling business.  I think it is worth disambiguationg "recycling bin" from "trash", and maybe including a bit more of a high-level overview of purpose and intended behaviour in relevant places (README, class-level JSDocs, etc.) to explain what's going on, because this was definitely the part of the code that I was least confident about my understanding of.

plugins/field-bitmap/src/field-bitmap.ts
gonfunko: This field is [dropped in v12](https://github.com/google/blockly/pull/8648) in favor of specifying cursors in CSS.

plugins/content-highlight/src/index.ts
gonfunko: It's needed to be able to do `contentChangeEvents.indexOf(event.type)` on line 187 below; without it it yells since event.type is `string`, but the list is `EventType[]`.

plugins/continuous-toolbox/src/ContinuousFlyout.ts
gonfunko: Added; FlyoutButton is actually used for both buttons and labels.

plugins/continuous-toolbox/src/ContinuousMetrics.ts
gonfunko: Done

plugins/continuous-toolbox/src/ContinuousFlyout.ts
gonfunko: Done

plugins/continuous-toolbox/src/ContinuousFlyout.ts
gonfunko: Done.

plugins/continuous-toolbox/src/ContinuousFlyout.ts
gonfunko: Done, and yes that's just there for the typechecker because pop() isn't guaranteed to return a value.

plugins/continuous-toolbox/src/ContinuousToolbox.ts
gonfunko: Done.

plugins/continuous-toolbox/src/ContinuousToolbox.ts
gonfunko: Done.

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
gonfunko: Done.

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
gonfunko: Done.

plugins/continuous-toolbox/src/ContinuousMetrics.ts
gonfunko: Done; I think this was probably just matching how it was done in the superclass, which should perhaps be similarly adjusted.

plugins/continuous-toolbox/src/ContinuousFlyout.ts
gonfunko: My understanding is that this is being run only by viewport changes affecting the flyout's workspace? This handler is registered on `this.getWorkspace()`, ie the flyout's own workspace. Experimentally, scrolling the main workspace does not trigger this as well.

plugins/continuous-toolbox/src/ContinuousFlyout.ts
gonfunko: Yep, noted – I think in this case the functional style makes the intent a bit clearer, but agreed it's not a huge difference either way.

plugins/continuous-toolbox/src/ContinuousFlyout.ts
gonfunko: Yes; added a comment noting this here as well.

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
gonfunko: Done.

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
gonfunko: Done.

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
gonfunko: Done.

plugins/continuous-toolbox/src/ContinuousFlyout.ts
gonfunko: Because the recycle bin needs to be populated in order for recycling to actually happen when the flyout is shown. If it were emptied first, there would be no recyclable blocks, so when the flyout is populated every block would have to be created from scratch.

plugins/continuous-toolbox/src/ContinuousToolbox.ts
gonfunko: Done.

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
gonfunko: Done.

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
gonfunko: Done.

plugins/continuous-toolbox/src/ContinuousToolbox.ts
gonfunko: They might be, but there's various methods that reference Item also, so I think the consistency wins out, particularly since none of this is new code.

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
gonfunko: Good call, done!

plugins/continuous-toolbox/src/ContinuousFlyout.ts
gonfunko: Added a comment, but because the flyout is a registry thing, this class gets used for all flyouts, e.g. the trashcan, or a simple flyout. The checks are to only run this in the case where it's a flyout associated with the toolbox.

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
```diff
@@ -113,7 +113,12 @@ export class RecyclableBlockFlyoutInflater extends Blockly.BlockFlyoutInflater {
     }
 
     // If the block needs to parse mutations, never recycle.
-    if (block.mutationToDom && block.domToMutation) return false;
+    if (
+      (block.mutationToDom && block.domToMutation) ||
+      (block.saveExtraState && block.loadExtraState)
+    ) {
```
cpcallen: I can't imagine why a block would get just one of `mutationToDom` and `domToMutation` (or `saveExtraState` and `loadExtraState`) but not the matching pair, but perhaps safer to just check if _any_ of them exist?

cpcallen (approved):
I still think it would be a good idea to provide a bit more internal documentation about the recycling business, and perhaps also make sure the external documentation mentions the issue with non-obviously mutable blocks, but otherwise this LGTM.

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
gonfunko: Fair enough, done.

plugins/continuous-toolbox/src/RecyclableBlockFlyoutInflater.ts
gonfunko: Added some info to the README; block types can be opted out by setting a custom recycling eligibility check function.

PR #2474, opened 2024-12-30
Fix: Migrating every theme to TypeScript.

### LINKED ISSUE #2158: Convert theme plugins to ts
### Check for duplicates

### Component

Themes (like @blocky/theme-dark)

### Description

There is no index.d.ts in module so we can't import it with Typescript (or we need to use //@ts-ignore)

### Reproduction steps

1. Prepare a typescript project
2. Install the package named "@blockly/theme-dark"

### Stack trace

```shell
[tsl] ERROR in C:\Users\user\Desktop\blocksk-typescript\src\index.ts(8,23)
      TS7016: Could not find a declaration file for module '@blockly/theme-dark'. 'C:\Users\esteb\Desktop\blocksk-typescript\node_modules\@blockly\theme-dark\dist\index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/blockly__theme-dark` if it exists or add a new declaration (.d.ts) file containing `declare module '@blockly/theme-dark';`
```

### Screenshots

image

tgbhy:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/2158

### Proposed Changes

Migrate every theme sample to TypeScript.

### Reason for Changes

It would be usable in TypeScript like explained in the fixed issue.

### Test Coverage

Manually tested with a Windows 10 PC and Brave Browser.
I tested on using a default installation of Blockly and recreating the schema of each readme-media/Theme.png


cpcallen (approved):

cpcallen:
Hi @tgbhy.  Thanks for this useful PR!  The changes look good but there are some additional steps needed before we can accept the PR:

* [X] You'll need to agree to the CLA.
* [ ] eslint is unhappy about all the properties with underscores in their name.  You'll need to disable this rule for these declarations.
* [ ] You'll also need to run `npm run format` to run Prettier and thereby fix the formatting.

tgbhy:
Working on it !

plugins/theme-deuteranopia/src/index.ts
```diff
@@ -11,79 +11,79 @@
 import * as Blockly from 'blockly/core';
 
 const defaultBlockStyles = {
-  colour_blocks: {
+  colourBlocks: {
```
cpcallen: So, renaming these properties will break the theme.  This is one case where you have _suppress_ the eslint errors, rather than changing the code so there isn't an error:

```suggestion
  colour_blocks: { // eslint-disable-line camelcase
```

(Here and many other places in this PR.)

plugins/theme-deuteranopia/package.json
```diff
@@ -51,5 +51,8 @@
   },
   "engines": {
     "node": ">=8.17.0"
+  },
+  "dependencies": {
+    "@blockly/theme-deuteranopia": "file:"
```
cpcallen: Not sure what you did while creating 3c49ad6 that caused this (and related changes to `package-lock.json`), but this change looks weird (a package should not in general depend upon itself!) and should probably be reverted.

(Here and in the other themes.)

cpcallen (changes requested):

plugins/theme-deuteranopia/package.json
tgbhy: Ye, I think it's my VSCode that bugged, let me correct this one in a few minutes.
Sorry for this thing I didn't saw it =x

plugins/theme-deuteranopia/src/index.ts
tgbhy: It doesn't seems to destroy the theme for me after my tests

tgbhy: So tell me if you want me to do it but the one or the other isn't breaking so...

plugins/theme-dark/package-lock.json
cpcallen, line 1: Please revert any changes to any `package-lock.json` file in this branch, since AFAICT there should be no changes to any `package.json` file to justify them.

(General dependency updates not resulting from a change to `package.json` should be done in a separate PR.)

plugins/theme-dark/package.json
```diff
@@ -51,5 +51,8 @@
   },
   "engines": {
     "node": ">=8.17.0"
+  },
+  "dependencies": {
+    "@blockly/theme-dark": "file:"
```
cpcallen: As discussed previously (in a different directory): please revert.

plugins/theme-deuteranopia/package-lock.json
cpcallen, line 1: Please revert.

plugins/theme-deuteranopia/src/index.ts
cpcallen: Please revert the property name changes.  The theme won't work if your rename them.

plugins/theme-hackermode/package-lock.json
cpcallen, line 1: Please revert.

plugins/theme-highcontrast/package-lock.json
cpcallen, line 1: Please revert.

plugins/theme-modern/package-lock.json
cpcallen, line 1: Please revert.

plugins/theme-tritanopia/package-lock.json
cpcallen, line 1: Please revert.

cpcallen (changes requested):

tgbhy:
This is fine now ?

plugins/theme-deuteranopia/src/index.ts
cpcallen: Please apply the suggested change above, and equivalent changes to each of the other names below.

cpcallen (changes requested):

tgbhy:
Wait, it's already the case ?
Where did you see that, it's already the case of being "colourBlocks" in every themes...

tgbhy:
Bump here

maribethb:
Hi @tgbhy we can't merge this because renaming the properties will break them. I'm not sure how you tested the change, but the way themes work are that they correlate a category name to a set of styles. The category names like `colour_blocks` are set in core, so changing them here and not in core would break the theme as there is no category name called `colourBlocks`. And we are not interested in changing the category names in core.

We can accept this PR after you address Christopher's comment here https://github.com/google/blockly-samples/pull/2474/files#r1929130761

Please let us know if you have further questions. Thanks for your work on this and for following up!

tgbhy:
oh okay, I don't know why it was working perfectly for me...
I'll do the change, you can come back in less than an hour I think ;)

maribethb:
You will need to fix this for all of the `defaultBlockStyles` and `categoryStyles`, not just the `colour_blocks`.

tgbhy:
So wait, I need to do this only for `defaultBlockStyle` (-> `default_block_style`) and `categoryStyles` (-> `category_style`) ?
After that it's finish ?

Because I really don't understand why you ask me that because it was working fine for me but whatever, just confirm me that and I'll do the changes.

maribethb:
You can check how a theme is supposed to look on the published plugin showcase page. Here's what the highcontrast theme is supposed to look like, for example: https://google.github.io/blockly-samples/plugins/theme-highcontrast/test/index.html

You can test your change by running `npm start` from the plugin's directory. When I checkout your PR and run `npm start` from the highcontrast plugin, here's what I see:

Screenshot 2025-05-12 at 9 17 32 AM

All of the blocks are black and the toolbox is missing the little color stripe for the category. This means your theme is broken.

You can read more about how themes work [here](https://developers.google.com/blockly/guides/configure/web/appearance/themes).

The properties in the theme like `defaultBlockStyle` and `categoryStyle` are supposed to be camel case. However, those properties are objects where the keys are arbitrary values set by the owner of the blocks. If an application owner creates their own block category and decides to call it `mY_SpEcIaL_BlOcKs`, that's what would have to go in as the object key, so that blockly can match that name from the toolbox definition to the theme definition. The themes you're editing provide values for the default blocks and categories that Blockly ships with (like `logic_blocks`, `math_blocks`, and so on). Since Blockly uses snake case for these values, the theme has to as well.

This PR shouldn't change any object keys as part of the change. You should revert all of those kinds of changes. If the key was previously `logic_blocks` it needs to remain that way. If the key was previously `defaultBlockStyles` it needs to remain that way. Does that help?

tgbhy:
Okay thank you, I will do all the edits right now.

tgbhy:
And done !

plugins/theme-dark/package.json
maribethb: This change needs to be removed from this PR.

tgbhy:
Give me a sec I have an idea
I'm really sorry for this ^^'

tgbhy:
And the work is done !
Sorry, I'll pay attention to that the next time ^^

plugins/theme-dark/package.json
tgbhy: All changes has been made for this.

tgbhy:
Any news here ?

cpcallen (approved):
Thanks for the changes; they look great—and sorry for the delayed review: we were busy with the Blockly  v12 release.

tgbhy:
No problem. Will it be merged now, or does it need more modifications?

PR #2485, opened 2025-02-03
fix: Fix scrolling and API issues with the continuous toolbox plugin.

gonfunko:
## The basics

### Resolves
This PR fixes bugs that (1) prevented scrolling the flyout if the first category was clicked, (2) caused jank when scrolling the flyout via the scrollbar and (3) reintroduces the `getCategoryScrollPosition()` method, which is used by Scratch.

### DIFF
--- a/plugins/continuous-toolbox/src/ContinuousFlyout.ts
+++ b/plugins/continuous-toolbox/src/ContinuousFlyout.ts
@@ -118,7 +118,8 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
    * Returns the scroll position for the given category name.
    *
    * @param name Category name.
-   * @returns Scroll position for given category, or null if not found.
+   * @returns Scroll position for given category in workspace units, or null if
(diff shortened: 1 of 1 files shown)


plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -114,6 +114,19 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
     );
   }
 
+  /**
+   * Returns the scroll position for the given category name.
+   * @param name Category name.
+   * @returns Scroll position for given category, or null if not found.
```
rachel-fenichel: Nit: in what units?

rachel-fenichel (approved):

plugins/continuous-toolbox/src/ContinuousFlyout.ts
gonfunko: Done.

PR #2492, opened 2025-02-21
fix: #2463 update event handling for field-bitmap to support touch 

### LINKED ISSUE #2463: Bitmap field editor freezes after one click on mobile devices
### Check for duplicates

### Component

field-bitmap

### Description

Our users have reported that the bitmap field editor stops working after one click on iPad.  We have repro'ed on iPhone and also in the macOS Simulator of an iPad. 

### Reproduction steps

1. Using a mobile device or simulator, go to https://google.github.io/blockly-samples/plugins/field-bitmap/test/index.html
2. Open any bitmap field and toggle any pixel in the bitmap editor.
3. Observe how the field editor and workspace are no longer responsive. However, the "random" and "clear" options still work.

### Stack trace

```shell
No errors
```

### Screenshots

_No response_

RoboErikG:
## The basics

### Resolves

Fixes #2463 by changing how the editor handles mouse and touch events.

### Proposed Changes

Updates the Bitmap field to properly support touch events. This change is based on code-dot-org/code-dot-org#61709 .

- Switches from handling mouse events to pointer events and moves the event handling up to the editor. This prevents the touch events from getting trapped in an individual pixel.
- Uses bind instead of conditionalBind. conditionalBind does extra work to account for multi-touch gestures, which causes issues for the editor and isn't needed.
- Prevents default handling of touchmove (instead of drag) within the editor. The default handling cancels the original gesture when it converts it to a drag, which prevents future move events.
- Adds a cancel listener to also cleanup the edit state when canceled.

[Screen recording 2025-02-21 9.59.31 AM.webm](https://github.com/user-attachments/assets/9987ed04-451a-49ed-94cb-1e8addbb96ec)

### Reason for Changes

Touch events in the bitmap editor would get canceled by the browser, causing the gesture to end early and the state to not be cleaned up properly. This would cause Blockly to become partially unresponsive.

### Test Coverage

Tested manually on a Chromebook with a touch screen.

### DIFF
--- a/plugins/field-bitmap/src/field-bitmap.ts
+++ b/plugins/field-bitmap/src/field-bitmap.ts
@@ -41,7 +41,7 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
   private valToPaintWith?: number;
   buttonOptions: Buttons;
   pixelSize: number;
-  pixelColours: { empty: string; filled: string };
+  pixelColours: {empty: string; filled: string};
   fieldHeight?: number;
 
   /**
@@ -60,8 +60,8 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
 
     this.SERIALIZABLE = true;
     this.CURSOR = 'default';
-    this.buttonOptions = { ...DEFAULT_BUTTONS, ...config?.buttons };
-    this.pixelColours = { ...DEFAULT_PIXEL_COLOURS, ...config?.colours };
+    this.buttonOptions = {...DEFAULT_BUTTONS, ...config?.buttons};
+    this.pixelColours = {...DEFAULT_PIXEL_COLOURS, ...config?.colours};
 
     // Configure value, height, and width
     const currentValue = this.getValue();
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Use `npm run format` to format and then push again.

RoboErikG:
Hi @mikeharv I made a fix to the bitmap editor based on your change. If you have some time would you mind taking a look to make sure I'm not missing anything?

plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -473,29 +470,63 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
...
 
   /**
-   * Called when a mousedown event occurs within the bounds of a pixel.
+   * Checks if a down event is on a pixel in this editor and if it is starts an
+   * edit gesture.
+   *
+   * @param e The down event.
+   */
+  private onPointerStart(e: PointerEvent) {
+    const currentElement = document.elementFromPoint(e.clientX, e.clientY);
```
gonfunko, line 479: Does `e.target` not work here?

gonfunko: And in `onPointerMove`

gonfunko (approved):

plugins/field-bitmap/src/field-bitmap.ts
RoboErikG: Touch events get captured by the element they start on. This means the target is always the element the event started on, so it won't change when you drag across another pixel cell.

RoboErikG:
For completeness, I did find an alternate to using elementFromPoint() would be to cast the target in the pointerdown event to an Element and call releasePointerCapture(). This might have a slight performance advantage, but both options work well so it's not worth revisiting unless we find an issue with the current approach.

```
      const elem = e.target as Element;
      if (elem && elem.hasPointerCapture(e.pointerId)) {
        elem.releasePointerCapture(e.pointerId);
      }
```

PR #2495, opened 2025-03-05
Fix: Add a migration step for importing Blockly correctly

### LINKED ISSUE #2349: migration script: fix `import Blockly from 'blockly'`
### Check for duplicates

### Component

migration

### Problem

_No response_

### Request

Using `import Blockly from 'blockly'` has never been recommended or explicitly supported. It won't work when we publish esmodules because we don't have a default export. The migration script should find and change these to `import * as Blockly from 'blockly'` if possible.

### Alternatives considered

_No response_

### Additional context

_No response_

RoboErikG:
## The basics

Adds a check for the Blockly import to the migration script. Due to how the script is set up this will run every time, even if it's already correct, but that should be a no-op.

### Resolves

Fixes #2349

### Proposed Changes

Adds a database entry to fix_imports.js for updating how Blockly gets imported
from:
import Blockly from 'blockly';
to:

Because of how the migration script is set up there's no easy way to check if the import is already correct, so it will always perform the replacement if there are references to Blockly in the file. This shouldn't cause any visible change, so is just inefficient. This shouldn't be a problem given the usage of the migration script.

I also changed the manual test files from .txt to .js because the script was previously updated to only run against .js and .ts files to avoid modifying something unexpectedly.

### Reason for Changes

import Blockly from 'blockly' isn't guaranteed to work correctly and should be import * as Blockly from 'blockly' instead.

### Test Coverage

Ran mocha tests and the manual test files to verify.


RoboErikG:
Working on disabling the lint checks for the test files...

gonfunko (approved):

PR #2504, opened 2025-05-06
chore(deps): Update `blockly` dev dependency to `blockly@12.0.0-beta.4`

cpcallen:
## The basics

- [x] I [validated my changes](https://developers.google.com/blockly/guides/contribute/samples#making_and_verifying_a_change)

### Proposed Changes

* Fix `clean:node` npm script that was inadvertently broken by a mistranscription in PR #2453.
* Bump `blockly` dev dependency to v12.0.0-beta.4.
* Fix type of `FieldColour.prototype.isFullBlockField`, as superclass method  becomes public in v12.0.0.
* Fix `FieldMultilineInput` to not depend on `Field.NBSP`, which was temporarily removed.
  * Include a `TODO` to reinstate use of `.NBSP` where appropriate.
* Fix `FieldGridDropdown` to not fail type checking due to the introduction of spacers.

### Reason for Changes

Keep samples working with v12.

### Test Coverage

Passes `npm test` (at least, I think it does—`FieldDate` tests always fail when tested outside the US!)

Have _not_ done manual testing of individual plugins.

PR #2508, opened 2025-05-08
fix!: Delete old `keyboard-navigation` plugin

cpcallen:
## The basics

### Proposed Changes

Delete `plugins/keyboard-navigation/` and contents.

### Reason for Changes

This plugin depends on too many things that no longer exist in Blockly, like `Cursor` and `BasicCursor`, and also on other things that will cease to exist before 12.0.0, like `ASTNode`.

This plugin depends on things that no longer exist in `blockly@v12.0.0-beta.4`, like `BasicCursor` and `Cursor`, and also
on other things that will cease to exist before 12.0.0, like `ASTNode`.

We need to update the samples repository to be able to test plugins with v12, and this plugin is soon to be replaced by
the new plugin being developed in [google/blockly-keyboard-experimentation](https://github.com/google/blockly-keyboard-experimentation), so the most expedient course of action is just to delete this one to allow upgrading this repo's `blockly` `devDependency`.

### Additional Information

(We could alternatively leave this plugin as an empty shell waiting to be filled with the contents of the `blockly-keyboard-experimentation` repo, but it seems preferable to not publish publish anything, thus leaving the latest version of the `@blockly/keyboard-navigation` NPM package useful though outdatedly dependent on `blockly@11`, rather than publish an empty plugin package that is of no use at all.

This has been broken out from #2504 so that that PR will not contain any breaking changes, since it affects multiple plugins.

**BREAKING CHANGE:**

Removes the keyboard-experimentation plugin, obviously.

The plugin will return soon, but in the meantime use the plugin being developed in [google/blockly-keyboard-experimentation](https://github.com/google/blockly-keyboard-experimentation), published as the `@blockly/keyboard-experiment` NPM package.


BenHenning:
I think removing this makes complete sense, but I want to defer to @maribethb or @rachel-fenichel to make the actual decision since this may affect things I'm not certain of in the external community.

maribethb (commented):
Also needs to be removed from the plugin showcase page.

Doesn't really matter in this exact case since you are deleting the changelog, but generally `chore` type commits shouldn't be breaking changes, because chores are not included in the release notes/changelog.

maribethb:
Also, could you update the PR description to contain info that's helpful for people coming here to find out what happened? e.g. point them at the experiment repo

cpcallen:
> Also, could you update the PR description to contain info that's helpful for people coming here to find out what happened? e.g. point them at the experiment repo

Done.

maribethb (approved):

PR #2509, opened 2025-05-08
chore(deps): Update `blockly` dev dependency to `blockly@12.0.0-beta.5`

cpcallen:
## The basics

- [x] I [validated my changes](https://developers.google.com/blockly/guides/contribute/samples#making_and_verifying_a_change)

### Proposed Changes

* Fix `clean:node` npm script that was inadvertently broken by a mistranscription in PR #2453.
* Bump `blockly` dev dependency to v12.0.0-beta.4.
* Fix type of `FieldColour.prototype.isFullBlockField`, as superclass method  becomes public in v12.0.0.
* Fix `FieldMultilineInput` to not depend on `Field.NBSP`, which was temporarily removed.
  * Include a `TODO` to reinstate use of `.NBSP` where appropriate.
* Fix `FieldGridDropdown` to not fail type checking due to the introduction of spacers in `MenuOptions`.
* Bump `blockly` dev dependency to v12.0.0-beta.5.
* Fix compilation failures for `FieldBitmap` and `FieldColour` by implementing abstract `getClass` method added in google/blockly#8992.
* Fix `FieldGridDropdown` to not fail type checking due to the introduction of `HTMLElement` in `MenuOptions`.

### Reason for Changes

Keep samples working with v12.

### Test Coverage

Passes `npm test` (at least, I think it does—`FieldDate` tests always fail when tested outside the US!)

Have _not_ done manual testing of individual plugins.

PR #2524, opened 2025-05-14
chore: update to v12 beta 6

maribethb:
My turn to do https://github.com/google/blockly-samples/pull/2523

- Disables the tests for shadow-block-converter as there's some jsdom problem with the test setup
- updates package-lock for shadow-block-converter as at some point it was committed with incorrect changes
- reverts https://github.com/google/blockly-samples/pull/2509/commits/96b299b16ef5e27b1903ed6abef2642f42cf2d74 now that `Blockly.Field.NBSP` exists again
- merges in master


maribethb:
this isn't ready for review yet, idk how to make it a draft after the fact

edit: ready now if tests pass

.github/workflows/publish_from_package.yml
```diff
@@ -0,0 +1,71 @@
+# This workflow can be manually triggered to publish all of the plugins
+# that have a version in their package.json that is newer than what
+# is available in the npm registry. This is useful if lerna publishing
+# failed after a package had its version updated on github but before
+# that version was published to npm.
+
+name: publish unpublished plugins
```
BenHenning, line 7:
```suggestion
name: Publish unpublished plugins
```

For consistency.

plugins/field-grid-dropdown/src/grid.ts
```diff
@@ -293,3 +299,25 @@ export class Grid {
...
 }
+
+/**
+ * Returns whether or not an object conforms to the ImageProperties
+ * interface.
+ *
+ * @param obj The object to test.
+ * @returns True iff the object conforms to ImageProperties.
+ */
+function isImageProperties(obj: any): obj is ImageProperties {
```
BenHenning, line 310: Perhaps this would be worth adding to Core proper in the future. Worth filing an issue?

plugins/shadow-block-converter/test/shadow_block_converter_test.mocha.js
```diff
@@ -11,7 +11,7 @@ const {shadowBlockConversionChangeListener} = require('../src/index');
 
 const assert = chai.assert;
 
-suite('shadowBlockConversionChangeListener', function () {
+suite.skip('shadowBlockConversionChangeListener', function () {
```
BenHenning, line 14: Should we perhaps file a TODO to re-enable these and add a line comment here pointing to that issue?

Or, do these pass now that the `package-lock.json` file has been corrected?

BenHenning (approved):
Thanks @maribethb! Just a few small comments, but approving since CI is green and this generally looks good.

.github/workflows/publish_from_package.yml
maribethb: This was done from merging in master, I don't want to muddy up actual changes into what should be a merge commit. Anyway, the publish action was already lower case so it's not really any more inconsistent than publish already is

plugins/shadow-block-converter/test/shadow_block_converter_test.mocha.js
maribethb: Filed https://github.com/google/blockly-samples/issues/2525

PR #2530, opened 2025-05-15
feat!: Update plugins to use Blockly v12

maribethb:
Updates all plugins to use Blockly v12. This is a breaking change for each plugin.

This PR does not update examples, as some examples rely on the newly-updated plugin being published first.

Verified that no plugin references a beta version of blockly.


gonfunko (approved):

rachel-fenichel (approved):
Diffs look good (all the correct size, no spurious changes in `package-lock`s.

PR #2534, opened 2025-05-16
revert: revert last publish with incorrect versions

maribethb:
Reverts https://github.com/google/blockly-samples/commit/332ddbcc50c81f77072e1043ad8557e3c1b4b556 due to incorrect version numbers.


rachel-fenichel:
Just read the discussion + looked at the other lerna change. Makes sense, but how is the "unpublish" side of this propagated to npm?

rachel-fenichel (approved):

maribethb:
It’s not, we shouldn’t do this. I was going to do this in tandem with npm
unpublish but as discussed that won’t work. We should do a force publish
with v11 instead.

PR #2541, opened 2025-05-17
chore: update package-locks probably

### LINKED ISSUE #2539: unbork the package-locks again
### Check for duplicates

### Component

_No response_

### Description

Image

Some of the problems:

- A bunch of plugins have other plugins listed in their `package-lock` files. This is bad because lerna is supposed to remove those. Not sure how they got in, either from people running `npm install` from within an individual plugin and committing that, or maybe from dependabot
- Some of the package-locks were out of date and had blockly-v11 listed there. I *think* these were packages that didn't have any dependencies to install (because the only dependencies were managed by lerna) so the package-locks were just not updated
- the continuous-toolbox plugin was updated to use typescript but didn't list typescript in its devDependencies

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

maribethb:
## The basics

### Resolves

Fixes #2539 I think?

### Proposed Changes

- Adds typescript as a devDependency of `continuous-toolbox` because it was missing
- Updates `theme-hackermode` to the latest version of dev-tools. this plugin is private so it doesn't automatically get breaking updates.
- Updates a bunch of package-locks to remove references to blockly v11
- Deletes some package-locks that I guess shouldn't exist?
- Incidentally updates a bunch of locked versions like dependabot would because I deleted package-locks and started over
- Adds error handling to the webpack callback in our `build` dev-script
- Updates the version of `eslint-webpack-plugin` in dev-scripts package-lock so that ajv has the right version

### Reason for Changes

I created this PR by

1. Removed all package-lock files because I couldn't get lerna to update the ones that had incorrect devDependencies 
2. Ran `npm run clean:node` to delete all the node_modules directories
3. Ran `npm install` at root

The package-locks that are deleted are ones that don't have any dependencies that are installed from npm. The only dependencies are managed by lerna. I think something must have changed but not sure where, that causes these to be not created in the first place instead of created but not have any dependencies. Since no dependencies are installed, I think it's fine for them to not have locks since there is nothing to lock.


gonfunko (approved):

maribethb:
Builds of some plugins were failing at the webpack step (before compilation) but these errors were not being shown. This caused build failure in the form of "cannot access property toJson of undefined" because webpack never got stats as it errored first. So, I updated the webpack callback to display errors and then exit with error. Got this code from the webpack docs: https://webpack.js.org/api/node/#webpack

This allowed me to see what the webpack error was, which was something around the ajv library that a bunch of dependencies use. Not sure exactly what happened here but I think the `eslint-webpack-plugin` was using a version of ajv it wasn't compatible with or something, so running `npm update eslint-webpack-plugin` from the `dev-scripts` package seems to have fixed things.

PTAL

gonfunko (approved):

PR #2547, opened 2025-05-22
fix(workspace-search): set button type to "button" to prevent unintended form submissions

### LINKED ISSUE #2510: workspace-search buttons need type=button
### Check for duplicates

### Component

workspace-search's `createBtn` private helper method

### Description

We are using Blockly inside a `<form>` element (for various reasons), and the existing `createBtn` function does not set the `<button>` to be `<button type='button'>` which means when users click them it erroneously causes a form-submit.

The fix should in theory works on my machine(tm), be as simple as changing the helper function to read something akin to:

```TypeScript

  private createBtn(className: string, text: string): HTMLButtonElement {
    // Create the button
    const btn = document.createElement('button');
    btn.type = "button"; //++added++
    btn.setAttribute('aria-label', text);
    return btn;
  }

```

### Reproduction steps

1. Have a Blockly instance nested within a `<form>`
2. Have the workspace-search plugin installed in the workspace
3. ctrl-f for something, then hit any of the three buttons (next, prev, exit)
4. Find that your HTML Form is now Submitted :( 

### Stack trace

```shell

```

### Screenshots

_No response_

viky-01:
## The basics

### Resolves

Fixes [#2510](https://github.com/google/blockly-samples/issues/2510)

### Proposed Changes

Updated the `createBtn` helper method in the `workspace-search` plugin to explicitly set the button’s `type` attribute to `"button"`.

```ts
private createBtn(className: string, text: string): HTMLButtonElement {
  const btn = document.createElement('button');
  btn.type = "button";
  btn.setAttribute('aria-label', text);
  return btn;
}

### DIFF
--- a/plugins/workspace-search/src/workspace_search.ts
+++ b/plugins/workspace-search/src/workspace_search.ts
@@ -273,6 +273,7 @@ export class WorkspaceSearch implements Blockly.IPositionable {
     // Create the button
     const btn = document.createElement('button');
     Blockly.utils.dom.addClass(btn, className);
+    btn.type = "button";
     btn.setAttribute('aria-label', text);
     return btn;
   }


BenHenning:
Thanks @viky-01. I was thinking it might be nice to test and/or comment this a bit more, but https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#notes makes it pretty clear that this is a required approach to solve this exact form problem. I think the code is reasonable as-is.

Please take a look at the failing tests and correct the code as needed @viky-01. I suspect you'll need to run either eslint's fix mode or prettier's auto-format to fix whichever style issues are causing the failure (these routines usually fix most style problems automatically).

BenHenning:
@viky-01 just checking: do you need any additional help with resolving the CI failure? If you're having difficulties, please don't hesitate to ask for help if needed. :)

It would be nice to get this PR wrapped up rather than keeping it open, if possible.

PR #2559, opened 2025-06-09
refactor: Make `FieldColour` inherit from `FieldGridDropdown`.

gonfunko:
## The basics

### Proposed Changes
This PR updates `FieldColour` to inherit from `FieldGridDropdown`. The latter was recently updated to support using abitrary HTML grid elements, which allows `FieldColour` to specify color swatches as grid items, and allows for the removal of a great deal of duplicative grid-handling code from `FieldColour`. There are two small behavioral changes:

* `FieldColour`'s dropdown now inherits its background color from its parent block, à la `FieldDropdown`, since it's now an (indirect) subclass thereof.
* The "kinda derpy" behavior (per the test comment) whereby setting a new list of colors without tooltip titles after having set a list of colors and tooltip titles preserves the tooltips for the new colors has been removed, because it seemed like a bug enshrined in the test suite rather than an intentional behavior, and was indeed kinda derpy, and would have been obnoxious to try to preserve.

### DIFF
--- a/plugins/field-colour-hsv-sliders/package-lock.json
+++ b/plugins/field-colour-hsv-sliders/package-lock.json
@@ -9,6 +9,8 @@
       "version": "6.0.1",
       "license": "Apache-2.0",
       "devDependencies": {
+        "jsdom": "^26.1.0",
+        "jsdom-global": "^3.0.2",
         "typescript": "^5.4.5"
       },
       "engines": {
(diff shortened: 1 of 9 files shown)


maribethb (approved):
Changes to `FieldColour` make sense. I think it's now somewhat weird that `FieldColourHsvSlider` is an indirect subclass of `FieldGridDropdown`, but I guess that's what we get for using inheritance / the same class to represent both the data and the UI. I was questioning if it still made sense, but I think it does since `FieldColour` has logic for things like is full block field, class validation, etc. that do still apply to the slider, it's just the ui that's different. so LGTM.

PR #2551, opened 2025-05-30
fix: set button type to 'button' and format workspace_search.ts

viky-01:
## The basics

- [x] I validated my changes

### Resolves

Fixes #2510

### Proposed Changes

Updated the `createBtn` helper method in the workspace-search plugin to explicitly set the button’s `type` attribute to `"button"`.

```ts
private createBtn(className: string, text: string): HTMLButtonElement {
  const btn = document.createElement('button');
  Blockly.utils.dom.addClass(btn, className);
  btn.type = "button";
  btn.setAttribute('aria-label', text);
  return btn;
}

PR #2563, opened 2025-06-17
fix: set button type to 'button' and format workspace_search.ts

### LINKED ISSUE #2510: workspace-search buttons need type=button

viky-01:
## The basics

- [x] I validated my changes

### Resolves

Fixes #2510

### Proposed Changes

Updated the `createBtn` helper method in the workspace-search plugin to explicitly set the button’s `type` attribute to `"button"`.

```ts
private createBtn(className: string, text: string): HTMLButtonElement {
  const btn = document.createElement('button');
  btn.type = "button";
  btn.setAttribute('aria-label', text);
  return btn;
}

### DIFF
--- a/examples/blockly-react/src/index.css
+++ b/examples/blockly-react/src/index.css
@@ -1,8 +1,8 @@
 body {
   margin: 0;
-  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen',
-    'Ubuntu', 'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue',
-    sans-serif;
+  font-family:
+    -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu',
+    'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
 }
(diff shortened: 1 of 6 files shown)


viky-01:
@maribethb kindly review

RoboErikG:
Thank you for the PR, @viky-01 .

In order to make it easier for us to review, please don't batch multiple changes together. Specifically, updating dependencies like prettier is something we do with bots because we sometimes need to manually validate or fix breaks caused by the version update.

Can you update this PR or submit a new one with just the change to button to make it easier to review and accept?

maribethb:
Closing this PR. Please use https://github.com/google/blockly-samples/pull/2551.

PR #2570, opened 2025-06-24
fix: prevent Enter/Space from propagating from Grid

microbit-matt-hillsdon:
## The basics

### Resolves

See https://github.com/google/blockly-keyboard-experimentation/issues/618
This is a narrower fix than the one on https://github.com/google/blockly-keyboard-experimentation/pull/619

### Proposed Changes

Prevent Enter/Space from propagating from Grid so it is not inappropriately handled by other listeners, e.g. the keyboard navigation plugin.

### DIFF
--- a/plugins/field-grid-dropdown/src/grid.ts
+++ b/plugins/field-grid-dropdown/src/grid.ts
@@ -189,6 +189,12 @@ export class Grid {
       case 'End':
         this.moveFocus(this.items.length - 1, false);
         break;
+      case 'Enter':
+      case 'Space':
+        // Handled via GridItem click handler, so we want its default but it
+        // must not propagate.
+        e.stopPropagation();
+        return;
       default:
         // Not a key the grid is interested in.
         return;


maribethb:
Can you do the `e.stopPropagation` inside the click event handler for the grid_item instead? That's what I was originally thinking, but I guess I don't know if that works for stopping the keyboard event too since that event being fired on enter is browser behavior due to being a button.

microbit-matt-hillsdon:
> Can you do the `e.stopPropagation` inside the click event handler for the grid_item instead? That's what I was originally thinking, but I guess I don't know if that works for stopping the keyboard event too since that event being fired on enter is browser behavior due to being a button.

I don't think so because the keydown event (including propagation) happens before the click event.
Demo: https://jsfiddle.net/hsatg5un/

This is part of why I think this change is OK if we want it in Blockly field editors, but it's not really a reasonable ask of all field editors.

maribethb:
Thanks for the demo, that's unfortunate. That definitely makes sense why we wouldn't want to ask everyone to do it. I'm not sure if we should even do it, but it's probably worth it to make it easier to write custom keyboard shortcuts.

cpcallen (approved):

cpcallen:
Sorry, I completely forgot that I also needed to hit merge!

PR #2571, opened 2025-06-25
feat!: cross-tab-copy-paste can copy more than just blocks

### LINKED ISSUE #2562: cross-tab-copy-paste needs keyCodes
### Check for duplicates

### Component

cross-tab-copy-paste

### Description

The cross-tab-copy-paste plugin registers keyboard shortcut handlers for cut/copy/paste, but it doesn't set the `keyCodes` property directly, it manually adds key combinations. we should update it to use keyCodes, because the keyboard-navigation plugin relies on the keyCodes being set in order override the behavior.

this is a blocker for code.org

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### LINKED ISSUE #2565: Remove Gesture checks from shortcut preconditions
### Check for duplicates

### Component

plugin-cross-tab-copy-paste

### Problem

If `keyboard-navigation` is also used, new context menu items are added which use the precondition from the old shortcuts. This means the precondition expects that there is not an in-progress gesture in order to enable the option. Because a gesture is always in progress when right-clicking a block, the option always ends up disabled.

### Request

Replace the gesture check with an in-progress drag check, similar to https://github.com/google/blockly/pull/9107

### Alternatives considered

_No response_

### Additional context

_No response_

maribethb:
## The basics

### Resolves

Fixes #2565 
Fixes #2562 

### Proposed Changes

Basically rewrote the plugin, perhaps easier to review this commit-wise.

- Converted plugin to typescript
- updated minimum version of Blockly to v12.1.0 because we need a new function only in that release
- updates all of the logic to be more similar to the logic in core, though with some bug fixes/improvements that were already in the plugin. notably, checks the type counts are allowed for blocks and handles event grouping. this work also enables workspace comments or other `ICopyable` items to be copied, not just blocks.
- allow customizing the local storage key
- use the `Blockly.Msg` values for clipboard shortcuts that now exist in core instead of hardcoded strings
- register the keys for shortcuts using `.keyCodes` property instead of manually calling `addKeyMapping`
- add workspace comments to the test page for easier testing.

### Reason for Changes

Makes this plugin compatible with keyboard-navigation.

### Test Coverage

Manual testing only because samples is not set up to use webdriver tests :(

### Documentation

readme still looks good


maribethb:
The test failures are with shadow-block-converter I guess as a result of updating to v12.1, so I'm going to fix that in a separate PR. Doesn't block review of this one :)

plugins/cross-tab-copy-paste/src/index.ts
```diff
@@ -8,24 +8,198 @@ import * as Blockly from 'blockly/core';
...
+ * Checks if the copy data represents that for a block.
+ *
+ * @param obj any ICopyData.
+ * @returns if the ICopyData is a BlockCopyData.
+ */
+function isBlockCopyData(
+  obj: Blockly.ICopyData,
+): obj is Blockly.clipboard.BlockCopyData {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return !!(obj as any).typeCounts;
```
gonfunko: I think `'typeCounts' in obj` might let you do this check without lint/TS yelling about it.

plugins/cross-tab-copy-paste/src/index.ts
```diff
@@ -8,24 +8,198 @@ import * as Blockly from 'blockly/core';
...
+  DISABLED = 'disabled',
+  HIDDEN = 'hidden',
+}
+
+/**
+ * Copy precondition called by both keyboard shortcut and context menu item.
+ * Allows copying out of the flyout, as long as they could be pasted
+ * into the main workspace.
+ *
+ * @param scope
```
gonfunko: Should this have a description of `scope`?

plugins/cross-tab-copy-paste/src/index.ts
```diff
@@ -8,24 +8,198 @@ import * as Blockly from 'blockly/core';
...
+    return ContextMenuState.ENABLED;
+  return ContextMenuState.DISABLED;
+}
+
+/**
+ * Copy callback called by both keyboard shortcut and context menu item.
+ * Copies the copy data to local storage.
+ *
+ * @param scope
+ * @param workspace
```
gonfunko: Here too

plugins/cross-tab-copy-paste/src/index.ts
```diff
@@ -8,24 +8,198 @@ import * as Blockly from 'blockly/core';
...
+  const focused = scope.focusedNode;
+  if (!focused || !Blockly.isCopyable(focused) || !isCopyable(focused))
+    return false;
+
+  if (!(workspace instanceof Blockly.WorkspaceSvg)) return false;
+
+  const targetWorkspace = workspace.isFlyout
+    ? workspace.targetWorkspace
+    : workspace;
+  if (!targetWorkspace) return false;
```
gonfunko: This seems fairly duplicative of the precondition check above – can they be factored out, or perhaps have this removed entirely? No worries if that's complicated due to shortcut/context menu interop or typing though.

gonfunko (approved):

maribethb:
note to self: remove formatting changes to colour-slider so it doesn't get a breaking change

plugins/cross-tab-copy-paste/src/index.ts
maribethb: This is the minimum set of checks required to get typescript to let you do all the work below, so they can't be removed (without doing a bunch of type assertions)

Factoring them out would be interesting, but the copy precondition currently looks really similar to the one in core, and I think that's more valuable to keep those easily comparable. Plus it's not obvious when we need to make changes to the precondition which conditions go in the "real" precondition vs "parts of the precondition that are required to convince typescript this whole operation is allowed" so I think it's actually more maintainable like this, but I could be persuaded otherwise.

gonfunko: Sounds good, fine to leave as-is then!

plugins/cross-tab-copy-paste/src/index.ts
maribethb: Nice, thanks

plugins/cross-tab-copy-paste/src/index.ts
maribethb: Done (we don't require them if it's obvious from context and I'm not sure my comment added anything useful, but added one nonetheless)

maribethb:
Tests should pass after #2574

PR #2576, opened 2025-07-01
feat(workspace-search): focus found block on close or restore focus

### LINKED ISSUE #2575: Keyboard navigation for workspace search
### Check for duplicates

### Component

workspace-search

### Problem

This is close to being usable by keyboard except it never focuses a match.

### Request

Browsers focus a match on Escape so we should probably match that behaviour.

### Alternatives considered

_No response_

### Additional context

_No response_

microbit-matt-hillsdon:
## The basics

### Resolves

Fixes  https://github.com/google/blockly-samples/issues/2575

### Proposed Changes

Focus the found block on close or restore focus. This matches browser in find in page behaviour where Escape closes find and moves focus to the highlighted item.

When there's no search performed or nothing ever matched we restore focus to the previous node.

Where there was a match but then the search changes so it no longer matched we still focus that last match, because we've scrolled it into view which may have moved the previous focus position out of view. In this scenario browsers seem to focus the element but not mark it as focus:visible but we don't have a similar distinction.

### Test Coverage

Added jsdom tests.

### Additional Information

Unsure if this behaviour is OK to just change without requiring option. More impact for keyboard user if we just do it but not sure if anyone will be affected by the focus change.

### DIFF
--- a/plugins/workspace-search/src/workspace_search.ts
+++ b/plugins/workspace-search/src/workspace_search.ts
@@ -43,6 +43,15 @@ export class WorkspaceSearch implements Blockly.IPositionable {
   protected blocks: Blockly.BlockSvg[] = [];
 
   /**
+   * The last highlighted block, which we focus on close.
+   *
+   * This block is focused on close even if the most recent search found nothing
+   * because we've centered on it and it's helpful for focus to be in sync with
+   * the scroll position.
+   */
+  private lastHighlighted: Blockly.BlockSvg | null = null;
+
+  /**
    * Index of the currently "selected" block in the blocks array.
    */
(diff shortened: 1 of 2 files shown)


microbit-matt-hillsdon:
Back to draft for a mo as I'm getting some unexpected errors testing this inside the keyboard nav demo environment.

microbit-matt-hillsdon:
> Back to draft for a mo as I'm getting some unexpected errors testing this inside the keyboard nav demo environment.

Aha, I had two versions of Blockly and needed to npm link it in the workspace search plugin too. Made the focus manager very confused. They work well together with that fixed.

gonfunko (approved):

PR #2578, opened 2025-07-07
fix: toolbox-search cannot focus in Blockly v12

### LINKED ISSUE #2577: Toolbox search broken in Blockly v12
### Check for duplicates

### Component

Toolbox search

### Description

Blockly v12.1.0

The demo is also broken https://google.github.io/blockly-samples/plugins/toolbox-search/test/index.html
No console error messages are showing either.

### Reproduction steps

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

kuangshen04:
## The basics

### Resolves

Fixes #2577 
The input in toolbox-search cannot focus because of the new FocusManager mechanism in Blockly v12.

### Proposed Changes
Rewrite the focus logic using the interface provided by FocusManager.

### DIFF
--- a/plugins/toolbox-search/src/toolbox_search.ts
+++ b/plugins/toolbox-search/src/toolbox_search.ts
@@ -137,33 +137,23 @@ export class ToolboxSearchCategory extends Blockly.ToolboxCategory {
     this.blockSearcher.indexBlocks([...availableBlocks]);
   }
 
-  /**
-   * Handles a click on this toolbox category.
-   *
-   * @param e The click event.
-   */
-  override onClick(e: Event) {
-    super.onClick(e);
-    e.preventDefault();
-    e.stopPropagation();
-    this.setSelected(this.parentToolbox_.getSelectedItem() === this);
+  /** See IFocusableNode.getFocusableElement. */
+  getFocusableElement(): HTMLElement | SVGElement {
+    if (!this.searchField) {
+      throw Error('This field currently has no representative DOM element.');
+    }
+    return this.searchField;
   }
 
-  /**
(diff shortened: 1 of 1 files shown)


xXBuilderBXx:
registerShortcut on line 90 also needs fixing, it seems to select the category but not focus the input box.

cpcallen:
Hi @kuangshen04 (and @xXBuilderBXx), and thanks for sending us this PR.

The changes you propose all look fairly reasonable, but found myself confused about _why_ they should be necessary.  It eems the plugin works fine with Blockly v12, so, per [the reasoning I outlined in my comment](https://github.com/google/blockly-samples/issues/2577#issuecomment-3052093448) on #2577 I believe the underlying issue is actually an unintentional breaking change in Blockly, and should be fixed there.

I'm going to leave this PR open for the time being, for two reasons:

* If my colleagues can convince me that the problem is not with Blockly, or cannot be fixed there, then this fix may well be the next best alternative.
* The change to use focus rather than selection is probably the direction we want to go anyway, so we should consider applying it anyway—but in that case I would want to consider it as a stand-alone refactor, once the underlying problem causing #2577 is fixed.

My colleagues will do triage on Friday, so I'll aim to revisit this next week.

maribethb:
@cpcallen I think we should merge this. The discussion about the Blockly change in #9082 shouldn't block fixing the broken plugin. Can you review (if necessary) and merge when ready?

cpcallen (approved):

cpcallen:
Thanks for sending this fix, and apologies that it has taken so long to get it reviewed and merged!

PR #2597, opened 2025-09-10
Remove backpack count from copy to message

### LINKED ISSUE #2599: Remove backpack count from copy to message
### Check for duplicates

### Component

backpack plugin

### Problem

> Unfortunately, that convention came from App Inventor and I have never personally been a fan of it. Having the number of blocks in the backpack listed in the menu option is just noise as it has nothing to do with the blocks to which the menu is actually attached. The backpack plugin took this bad design decision from App Inventor. If we have our choice between showing two numbers in this plugin the appropriate number to show would be the number of workable blocks since that's how much the backpack will grow by (which is what is implemented here).
@ewpatton https://github.com/mit-cml/workspace-multiselect/pull/43#issuecomment-1971473289

### Request

The request is to remove the count of blocks already in the backpack from the copy to backpack message as it doesn't seem relevant.

### Additional context

https://github.com/google/blockly-samples/pull/2597
https://github.com/mit-cml/workspace-multiselect/pull/43

mjgallag:
## The basics

### Resolves
Resolves #2599.
### Proposed Changes

Removes backpack count from copy to backpack message.

### Reason for Changes

https://github.com/mit-cml/workspace-multiselect/pull/43

### DIFF
--- a/plugins/workspace-backpack/src/backpack_helpers.ts
+++ b/plugins/workspace-backpack/src/backpack_helpers.ts
@@ -108,11 +108,7 @@ function registerCopyToBackpack(disablePreconditionContainsCheck: boolean) {
       if (!scope.block) {
         return '';
       }
-      const backpack = scope.block.workspace
-        .getComponentManager()
-        .getComponent('backpack') as Backpack;
-      const backpackCount = backpack.getCount();
-      return `${Blockly.Msg['COPY_TO_BACKPACK']} (${backpackCount})`;
+      return Blockly.Msg['COPY_TO_BACKPACK'];
     },
     preconditionFn: function (scope: Blockly.ContextMenuRegistry.Scope) {
       if (!scope.block) return 'hidden';


BenHenning:
Thanks for creating this @mjgallag!

We're not quite sure about this solution. From looking at the reference issue and the proposed code here, it seems to me that there are two parts to this:
1. The actual message for the copy label seems incorrect. It ought to include a `%1` interpolation placeholder that gets filled in with the count using `replace` rather than the suffix being attached to the label string. This allows for proper ordering and contextualization during translations for better internationalization support. This is an existing problem with the plugin.
2. We would generally expect the message string itself to be replaced in your application outside the plugin to a preferred string that doesn't include the count since that's not wanted.

(2) is the main emphasis here since we generally don't want to change a plugin's behavior for a single use case since it affects everyone else who uses it.

However, if there's a reason why (2) doesn't work for you or there's a more complex constraint that needs consideration, could you please file an issue that provides that context? We can then triage that issue and ask follow-up questions, and possibly provide other suggestions in case there are alternative solutions. This will probably be easier for us when considering this & other solutions rather than discussing everything in a PR thread.

Closing this for now since the discussion is better continued in an issue.

mjgallag:
@BenHenning thanks for the response, I created https://github.com/google/blockly-samples/issues/2599 to continue the discussion.

BenHenning:
Deferring to @maribethb but happy to take another pass on this if/when needed.

maribethb (approved):

Issue #2601, opened 2025-09-17
error in shareable-procedures mutator "allow statements" checkbox


bbrudastyy:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Component

block-shareable-procedures

### Description

Hello everyone, I noticed a problem that when there is an inserted block in the input of the procedure block and switching the block insertion resolution mode, an error appears that breaks the logic of the library

https://github.com/user-attachments/assets/be5e0eea-2560-431b-909a-236d205805cd

### Reproduction steps

1. I showed it in the video
2. 
3.

### Stack trace

```shell
Uncaught TypeError: _a.call is not a function
    at BlockSvg$$module$build$src$core$block_svg.setStatements_ (blocks.ts:364:32)
    at BlockSvg$$module$build$src$core$block_svg.compose (blocks.ts:581:12)
```

PR #2602, opened 2025-09-17
fix: check statementConnection_ is a function before calling

bbrudastyy:
## The basics

- [ x] I [validated my changes](https://developers.google.com/blockly/guides/contribute/samples#making_and_verifying_a_change)

### Resolves

https://github.com/google/blockly-samples/issues/2601
Fixes

### DIFF
--- a/plugins/block-shareable-procedures/src/blocks.ts
+++ b/plugins/block-shareable-procedures/src/blocks.ts
@@ -361,7 +361,9 @@ const procedureDefUpdateShapeMixin = {
         this.moveInputBefore('STACK', 'RETURN');
       }
       // Restore the stack, if one was saved.
-      this.statementConnection_?.(this, 'STACK');
+      if (typeof this.statementConnection_ === 'function') {
+        this.statementConnection_(this, 'STACK');
+      }
       this.statementConnection_ = null;
     } else {
       // Save the stack, then disconnect it.


bbrudastyy:
I also noticed a complicated code, so I suggest making it easier. I can add a commit in the same PR.
image
I suggest using the usual negation `this.setStatements_(state['hasStatements'] !== false);`

BenHenning:
Thanks @bbrudastyy.

To start, we'll want to triage #2601 on our side in case there's additional clarity needed for the issue.

As for the fix, it would be really helpful to understand why this is the best approach for fixing the problem. While it does mitigate the direct error that you encountered, it's not clear if it's actually the correct thing to do. It may be the case, for instance, that `statementConnection_` should only ever be a function or null and that this change is hiding a problem happening elsewhere in the plugin or core Blockly.

Could you dig into this a bit more? I think something that would help is identifying what the type actually is (if it isn't a function), why is it that type (e.g. by tracing when it changes), and whether that's correct (i.e. is something assigning it the incorrect value). If it turns out that it can sometimes not be a function, then we probably need to consider all the values it can have and ensure the plugin does the correct thing for each one. All of this may still end up with essentially the solution you landed at, but we can then explain why it's the correct approach by pointing to the places in code which assign it to a non-function value and why it's okay to ignore that in this case (or the code will be updated to account for all the different legitimate values).

BenHenning:
Also: please fill in the PR description fully and correctly. There should be no space before the 'x' for the checkbox, the issue number should go after 'Fixes' (so that GitHub knows to link the issue to the PR), and the other sections should be filled in (I can provide guidance on how to do that if it helps, though I also suspect the investigation I requested in my previous comment will give you a lot of useful details to add to those sections).

BenHenning:
@bbrudastyy since this has been open for a few weeks, I'm going to go ahead and close this. If you'd like to get it merged, please feel free to reopen and address the requested changes in my previous message above.

PR #2600, opened 2025-09-17
Fix compilation issue with vite

rlconst:
Without explicit type export I see an error
```
node_modules/@blockly/continuous-toolbox/src/index.ts (14:26): "LabelFlyoutItem" is not exported by "node_modules/@blockly/continuous-toolbox/src/ContinuousFlyout.ts", imported by "node_modules/@blockly/continuous-toolbox/src/index.ts".
```

[Description](https://github.com/evanw/esbuild/issues/314#issuecomment-668401819)

PR #2607, opened 2025-09-27
fix: use type-only export for LabelFlyoutItem to fix build error

### LINKED ISSUE #2608: Build error when importing LabelFlyoutItem as a value
### Check for duplicates

### Component

blockly-samples/tree/master/plugins/continuous-toolbox/src/index.ts
(ContinuousFlyout / TypeScript exports)

### Description

When importing LabelFlyoutItem as a value from ContinuousFlyout.ts, the build fails because LabelFlyoutItem is a TypeScript interface and does not exist at runtime.

Expected behavior:
- LabelFlyoutItem should be exported as a type-only export to avoid build errors.

Actual behavior:
- Rollup, Vite, and other build tools fail with:
  "LabelFlyoutItem" is not exported by ContinuousFlyout.ts

### Reproduction steps

1. Import LabelFlyoutItem as a value in another TypeScript file:
   import { LabelFlyoutItem } from './ContinuousFlyout';
2. Run a build using Vite or Rollup.
3. Observe the build error: "LabelFlyoutItem" is not exported by ContinuousFlyout.ts

### Stack trace

```shell
"LabelFlyoutItem" is not exported by ContinuousFlyout.ts
```

### Screenshots

Image

coririri:
LabelFlyoutItem is a TypeScript interface and should be exported as a type-only export. 
Exporting it as a value causes build tools (Rollup, Vite, etc.) to fail because the value does not exist at runtime.

## The basics

- [x] I validated my changes by running a Vite build and verifying no errors occurred

### Resolves

Fixes build error when importing LabelFlyoutItem as a value from ContinuousFlyout.ts
(`"LabelFlyoutItem" is not exported by ContinuousFlyout.ts`)

Fixes #2608

### Proposed Changes

- Changed `index.ts` to export LabelFlyoutItem as a type-only export:
```ts
import { ContinuousFlyout } from './ContinuousFlyout';
import type { LabelFlyoutItem } from './ContinuousFlyout';

export { ContinuousFlyout } from './ContinuousFlyout';
export type { LabelFlyoutItem } from './ContinuousFlyout';

### DIFF
--- a/plugins/continuous-toolbox/src/index.ts
+++ b/plugins/continuous-toolbox/src/index.ts
@@ -11,8 +11,8 @@
 import * as Blockly from 'blockly/core';
 
 import {ContinuousCategory} from './ContinuousCategory';
-import { ContinuousFlyout } from './ContinuousFlyout';
-import type { LabelFlyoutItem } from './ContinuousFlyout';
+import {ContinuousFlyout} from './ContinuousFlyout';
+import type {LabelFlyoutItem} from './ContinuousFlyout';
 import {ContinuousMetrics} from './ContinuousMetrics';
 import {ContinuousToolbox} from './ContinuousToolbox';
 import {RecyclableBlockFlyoutInflater} from './RecyclableBlockFlyoutInflater';
--- a/plugins/continuous-toolbox/src/index.ts
+++ b/plugins/continuous-toolbox/src/index.ts
@@ -11,7 +11,8 @@
 import * as Blockly from 'blockly/core';
 
 import {ContinuousCategory} from './ContinuousCategory';
-import {ContinuousFlyout, LabelFlyoutItem} from './ContinuousFlyout';
+import { ContinuousFlyout } from './ContinuousFlyout';
+import type { LabelFlyoutItem } from './ContinuousFlyout';
 import {ContinuousMetrics} from './ContinuousMetrics';
 import {ContinuousToolbox} from './ContinuousToolbox';
 import {RecyclableBlockFlyoutInflater} from './RecyclableBlockFlyoutInflater';

### CODE REVIEW

**Comment by BenHenning** [comment]
Thanks @coririri! Could you please do three things:

1. Fix the format failure that was caught in CI? This should be as simple as running `npm run format`.
2. File an issue that includes sample code or references and the failure that you're fixing here? Then, that issue can be marked as fixed by this PR. This helps us better catalog the specific problems and what each PR is attempting to fix.
3. Update this PR to be 'fix' instead of 'feat' since it seems to actually be fixing an error rather than introducing new functionality.

**Comment by coririri** [comment]
Hi @BenHenning,

Thanks for the review!

1. The format issues caught by CI have been fixed.

2. The related issue with sample code and references has been created. (https://github.com/google/blockly-samples/issues/2608)

3. The PR title is already using fix:, so the third point is already addressed.

All set from my side—this PR is ready to be merged.
Best regards,
coririri

**Review by BenHenning — APPROVED** [review]
Thanks @coririri! I think we just missed this when migrating the plugin to TypeScript.

Also apologies on mixing up the title. I seem to have read 'fix' as 'feat' and gave confusing advice that wasn't needed. :)

This looks great--approving.

File: plugins/continuous-toolbox/src/index.ts [hunk 1]
```diff
@@ -11,7 +11,8 @@
 import * as Blockly from 'blockly/core';
 
 import {ContinuousCategory} from './ContinuousCategory';
-import {ContinuousFlyout, LabelFlyoutItem} from './ContinuousFlyout';
+import {ContinuousFlyout} from './ContinuousFlyout';
+import type {LabelFlyoutItem} from './ContinuousFlyout';
```
cpcallen, line 15: This could have been
```JS
import {ContinuousFlyout, type LabelFlyoutItem} from './ContinuousFlyout';
```

(See #2600 for example.)

PR #2621, opened 2025-10-23
docs: Update context menu codelab for v12 (re-try)

rpbourret:
## The basics

### Resolves

v12 made two major changes to context menus:

*   Added a focusedNode property to the scope argument
*   Added support for separators

The codelab also was inconsistent in its use of the terms item/registry item (the thing in the registry) and menu option (the thing in the context menu).

### Proposed Changes

In the codelab doc:

*   Remove scopeType and replace uses of scopeType with scope.focusedNode
*   Remove the section Scope/Block scope; restricting menu options to blocks is now illustrated in the Precondition: Node type section.
*   Add a section on Separators
*   Clarify/clean up uses of the terms item and option.
*   Minor formatting cleanup. (Copied code is now indented properly for pasting.)

In the codelab code:

*   Update starter-code/index.js and complete-code/index.js accordingly.

### Reason for Changes

Current codelab is not up to date.

### Test Coverage

Tested by hand.

### Documentation

PR includes documentation fixes.

### DIFF
--- a/codelabs/context_menu_option/context_menu_option.md
+++ b/codelabs/context_menu_option/context_menu_option.md
@@ -138,7 +138,7 @@ Each registry item has a `preconditionFn`. It is called by Blockly to decide whe
 
 ### The scope argument
 
-The `scope` argument is an object that is passed to `preconditionFn`. You'll use the `scope.focusedNode` property to determine which object the context menu was invoked on. Why a focused node? Why a focused node? Because all Blockly components that support context menus implement the `IFocusableNode` interface, so it's a handy way to pass objects (like workspaces, blocks, and comments) that otherwise have nothing in common.
+The `scope` argument is an object that is passed to `preconditionFn`. You'll use the `scope.focusedNode` property to determine which object the context menu was invoked on. Why a focused node? Because Blockly keeps track of where the user is -- that is, what node (component) the user is focused on -- and opens the context menu on that node.
 
 ### Return value
 
--- a/codelabs/context_menu_option/context_menu_option.md
+++ b/codelabs/context_menu_option/context_menu_option.md
@@ -152,8 +152,10 @@ You can now test `scope.focusedNode` to display the "Hello World" option in work
   const helloWorldItem = {
     ...
     preconditionFn: function (scope) {
-      if (scope.focusedNode instanceof Blockly.WorkspaceSvg ||
-          scope.focusedNode instanceof Blockly.BlockSvg) {
+      if (
+        scope.focusedNode instanceof Blockly.WorkspaceSvg ||
+        scope.focusedNode instanceof Blockly.BlockSvg
+      ) {
         return 'enabled';
       }
       return 'hidden';
--- a/codelabs/context_menu_option/context_menu_option.md
+++ b/codelabs/context_menu_option/context_menu_option.md
@@ -81,16 +81,16 @@ Add a function to `index.js` named `registerHelloWorldItem`. Create a new regist
 
 ```js
 function registerHelloWorldItem() {
-    const helloWorldItem = {
-      displayText: 'Hello World',
-      preconditionFn: function(scope) {
-        return 'enabled';
-      },
-      callback: function(scope) {
-      },
-      id: 'hello_world',
-      weight: 100,
-    };
+  const helloWorldItem = {
+    displayText: 'Hello World',
+    preconditionFn: function(scope) {
+      return 'enabled';
+    },
+    callback: function(scope) {
+    },
+    id: 'hello_world',
+    weight: 100,
+  };
 }
 ```
 
@@ -232,14 +232,14 @@ It is also passed a `PointerEvent` which is the original event that triggered op
 As an example, update the help item's `callback` to add a block to the workspace when clicked:
 
 ```js
-callback: function(scope) {
-  Blockly.serialization.blocks.append({
-    'type': 'text',
-    'fields': {
-      'TEXT': 'Now there is a block'
-    }
-  }, scope.focusedNode);
-}
+    callback: function(scope) {
+      Blockly.serialization.blocks.append({
+        'type': 'text',
+        'fields': {
+          'TEXT': 'Now there is a block'
+        }
+      }, scope.focusedNode);
+    },
 ```
 
 ### Test it
@@ -309,7 +309,7 @@ Every registry item has an `id` that can be used to unregister it. You can use t
 For instance, you can remove the item that deletes all blocks on the workspace:
 
 ```js
-Blockly.ContextMenuRegistry.registry.unregister('workspaceDelete');
+  Blockly.ContextMenuRegistry.registry.unregister('workspaceDelete');
 ```
 
 ### Default items
@@ -343,7 +343,7 @@ function registerSeparators() {
     separator: true,
   }
   Blockly.ContextMenuRegistry.registry.register(blockSeparator);
-};
+}
 ```
 
 As usual, remember to call `registerSeparators()` from your `start` function.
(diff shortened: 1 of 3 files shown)


codelabs/context_menu_option/context_menu_option.md
```diff
@@ -55,51 +56,51 @@ To run the code, simple open `starter-code/index.html` in a browser. You should
...
-## Add a context menu option to the workspace
+## Add a context menu item
 
-In this section you will create a very basic `Blockly.ContextMenuRegistry.RegistryItem`, then register it to display when you right-click on the workspace.
+In this section you will create a very basic `Blockly.ContextMenuRegistry.RegistryItem`, then register it to display when you right-click on the workspace, a block, or a comment.
 
 ### The RegistryItem
 
-Blockly stores context menu options as items in a registry. When the user right-clicks, Blockly queries the registry for a list of context menu options that should be displayed.
+A context menu consists of one or more menu options that a user can click. Blockly stores information about menu option as items in a registry. You can think of the _registry items_ as templates for constructing _menu options_. When the user right-clicks, Blockly retrieves all of the registry items that apply to the current context and uses them to construct a list of menu options.
```
maribethb: nit: consider `select` instead of `click` to reflect their keyboard-navigability?

similarly `opens the context menu` rather than `right-clicks`

codelabs/context_menu_option/context_menu_option.md
```diff
@@ -113,165 +114,136 @@ function start() {
...
-  blockItem.id = 'hello_world_block';
-  Blockly.ContextMenuRegistry.registry.register(blockItem);
-```
+Reload your web page and right-click on the workspace. You should see a new option labeled "Hello World" at the bottom of the context menu.
 
-Notice that this code uses the JavaScript spread operator to copy the original item object, then replaces the scope type and id. Simply updating `workspaceItem` and re-registering it would modify the original registry item in place, leading to unintended behaviour.
+![A context menu. The last option says "Hello World".](hello_world.png)
 
-### Test it
+Next, drag a block onto the workspace and right-click on the block. You'll see "Hello World" at the bottom of the block's context menu. Finally, right-click on the workspace and create a comment, then right-click on the comment's header. "Hello World" should be at the bottom of the context menu.
```
maribethb: the context menu options for adding workspace comments [aren't present by default](https://developers.google.com/blockly/guides/configure/web/workspace_comment#enable_workspace_comments), so you'll either need to add them to the starter/complete code or remove this part. I think I would remove this part and have your next example only work on the workspace and not blocks.

codelabs/context_menu_option/context_menu_option.md
```diff
@@ -113,165 +114,136 @@ function start() {
...
 
-![An if block with a context menu with five items. The last item says "Hello World".](hello_world_block.png)
+Each registry item has a `preconditionFn`. It is called by Blockly to decide whether and how to display an option on a context menu. You'll use it to display the "Hello, World" option on workspace and block context menus, but not on comment context menus.
 
-## Precondition
+### The scope argument
 
-Each registry item has a `preconditionFn`. This function takes in a scope and returns a string
-indicating whether and how to display the context menu option. We will discuss the scope in the next section.
+The `scope` argument is an object that is passed to `preconditionFn`. You'll use the `scope.focusedNode` property to determine which object the context menu was invoked on. Why a focused node? Because all Blockly components that support context menus implement the `IFocusableNode` interface, so it's a handy way to pass objects (like workspaces, blocks, and comments) that otherwise have nothing in common.
```
maribethb: Also, if the context menu is being opened on an object, then it has focus. So it's the thing being acted upon by the context menu.

codelabs/context_menu_option/context_menu_option.md
```diff
@@ -113,165 +114,136 @@ function start() {
...
+Next, drag a block onto the workspace and right-click on the block. You'll see "Hello World" at the bottom of the block's context menu. Finally, right-click on the workspace and create a comment, then right-click on the comment's header. "Hello World" should be at the bottom of the context menu. 
 
-Drag a block into the workspace and right-click it. You should see a "Hello world" option on the block context menu.
+## Precondition: Node type
 
-![An if block with a context menu with five items. The last item says "Hello World".](hello_world_block.png)
+Each registry item has a `preconditionFn`. It is called by Blockly to decide whether and how to display an option on a context menu. You'll use it to display the "Hello, World" option on workspace and block context menus, but not on comment context menus.
 
-## Precondition
+### The scope argument
```
maribethb, line 139: The other thing you might want to explain somewhere in here is that the context menus may be opened on a wide variety of components. for example, the keyboard-navigation plugin may allow users to open context menus on connections. so if they're restricting the scope in which a menu option appears, they should use an allow-list style instead of a deny-list style. in other words, check that the focused node IS a block rather than IS NOT a workspace. because they may not be able to anticipate all places a menu will be opened.

maribethb (commented):
Mostly some suggestions you should feel free to push back on, but you will have to decide one way or the other about the workspace comment bit.

codelabs/context_menu_option/context_menu_option.md
rpbourret: Agreed.

codelabs/context_menu_option/context_menu_option.md
rpbourret: Both the starter code and the complete code call `registerCommentOptions()`. I didn't call it out in the codelab text because the codelabs don't generally discuss starter code.

I deliberately included comments here and left them out in the next section because I wanted people to see we were testing for BlockSvg (as opposed to, say, Block). This is based on the guess that people are more likely to change the context menus for workspaces and blocks than other things.

codelabs/context_menu_option/context_menu_option.md
rpbourret: I'll clarify that.

To be clear, I consider both of the following to be allow-lists:

```
  if (scope.focusedNode instanceof BlockSvg) {
    // do something
  }
```

```
  if (!(scope.focusedNode instanceof BlockSvg)) return;
  // do something
```

Any objections to the latter? I use both.

maribethb: Yeah, the latter is fine (and I tend to prefer it since it reduces the overall amount of nesting)

I think it would be good to explain this approach directly and not only rely on it being how the examples are done, is all

codelabs/context_menu_option/context_menu_option.md
maribethb: Oh, cool. I have no idea how I missed that you already added them. Sounds good then!

codelabs/context_menu_option/context_menu_option.md
rpbourret: I replaced my more technical explanation with something closer to what you suggest. Mostly, I expect readers to have no idea what focus is -- we barely discuss it in the docs -- and I'm just trying to explain the property name.

codelabs/context_menu_option/context_menu_option.md
```diff
@@ -164,9 +164,11 @@ You can now test `scope.focusedNode` to display the "Hello World" option in work
   };
 ```
 
+Notice that the code tests for where context menus are allowed, rather than where they are not allowed. This is because custom code (such as a plugin) can add context menus to any Blockly component that can be focused. Thus, testing for what something isn't may result in allowing context menus on more components than you anticipated.
```
maribethb: > Thus, testing for what something isn't

How about... Thus, testing for specific types rather than allowing all (or all but certain types) will ensure that context menus are not shown on more components than you anticipated.

I'm not sure if that's actually better. If you don't like that, maybe putting `isn't` in italics would help make it read the appropriate emphasis?

maribethb (approved):
LGTM with minor wording quibble

codelabs/context_menu_option/context_menu_option.md
rpbourret: Fixed wording per email suggestion

PR #2646, opened 2025-12-04
Add a CSS codelab

rpbourret:
## The basics

### Resolves

Adds a CSS codelab.

### Proposed Changes

Adds a CSS codelab in the form of a Markdown file, PNG screenshots, and starter and completed code.

### Reason for Changes

We need a CSS codelab.

### Test Coverage

Manually tested.

### Documentation

Included.


BenHenning:
I think @maribethb is planning to look at this, so de-assigning myself unless I'm told that I need to review this.

codelabs/css/css.md
```diff
@@ -0,0 +1,734 @@
...
+### What you'll build
+
+A simple Blockly workspace that uses the same Halloween colours as the
+[Customizing your themes](https://blocklycodelabs.dev/codelabs/theme-extension-identifier/index.html#0)
+codelab.
+
+### What you'll need
+
+-   A browser
+-   Basic knowledge of HTML, CSS, and JavaScript.
```
rachel-fenichel: Add: Basic knowledge of SVG elements.

codelabs/css/css.md
```diff
@@ -0,0 +1,734 @@
...
+
+![A Blockly editor with Halloween colors.](components-halloween.png)
+
+### fill vs background-color
+
+You might have noticed that some rules use `background-color` and `color` and
+others use `fill` and `stroke`. This is because `background-color` and `color`
+apply to HTML elements, like the `<div>` used by the toolbox, and `fill` and
+`stroke` apply to SVG elements, like the `<path>` used by the flyout background.
+(An exception to this is the `<svg>` element, which uses `background-color` and
```
rachel-fenichel: It's confusing to say that fill and stroke apply to SVG elements, except for the `<svg>` element. 

How about:

"An exception to this is the top-level `<svg>` element that contains Blockly, which uses `background-color` and `color`."

codelabs/css/css.md
```diff
@@ -0,0 +1,734 @@
...
+`color`.)
+
+### The !important declaration
+
+You might have also noticed that some rules use an `!important` declaration
+while others don't. This is because Blockly sets colours in several different
```
rachel-fenichel, line 283: "This is because Blockly sets colours in several different ways" -> "This is because Blockly dynamically creates and sets colours of some elements based on block definitions."

codelabs/css/css.md
```diff
@@ -0,0 +1,734 @@
...
+
+The easiest way to determine how a rule is set is to highlight the appropriate
+element in the element inspector and look at the corresponding style
+information. In a few cases, this isn't possible. For example, an insertion
+marker is created only when you drag a child near its parent and is deleted when
+you let go of the parent to highlight the insertion marker's element. In these
+cases, you will need to
+[search Blockly's rules](https://developers.google.com/blockly/guides/configure/web/appearance/css#blockly_css_rules).
+
+A simpler rule of thumb is that if your rule doesn't work, you probably need an
```
rachel-fenichel: Oof. I'd prefer not to give this advice, since CSS [best practices](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/important#best_practices) involve not using `!important`

I would prefer no advice here, just the listing of different ways that styles are set (as above).

codelabs/css/css.md
```diff
@@ -0,0 +1,734 @@
...
+    {
+      kind: 'category',
+      name: 'Logic',
+      categorystyle: 'logic_category',
+      contents: [...],
+    },
+```
+
+The `categorystyle` property assigns a style that is used by a theme. Because
+you're not using themes to assign category colours, replace this with a
```
rachel-fenichel: "Because you're not using themes to assign category colours, you don't need the `categorystyle` property. Delete it and add a `cssConfig` property that [...]"

codelabs/css/css.md
```diff
@@ -0,0 +1,734 @@
...
+          <div class="blocklyToolboxCategoryContainer">
+            <div class="blocklyToolboxCategory" id="blockly-1">
+```
+
+Unfortunately, the only thing that distinguishes this `<div>` from other
+category `<div>`s is a generated `id` attribute (`blockly-1`). This isn't stable
+enough to use in a CSS rule -- for example, if you switched the order of two
+categories you'd also have to switch the selectors in their rules.
+
+To solve this problem, you'll need to add a class. Open the `toolbox.js` file
```
rachel-fenichel: "To solve this problem, you'll need to add a class to your `<div>` in your toolbox definition."

codelabs/css/css.md
```diff
@@ -0,0 +1,734 @@
...
+
+#### Choose an element to use
+
+Next, you need to decide what element to use in your colour rule. The `<g>`
+element identifies the block but doesn't draw it. Instead, you can use the `<g>`
+element's first child. This is a `<path>` element with `fill` and `stroke`
+presentation attributes, which are easily overridden.
+
+(Note that different renderers use different numbers of `<path>` elements to
+draw a block. The renderer used in this codelab (thrasos) uses a single `<path>`
```
rachel-fenichel: Capitalize Thrasos and Zelos, and add a link to the renderer documentation.

rachel-fenichel: Or perhaps move this note to the top of the block colouring section, since you also mention renderer-specific changes in the next subsection.

codelabs/css/css.md
```diff
@@ -0,0 +1,734 @@
...
+/**************/
+/* COMPONENTS */
+/**************/
+
+.blocklySvg {
+  background-color: #ff7518;
+}
+```
+
+Refresh your page and notice the change in colour:
```
rachel-fenichel: "and notice that the workspace background is now orange:"

codelabs/css/css.md
```diff
@@ -0,0 +1,734 @@
...
+Next, add a CSS rule to set the workspace's background colour:
+
+```css
+.blocklyMutatorBackground {
+  fill: #ff7518;
+}
+```
+
+To see your new colours, reload the web page, drag the `if do` block out again,
+and reopen the mutator. You should see something like this:
```
rachel-fenichel: "You should see that the mutator workspace background colour is orange, matching the main workspace background:"

codelabs/css/css.md
```diff
@@ -0,0 +1,734 @@
...
+}
+
+.blocklyInsertionMarker > .blocklyPath {
+  fill: #fff !important;
+  fill-opacity: 0.3 !important;
+  stroke: none;
+}
+```
+
+Now reload your page. You should see a Blockly editor with Halloween colours:
```
rachel-fenichel: "You should see a Blockly editor with a yellow toolbox and yellow scrollbars:"

codelabs/css/css.md
```diff
@@ -0,0 +1,734 @@
...
+  stroke: #c5eaff;
+}
+
+.logic_blocks.blocklyShadow > .blocklyPath {
+  fill: #ff0000;
+  stroke: none;
+}
+```
+
+Refresh your web page and open the **Logic** category. You should see something
```
rachel-fenichel: "You should see that the logic blocks are now rendered in an autumnal brown instead of blue:"

codelabs/css/css.md
rpbourret: Done

codelabs/css/css.md
rpbourret: Done

codelabs/css/css.md
rpbourret: I'm not sure I understand this comment. The point here is that Blockly uses several different techniques to set colours and you have to understand which one is being used in order to override it. It doesn't matter if the colour was set dynamically or statically -- the user is looking at/overriding the end state. Also, this applies to all colours, not just those used in blocks.

codelabs/css/css.md
rpbourret: Removed.

codelabs/css/css.md
rpbourret: Done

codelabs/css/css.md
rpbourret: Fixed with slightly different wording.

codelabs/css/css.md
rpbourret: Added link and capitals. Moved the more general part of the note to the start of the block fill and stroke section.

codelabs/css/css.md
rpbourret: Done

codelabs/css/css.md
rpbourret: Done

codelabs/css/css.md
rpbourret: Done

codelabs/css/css.md
rpbourret: Done

codelabs/css/css.md
rachel-fenichel: Fair, the current wording is fine.

rachel-fenichel (approved):
Thank you!

PR #2648, opened 2025-12-05
fix: toolbox search final trigram

### LINKED ISSUE #2645: Toolbox search: long queries drop the last trigram, causing false positives (e.g. “conso” matches “math constrain”)
### Check for duplicates

### Component

toolbox-search plugin (plugins/toolbox-search), version 3.1.3 (current main)

### Description

When searching in the toolbox search category, queries longer than three characters can match blocks that don’t contain the full query. Example: typing “conso” returns the built-in math_constrain block even though its text is “math constrain”. The trigram generator in src/block_searcher.ts stops one character early:

```
for (let start = 0; start < normalizedInput.length - 3; start++)
```

so “conso” produces only ["con", "ons"] (missing nso). Because matches are the intersection of generated trigrams, dropping the final trigram broadens results and yields false positives. Expected: include the final trigram (loop should use <=), so “conso” would not match “math constrain”.

I’m happy to work on a fix; the minimal change is to adjust the trigram loop in src/block_searcher.ts to include the final trigram (<= normalizedInput.length - 3).

### Reproduction steps

1. Install @blockly/toolbox-search and include a toolbox that contains the default math blocks (so math_constrain is present).
2. Add the search category to the toolbox: { kind: 'search', name: 'Search', contents: [] }.
3. Open the search category and type conso.
4. Observe that math_constrain appears in the results.

**Expected result**
No matches for “conso” unless a block actually contains that substring.

**Actual result**
math_constrain (and any block with “con”+“ons” but not the full “conso”) appears.

### Stack trace

```shell

```

### Screenshots

_No response_

goughjo02:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly-samples/issues/2645

### Proposed Changes

- Include the final trigram when generating search tokens to prevent partial matches on long queries.
- Add a regression test covering the “conso” false positive case.
- Expand BlockSearcher tests for short/empty input, underscore normalization, dropdown alt text indexing, and disambiguation between similar blocks.

### Reason for Changes

- Prevent toolbox search from returning false positives for queries longer than three characters and improve overall search accuracy.

### Test Coverage

- `npm test -- --runInBand` (blockly-scripts test)

### Documentation

- No documentation updates needed.

### Additional Information

- Additional tests define temporary blocks/dropdowns to exercise edge cases; no user-facing behavior changes beyond correcting search results.

### DIFF
--- a/plugins/toolbox-search/test/tests.mocha.js
+++ b/plugins/toolbox-search/test/tests.mocha.js
@@ -57,6 +57,23 @@ suite('BlockSearcher', () => {
     assert.sameMembers(ransomNoteMatches, [listCreateWithBlock]);
   });
 
+  test('requires the final trigram when matching longer queries', () => {
+    const searcher = new BlockSearcher();
+    const mathConstrainBlock = {
+      kind: 'block',
+      type: 'math_constrain',
+    };
+    searcher.indexBlocks([mathConstrainBlock]);
+
+    const matches = searcher.blockTypesMatching('conso');
+
+    assert.notInclude(
+      matches,
+      mathConstrainBlock,
+      'query missing trailing trigram should not match',
+    );
+  });
+
   test('returns an empty list when no matches are found', () => {
     const searcher = new BlockSearcher();
     assert.isEmpty(searcher.blockTypesMatching('abc123'));
--- a/plugins/toolbox-search/test/tests.mocha.js
+++ b/plugins/toolbox-search/test/tests.mocha.js
@@ -15,6 +15,15 @@ suite('Toolbox search', () => {
 });
 
 suite('BlockSearcher', () => {
+  test('generateTrigrams handles empty and short input', () => {
+    const searcher = new BlockSearcher();
+    const generateTrigrams = searcher.generateTrigrams.bind(searcher);
+
+    assert.deepEqual(generateTrigrams(''), []);
+    assert.deepEqual(generateTrigrams('a'), ['a']);
+    assert.deepEqual(generateTrigrams('abc'), ['abc']);
+  });
+
   test('indexes the default value of dropdown fields', () => {
     const searcher = new BlockSearcher();
     const blocks = [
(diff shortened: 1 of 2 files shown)


plugins/toolbox-search/test/tests.mocha.js
```diff
@@ -57,6 +66,119 @@ suite('BlockSearcher', () => {
...
+
+    const searcher = new BlockSearcher();
+    const blockInfo = {
+      kind: 'block',
+      type: 'searcher_underscore_block',
+    };
+    searcher.indexBlocks([blockInfo]);
+
+    assert.sameMembers(
+      searcher.blockTypesMatching('custom block with underscore'),
```
gonfunko: If the test is verifying that underscores in the block type get treated as spaces, should this be 'searcher underscore block'? As written this would match the unmodified message, but not verify the normalization of the block type.

plugins/toolbox-search/test/tests.mocha.js
```diff
@@ -57,6 +66,119 @@ suite('BlockSearcher', () => {
...
+          type: 'searcher_dropdown_alt',
+          message0: 'weather %1',
+          args0: [
+            {
+              type: 'field_dropdown',
+              name: 'WEATHER',
+              options: [
+                [
+                  {
+                    src: 'data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEA',
```
gonfunko: This doesn't appear to actually be a valid image; if it doesn't otherwise cause problems, can you set `src` to `''` here and below to avoid confusion as to what this is and whether the particular value is relevant to the test case?

gonfunko:
Thank you for identifying and reporting this issue and sending a fix! I really appreciate the improved tests as well. I just had two minor questions, and if you could run `npm run format` and commit the results of that to make the linter happy I think we should be good to merge this.

plugins/toolbox-search/test/tests.mocha.js
goughjo02: I have update the code to verify the block type `searcher_underscore_block ` gets normalised

plugins/toolbox-search/test/tests.mocha.js
goughjo02: i have deleted the contents of src strings. they are empty strings now

gonfunko:
Thank you for making these changes, I really appreciate the fix!

PR #2659, opened 2025-12-22
chore: update code owners to enable auto code review assignment

maribethb:
## The basics

Update CODEOWNERS file so that auto review assignment works again.


rachel-fenichel (approved):
Oh, is that why Ben was getting everything?

maribethb:
> Oh, is that why Ben was getting everything?

I don't think it was assigning anyone automatically.

PR #2664, opened 2026-01-05
chore: add a workflow input to publish from-package

maribethb:
## The basics

### Resolves

Works on publishing blockly-samples

### Proposed Changes

- adds more logging because I can't figure out why OIDC isn't working :(
- adds a workflow option to try publishing from-package so that I can use that when publishing fails

### Test Coverage

Ran this through a github actions validator

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,12 @@ name: publish
 
 on:
   workflow_dispatch: # Manually trigger. Colon is required.
+    inputs:
+      from-package:
+        description: 'Check to publish from-package instead of the default publish. This is used to publish packages that have already been version bumped and tagged by a previous workflow run, but failed to publish for some reason.'
+        required: false
+        type: boolean
+        default: false
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -55,7 +61,14 @@ jobs:
       - name: Publish
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: cd plugins && npx lerna publish --no-private --conventional-commits --create-release github --yes
+          FROM_PACKAGE: ${{ github.event.inputs['from-package'] }}
+        run: |
+          cd plugins
+          if [ "${FROM_PACKAGE}" = "true" ]; then
+            npx lerna publish from-package --yes --loglevel silly
```
gonfunko, line 68: Does this need the `--no-private` flag as well?

gonfunko (approved):

.github/workflows/publish.yml
maribethb: No, it publishes packages that have a version in the npm registry different than the version in git, and private packages are not in the npm registry

PR #2666, opened 2026-01-08
chore: make manual publishing a bit better

maribethb:
## The basics

### Proposed Changes

- Runs the publish:prepare script automatically before running any of the other publish scripts.
- Removes the `npm login` step from the prepare script since logging in no longer helps. You have to use an access token instead. I am updating some internal documentation about how to publish with more information.

### Reason for Changes

Makes publishing manually a bit less error-prone since you can't forget to run the step that builds.

### Test Coverage

Ran `npm run publish:checkVersions` and ensured it works correctly. Did not try to publish manually though.

### Documentation

Updated scripts.md and I'm working on adding some internal documentation about how to publish.

### DIFF
--- a/scripts.md
+++ b/scripts.md
@@ -52,6 +52,50 @@ This script builds all files needed to deploy plugins and examples to GitHub Pag
 
 This script installs a beta version of Blockly, builds all files needed to deploy plugins and examples to GitHub Pages, then starts a local server with that content.
 
+### `npm run publish:manual`
+
+This script will first run `npm run publish:prepare`. It will
+publish all of the changed plugins since the last release, using the `dist`
+directory. It runs the lerna command that uses conventional commits to determine
+a new version number for each plugin, and publishes the new versions to npm and
+to a github release and tag.
+
+If there is some error with npm while running this command, you may end up in a
+state where some plugins have been published to npm and not others, after lerna
+has already tagged the new releases. You can recover from this state by fixing
+the error, and then running either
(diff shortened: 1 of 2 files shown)


scripts.md
```diff
@@ -109,6 +96,21 @@ latest tags pulled. This is taken care of by the `publish:prepare` script.
 
 ## Other Scripts
 
+### `npm run publish:prepare`
+
+This script will clone a copy of blockly-samples to a directory called `dist`,
+run `npm ci`, then build and test all plugins.
+
+Note that this script does not handle npm authentication. You will need to
```
BenHenning: Is this relevant for this script or more for the publish ones? Since `prepare` is just building and never touches npm, this feels out of place (and potentially concerning--as a developer I'm now wondering if it _does_ interact with npm somehow).

I suggest clarifying this for the publish scripts that would presumably fail without this.

Separately, is there a particular set of actions a developer needs to take to mitigate this? It might be nice to clarify that here (or in its own section referenced by the other sections).

BenHenning (approved):
Thanks @maribethb! Nice robustness improvement. I had a few thoughts on the documentation but broadly the changes LGTM.

scripts.md
maribethb: There is more information about how to publish located within the team internal docs. It's there because it deals with authentication details that make more sense to host with the internal team processes. 

Initially the prepare script did run `npm login` so that's why I added the note there, but you're right it doesn't make sense anymore. I rearranged the file a bit to add more details and deter people from trying to publish on the command line anyway.

PR #2667, opened 2026-01-08
chore: remove outdated github actions

maribethb:
## The basics

### Proposed Changes

Removes some outdated github actions that we no longer need.

### Reason for Changes

- cleanup after dependabot: no longer needed as we are not using lerna's bootstrap method so dependabot already behaves correctly.
- publish from package: handled by the main publish action now. we can only have one publish action that uses trusted publishing.
- clean install examples: no longer needed since examples are using npm workspaces and don't need to be force installed.


BenHenning (approved):
Nice clean-up and CI simplification!

PR #2671, opened 2026-01-14
fix: Remove unsafe cast in ContinuousFlyout

### LINKED ISSUE #8884: Custom Toolbox Categories don't work if you have only custom categories
### Check for duplicates

### Description

When using Blockly v11, I've encountered a limitation where setting all categories in a toolbox to use kind="custom_category" causes the toolbox to fail completely. If at least one category is set to the default category type (without kind="custom_category"), the toolbox renders correctly and functions normally.

This appears to be a regression introduced in Blockly v11, as the same configuration works without issues in earlier versions.

### Expected Behavior
All categories marked as `kind="custom_category"` should display and function normally with their custom icons and behaviors.
### Actual Behavior
When all categories are defined as `custom_category`, the toolbox fails to initialize and throws the following TypeError:
```
TypeError: Cannot read properties of null (reading 'getSelectedItem')
```

### Reproduction steps

1. Create a Custom Category Implementation.

```
class CustomToolboxCategory extends Blockly.ToolboxCategory {
    private iconClass_: string;

    constructor(categoryDef: any, toolbox: Blockly.Toolbox, opt_parent?: any) {
        super(categoryDef, toolbox, opt_parent);

        if (categoryDef['xml'] && categoryDef['xml'].getAttribute) {
            this.iconClass_ = categoryDef['xml'].getAttribute('iconClass') || '';
        } else if (categoryDef['iconClass']) {
            this.iconClass_ = categoryDef['iconClass'];
        }
    }

    public override createDom_(): HTMLDivElement {
        const categoryDom = super.createDom_();
        const labelElement = categoryDom.querySelector('.blocklyTreeLabel') as HTMLElement;
        
        if (labelElement && this.iconClass_) {
            const iconElement = document.createElement('i');
            iconElement.className = this.iconClass_;
            iconElement.style.marginRight = '8px';
            labelElement.insertBefore(iconElement, labelElement.firstChild);
        }

        return categoryDom as HTMLDivElement;
    }
}

```

2.  Register it

```
Blockly.registry.register(
    Blockly.registry.Type.TOOLBOX_ITEM,
    'custom_category',
    CustomToolboxCategory,
    true
);
```

3. Inject the workspace.

```
  const injectionOptions: Blockly.BlocklyOptions = {
    toolbox,
    trashcan: false,
    sounds: false,
    move: { scrollbars: true },
    zoom: { startScale: startScale, controls: false, wheel: false },
  };
  this.workspace = inject(this.blocklyDiv.nativeElement, injectionOptions);
}
```

3. Define a toolbox with only custom categories::

``` 
<xml id="toolbox">
  <category name="Category1" colour="#ff0000" kind="custom_category" iconClass="fa-solid fa-image"></category>
  <category name="Category2" colour="#00ff00" kind="custom_category" iconClass="fa-solid fa-image"></category>
  <category name="Category3" colour="#0000ff" kind="custom_category" iconClass="fa-solid fa-image"></category>

</xml>
```
**Result**: Toolbox fails to render with the mentioned TypeError.

4. Add at least one default category (no `kind="custom_category"`):`:

```
<xml id="toolbox">
  <category name="Category1" colour="#ff0000" kind="custom_category" iconClass="fa-solid fa-image"></category>
  <category name="Category2" colour="#00ff00"></category> <!-- default category -->
  <category name="Category3" colour="#0000ff" kind="custom_category" iconClass="fa-solid fa-image"></category>
</xml>
```
**Result**: Toolbox renders correctly and works without errors.

### Stack trace

```shell
free-world.page.ts:89 ERROR TypeError: Cannot read properties of null (reading 'getSelectedItem')
    at ContinuousFlyout.show (ContinuousFlyout.js:227:34)
    at init$$module$build$src$core$inject (blockly_compressed.js:430:277)
    at inject$$module$build$src$core$inject (blockly_compressed.js:423:278)
    at ************.renderBlockly (ngx-blockly.component.ts:119:28)
    at ************.ngOnChanges (ngx-blockly.component.ts:85:10)
    at ************.rememberChangeHistoryAndInvokeOnChangesHook (core.mjs:3139:14)
    at callHookInternal (core.mjs:4207:14)
    at callHook (core.mjs:4238:9)
    at callHooks (core.mjs:4188:17)
    at executeInitAndCheckHooks (core.mjs:4138:9)
```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/8884

### Proposed Changes
This PR removes an unsafe cast that was assuming the parent toolbox is always non-null. I wasn't able to reproduce the behavior described in [#8884](https://github.com/RaspberryPiFoundation/blockly/issues/8884), but the stacktrace indicates that the toolbox was null on one of these lines, so handling that case should resolve it.

### DIFF
--- a/plugins/continuous-toolbox/src/ContinuousFlyout.ts
+++ b/plugins/continuous-toolbox/src/ContinuousFlyout.ts
@@ -76,10 +76,10 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
     const toolbox = this.targetWorkspace.getToolbox();
     if (!toolbox || toolbox instanceof ContinuousToolbox) return toolbox;
 
-    console.log(
+    console.warn(
       'Expected a `ContinuousToolbox` instance but did not find one. ' +
-      'Make sure `registerContinuousToolbox()` has been called and the ' +
-      'continuous toolbox has been injected.'
+        'Make sure `registerContinuousToolbox()` has been called and the ' +
+        'continuous toolbox has been injected.',
     );
 
     return null;
--- a/plugins/continuous-toolbox/src/ContinuousFlyout.ts
+++ b/plugins/continuous-toolbox/src/ContinuousFlyout.ts
@@ -72,8 +72,17 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
    *
    * @returns Toolbox that owns this flyout.
    */
-  private getParentToolbox(): ContinuousToolbox {
-    return this.targetWorkspace.getToolbox() as ContinuousToolbox;
+  private getParentToolbox(): ContinuousToolbox | null {
+    const toolbox = this.targetWorkspace.getToolbox();
+    if (!toolbox || toolbox instanceof ContinuousToolbox) return toolbox;
+
+    console.log(
+      'Expected a `ContinuousToolbox` instance but did not find one. ' +
+      'Make sure `registerContinuousToolbox()` has been called and the ' +
+      'continuous toolbox has been injected.'
+    );
+
+    return null;
   }
 
   /**
@@ -110,7 +119,7 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
       // Note that `FlyoutButton` represents both buttons and labels.
       element instanceof Blockly.FlyoutButton &&
       element.isLabel() &&
-      this.getParentToolbox().getCategoryByName(element.getButtonText())
+      this.getParentToolbox()?.getCategoryByName(element.getButtonText())
     );
   }
 
@@ -146,7 +155,7 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
       ...this.scrollPositions.entries(),
     ].reverse()) {
       if (scaledPosition >= position) {
-        this.getParentToolbox().selectCategoryByName(name);
+        this.getParentToolbox()?.selectCategoryByName(name);
         return;
       }
     }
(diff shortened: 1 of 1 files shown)


plugins/continuous-toolbox/src/ContinuousFlyout.ts
```diff
@@ -72,8 +72,17 @@ export class ContinuousFlyout extends Blockly.VerticalFlyout {
    *
    * @returns Toolbox that owns this flyout.
    */
-  private getParentToolbox(): ContinuousToolbox {
-    return this.targetWorkspace.getToolbox() as ContinuousToolbox;
+  private getParentToolbox(): ContinuousToolbox | null {
+    const toolbox = this.targetWorkspace.getToolbox();
+    if (!toolbox || toolbox instanceof ContinuousToolbox) return toolbox;
+
+    console.log(
```
BenHenning: Should this maybe be a warning or error, instead? It seems like a case where the configuration is actually broken and we're trying to sort of recover as best as we can (which is to basically do nothing I suppose).

BenHenning (approved):
Thanks @gonfunko! This seems a reasonable defensive thing to do, though I wonder if the console log should be a bit more warning/error esque to represent the severity of the fact that the configuration is pretty clearly broken in such a case.

BenHenning:
Also: I suppose it's possible that the flyout is simply null and not the wrong type, right? The stacktrace, as far as I can tell, is just failing on it being null either due to mismatched type or the value being null (unless I'm misreading it...).

plugins/continuous-toolbox/src/ContinuousFlyout.ts
gonfunko: Bumped it to a warning, good call.

gonfunko:
Yes, it could just be null, in which case this should still not throw. All it was doing was trying to select the first category, and if there is no toolbox that just becomes a no-op now which seems reasonable.

PR #2674, opened 2026-02-05
fix: getting started example missing style

joriszwart:


gonfunko (approved):
Thanks for catching and fixing this, we really appreciate it!

PR #2680, opened 2026-03-20
fix: strip disabledReasons when saving blocks to backpack

B-head:
## Summary
- Strip `disabledReasons` property when saving blocks to the backpack, preventing
  disabled blocks from becoming impossible to copy out of the backpack
- Apply the same stripping to blocks added via the `Backpackable` interface
  (`addBackpackables`), which was bypassing `blockToJsonString`
- Refactor: extract key-stripping logic into reusable `cleanFlyoutInfo` method
- Fix: null dereference in `blockToJsonString` when `blocks.save()` returns null

## Test plan
- Disable a block (right-click > Disable Block) and add it to the backpack
- Verify the block can be copied out of the backpack in an enabled state
- Verify non-disabled blocks still work with the backpack as before

Note: Safari has not been tested (no macOS environment available).

### DIFF
--- a/plugins/workspace-backpack/src/backpack.ts
+++ b/plugins/workspace-backpack/src/backpack.ts
@@ -467,32 +467,47 @@ export class Backpack
    * @returns The JSON object as a string.
    */
   private blockToJsonString(block: Blockly.Block): string {
-    const json = Blockly.serialization.blocks.save(block);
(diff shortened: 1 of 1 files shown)


maribethb (approved):
Thanks for the PR, this looks great!

PR #2683, opened 2026-04-01
chore: remove individual plugin package-locks

maribethb:
## The basics

### Resolves

Fixes having unnecessary package-locks in blockly-samples

### Proposed Changes

removes individual package-locks

### Reason for Changes

they aren't being updated anymore because when we switched to using npm workspaces, only the top-level package-lock matters, the individual ones aren't used so they also aren't updated when you add or remove dependencies from a subpackage

this is standard practice to not have the individual ones anymore

### Test Coverage

install still works


mikeharv (approved):
>-19,955

Nice!

PR #2696, opened 2026-06-12
fix!: Add ARIA labels for block-plus-minus

gonfunko:
## The basics

### Proposed Changes
This PR updates the block-plus-minus plugin to add ARIA labels for the plus and minus buttons that describe what they will add and remove.

### Breaking Changes
If you are using the plus and minus buttons on custom blocks, ensure that you pass an accessibility label as the first argument to the methods used to create them. This should describe the action that the button will perform.

### DIFF
--- a/plugins/block-plus-minus/src/field_minus.js
+++ b/plugins/block-plus-minus/src/field_minus.js
@@ -18,8 +18,8 @@ import {getExtraBlockState} from './serialization_helper';
  *     is clicked.
  * @returns {Blockly.FieldImage} The minus field.
  */
-export function createMinusField(args = undefined) {
-  const minus = new Blockly.FieldImage(minusImage, 15, 15, undefined, onClick_);
+export function createMinusField(args = undefined, altText = '') {
+  const minus = new Blockly.FieldImage(minusImage, 15, 15, altText, onClick_);
   /**
    * Untyped args passed to block.minus when the field is clicked.
    * @type {?(Object|undefined)}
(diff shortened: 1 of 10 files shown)


plugins/block-plus-minus/src/field_minus.js
```diff
@@ -18,8 +18,8 @@ import {getExtraBlockState} from './serialization_helper';
  *     is clicked.
  * @returns {Blockly.FieldImage} The minus field.
  */
-export function createMinusField(args = undefined) {
-  const minus = new Blockly.FieldImage(minusImage, 15, 15, undefined, onClick_);
+export function createMinusField(args = undefined, altText = '') {
```
maribethb: pls add the jsdoc the linter is warning you about

maribethb: i wonder if it makes sense to set a default of "Remove branch"? "branch" isn't really a super technical term but we do use it in other announcements like saying an if block has 3 branches. it's not all that accurate for procedures. maybe just "remove" and "add"? i don't like not having a default even if everyone should specify a value.

or even make this non-optional to specify and mark this as a breaking change. you've already updated all the callers of it here so it's only breaking for people using it in custom block definitions, but that's fine, because we really want those people to pass a value here.

maribethb (approved):

plugins/block-plus-minus/src/field_minus.js
gonfunko: Went ahead and made it mandatory, which also required swapping the order of the arguments, but since it's breaking anyway and makes the API a bit cleaner that seems fine.

PR #2698, opened 2026-06-12
fix: Improve accessibility of `FieldDate`

gonfunko:
## The basics

### Proposed Changes
This PR improves the accessibility of `FieldDate`. Specifically, it:

* Indicates the field's type as "date"
* Customizes the ARIA label to use a localized, long-form description of the date, like "January 23, 2024" rather than "1/23/2024" (or local equivalent) which may be read out as "one slash twenty three slash two thousand twenty four", to improve understandability.

[9986](https://github.com/RaspberryPiFoundation/blockly/pull/9986) is also needed to make this field accessible; without it, the value cannot be committed via the keyboard.

### DIFF
--- a/plugins/field-date/src/field_date.ts
+++ b/plugins/field-date/src/field_date.ts
@@ -94,6 +94,32 @@ export class FieldDate extends Blockly.FieldTextInput {
   }
 
   /**
+   * Returns a description of the type of this field for screenreaders.
+   */
+  override getAriaTypeName() {
+    return Blockly.Msg['ARIA_TYPE_FIELD_DATE'];
+  }
+
+  /**
+   * Returns a description of the current date for use by screenreaders.
+   */
+  override getAriaValue() {
+    const stringValue = this.getValue();
+    if (!stringValue) return super.getAriaValue();
+
+    const date = new Date(stringValue);
+    // Use a localized long-form description of the date, e.g. January XX, 20XX,
+    // rather than a short-form/ISO version of the date which may be read out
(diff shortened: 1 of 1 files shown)


plugins/field-date/src/field_date.ts
```diff
@@ -93,6 +93,32 @@ export class FieldDate extends Blockly.FieldTextInput {
...
+   */
+  override getAriaValue() {
+    const stringValue = this.getValue();
+    if (!stringValue) return super.getAriaValue();
+
+    const date = new Date(stringValue);
+    // Use a localized long-form description of the date, e.g. January XX, 20XX,
+    // rather than a short-form/ISO version of the date which may be read out
+    // with slashes or the like.
+    return date.toLocaleDateString(undefined, {
```
maribethb, line 114: I'm mildly concerned that this gets weird with time zones and the locale? is there any risk it reads out a different date due to specifying UTC? Or no because there's not a time picker in the original date anyway

maribethb (approved):
approved with behavioral question

plugins/field-date/src/field_date.ts
gonfunko: That was the issue that this resolved; `Date` and the internal storage uses UTC AFAICT, and specifying the timezone in the options informs it of that (it otherwise treats it as the local timezone). Likewise, specifying undefined as the first arg makes it use the local timezone/locale for formatting, so this is basically "take this date which is in UTC and output it nicely in the current locale"

PR #2713, opened 2026-06-16
fix: Improve accessibility of toolbox search

### LINKED ISSUE #2613: make toolbox search compatible with keyboard navigation
### Check for duplicates

### Component

toolbox-search

### Problem

_No response_

### Request

Toolbox search does not work with keyboard navigation currently.

When you use the arrow keys to navigate to the search category, the input box has focus. There is no way to use keyboard navigation to then move to the flyout that has the search results. If you press escape, currently the search is cleared. There is also no way to use the arrow keys to move to a different toolbox category once the search has focus.

To solve this, we'll likely need a 2 step focus system, where you can be focused on the toolbox category separately from the search input bar, so that you can navigate separately from the text input. This would be similar to how we have things set up for comments.

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

## The details
Fixes #2613

### Proposed Changes

This PR fixes and improves a number of keyboard and screenreader issues with the toolbox search plugin:

* Resolves some situations where the focus manager could get into a bad state
* Clarifies the placeholder text so the purpose of the field is more evident to screenreader users
* Improves keyboard navigation: left and right arrows can navigate character by character, moving focus to the matching blocks once the end of the search string is reached; up and down move to the beginning/end of the search string and then to the previous/next toolbox category
* Fixes a bug that prevented clicking the clear button in the field in Safari
* Announces the number of matching blocks to screenreaders so that users know when/if they should navigate to the results list
* Removes some now-dead code

### DIFF
--- a/plugins/toolbox-search/src/toolbox_search.ts
+++ b/plugins/toolbox-search/src/toolbox_search.ts
@@ -53,14 +53,40 @@ export class ToolboxSearchCategory extends Blockly.ToolboxCategory {
     this.searchField = document.createElement('input');
     this.searchField.id = this.SEARCH_INPUT_ID;
     this.searchField.type = 'search';
-    this.searchField.placeholder = 'Search';
+    this.searchField.placeholder = 'Search for blocks';
     this.workspace_.RTL
       ? (this.searchField.style.marginRight = '8px')
       : (this.searchField.style.marginLeft = '8px');
-    this.searchField.addEventListener('keyup', (event) => {
-      if (event.key === 'Escape') {
-        this.parentToolbox_.clearSelection();
+    this.searchField.addEventListener('keydown', (event) => {
+      if (event.key === 'ArrowUp' && this.searchField?.selectionStart === 0) {
+        const previous = this.parentToolbox_.getNavigator().getPreviousNode();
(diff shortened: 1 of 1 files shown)


plugins/toolbox-search/src/toolbox_search.ts
```diff
@@ -53,14 +53,40 @@ export class ToolboxSearchCategory extends Blockly.ToolboxCategory {
     this.searchField = document.createElement('input');
     this.searchField.id = this.SEARCH_INPUT_ID;
     this.searchField.type = 'search';
-    this.searchField.placeholder = 'Search';
+    this.searchField.placeholder = 'Search for blocks';
```
mikeharv, line 56: We don't use translations for this?

mikeharv (approved):

plugins/toolbox-search/src/toolbox_search.ts
gonfunko: We ideally would, but we've already cut v13, and it wasn't previously localized. Once we convert to a monorepo this'll be much easier, and I think we can/should do a localization pass across all the plugins at that point.

PR #2714, opened 2026-06-17
fix: Fix block-test under v13

gonfunko:
## The basics

### Proposed Changes

This PR fixes the behavior of some test blocks under Blockly v13. Specifically, it:

* Makes the duplicate-on-drag compatible with the updated APIs and more robust in general
* Uses non-deprecated variable-related methods
* Fixes an existing bug where the block that was supposed to be limited to 3 instances was not

### DIFF
--- a/plugins/block-test/src/drag.js
+++ b/plugins/block-test/src/drag.js
@@ -94,7 +94,10 @@ class DragToDupe extends Blockly.dragging.BlockDragStrategy {
    */
   revertDrag() {
     super.revertDrag();
-    if (this.draggingBlock.type === 'drag_to_dupe') {
+    if (
+      'isCloningAllowed' in this.draggingBlock &&
+      typeof this.draggingBlock.isCloningAllowed === 'function'
+    ) {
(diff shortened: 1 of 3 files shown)


plugins/block-test/src/drag.js
```diff
@@ -18,36 +18,99 @@ Blockly.Blocks['drag_to_dupe'] = {
...
   }
 
-  endDrag(e) {
-    this.baseStrat.endDrag(e);
+  /**
+   * Handles a drag being reverted/cancelled.
+   */
+  revertDrag() {
+    super.revertDrag();
+    if (this.draggingBlock.type === 'drag_to_dupe') {
```
maribethb: this strikes me as odd, do you really want to hardcode in the block type here? or should it be if cloning was allowed?

maribethb (approved):

plugins/block-test/src/drag.js
gonfunko: Switched to checking if the block is cloneable.

PR #2724, opened 2026-06-23
fix: Improve accessibility of grid and colour fields

gonfunko:
## The basics

### Proposed Changes
This PR improves the accessibility of the grid and colour fields. Specifically, it:

* Restructures the markup for grid cells such that the button is inside of a div with `role="gridcell"`, rather than itself being the grid cell. This improved readout in several ways:
  * VoiceOver no longer describes the focused cell as being "blank" if it has no textual content (i.e. color swatches)
  * The full textual content of the current row is not read out, in favor of a row/column count
  * Each focused item is described as a button
  * In Safari, items are consistently read out as they are navigated betweem
* `FieldGridDropdown` adds ARIA labels to its items when the `MenuGenerator` it is initialized with includes them
* Describes `FieldGridDropdown` as having a grid popup, rather than a listbox popup
* Gives names to all of the default colors in `FieldColour`, which are passed through as accessibility labels
* Denotes `FieldColour`'s type as "colour" when the field is being read out
* Updates the ARIA label for `FieldColour` in response to initialization/changing the selected color value
* Uses the Blockly keyboard focus highlight ring to indicate the focused color swatch in the grid

### DIFF
--- a/plugins/field-colour/src/field_colour.ts
+++ b/plugins/field-colour/src/field_colour.ts
@@ -20,7 +20,7 @@ import {
  * An array of colour strings for the palette.
  * Copied from goog.ui.ColorPicker.SIMPLE_GRID_COLORS
  */
-const DEFAULT_COLOURS = {
+const DEFAULT_COLOURS: Record<string, string> = {
   // greys
   '#ffffff': 'White',
   '#cccccc': 'Light Grey',
@@ -116,6 +116,8 @@ export class FieldColour extends FieldGridDropdown {
   // eslint-disable-next-line @typescript-eslint/naming-convention
   protected override isDirty_ = false;
 
+  protected override ariaTypeName = Blockly.Msg['ARIA_TYPE_FIELD_COLOUR'];
+
   /**
    * @param value The initial value of the field.  Should be in '#rrggbb'
    *     format.  Defaults to the first value in the default colour array.  Also
@@ -137,9 +139,7 @@ export class FieldColour extends FieldGridDropdown {
   ) {
     const swatches = makeSwatches(
       config?.colourOptions ?? Object.keys(DEFAULT_COLOURS),
-      config?.colourTitles?.length
-        ? config.colourTitles
-        : Object.values(DEFAULT_COLOURS),
+      config?.colourTitles,
     );
     super(swatches, validator, {...config, columns: config?.columns ?? 7});
 
@@ -195,14 +195,6 @@ export class FieldColour extends FieldGridDropdown {
     }
 
     this.recomputeAriaContext();
-  }
-
-  /**
-   * Returns a description of the type of this field for use in screenreader
-   * labels.
-   */
-  override getAriaTypeName(): string | null {
-    return Blockly.Msg['ARIA_TYPE_FIELD_COLOUR'];
   }
 
   /**
@@ -429,8 +421,7 @@ function makeSwatches(
     if (titles && index < titles.length) {
       swatch.title = titles[index];
     }
-
-    return [swatch, colour, titles?.[index]];
+    return [swatch, colour, titles?.[index] ?? DEFAULT_COLOURS[colour]];
   });
 }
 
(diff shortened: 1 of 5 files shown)


plugins/field-colour/src/field_colour.ts
```diff
@@ -179,10 +186,23 @@ export class FieldColour extends FieldGridDropdown {
...
+
+    this.recomputeAriaContext();
+  }
+
+  /**
+   * Returns a description of the type of this field for use in screenreader
+   * labels.
+   */
+  override getAriaTypeName(): string | null {
+    return Blockly.Msg['ARIA_TYPE_FIELD_COLOUR'];
```
mikeharv: Typically, this method will prioritize `this.ariaTypeName` if one is present, so that a particular block's field instance can specify its own name with more context (possible examples, "background color", "LED color").

plugins/field-colour/src/field_colour.ts
```diff
@@ -133,7 +135,12 @@ export class FieldColour extends FieldGridDropdown {
     validator?: FieldColourValidator,
     config?: FieldColourConfig,
   ) {
-    const swatches = makeSwatches(config?.colourOptions ?? DEFAULT_COLOURS);
+    const swatches = makeSwatches(
+      config?.colourOptions ?? Object.keys(DEFAULT_COLOURS),
+      config?.colourTitles?.length
+        ? config.colourTitles
+        : Object.values(DEFAULT_COLOURS),
```
mikeharv: I think the fallback logic here only works if there is no `colourOptions`.
It feels pretty risky to pass `Object.values(DEFAULT_COLOURS)` here, because there's no guarantee that they're going to align 1:1 with `config.colourOptions` if it exists. I think it would be safer to just pass `config.colourTitles`, (even if it's `undefined`) and attempt to look up missing titles inside `makeSwatches`.

plugins/field-colour/src/field_colour.ts
```diff
@@ -410,7 +430,7 @@ function makeSwatches(
       swatch.title = titles[index];
     }
 
-    return [swatch, colour];
+    return [swatch, colour, titles?.[index]];
```
mikeharv: To handle the case where don't have `titles` from the config, I think we could do:

`const title = titles?.[index] ?? DEFAULT_COLOURS[colour]`

For custom palettes, this would support adding labels for the colors that we do know, without mangling ones that we don't.

mikeharv (commented):
I think we'll want to check this with blocks that provide their own ARIA type name for the field, as well as for fields that supply color options but without titles. If not with unit tests, then perhaps manually.

plugins/field-colour/src/field_colour.ts
gonfunko: Switched to just overriding that attribute rather than the method

plugins/field-colour/src/field_colour.ts
gonfunko: Good call, I wasn't aiming at all to backfill color names if unspecified, just to provide them if we were using the defaults. But your suggestion below about resolving this in `makeSwatches` is better, so moved to that.

plugins/field-colour/src/field_colour.ts
gonfunko: Done

plugins/field-colour/src/field_colour.ts
```diff
@@ -116,6 +116,8 @@ export class FieldColour extends FieldGridDropdown {
   // eslint-disable-next-line @typescript-eslint/naming-convention
   protected override isDirty_ = false;
 
+  protected override ariaTypeName = Blockly.Msg['ARIA_TYPE_FIELD_COLOUR'];
```
mikeharv, line 119: Mostly checking my own understanding, but I think this works because it can still be overridden by the base `Field.configure_` method per instance.

mikeharv (approved):

plugins/field-colour/src/field_colour.ts
gonfunko: Yeah, that should work fine.

PR #2727, opened 2026-06-24
fix: Improve keyboard accessibility of slider field

gonfunko:
## The basics

### Proposed Changes
This PR improves the keyboard accessibility of the slider field. Specifically, it:

* Waits for the dropdown to animate in before focusing the slider, which fixes screenreader readouts
* Adds a keyboard listener that commits the value on enter and reverts it on Escape

### DIFF
--- a/plugins/field-slider/src/field_slider.ts
+++ b/plugins/field-slider/src/field_slider.ts
@@ -124,9 +124,13 @@ export class FieldSlider extends Blockly.FieldNumber {
 
     // Focus on the slider field, unless quietInput is passed.
     if (!quietInput) {
-      (editor.firstChild as HTMLInputElement).focus({
-        preventScroll: true,
-      });
+      // Wait for the dropdown to animate in before attempting to focus the
+      // slider.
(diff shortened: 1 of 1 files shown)


plugins/field-slider/src/field_slider.ts
```diff
@@ -153,7 +157,27 @@ export class FieldSlider extends Blockly.FieldNumber {
...
+      } else if (e.key === 'Escape') {
+        this.isBeingEdited_ = false;
+        this.setValue(
+          sliderInput.getAttribute('data-untyped-default-value'),
+          false,
+        );
+        handled = true;
+      }
+
+      if (handled) {
```
maribethb, line 175: Does this handle firing the final block change event too? I expected there to be more to do about setting the final value

gonfunko: Yep, the superclass' `widgetDispose_` handler takes care of that when it's automatically invoked by the `WidgetDiv` being dismissed here.

maribethb (approved):

PR #2728, opened 2026-06-24
fix: Make zoom to fit plugin keyboard and screenreader accessible

### LINKED ISSUE #2712: Zoom to fit plugin is not keyboard navigable
The zoom to fit button can't be focused with the keyboard. It should also have appropriate ARIA labels and roles added.

gonfunko:
## The basics

### Proposed Changes

This PR fixes #2712 by making the zoom to fit control keyboard navigable and screenreader accessible following a similar pattern to the built-in zoom controls. This depends on a new release of core that adds `Capability.FOCUSABLE`. With these changes, the control can be tabbed to, enter/space will activate it, it has a screenreader label and role of button, and focusing it will draw a highlight ring and darken the icon in the same manner as hovering over it.

### DIFF
--- a/plugins/zoom-to-fit/src/index.ts
+++ b/plugins/zoom-to-fit/src/index.ts
@@ -112,7 +112,7 @@ export class ZoomToFitControl
     Blockly.utils.aria.setState(
       this.svgGroup,
       Blockly.utils.aria.State.LABEL,
-      'Zoom to fit',
+      Blockly.Msg['ZOOM_TO_FIT_ARIA_LABEL'],
     );
     Blockly.utils.aria.setRole(this.svgGroup, Blockly.utils.aria.Role.BUTTON);
 
--- a/plugins/zoom-to-fit/src/index.ts
+++ b/plugins/zoom-to-fit/src/index.ts
@@ -9,7 +9,9 @@ import * as Blockly from 'blockly/core';
 /**
  * Class for zoom to fit control.
  */
-export class ZoomToFitControl implements Blockly.IPositionable {
+export class ZoomToFitControl
+  implements Blockly.IComponent, Blockly.IPositionable, Blockly.IFocusableNode
+{
   /**
    * The unique id for this component.
    */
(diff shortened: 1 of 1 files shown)


plugins/zoom-to-fit/src/index.ts
```diff
@@ -97,15 +102,46 @@ export class ZoomToFitControl implements Blockly.IPositionable {
...
+      height: `${this.height}px`,
+      width: `${this.width}px`,
+      class: 'zoomToFit',
+      id: Blockly.utils.idGenerator.getNextUniqueId(),
+      tabindex: '0',
+    });
+    Blockly.utils.aria.setState(
+      this.svgGroup,
+      Blockly.utils.aria.State.LABEL,
+      'Zoom to fit',
```
mikeharv: Translation?

plugins/zoom-to-fit/src/index.ts
```diff
@@ -97,15 +102,46 @@ export class ZoomToFitControl implements Blockly.IPositionable {
...
+        y: -4,
+        rx: 2,
+        ry: 2,
+        fill: 'none',
+        class: 'blocklyFocusRing',
+      },
+      this.svgGroup,
+    );
+
+    const image = Blockly.utils.dom.createSvgElement(
```
mikeharv, line 134: Do we need to hide this element from the accessibility tree?

mikeharv (approved):
Looks good. My questions do not block approval.

plugins/zoom-to-fit/src/index.ts
gonfunko: Experimentally it seemed fine, the screenreader wasn't attempting to read it

plugins/zoom-to-fit/src/index.ts
gonfunko: Yes, but I think we're better doing a pass across all of samples since the translation situation overall is pretty poor. This will become much less painful once it's a monorepo as well.

gonfunko: Updated to a localized string since we snuck some into 13.1.

PR #2729, opened 2026-06-24
feat!: make field-multilineinput keyboard navigable and accessible

maribethb:
## The basics

### Resolves

Fixes field-multilineinput not keyboard accessible

### Proposed Changes

- Makes `Enter` commit edits, like all other input fields (this is a breaking behavioral change).
- Adds a `Shift+Enter` handler that adds newlines.
- Adds a hint UI that explains these options, with translatable hints + fallbacks because these aren't in core yet.
- Adds 2 new APIs that allow you to control this behavior. if you want enter to add a newline like it used to, you can do that, in which case shift+enter will commit the change.
- Updates the readme with this information.
- Adds the missing keyboard focus ring indicator.
- Adds 2 options in the test playground page UI that control the 2 settings.
- Edits the eslint config to allow UPPER_CASE for static readonly class members, and removes some eslint silencing on existing ones

### Reason for Changes

This plugin couldn't be used with only the keyboard as there was no way to commit a change.

### Test Coverage

Tested manually and added additional unit tests.

### Documentation

updated the readme already

### Additional Information

Screenshot 2026-06-24 at 1 10 39 PM

### DIFF
--- a/plugins/field-multilineinput/src/field_multilineinput.ts
+++ b/plugins/field-multilineinput/src/field_multilineinput.ts
@@ -18,7 +18,7 @@ export class FieldMultilineInput extends Blockly.FieldTextInput {
    * Minimum editor width (in SVG/workspace units) when the field is open.
    * Prevents narrow editors when the initial text is very short.
    */
-   
+
   static readonly EDITOR_MIN_WIDTH = 150;
 
   /**
--- a/plugins/field-multilineinput/src/field_multilineinput.ts
+++ b/plugins/field-multilineinput/src/field_multilineinput.ts
@@ -466,15 +466,7 @@ export class FieldMultilineInput extends Blockly.FieldTextInput {
     const paddingBottomPx = FieldMultilineInput.showHint
       ? paddingY + hintHeightPx
       : paddingY;
-    htmlInput.style.padding =
-      paddingY +
-      'px ' +
-      paddingX +
-      'px ' +
-      paddingBottomPx +
-      'px ' +
-      paddingX +
-      'px';
+    htmlInput.style.padding = `${paddingY}px ${paddingX}px ${paddingBottomPx}px ${paddingX}px`;
     htmlInput.style.lineHeight = lineHeight * scale + 'px';
     div.appendChild(htmlInput);
 
@@ -625,8 +617,8 @@ export class FieldMultilineInput extends Blockly.FieldTextInput {
   protected override onHtmlInputKeyDown_(e: KeyboardEvent) {
     if (e.key === 'Enter') {
       if (e.isComposing) {
-        // Let the IME finalize its composition naturally; don't commit or
-        // insert a newline ourselves.
+        // Let the IME (input method editor) finalize its composition naturally;
+        // don't commit or insert a newline ourselves.
         return;
       }
       const shiftPressed = e.shiftKey;
--- a/plugins/field-multilineinput/src/field_multilineinput.ts
+++ b/plugins/field-multilineinput/src/field_multilineinput.ts
@@ -166,6 +166,9 @@ export class FieldMultilineInput extends Blockly.FieldTextInput {
       },
       this.fieldGroup_,
     );
+    if (this.fieldGroup_) {
+      Blockly.utils.dom.addClass(this.fieldGroup_, 'blocklyField');
+    }
   }
 
   /**
--- a/plugins/field-multilineinput/src/field_multilineinput.ts
+++ b/plugins/field-multilineinput/src/field_multilineinput.ts
@@ -15,6 +15,12 @@ import * as Blockly from 'blockly/core';
  */
 export class FieldMultilineInput extends Blockly.FieldTextInput {
   /**
+   * Minimum editor width (in SVG/workspace units) when the field is open.
+   * Prevents narrow editors when the initial text is very short.
+   */
+  static readonly EDITOR_MIN_WIDTH = 150;
+
+  /**
    * The SVG group element that will contain a text element for each text row
    *     when initialized.
    */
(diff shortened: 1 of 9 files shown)


plugins/field-multilineinput/src/field_multilineinput.ts
```diff
@@ -401,12 +457,24 @@ export class FieldMultilineInput extends Blockly.FieldTextInput {
...
+      : paddingY;
+    htmlInput.style.padding =
+      paddingY +
+      'px ' +
+      paddingX +
+      'px ' +
+      paddingBottomPx +
+      'px ' +
+      paddingX +
+      'px';
```
gonfunko: A backtick string would probably be a bit more readable here

plugins/field-multilineinput/src/field_multilineinput.ts
```diff
@@ -421,9 +489,102 @@ export class FieldMultilineInput extends Blockly.FieldTextInput {
 
     this.bindInputEvents_(htmlInput);
 
+    if (FieldMultilineInput.showHint) {
+      this.hintElement = this.createHint(hintHeightPx);
+      this.hintElement.style.fontFamily = constants.FIELD_TEXT_FONTFAMILY;
+      div.appendChild(this.hintElement);
```
gonfunko, line 488: This seems like it should be done before measuing the hintElement above since presumably the font family could change its metrics? Also does it need to be created every time this is called, or could it just be created if it's currently undefined?

plugins/field-multilineinput/src/field_multilineinput.ts
```diff
@@ -421,9 +489,102 @@ export class FieldMultilineInput extends Blockly.FieldTextInput {
...
+  }
+
+  /**
+   * Creates a keycap element displaying a single key label.
+   *
+   * @param label The key symbol to display.
+   * @returns The keycap element.
+   */
+  private createHintKeycap(label: string): HTMLSpanElement {
+    const key = document.createElement('span');
```
gonfunko, line 557: Could use the [`<kbd>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/kbd) to be fancy

plugins/field-multilineinput/src/field_multilineinput.ts
```diff
@@ -451,18 +612,59 @@ export class FieldMultilineInput extends Blockly.FieldTextInput {
...
+   * All other keys are handled by the parent class (Escape reverts, Tab navigates).
    *
    * @param e Keyboard event.
    */
   // eslint-disable-next-line @typescript-eslint/naming-convention
   protected override onHtmlInputKeyDown_(e: KeyboardEvent) {
-    if (e.key !== 'Enter') {
+    if (e.key === 'Enter') {
+      if (e.isComposing) {
+        // Let the IME finalize its composition naturally; don't commit or
```
gonfunko: What is an IME?

plugins/field-multilineinput/test/field_multilineinput_test.mocha.js
```diff
@@ -119,6 +120,152 @@ suite('FieldMultilineInput', function () {
...
 
+  suite('Keyboard behavior', function () {
+    setup(function () {
+      this.jsdomCleanup = require('jsdom-global')('');
+      this.field = new FieldMultilineInput('hello');
+      this.textarea = document.createElement('textarea');
+      this.textarea.value = 'hello';
+      this.textarea.selectionStart = 5;
+      this.textarea.selectionEnd = 5;
+      this.field.htmlInput_ = this.textarea;
```
gonfunko: Rather than making a textarea and patching it into the field, why not retrieve the field's actual input and manipulate it? Would be nice if this wasn't directly accessing a protected field too

plugins/field-multilineinput/test/field_multilineinput_test.mocha.js
```diff
@@ -119,6 +120,152 @@ suite('FieldMultilineInput', function () {
...
+      this.jsdomCleanup();
+    });
+
+    test('Enter calls super to commit the value', function () {
+      const event = new KeyboardEvent('keydown', {
+        key: 'Enter',
+        ctrlKey: false,
+        metaKey: false,
+      });
+      this.field.onHtmlInputKeyDown_(event);
```
gonfunko: Can the event be dispatched rather than directly calling the handler and passing it in?

plugins/field-multilineinput/test/field_multilineinput_test.mocha.js
```diff
@@ -119,6 +120,152 @@ suite('FieldMultilineInput', function () {
...
+      this.field['insertNewline']();
+      assert.equal(this.textarea.value, 'hello\n world');
+      assert.equal(this.textarea.selectionStart, 6);
+    });
+
+    test('insertNewline replaces a selection with a newline', function () {
+      this.textarea.value = 'hello world';
+      this.textarea.selectionStart = 5;
+      this.textarea.selectionEnd = 11;
+      this.field['insertNewline']();
```
gonfunko: Is this tortured syntax to get around it being private?

plugins/field-multilineinput/test/field_multilineinput_test.mocha.js
```diff
@@ -119,6 +120,152 @@ suite('FieldMultilineInput', function () {
...
+      });
+
+      test('defaults to true', function () {
+        assert.isTrue(FieldMultilineInput.enterCommits);
+      });
+
+      test('can be set to false globally', function () {
+        FieldMultilineInput.enterCommits = false;
+        assert.isFalse(FieldMultilineInput.enterCommits);
+      });
```
gonfunko: This kind of feels like it's just testing that Javascript exists?

plugins/field-multilineinput/src/field_multilineinput.ts
maribethb: We aren't measuring the hint element in this function, the computation above that uses the hint height is using the lineheight * scale.

maribethb: and i could be wrong, but if widgetCreate is being called in the first place, it means the dom doesn't exist. i think the editor is fully torn down every time the editor is closed and opened.

gonfunko: Oh sorry, github's terrible diff collapsing across method boundaries confused me. Yes this seems fine, I though this was being done in `updateSize`

plugins/field-multilineinput/src/field_multilineinput.ts
maribethb: it's input method editor, basically a way to type foreign characters into text inputs. best practice is not to take action on keydown events during composition in case they're using the key you're acting on as part of a multi-key input. should i explain this here? i also wasn't super familiar with the idea it's not really unique to this handler. idk

gonfunko: Yeah I think just spelling it out would be helpful vs the acronym.

plugins/field-multilineinput/test/field_multilineinput_test.mocha.js
maribethb: lol yeah which tbh i prefer over the `TEST_ONLY` nonsense we sometimes do

plugins/field-multilineinput/test/field_multilineinput_test.mocha.js
maribethb: fair, remvoed these two suites entirely

gonfunko (approved):

PR #2730, opened 2026-06-25
feat: fix keyboard nav for minimap

maribethb:
## The basics

### Resolves

Fixes buggy minimap keyboard nav where you could keyboard nav into that tiny workspace

### Proposed Changes

- Prevents navigating into the mini workspace
- Adds the ability to use arrow keys to pan the workspace while you're focused on the minimap
- Adds appropriate aria label and role
- Adds tests for new functionality

### Reason for Changes

accessibility!

### Test Coverage

added and manually tested

### Documentation

updated readme

### DIFF
--- a/plugins/workspace-minimap/src/minimap.ts
+++ b/plugins/workspace-minimap/src/minimap.ts
@@ -142,13 +142,13 @@ export class Minimap {
       this.minimapWrapper,
       'focus',
       this,
-      this.onWrapperFocus,
+      this.onMinimapFocus,
     );
     this.onBlurWrapper = Blockly.browserEvents.bind(
       this.minimapWrapper,
       'blur',
       this,
-      this.onWrapperBlur,
+      this.onMinimapBlur,
     );
 
     this.minimapWorkspace.scrollbar?.setContainerVisible(false);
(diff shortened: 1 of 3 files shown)


plugins/workspace-minimap/src/minimap.ts
```diff
@@ -277,6 +351,80 @@ export class Minimap {
     this.primaryScroll(event);
   }
 
+  private onWrapperFocus(): void {
```
lizschwab: Shouldn't `onFocusWrapper` and `onBlurWrapper` be wrappers for onFocus and onBlur, rather than for `onWrapperFocus` and `onWrapperBlur`? These names imply we're focusing/blurring a wrapper, not wrapping a function.

Either way, I'm not going to gate check-in on naming, just found it a bit confusing.

lizschwab (approved):

plugins/workspace-minimap/src/minimap.ts
maribethb: haha it's actually both. the container we're putting this event on is `minimapWrapper`. the actual focus handler is called `onWrapperFocus` which is accurate (it's what happens when you focus the wrapper). the thing we pass to `Events.bind` is called `onFocusWrapper` which matches what the other handlers are called, because they are a wrapper around the `onFocus` event.

i'm going to rename `onWrapperFocus` to `onMinimapFocus` and leave the other one alone to match the existing pattern

maribethb: (the reason I didn't just do `onFocus` and `onBlur` is because those are names in the `FocusableTree` interface and i wanted to not imply that that's how those events worked)

PR #2731, opened 2026-06-25
fix: Improve the keyboard and screenreader accessibility of the backpack

### LINKED ISSUE #2711: Backpack has limited keyboard accessibility
The backpack plugin suffers from several issues when using keyboard navigation:

* It can't be tabbed to and doesn't have a hotkey to focus/open it
* Wide blocks can't be added to it, because the top left corner needs to be positioned over the backpack to drop it, and keyboard move mode scrolls the workspace to keep a moving block fully in frame, so wide blocks can't have their corner placed atop the backpack
* Escape doesn't dismiss the backpack flyout when focused
* Because the backpack can't be focused, its contextual menu can't be displayed with the keyboard
* ARIA labels/roles should be added as appropriate

Navigating the flyout and inserting blocks from it does work as expected once it's manually focused, although blocks are inserted at the far left next to the normal flyout rather than adjacent to the backpack flyout.

gonfunko:
## The basics

### Resolves

Fixes #2711

### Proposed Changes
This PR improves the keyboard navigability and screenreader accessibility of the backpack plugin. Now,

* The backpack can be tabbed to
* The contextual menu can be shown via the keyboard
* The backpack has a button role and appropriate ARIA label depending on its current state
* When focused, the backpack has a focus ring and darkens as it does on hover
* When opened, the first block in the flyout is automatically focused


plugins/workspace-backpack/src/backpack.ts
```diff
@@ -219,9 +226,33 @@ export class Backpack
...
+    Blockly.utils.dom.createSvgElement(
+      Blockly.utils.Svg.RECT,
+      {
+        width: this.WIDTH_ + 8,
+        height: this.HEIGHT_ + 8,
+        class: 'blocklyFocusRing',
+        x: -4,
+        y: -4,
+        rx: 2,
+        ry: 2,
```
mikeharv: Worth explaining magic numbers (or using constants)?

mikeharv (approved):

plugins/workspace-backpack/src/backpack.ts
gonfunko: Done

PR #2735, opened 2026-06-29
chore(deps): Update to Blockly v13.1.0

gonfunko:
## The basics

### Proposed Changes
This PR updates the v13 branch to use Blockly v13.1.0 and fixes a couple of test failures.

### DIFF
--- a/plugins/workspace-search/test/workspace_search_test.mocha.js
+++ b/plugins/workspace-search/test/workspace_search_test.mocha.js
@@ -72,6 +72,7 @@ suite('WorkspaceSearch', function () {
     this.workspaceSearch = new WorkspaceSearch(this.workspace);
     // See https://github.com/RaspberryPiFoundation/blockly-samples/issues/2528 for context.
     global.SVGElement = window.SVGElement;
+    global.requestAnimationFrame = (callback) => setTimeout(callback, 0);
   });
 
   teardown(function () {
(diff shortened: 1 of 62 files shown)


plugins/workspace-search/test/workspace_search_test.mocha.js
```diff
@@ -492,7 +493,13 @@ suite('WorkspaceSearch', function () {
       this.workspaceSearch.searchAndHighlight('nothingMatchesThis', false);
       this.workspaceSearch.close();
 
-      assert.equal(Blockly.getFocusManager().getFocusedNode(), this.workspace);
+      // As of Blockly v13.1.0 the workspace is focused via a dedicated focus
+      // target node (rather than the workspace itself being the focused node),
+      // so focusing the workspace as a whole lands on that target.
+      assert.equal(
+        Blockly.getFocusManager().getFocusedNode(),
+        this.workspace.getWorkspaceFocusTarget(),
```
maribethb: Does `getFocusableNode` on the workspace return this element correctly? If so then using that would sort of future-proof this test since we don't care what the specific element that's focused, just that it's the one that represents the workspace being focused

maribethb (approved):

plugins/workspace-search/test/workspace_search_test.mocha.js
gonfunko: No, the workspace focus target is a separate focusable node and DOM element from the workspace, and the workspace's focusable element is still its root SVG group. The focus target gets used here because it's returned from `WorkspaceSvg.getRestoredFocusableNode()`; I updated the test to use that rather than hardcoding the focus target.

PR #2739, opened 2026-06-30
chore: update developer-tools for ts 6 and webpack 7

maribethb:
## The basics

### Resolves

Fixes developer-tools not building

### Proposed Changes

- sets strict mode to off because hella errors with strict mode on
- sets node env via environment variable instead of flag
- fixes type error with v13 locale
- adds declaration for css file that ts wants now

### Reason for Changes

samples was updated to typescript version 6, which among other things set strict mode to true by default. there are lots of errors with strict mode so turning it off for now.

also was updated to webpack 7 which removed support for the `--node-environment` flag we were using.

### Test Coverage

developer-tools builds

### Additional Information

I think the issue with `setLocale` needing the `any` cast is a legit problem in v13, will investigate and fix separately.

### DIFF
--- a/examples/developer-tools/package.json
+++ b/examples/developer-tools/package.json
@@ -6,9 +6,9 @@
   "main": "index.js",
   "scripts": {
     "start": "webpack serve --open --mode=development",
-    "build": "webpack --mode=production --node-env=production",
+    "build": "NODE_ENV=production webpack --mode=production",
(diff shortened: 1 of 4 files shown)


gonfunko (approved):

maribethb:
test failures unrelated and fixed by https://github.com/RaspberryPiFoundation/blockly-samples/pull/2744

PR #2751, opened 2026-07-16
fix(block-dynamic-connection): recache connection pairs after dynamically adding inputs

### LINKED ISSUE #2708: Connections added by block-dynamic-connection are not available in keyboard move mode
When using block-dynamic-connection with Blockly v13, and moving blocks via the keyboard, new connections are correctly added to blocks, but because move mode precaches the list of available connections at the start of a keyboard-driven move, they are not offered as connection points in constrained move mode.

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly-samples/issues/2708
Depends upon:
- https://github.com/RaspberryPiFoundation/blockly/pull/10149
### Proposed Changes

After `onPendingConnection()`, if the dynamic block has new inputs, re-cache the compatible connection pairs for the currently dragging block. 

### Reason for Changes

As stated in the issue:
> When using `block-dynamic-connection` with Blockly v13, and moving blocks via the keyboard, new connections are correctly added to blocks, but because move mode pre-caches the list of available connections at the start of a keyboard-driven move, they are not offered as connection points in constrained move mode.

This fixes this issue.

### Test Coverage

No plugin-level tests were added for this change. A headless test would only verify mocked interactions and would not meaningfully exercise constrained movement. Testing the complete behavior with rendered blocks would require substantial browser, rendering, and drag-state setup disproportionate to the small plugin change.

Instead, Blockly core tests cover the important behavioral contract: after `cacheAllConnectionPairs()` is called during a move, newly added connections are included in the cache and can be visited in the expected traversal order.

### Additional Information

This depends upon `BlockDragStrategy.cacheAllConnectionPairs` being made public, so this should only be merged after https://github.com/RaspberryPiFoundation/blockly/pull/10149 is merged and, ideally, released.

### DIFF
--- a/plugins/block-dynamic-connection/src/connection_previewer.ts
+++ b/plugins/block-dynamic-connection/src/connection_previewer.ts
@@ -10,10 +10,10 @@ interface ConnectionPreviewerConstructor {
   new (draggedBlock: Blockly.BlockSvg): Blockly.IConnectionPreviewer;
 }
 
-interface DynamicBlock extends Blockly.BlockSvg {
+type DynamicBlock = Blockly.BlockSvg & {
   onPendingConnection(connection: Blockly.RenderedConnection): void;
   finalizeConnections(): void;
-}
+};
 
 /**
  * A type guard that checks if the given block fulfills the DynamicBlock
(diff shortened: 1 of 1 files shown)


maribethb (approved):
clever way to make this backwards compatible. I think I'd either add a comment saying to simplify this once plugin uses v13.2, or just hold this PR until we release v13.2 and update the blockly deps at the same time (or first). there is less need to keep the backwards-compat hacks like these around longer-term since we're moving to fixed versioning.

PR #2754, opened 2026-07-23
feat(field-bitmap): make the bitmap editor keyboard and screen-reader accessible

### LINKED ISSUE #2752: FieldBitmap is not accessible
### Check for duplicates

### Component

field-bitmap

### Description

`FieldBitmap` is not keyboard navigable or screen reader compatible. `FieldGridDropdown` does have this capability so it should be possible to use the grid classes from it. Note that the former class cannot extend the latter as it is not meant to behave like a dropdown field with a single selectable value.

### Reproduction steps

1.
2.
3.

### Stack trace

```shell

```

### Screenshots

_No response_

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly-samples/issues/2752
### Proposed Changes

Make `@blockly/field-bitmap`’s dropdown editor keyboard- and screen-reader-accessible.

**Editor accessibility** (similar implementation to `FieldGridDropdown`)
- Builds an ARIA `grid` / `row` / `gridcell` structure with real `<button>` pixels.
- Keyboard: arrows (hard edges, RTL-aware), Home/End/PageUp/PageDown, Enter/Space to toggle pixel.
- Arrow keys at a hard edge play Blockly’s error beep.
- Field-level ARIA: `aria-haspopup="grid"`, `ariaTypeName` from `ARIA_TYPE_FIELD_BITMAP`, `getAriaValue()` summarizing size and filled-pixel count.

**Pointer painting**
- Uses `setPointerCapture` so drag can continue outside the editor; `pointerup` / `pointerleave` / `pointercancel` end the gesture.
- Pixel `click` is suppressed so it does not double-toggle after pointer paint-drag. 
- Focus: yellow Blockly focus ring; focus follows pointer on hover and while drag-painting; initial focus on first pixel when the editor opens.
- Not technically possible with keyboard since each pixel toggle requires a key press.

**Messages**
- New messages, coupled either with English hard-coded backup (for new strings).
- Keep reading legacy `BUTTON_LABEL_RANDOMIZE` / `BUTTON_LABEL_CLEAR` if present for backwards compatibility.

**DOM / CSS**
- Replace `.pixelContainer` / `.pixelRow` with `.bitmapPixelGrid` / `.bitmapPixelRow` (CSS grid + `display: contents` rows).
- Add `blocklyField` on the on-block field group so the cursor matches other fields (default arrow, not grabby hand).
- On-block focus indicator: create a transparent `.blocklyFieldRect` above the pixel grid so the yellow stroke is visible (SVG strokes are centered; pixels would otherwise cover the inner half).

**Related Blockly core (separate PR in core)**  
New/updated Msg strings used by this plugin, including `FIELD_BITMAP_PIXEL_LABEL = '%1, row %2, column %3';` (bit value spoken first) and `FIELD_BITMAP_ARIA_VALUE = '%1 by %2, %3 pixels on'` (dimensions are spoken "by" rather than "x"). 
- https://github.com/RaspberryPiFoundation/blockly/pull/10180

Block label:
image

Field label with focus indicator:
image

Initial cell label with focus indicator:
image
After pressing the right arrow:
image

### Reason for Changes

The bitmap editor was pointer-only and not exposed as an accessible control. Users who rely on keyboard or assistive tech could not navigate or toggle pixels reliably, and the field did not announce type/value in line with other Blockly fields.

The changes here are kept local to the plugin and no new peer dependencies are established. Before settling on this approach, I entertained the idea of refactoring `@blockly/field-grid-dropdown`’s `Grid`/`GridItem` so that `FieldBitmap` could use them too. This could have had some long-term maintenance benefits but sharing those components was resulting in a lot of API churn that didn't feel worthwhile (effectively two separate modes). Prior to this, I briefly considered making `FieldBitmap` extend `FieldGridDropdown` but this wouldn't work as it doesn't really behave like a dropdown (single-choice selection). 

### Test Coverage

Manually tested changes:
- Open/close editor (Esc, click-away); focus returns via `DropDownDiv` ephemeral focus.
- Keyboard: arrow navigation at edges (no wrap is intentional), Home/End, Enter/Space toggle.
- Pointer: single-pixel toggle, drag-paint / drag-clear, paint while dragging outside the grid.
- Hover / mid-drag focus ring tracking.
- Screen reader / a11y tree: pixel labels (state before coordinates), field summary ("5 by 5, 10 pixels on" when Msg/fallback loaded).
- Randomize / Clear still work; labels resolve with core messages when linked locally.
- On-block field cursor matches other fields (`blocklyField`).
- Tested changes across all renderers and in RTL. Note that https://github.com/RaspberryPiFoundation/blockly-samples/issues/2021 is still present.

### Documentation

- Note DOM/CSS class renames (`.pixelContainer` / `.pixelRow` -> `.bitmapPixelGrid` / `.bitmapPixelRow`) and `<button>` pixels for anyone styling the editor.

### DIFF
--- a/plugins/field-bitmap/src/field-bitmap.ts
+++ b/plugins/field-bitmap/src/field-bitmap.ts
@@ -338,6 +338,7 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
     this.bindEvent(dropdownEditor, 'pointerleave', this.onPointerEnd);
     this.bindEvent(dropdownEditor, 'pointerdown', this.onPointerStart);
     this.bindEvent(dropdownEditor, 'pointercancel', this.onPointerEnd);
+    this.bindEvent(dropdownEditor, 'keydown', this.onEditorKeyDown);
     // Stop the browser from handling touch events and cancelling the event.
     this.bindEvent(dropdownEditor, 'touchmove', (e: Event) => {
       e.preventDefault();
@@ -421,6 +422,26 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
       }
     }
     return grid;
+  }
+
+  /**
+   * Handles editor-level keyboard shortcuts.
+   * Ctrl/Cmd+Enter commits and closes; Escape reverts and closes.
+   *
+   * @param e The keydown event.
+   */
+  private onEditorKeyDown(e: KeyboardEvent) {
+    const isEscape = e.key === 'Escape';
+    const isCommit = e.key === 'Enter' && (e.ctrlKey || e.metaKey);
+    if (!isEscape && !isCommit) return;
+
+    if (isEscape && this.initialValue !== null) {
+      this.setValue(this.initialValue, false);
+    }
+    Blockly.DropDownDiv.hideIfOwner(this);
+    Blockly.getFocusManager().focusNode(this);
+    e.preventDefault();
+    e.stopPropagation();
   }
 
   /**
@@ -822,11 +843,10 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
    * Resets pointer state (e.g. After either a pointerup event or if the
    * gesture is canceled).
    *
-   * @param e The pointer event that ended the gesture, when available.
+   * @param e The pointer event that ended the gesture.
    */
-  private onPointerEnd(e?: PointerEvent) {
+  private onPointerEnd(e: PointerEvent) {
     if (
-      e &&
       e.currentTarget instanceof HTMLElement &&
       e.currentTarget.hasPointerCapture?.(e.pointerId)
     ) {
@@ -918,10 +938,10 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
    * @param eventName Name of the event to bind.
    * @param callback Function to be called on specified event.
    */
-  private bindEvent(
+  private bindEvent<E extends Event>(
     element: HTMLElement,
     eventName: string,
-    callback: (e: PointerEvent) => void,
+    callback: (e: E) => void,
   ) {
     this.boundEvents.push(
       Blockly.browserEvents.bind(element, eventName, this, callback),
--- a/plugins/field-bitmap/src/field-bitmap.ts
+++ b/plugins/field-bitmap/src/field-bitmap.ts
@@ -606,6 +606,11 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
    * Initializes the on-block display.
    */
   override initView() {
+    this.createBorderRect_();
+    // Invisible fill so the rect only provides keyboard-nav focus stroke.
+    this.getBorderRect().style.fill = 'none';
+    // Don't intercept clicks meant for the pixel rects beneath.
+    this.getBorderRect().style.pointerEvents = 'none';
     if (this.fieldGroup_) {
       Blockly.utils.dom.addClass(this.fieldGroup_, 'blocklyField');
     }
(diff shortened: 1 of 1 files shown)


plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -114,6 +117,20 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
     return this.imgHeight;
   }
 
+  /**
+   * Updates the ARIA roles and label for this field.
+   */
+  override recomputeAriaContext(): boolean {
```
mikeharv, line 123: Identical to https://github.com/RaspberryPiFoundation/blockly-samples/blob/bitmacc/plugins/field-grid-dropdown/src/index.ts#L133-L141

plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -301,66 +336,287 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
...
+    return grid;
+  }
+
+  /**
+   * Handles keyboard navigation and activation inside the pixel grid.
+   *
+   * @param e The keydown event.
+   * @param rtl Whether the workspace is RTL.
+   */
+  private onPixelGridKeyDown(e: KeyboardEvent, rtl: boolean) {
```
mikeharv, line 453: Similar to: https://github.com/RaspberryPiFoundation/blockly-samples/blob/bitmacc/plugins/field-grid-dropdown/src/grid.ts#L161-L205

plugins/field-bitmap/src/field-bitmap.ts
gonfunko: Probably good to add the expanded state change from #2757

plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -301,66 +336,287 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
...
+      }
+      case 'PageUp':
+      case 'Home':
+        this.focusPixelAt(0);
+        break;
+      case 'PageDown':
+      case 'End':
+        this.focusPixelAt(length - 1);
+        break;
+      case 'Enter':
```
gonfunko, line 496: It'd be nice for cmd/ctrl enter to commit the value, and escape to cancel/revert to the value before the editor was opened

plugins/field-bitmap/src/field-bitmap.ts
```diff
@@ -301,66 +336,287 @@ export class FieldBitmap extends Blockly.Field<number[][]> {
...
+   * @param c Column index.
+   * @param pixelValue Pixel value (0 or 1).
+   */
+  private updateEditorPixelDisplay(r: number, c: number, pixelValue: number) {
+    const button = this.editorPixels?.[r]?.[c];
+    if (!button) return;
+    button.style.background = pixelValue
+      ? this.pixelColours.filled
+      : this.pixelColours.empty;
+    button.setAttribute('aria-pressed', String(!!pixelValue));
```
gonfunko, line 628: Would be good to file a bug to add this to utils.aria.STATE, and update this line after that's done

gonfunko (approved):

plugins/field-bitmap/src/field-bitmap.ts
mikeharv: https://github.com/RaspberryPiFoundation/blockly/issues/10201

plugins/field-bitmap/src/field-bitmap.ts
mikeharv: https://github.com/RaspberryPiFoundation/blockly-samples/pull/2754/commits/5492dba33a97f9ebab77cf956e4e6fc57dc6018c

mikeharv: Reverting via https://github.com/RaspberryPiFoundation/blockly-samples/pull/2760

PR #2755, opened 2026-07-26
fix: Wrong code generation for FieldImage

### LINKED ISSUE #2691: block factory generates incorrect javascript definitions for FieldImage
### Check for duplicates

### Component

block factory

### Description

See https://github.com/RaspberryPiFoundation/blockly/pull/9942

the JS definition generated for blocks that contain a FieldImage is wrong. it has been wrong for a long time including in the legacy factory. the 4th argument is the config arg whereas the FieldImage constructor takes that as the seventh argument (and there's no need for it anyway as it could pass the alt text and flipRtl directly as the 4th and 5th arguments to the constructor)

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

TannerGabriel:
## The basics

### Resolves

Fixes #2691

### Other optional implementations

Alternatively, we could pass the `alt` and `fieldRtl` into the `config`, which is now the seventh, but I think this is the clearer approach.

```
new Blockly.FieldImage(src, width, height, undefined, undefined, undefined, {alt: 'star', flipRtl: true})
```

### DIFF
--- a/examples/developer-tools/src/output-generators/fields/image.ts
+++ b/examples/developer-tools/src/output-generators/fields/image.ts
@@ -48,7 +48,12 @@ javascriptDefinitionGenerator.forBlock['field_image'] = function (
   const alt = generator.quote_(block.getFieldValue('ALT'));
   const flipRtl = block.getFieldValue('FLIP_RTL') === 'TRUE';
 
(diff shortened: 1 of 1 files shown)


gonfunko (approved):

gonfunko:
Thanks for the fix for this!

PR #2757, opened 2026-07-27
feat(field-grid-dropdown): wrap vertical keyboard focus across columns

### LINKED ISSUE #2678: keyboard navigation for field-grid
### Check for duplicates

### Component

field-grid

### Description

Grids have been somewhat tricky for keyboard/screenreader navigation. If they are announced as a list, users tend to only go down the first column and not ever discover there are more columns

To help with this, the navigation should wrap between columns, but not back around to the beginning. So if you press "down" at the bottom of a column, you'll go to the top of the next column. But if you're at the very end of the grid, pressing down will *not* wrap you back to the beginning of the grid. This gives the best combination of hard stops and column discovery.

Also make sure there's an appropriate role on the field. It currently says the field will open a listbox which is not correct.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly-samples/issues/2678

### Proposed Changes

`field-grid-dropdown`:
Ensure aria role and field type name are accurate, including when expanding/collapsing. Use an existing translation for the type name.
Change type of `FieldGridDropdown.grid` to `Grid | null` to better align to `FieldDropdown.menu`.
Rework `Grid.moveFocus()` to support vertical wrapping:
- Pressing up or down now passes a `movementIndex` of 1/-1 (previously it was a number with a larger absolute value based on the number of columns) and `horizontal = false`.
- Helper method `getVerticalTargetIndex` which handles the actual wrapping logic.
- Error beeps added when no valid `targetItem` can be found.

image

In the screenshot above, up/down can now be used to navigate between M/B and K/C, but not L/A.

### Reason for Changes

Grids have been somewhat tricky for keyboard/screenreader navigation. If they are announced as a list, users tend to only go down the first column and not ever discover there are more columns.

### DIFF
--- a/plugins/field-grid-dropdown/src/grid.ts
+++ b/plugins/field-grid-dropdown/src/grid.ts
@@ -6,10 +6,12 @@
 
 import {
   browserEvents,
+  common,
   FieldDropdown,
   ImageProperties,
   MenuOption,
   utils,
+  WorkspaceSvg,
 } from 'blockly/core';
 import {GridItem} from './grid_item';
 
@@ -171,10 +173,10 @@ export class Grid {
 
     switch (e.key) {
       case 'ArrowUp':
-        this.moveFocus(-1 * this.columns, true);
+        this.moveFocus(-1, true, false);
         break;
       case 'ArrowDown':
-        this.moveFocus(this.columns, true);
+        this.moveFocus(1, true, false);
         break;
       case 'ArrowLeft':
         this.moveFocus(-1 * (this.rtl ? -1 : 1), true);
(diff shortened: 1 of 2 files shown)


plugins/field-grid-dropdown/src/grid.ts
```diff
@@ -244,21 +246,41 @@ export class Grid {
...
+   * @param horizontal True for a left/right move (default). False for an
+   *     up/down move. Vertical moves wrap to the next/previous column at
+   *     column edges, but do not wrap around the ends of the grid.
    */
-  private moveFocus(index: number, relative: boolean) {
-    let targetIndex = index;
+  private moveFocus(
+    movementIndex: number,
+    relative: boolean,
+    horizontal = true,
```
gonfunko: I'd make this mandatory and update the other callsites to pass it, no need to maintain backwards compatibility since it's private.

gonfunko (approved):

PR #2761, opened 2026-07-28
Cross origin copy

### LINKED ISSUE #2496: Add support for cross-origin cross-tab copy and paste
### Check for duplicates

### Component

cross-tab copy paste

### Problem

_No response_

### Request

The cross-tab copy-paste plugin currently only works for sites on the same domain, since it uses localstorage to persist serialized blocks, which has a same origin policy enforced by the browser. It would be nice if it were possible to copy blocks across domains; most likely this would require coordination/allowlisting specific domains and communicating over `postMessage`, but it should in theory be possible as an additional option. One usecase (staging vs prod environments) was outlined here: https://groups.google.com/g/blockly/c/qK_jP9mG_pw.

### Alternatives considered

- Consider using system clipboard using old tricks to copy with a hidden text field
- Consider using new clipboard web apis

### Additional context

_No response_

TannerGabriel:
## The basics

- [x] I [validated my changes](https://docs.blockly.com/guides/contribute/samples#making_and_verifying_a_change)

### Resolves

Fixes #2496

### Proposed Changes

I personally use the system clipboard for copying between different origins and thought I would share the implementation. Let me know if this is what you had in mind.

### DIFF
--- a/plugins/cross-tab-copy-paste/README.md
+++ b/plugins/cross-tab-copy-paste/README.md
@@ -51,6 +51,7 @@ Blockly.ContextMenuRegistry.registry.getItem('blockPasteFromStorage').weight =
 
 - `contextMenu` `{boolean}`: Register copy, cut, and paste commands in the Blockly context menu.
 - `shortcut` `{boolean}`: Register cut (ctr + x), copy (ctr + c) and paste (ctr + v) in the keybord shortcut.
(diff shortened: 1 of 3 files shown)


maribethb:
Hey, thanks for the PR! I think there's a couple problems with this that would prevent us from accepting as is, but before I do a detailed review I want to check with @gonfunko because he's looked at this lately and may have some work in progress, other review thoughts, or we might think this isn't worth putting in the plugin if it's going to move into core.

Blocking issues with this approach:
- needs to respect things like the block limits (the current local storage approach does but not if pasted from system clipboard, if I'm reading correctly)

PR #2764, opened 2026-08-26
chore: update github pages for plugins

### LINKED ISSUE #7767: Update github pages site on samples
This is part of the [monorepo project](https://github.com/google/blockly/issues/7755).

- Continue to host the page from `blockly-samples`, but have the build scripts pull from the `blockly` repository.
- Update the github pages action to either only be manually run, or be triggered from the `blockly` repository, if that’s possible

This is a good candidate for a 20%er to work on.

lizschwab:
- [x] I [validated my changes](https://docs.blockly.com/guides/contribute/samples#making_and_verifying_a_change)

### Resolves

Fixes [#7767](https://github.com/RaspberryPiFoundation/blockly/issues/7767)

### Proposed Changes

This updates the Blockly-Samples github pages to pull in the code from the core repo but still host on blockly-samples. There are no changes to how examples are hosted or built.

### Reason for Changes

The plugins have moved to core, so we need to pull the code from there rather than from the stale and soon-to-be-deleted directory on the samples repo.

### DIFF
--- a/.github/workflows/update_gh_pages.yml
+++ b/.github/workflows/update_gh_pages.yml
@@ -26,20 +26,20 @@ jobs:
 
     steps:
       - name: Checkout Blockly-Samples
-        uses: actions/checkout@v3
+        uses: actions/checkout@v7
         with:
           ref: ${{ inputs.samples_branch || 'main' }}
           path: blockly-samples
 
       - name: Checkout Blockly
-        uses: actions/checkout@v3
+        uses: actions/checkout@v7
         with:
           repository: 'RaspberryPiFoundation/blockly'
           ref: ${{ inputs.core_branch || 'main' }}
           path: blockly
 
       - name: Setup Node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v7
         with:
           node-version: 24
 
@@ -48,6 +48,13 @@ jobs:
         working-directory: blockly-samples
         run: |
           npm ci
+
+      - name: Pepare Plugins for deployment
+        working-directory: blockly
+        run: |
+            npx nx reset && lerna run clean --scope '@blockly/*' && \
+            lerna run build --scope '@blockly/*' && \
+            lerna run predeploy --scope '@blockly/*' 
 
       - name: Deploy to GH Pages
         env:
--- a/.github/workflows/update_gh_pages.yml
+++ b/.github/workflows/update_gh_pages.yml
@@ -5,8 +5,13 @@ name: Update GitHub Pages
 on:
   workflow_dispatch: # Manually trigger. Colon is required.
     inputs:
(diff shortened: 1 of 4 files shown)


.github/workflows/update_gh_pages.yml
```diff
@@ -5,8 +5,13 @@ name: Update GitHub Pages
...
     inputs:
-      branch:
-        description: 'Branch to publish from'
+      samples_branch:
+        description: 'Blockly-Samples branch to publish from (for examples)'
+        required: true
+        default: 'main'
+        type: string
+      core_branch:
+        description: 'Blockly core branch to publish from (for plugins)'
```
maribethb, line 14: I think with the way you're passing this in, it can also be a tag right? I think the default should be leaving this blank and having it use the latest published version of the plugins (so if the user didn't specify a branch/tag, you'd add a step finding the most recent git tag and use that as the checkout ref) because `main` may contain unreleased changes. but this could be a follow-up issue because this is really a feature request on top of this PR

package.json
```diff
@@ -9,7 +9,7 @@
     "clean": "lerna run clean --scope '@blockly/*'",
     "deploy:prepare": "npm run deploy:prepare:plugins && npm run deploy:prepare:examples && gulp predeploy",
     "deploy:prepare:examples": "npm install --workspace=examples && npm run prepare-examples && npm run predeploy --if-present --workspace=examples",
-    "deploy:prepare:plugins": "npm run clean && npm run build && lerna run predeploy --scope '@blockly/*'",
+    "deploy:prepare:plugins": "cd ../blockly && npx nx reset && lerna run clean --scope '@blockly/*' && lerna run build --scope '@blockly/*' && lerna run predeploy --scope '@blockly/*'",
```
maribethb: This is kind of sketchy to me because it presupposes that blockly is at `../blockly` and there's no inherent reason that must be true (especially when running locally, but could also get out of sync if the workflow file ever changed the checkout location). I think these steps might belong in the workflow file itself for deploying on CI. and deploying locally might need some other solution

.github/workflows/update_gh_pages.yml
```diff
@@ -20,10 +25,18 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout
+      - name: Checkout Blockly-Samples
+        uses: actions/checkout@v3
```
maribethb: may be worth updating the version of the checkout command action while you're here, i think they're on 4 or 5 now

package.json
lizschwab: Created an issue for the local side of this: https://github.com/RaspberryPiFoundation/blockly-samples/issues/2766

maribethb (approved):

PR #2768, opened 2026-09-11
fix(minimap): preserve field editor focus when tabbing between fields

### LINKED ISSUE #10429: Minimap affecting focus
### Check for duplicates

### Component

Minimap plugin

### Description

When editing a text field within a block, pressing tab will sometimes fail to move focus to the next text field.

### Reproduction steps

1. Open the minimap test page here https://raspberrypifoundation.github.io/blockly-samples/plugins/workspace-minimap/test/index.html
2. Drag in a "count with i from" block
3. Click on the first number "1" and change that to "2".
4. Press tab.

The focus should shift to the second number "10", but instead, the "10" briefly flashes and focus is lost.

### Stack trace

```shell

```

### Screenshots

_No response_

sharadvc:
## Summary

Fixes RaspberryPiFoundation/blockly#10429.

When editing a numeric field on a block and pressing Tab to move to the next field, focus briefly landed on the next field and was then lost. This happened because the minimap mirrored `BLOCK_CHANGE` events and then called `zoomToFit()` after queued renders finished. That async `zoomToFit()` could run while a field editor still held ephemeral focus, steal DOM focus, and close the editor.

## Fix

Skip `zoomToFit()` while Blockly's `FocusManager` reports ephemeral focus (for example, while a field editor is open). The minimap still mirrors block changes immediately; only the deferred fit/zoom is deferred until editing finishes.

## Testing

- Added a regression test that reproduces the reported flow (`controls_for` with `math_number` inputs, edit the first value, press Tab).
- All existing `@blockly/workspace-minimap` tests pass.


gonfunko:
Thanks for the PR! This broadly seems reasonable, but we've recently moved the plugins into the [core Blockly repository](https://github.com/RaspberryPiFoundation/blockly). Do you mind retargeting it to that repo? The minimap plugin now lives here: https://github.com/RaspberryPiFoundation/blockly/tree/main/packages/plugins/workspace-minimap

